PR #10, opened 2026-01-17
Add sticky caching behaviour

JKamsker:
Stick conversations to accounts unless the account is empty, then it reassignes the thread. 

# Why?
Not sure if constantly using different accoutns for the same convo affects caching. 
Also on OAI side this must look very suspicious (Same thread from different accounts)

### DIFF
--- a/tests/integration/test_proxy_sticky_sessions.py
+++ b/tests/integration/test_proxy_sticky_sessions.py
@@ -40,11 +40,13 @@ def _make_auth_json(account_id: str, email: str) -> dict:
     }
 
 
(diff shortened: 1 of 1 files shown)


JKamsker:
Hey! Sure, but what's the deal? 

Depleted accounts will be ignored and are what we want. Evenly used accounts can lead to accounts being reset before all tokens are used up.

In combination with logic that is not yet in this branch (it's in the pile) - this is exactly what the user wants - to use up accounts that get reset earlier.

JKamsker:
@Soju06 done 
image

PR #12, opened 2026-01-17
feat: add cached input tokens handling and update related metrics in …

JKamsker:
image

Bit ugly maybe - maybe we should add separate sections for cached displays (separate card and separate column?)


Soju06:
@codex review

Soju06 (commented):
Thanks for the contribution!
Could you please check the comment I left below?

PR #14, opened 2026-01-18
Support multiple accounts per Team/Business workspace

Quack6765:
## Summary
This PR updates account identification and persistence to support multiple ChatGPT accounts under the same Team/Business workspace, where the upstream account ID can be shared across multiple members.

## Why
In Team/Business workspaces, multiple users can appear with the same upstream account identifier. Treating that value as a globally-unique primary key causes collisions (one member overwriting another) and prevents importing multiple accounts from the same workspace.

## What changed
- **Unique account IDs for Team/Business**: generate a unique internal `Account.id` by appending an email-hash suffix when both `chatgpt_account_id` and a real email are available (`<chatgpt_account_id>_<8-char-hash>`). This avoids collisions while keeping the upstream ID readable.
- **Persist the upstream ID separately**: add `Account.chatgpt_account_id` to store the raw upstream identifier while keeping `Account.id` as the unique primary key.
- **SQLite migration on startup**: when using SQLite, automatically adds the `chatgpt_account_id` column if it’s missing.
- **Repository updates**: ensure `chatgpt_account_id` is written/updated during upserts and token refresh updates.

## Testing
- `uv run ruff check .`
- `uv run pytest`

## Rollout / compatibility
- Existing accounts keep their current `Account.id` behavior.
- New imports for Team/Business members will no longer overwrite each other when the upstream account ID is shared.


Quack6765:
I'll need to update my branch and reconcile the changes since the PR was done before @Soju06 implemented a lot of changes in his main branch a couple of hours later which caused conflict with my changes.

Quack6765:
Also included: Fixed `usage_limit_reached` errors being misclassified as `QUOTA_EXCEEDED`. These are actually rolling time-window limits with a `resets_at` timestamp, so they're now treated as `RATE_LIMITED`. Previously it made the account marked as rate limited not skipped by the routing process. Added `reset_at`column to persist the reset time across requests so accounts are properly skipped until the limit resets.

PR #58, opened 2026-02-04
chore(main): release 0.6.0

Soju06:
:robot: I have created a release *beep* *boop*
---

### Features

* **api:** OpenAI compatibility layers for Responses support ([#56](https://github.com/Soju06/codex-lb/issues/56)) ([3e95eb1](https://github.com/Soju06/codex-lb/commit/3e95eb134fc6066c6891830d6dd62a876b4526ee))
* **dashboard:** refactor load path and usage refresh ([#59](https://github.com/Soju06/codex-lb/issues/59)) ([899de74](https://github.com/Soju06/codex-lb/commit/899de74e48c8bace2fbbac92a0f9f6b5c699d15f))
* TOTP AUTH FOR WEB PANEL ([#61](https://github.com/Soju06/codex-lb/issues/61)) ([d05df1e](https://github.com/Soju06/codex-lb/commit/d05df1e6f658f6397c2ddaf7c0297814722839f0)), closes [#62](https://github.com/Soju06/codex-lb/issues/62)

### Documentation

* add dwnmf as a contributor for code, and test ([#63](https://github.com/Soju06/codex-lb/issues/63)) ([26bd133](https://github.com/Soju06/codex-lb/commit/26bd1334e727129a0e51168e222753ce485c737e))
* **openspec:** add context docs policy ([#57](https://github.com/Soju06/codex-lb/issues/57)) ([8a491f8](https://github.com/Soju06/codex-lb/commit/8a491f88637d3b4eb28e24aa5063f495350ecca1))

---
This PR was generated with [Release Please](https://github.com/googleapis/release-please). See [documentation](https://github.com/googleapis/release-please#release-please).

PR #59, opened 2026-02-05
feat(dashboard): refactor load path and usage refresh

Soju06:
## Summary
- move usage refresh to background scheduler and remove request-path refresh
- add dashboard overview endpoint and supporting dashboard module
- slim account/usage payloads to UI-needed fields only
- refactor frontend to use overview endpoint and lazy-load options/settings

## Testing
- .venv/bin/python -m pytest

### DIFF
--- a/app/static/index.html
+++ b/app/static/index.html
@@ -515,7 +515,9 @@
             <p class="text-muted">Toggle routing features. When both options are off, accounts are selected by
               balancing usage evenly.</p>
 
-            <fieldset class="settings-fieldset">
+            <p class="text-muted" x-show="settings.isLoading" x-transition.opacity>Loading settings…</p>
+
+            <fieldset class="settings-fieldset" :disabled="settings.isLoading || settings.isSaving || !settings.hasLoaded">
               <legend>Sticky threads</legend>
               <input id="sticky-threads-toggle" type="checkbox" x-model="settings.stickyThreadsEnabled"
                 aria-describedby="sticky-threads-help">
@@ -527,7 +529,7 @@
               </div>
             </fieldset>
 
-            <fieldset class="settings-fieldset">
+            <fieldset class="settings-fieldset" :disabled="settings.isLoading || settings.isSaving || !settings.hasLoaded">
               <legend>Reset priority</legend>
               <input id="reset-priority-toggle" type="checkbox" x-model="settings.preferEarlierResetAccounts"
                 aria-describedby="reset-priority-help">
@@ -539,7 +541,7 @@
             </fieldset>
 
             <div class="inline-actions" style="margin-top: 12px">
-              <button type="button" @click="saveSettings" :disabled="settings.isSaving">
+              <button type="button" @click="saveSettings" :disabled="settings.isLoading || settings.isSaving || !settings.hasLoaded">
                 <span>Save</span>
               </button>
             </div>
--- a/app/static/index.js
+++ b/app/static/index.js
@@ -1484,6 +1484,7 @@
 			isLoading: true,
 			hasInitialized: false,
 			refreshPromise: null,
+			settingsLoadPromise: null,
 
 			async init() {
 				if (this.hasInitialized) {
@@ -1580,24 +1581,33 @@
 			},
 
 			async ensureSettingsLoaded() {
-				if (this.settings.hasLoaded || this.settings.isLoading) {
+				if (this.settings.hasLoaded) {
 					return;
 				}
+				if (this.settingsLoadPromise) {
+					return this.settingsLoadPromise;
+				}
 				this.settings.isLoading = true;
+				this.settingsLoadPromise = (async () => {
+					try {
+						const settings = await fetchSettings();
+						this.settings.stickyThreadsEnabled = settings.stickyThreadsEnabled;
+						this.settings.preferEarlierResetAccounts =
+							settings.preferEarlierResetAccounts;
+						this.settings.hasLoaded = true;
+					} catch (err) {
+						console.error("Failed to load settings:", err);
+						this.openMessageBox({
+							tone: "error",
+							title: "Settings load failed",
+							message: err?.message || "Failed to load settings.",
+						});
+					}
+				})();
 				try {
-					const settings = await fetchSettings();
-					this.settings.stickyThreadsEnabled = settings.stickyThreadsEnabled;
-					this.settings.preferEarlierResetAccounts =
-						settings.preferEarlierResetAccounts;
-					this.settings.hasLoaded = true;
-				} catch (err) {
-					console.error("Failed to load settings:", err);
-					this.openMessageBox({
-						tone: "error",
-						title: "Settings load failed",
-						message: err?.message || "Failed to load settings.",
-					});
+					await this.settingsLoadPromise;
 				} finally {
+					this.settingsLoadPromise = null;
 					this.settings.isLoading = false;
 				}
 			},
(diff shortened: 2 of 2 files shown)


Soju06:
@codex review

vitalNohj:
@nohjbot review

Both defects you found on `eccf9cc1` are confirmed fixed, and one of them was only half-working - detail below. New head is `dab0f32b`.

Your two findings:

1. **Models cache TTL inert** - fixed by the captain's `fbddc891` (config-keyed client accessors on the catalog, dashboard and service paths). I found a follow-up: a *deleted* endpoint's cached client kept its decrypted API key alive for the process lifetime, since deletion produces no config to evict the entry with. Now reconciled against the live endpoint set on every settings read.

2. **In-band SSE error before `[DONE]` classified as success** - fixed by `fbddc891`, but it was **inert for any CRLF upstream**: the decoder split only on `\n\n`, so a CRLF stream buffered to EOF and the `error` event was never parsed at all. Fixed the framing, and added `test_a_crlf_framed_error_before_done_is_still_seen` so the classification cannot silently stop working again.

I also worked through Kody's 14 threads: 11 fixed, 2 answered as not-real with evidence, 2 already fixed. Two of your earlier positions are relevant:

- Your read that the cross-chunk UTF-8 issue only perturbs content the parser ignores was **mostly** right - raw bytes do pass through to the client untouched. But a multi-byte character landing inside the event carrying `usage` corrupts that JSON, which loses the number the reservation settles against. Fixed since the decoder was being rewritten for the framing anyway.
- Your read on the "authorization bypass" and "missing credential validation" findings both held up; I answered those threads as not-real with the code paths and tests.

Worth your attention on this head:
- `_settle_stream_deferring_cancellation` in both dispatchers - PR 58's pattern applied to the two sidecars that were cloned before it landed.
- `update_operational_json_column_if_unchanged` - a column-level CAS, chosen over the settings-version CAS because the version CAS would stale an open Settings form (an existing test pins that).
- `asyncio.gather(..., return_exceptions=True)` in `_build_models_response_body` - the bare gather regressed per-endpoint failure isolation.

Full suite on this head is down to the 2 failures that already exist on `main`; the 2 the PR introduced (unregenerated settings reference, over-strict provider-order guard) are fixed. `ty` is back to main's exact 645 diagnostics.

vitalNohj:
@kody review --force

Worked through all 14 threads on the new head `dab0f32b`. Per-thread replies are posted with the reasoning and the regression tests; summary:

**Fixed (11)** - each with a regression that fails before the fix and passes after:
- SSE CRLF framing + cross-chunk UTF-8, both dispatchers. Reused the decoder already proven for this on `opencode_go_sidecar_dispatch.py` rather than the suggested `.replace()` normalization, which would also rewrite CR/LF bytes inside a `data:` JSON payload.
- Cancellation-unsafe settlement, both dispatchers. Applied PR #58's `_await_result_deferring_cancellation` pattern instead of the detached-task registry your rule suggests - the maintainer explicitly declined that registry rule on #58, and the deferral achieves the same guarantee with nothing process-wide to drain.
- Base URL validation. Validated the *shape* (userinfo, query, fragment, dot segments, scheme, host) at save time and again at client construction, rather than pinning NVIDIA's canonical host: this is the generic OpenAI-compat endpoint list, so arbitrary hosts are the feature.
- Client cache retention, bearer redaction, lost-update race, N+1 usage query, serial `/v1/models` fanout.

**Answered as not real (2)**, left open for a human to judge:
- *Model-visibility authorization bypass.* Authorization is `validate_model_access` / `_access_identities_match`, which does include provider ownership and is enforced at dispatch, with dedicated cross-provider tests in `test_responses_model_access_routing.py`. `_model_visible_for_api_key` is only the `/v1/models` listing filter and is slug-keyed identically for all eight providers, so these two are consistent with the established contract rather than diverging from it.
- *Missing credential validation before dispatch.* Byte-for-byte identical to the OpenRouter and OrcaRouter dispatchers, and claiming the route without a credential is the documented "ownership, deliberately NOT usability" convention - dropping the entry would route the prompt to a different provider entirely.

**Already fixed before this review run (2)** by `fbddc891`: the NVIDIA models cache and the SSE error classification. Note the classification fix was inert for CRLF upstreams until the framing thread above was fixed.

Two things your findings led to that are worth flagging:
- Your "isolate each endpoint failure" caveat on the fanout thread was load-bearing: a bare `asyncio.gather` propagates the first exception and abandons the batch, so one unreachable endpoint would have emptied the catalog for all the others - worse than the serial loop. Now uses `return_exceptions=True`.
- The lost-update fix could not use the existing settings-version CAS: it bumps `version` and an existing test pins that a test-connection must not stale an open Settings form. Added a column-level compare-and-set instead.

Also fixed two defects the suite caught that no thread covered: the settings reference page was never regenerated after the NVIDIA settings were added, and the opencode-go provider-order guard asserted an exhaustive list where its own docstring specifies only relative order.

Validation on this head: full suite down to the 2 failures that pre-exist on `main`; `ruff` clean; `ty` back to main's exact 645 diagnostics (no new ones); architecture check passes.

vitalNohj:
Head is now `8037ff6a`. Two of your six new findings were real and are fixed; the other four are restatements of findings already answered on this PR, and I have replied on each thread.

**Fixed:**
- **Encoded-traversal bypass in the base URL validator** - a genuine gap in the validation I added on the previous head. The literal dot-segment check missed `%2e%2e`, and since the HTTP client canonicalizes *after* validation, `https://host/v1/%2e%2e/%2e%2e/admin` + `/models` left as `https://host/admin/models` with the bearer token on it. Verified against `yarl` rather than reasoned about. Percent-encoded paths and backslashes now refused outright.
- **Dashboard `GET /api/models` serial fanout** - I fixed this for `/v1/models` and missed the identical loop in the dashboard model picker. Now gathered, with `return_exceptions=True` to preserve the per-endpoint isolation its previous per-endpoint `try` provided.

**Answered, not changed** (each with the code path, test or spec on the thread): NVIDIA host pinning, which would turn a deliberately configurable self-hosted-NIM endpoint into a constant and is inconsistent with every other sidecar on `main`; the detached-task registry, where the `ensure_future` is awaited on the very next line inside a shield so no task outlives the request; pre-dispatch credential validation, which is byte-for-byte the OpenRouter/OrcaRouter behaviour and the documented "ownership, not usability" convention; and the OpenSpec wording, where provider ownership is specified once in the `api-keys` capability and already enforced by `validate_model_access`.

@nohjbot review

vitalNohj:
@kody review --force

vitalNohj:
Head is now `8037ff6a`. Two of your six new findings were real and are fixed; the other four are restatements of findings already answered on this PR, and I have replied on each thread.

**Fixed:**
- **Encoded-traversal bypass in the base URL validator** - a genuine gap in the validation I added on the previous head. The literal dot-segment check missed `%2e%2e`, and since the HTTP client canonicalizes *after* validation, `https://host/v1/%2e%2e/%2e%2e/admin` + `/models` left as `https://host/admin/models` with the bearer token on it. Verified against `yarl` rather than reasoned about. Percent-encoded paths and backslashes now refused outright.
- **Dashboard `GET /api/models` serial fanout** - I fixed this for `/v1/models` and missed the identical loop in the dashboard model picker. Now gathered, with `return_exceptions=True` to preserve the per-endpoint isolation its previous per-endpoint `try` provided.

**Answered, not changed** (each with the code path, test or spec on the thread): NVIDIA host pinning, which would turn a deliberately configurable self-hosted-NIM endpoint into a constant and is inconsistent with every other sidecar on `main`; the detached-task registry, where the `ensure_future` is awaited on the very next line inside a shield so no task outlives the request; pre-dispatch credential validation, which is byte-for-byte the OpenRouter/OrcaRouter behaviour and the documented "ownership, not usability" convention; and the OpenSpec wording, where provider ownership is specified once in the `api-keys` capability and already enforced by `validate_model_access`.

@nohjbot review

PR #65, opened 2026-02-10
Firewall-first update with stronger dashboard auth, SSL support, and proxy stability fixes

dwnmf:
Hi Soju, this PR is mainly about introducing a proper API firewall first, including middleware enforcement and dashboard endpoints to manage rules safely. On top of that, it strengthens dashboard authentication with TOTP and setup-token flow fixes, adds SSL support for the API server, and improves proxy and OpenAI-compat behavior for better reliability under real usage. It also includes the required DB migrations, tighter typed contracts in service and repository layers, and broader unit and integration coverage so the new behavior is easier to trust and maintain. Validation on my side with py: `py -m ty check .` passes, `py -m ruff check .` passes, and `py -m pytest -q` passes (`292 passed, 1 skipped` on Windows).

### DIFF
--- a/app/modules/firewall/repository.py
+++ b/app/modules/firewall/repository.py
@@ -1,9 +1,14 @@
 from __future__ import annotations
 
 from sqlalchemy import delete, select
+from sqlalchemy.exc import IntegrityError
 from sqlalchemy.ext.asyncio import AsyncSession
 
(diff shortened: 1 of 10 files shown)


dwnmf:
@Soju06 What do you think?

PR #100, opened 2026-02-25
fix(app-routing): add routing strategy setting and fix true round-robin runtime rotation

joeblack2k:
## Summary
- add configurable dashboard routing strategy: `usage_weighted` (default) or `round_robin`
- plumb `routingStrategy` through settings API, service, repository, schemas, frontend forms, and status UI
- add Alembic migration `013_add_dashboard_settings_routing_strategy`
- implement round-robin selection in balancer using least recently selected account
- fix runtime behavior so round-robin state persists across requests by making `ProxyService` dependency singleton-scoped
- add selection logging in `LoadBalancer` for easier production verification

## Why
The new setting was present but real-world rotation could still collapse because the load-balancer runtime state was reset per request. This patch makes round-robin deterministic and observable in production.

## Tests
- `uv run pytest tests/unit/test_load_balancer.py tests/integration/test_settings_api.py tests/integration/test_migrations.py -q`
  - result: `26 passed`

## Notes
- frontend tests were not run in this environment.

### DIFF
--- a/app/core/balancer/logic.py
+++ b/app/core/balancer/logic.py
@@ -2,8 +2,7 @@ from __future__ import annotations
 
 import time
 from dataclasses import dataclass
-from typing import Literal
-from typing import Iterable
+from typing import Iterable, Literal
 
 from app.core.balancer.types import UpstreamError
 from app.core.utils.retry import backoff_seconds, parse_retry_after


Soju06:
@all-contributors please add @joeblack2k for code, bug, test

PR #143, opened 2026-03-08
fix(proxy): preserve Codex session affinity for compact to stop 502s

aaiyer:
## Summary
- pin backend Codex routing to the inbound `session_id` contract so compact stays on the same account as the owning thread
- scope `session_id` affinity to `/backend-api/codex/*` instead of treating it as a proxy-wide routing signal
- add regression coverage for backend Codex stickiness and the negative `/v1/*` case

## Why
We were seeing `502` failures from compact when load balancing moved `/backend-api/codex/responses/compact` to a different upstream account than the account that owned the existing Codex thread. The original Codex client already sends the stable session/thread identifier, so the proxy needs to preserve that account affinity for backend Codex requests.

## Testing
- `.venv/bin/pytest tests/integration/test_proxy_sticky_sessions.py -q`
- `git diff --check`

### DIFF
--- a/app/modules/proxy/api.py
+++ b/app/modules/proxy/api.py
@@ -95,7 +95,7 @@ async def responses(
     context: ProxyContext = Depends(get_proxy_context),
     api_key: ApiKeyData | None = Security(validate_proxy_api_key),
 ) -> Response:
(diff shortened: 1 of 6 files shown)


Soju06:
@codex review

PR #145, opened 2026-03-08
feat(db): optimize SQLite startup and query paths

aaiyer:
## Summary
- add configurable SQLite startup validation with `quick` as the default mode
- replace the startup critical-path full integrity scan with mode-driven validation
- optimize `usage_history` primary-window queries and `request_logs` latest-first listing with new indexes and query-shape changes
- add manual migration drift coverage for the new expression index on SQLite

## Why
SQLite-backed instances were paying a full integrity scan on every startup, and the hot query paths still had avoidable inefficiencies around legacy primary-window reads and request-log listing. This keeps existing behavior while improving startup latency and SQLite query execution.

## Testing
- `.venv/bin/pytest tests/unit/test_db_session.py tests/unit/test_db_migrate.py tests/unit/test_request_logs_repository.py tests/integration/test_usage_repository.py tests/integration/test_request_logs_list_count.py tests/integration/test_request_logs_filters.py tests/integration/test_request_logs_api.py tests/integration/test_repositories.py tests/integration/test_usage_api.py tests/integration/test_migrations.py -q`
- `git diff --check`

### DIFF
--- a/tests/integration/test_usage_repository.py
+++ b/tests/integration/test_usage_repository.py
@@ -1,9 +1,11 @@
 from __future__ import annotations
 
+import json
 from datetime import timedelta
 
 import pytest
 from sqlalchemy import text
+from sqlalchemy.ext.asyncio import AsyncSession
 
 from app.core.crypto import TokenEncryptor
 from app.core.utils.time import utcnow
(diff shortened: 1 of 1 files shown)


Soju06:
This is a perfect PR. Thanks a lot for the thoughtful work here.

PR #150, opened 2026-03-09
fix(proxy): align compact retry account header after refresh

aaiyer:
## Summary
- fix compact retry to forward the refreshed provider account header on retry
- add regression coverage for a 401 refresh plus compact retry
- update the existing Codex session-routing OpenSpec change to document the retry contract

## Why
The earlier compact routing fix stopped mid-thread account hopping and was merged upstream in #143. There was still one remaining compact 502 path: if the selected account refreshed after a 401, the retry reused the pre-refresh `chatgpt-account-id` header instead of the refreshed account identity. That can still produce provider-side mismatch failures during compaction in multi-account setups.

## Testing
- /mnt/scratch/home/aaiyer/codex-lb/.venv/bin/pytest tests/integration/test_proxy_compact.py tests/integration/test_proxy_sticky_sessions.py -q
- /mnt/scratch/home/aaiyer/codex-lb/.venv/bin/ruff check app/modules/proxy/service.py tests/integration/test_proxy_compact.py
- git diff --check

### DIFF
--- a/app/modules/proxy/service.py
+++ b/app/modules/proxy/service.py
@@ -125,11 +125,11 @@ class ProxyService:
                 openai_error("no_accounts", selection.error_message or "No active accounts available"),
             )
         account = await self._ensure_fresh(account)
-        account_id = _header_account_id(account.chatgpt_account_id)
         request_service_tier = _service_tier_from_compact_payload(payload)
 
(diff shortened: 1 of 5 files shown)


Soju06:
@codex review

Soju06:
Thanks a lot for this fix, @aaiyer. Really appreciate the fast follow-up here.

PR #175, opened 2026-03-10
fix(proxy): preserve dedicated responses compact contract

defin85:
## Summary
- preserve `/responses/compact` as a dedicated direct upstream contract instead of reconstructing it through normal `/responses`
- return the canonical upstream compaction window unchanged, omit `store` upstream, and keep compact failures inside the compact contract
- add bounded same-contract retry plus compact-specific failure diagnostics and spec/test coverage

## Verification
- `/home/egor/services/codex-lb-defin85/.venv/bin/pytest -q tests/unit/test_openai_requests.py tests/unit/test_proxy_utils.py tests/integration/test_proxy_compact.py`
- `/home/egor/services/codex-lb-defin85/.venv/bin/pytest -q tests/integration/test_proxy_sticky_sessions.py tests/integration/test_proxy_api_extended.py tests/integration/test_api_keys_api.py`
- `openspec validate --specs`

## Related
- Follow-up to #169

### DIFF
--- a/tests/integration/test_proxy_websocket_responses.py
+++ b/tests/integration/test_proxy_websocket_responses.py
@@ -333,16 +333,6 @@ def test_backend_responses_websocket_forwards_non_create_events_to_active_upstre
                     separators=(",", ":"),
                 ),
             ),
-            _FakeUpstreamMessage(
-                "text",
-                text=json.dumps(
-                    {
-                        "type": "response.completed",
-                        "response": {"id": "resp_ws_control", "object": "response", "status": "completed"},
-                    },
-                    separators=(",", ":"),
-                ),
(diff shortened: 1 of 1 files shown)


defin85:
Reviewer notes:
- Review this PR as a compact-contract fix, not as a transport/protocol expansion. The goal is to keep `/responses/compact` semantically separate from normal `/responses`.
- Key invariants: raw compact payload is returned unchanged, `store` is omitted upstream, no surrogate fallback to `/codex/responses`, and retries stay `compact -> compact` only.
- Suggested review focus: compact payload typing/parsing, fail-closed boundaries, safe retry conditions, and compact-specific diagnostics/tests.

PR #237, opened 2026-03-20
Fix: accounts were marked as quota_exceeded as soon as the 7d window hit 100%, even when paid credits were available.

mws-weekend-projects:
- Quota decision now checks credits (has, unlimited, balance) before blocking an account.
  - If credits are usable, the account stays/returns active.
  - Added unit tests for both cases (with credits vs without credits).

  Result:
  Accounts with valid credits are no longer falsely blocked, while true quota-exhausted accounts are still blocked correctly.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -22,3 +22,7 @@ htmlcov/
 refs/
 docs/
 tests/
+
+# Explicit frontend excludes to keep Docker context small
+frontend/node_modules/
+**/node_modules/
--- a/app/core/usage/quota.py
+++ b/app/core/usage/quota.py
@@ -15,6 +15,9 @@ def apply_usage_quota(
     runtime_reset: float | None,
     secondary_used: float | None,
     secondary_reset: int | None,
+    credits_has: bool | None = None,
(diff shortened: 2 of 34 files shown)


L1st3r:
Is this something we still want? I can try to pick it up and get it ready for merging if so...

mws-weekend-projects:
I still think this would be very useful. Without it, it’s a bit like driving through a traffic-calmed street in the city with speed cameras but no dashboard — you really want to know your actual speed.

That said, the implementation is a bit tricky in practice. So for now, it might be better to leave this as a feature request unless someone comes up with a cleaner implementation. I may also improve it later on.

Also, the current logic is still pretty limited because until now I had only used Plus accounts. I’m now also using token-based billing (1000 for USD 40), and there are also USD 200 Pro accounts, so there are clearly more account/usage models to consider.

Soju06:

Komzpa:
Revived on a maintainer-owned branch as #764 because this PR head has maintainerCanModify=false and cannot be updated here. Keeping credit to the original implementation in the replacement PR.

PR #243, opened 2026-03-21
refactor(proxy): reduce bridge and balancer lock contention

aaiyer:
## Summary
- remove HTTP bridge global session-creation serialization by singleflighting same-key creation while allowing different keys to create in parallel
- narrow load-balancer runtime locking so non-sticky selection runs outside the lock and persistence happens after the in-memory commit
- add concurrency regressions for bridge session creation and load-balancer selection behavior

## Verification
- `.venv/bin/python -m py_compile app/modules/proxy/service.py app/modules/proxy/load_balancer.py tests/integration/test_http_responses_bridge.py tests/unit/test_proxy_load_balancer_refresh.py`
- `uv run ty check app/modules/proxy/service.py app/modules/proxy/load_balancer.py tests/integration/test_http_responses_bridge.py tests/unit/test_proxy_load_balancer_refresh.py`
- `.venv/bin/python -m pytest -q tests/unit/test_proxy_load_balancer_refresh.py tests/integration/test_http_responses_bridge.py tests/integration/test_proxy_websocket_responses.py tests/integration/test_proxy_api_extended.py`

## Notes
- keeps the downstream websocket `response_create_gate` in place because the current `response.created` matching logic still depends on a single unresolved pre-ID request per socket
- sticky-key routing remains on the conservative locked path; this change only relaxes the non-sticky hot path

### DIFF
--- a/app/modules/proxy/service.py
+++ b/app/modules/proxy/service.py
@@ -1441,6 +1441,7 @@ class ProxyService:
         while True:
             sessions_to_close: list[_HTTPBridgeSession] = []
             inflight_future: asyncio.Future[_HTTPBridgeSession] | None = None
+            capacity_wait_future: asyncio.Future[_HTTPBridgeSession] | None = None
             owns_creation = False
             continuity_error: ProxyResponseError | None = None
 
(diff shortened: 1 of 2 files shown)


aaiyer:
@Soju06 could you take a look at this and merge it if it looks good?

This one is worth landing because it removes the main global contention points in the proxy hot path. In practice that means better throughput and lower tail latency under concurrent Codex traffic, especially for HTTP `/responses` bridge requests and balancer-heavy workloads. It also includes the follow-up correctness fixes needed to make the concurrency changes safe: bridge continuity/replay handling, stale-selection retries, request-id observability, and startup failure normalization.

Net effect: fewer requests blocked behind unrelated work, fewer spurious bridge failures under retry/reconnect pressure, and better behavior under bursty multi-session load.

PR #249, opened 2026-03-22
fix(api-keys): normalize timezone-aware expirations before persistence

yigitkonur:
## Summary
- normalize timezone-aware `expiresAt` values to UTC-naive datetimes before API key create/update writes
- add unit regression coverage for create and update flows with offset-aware datetimes
- document the accepted `expiresAt` behavior in OpenSpec

## Notes
- split out from the broader work discussed in #211
- does not change the database backend contract
- keeps sqlite as the default backend and postgres as optional

## Testing
- `uv run pytest tests/unit/test_api_keys_service.py -q`
- `uv run ruff check app/modules/api_keys/service.py tests/unit/test_api_keys_service.py`
- `python3 -m compileall app/modules/api_keys/service.py tests/unit/test_api_keys_service.py`
- `uv run openspec validate --specs` *(fails in this environment because the `openspec` executable is not installed)*

### DIFF
--- a/app/modules/api_keys/service.py
+++ b/app/modules/api_keys/service.py
@@ -13,7 +13,7 @@ from app.core.usage.pricing import (
     calculate_cost_from_usage,
     get_pricing_for_model,
 )
-from app.core.utils.time import utcnow
+from app.core.utils.time import to_utc_naive, utcnow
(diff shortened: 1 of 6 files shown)


Soju06:
@codex review

PR #294, opened 2026-04-01
Enhancement: sticky session mass delete

L1st3r:
I realize the functionality was already created, but I felt I could polish it a little bit...

1. Better backend delete result contract
     Upstream returned only:
      - deletedCount

     This now returns:
      - deletedCount
      - deleted
      - failed with per-row reasons

     Reference:
      - app/modules/sticky_sessions/api.py
      - app/modules/sticky_sessions/service.py
      - app/modules/sticky_sessions/schemas.py
  2. Best-effort partial failure handling
     Upstream bulk delete assumed success as a count.
     This explicitly reports rows that could not be deleted instead of collapsing everything into one number.
  3. Better frontend behavior after bulk delete
     This branch adds:
      - keeping failed rows selected
      - page fallback when deletion empties the current page
      - preserving pagination/filter context more intentionally

     Reference:
      - frontend/src/features/sticky-sessions/components/sticky-sessions-section.tsx
      - frontend/src/features/sticky-sessions/hooks/use-sticky-sessions.ts
  4. More explicit UX wording
     Changed the bulk action from “Remove selected” to “Delete Sessions” and made the confirmation mention failure reporting.
  5. OpenSpec coverage
    Added a proper spec change under:
      - openspec/changes/bulk-delete-sticky-sessions

### DIFF
--- a/frontend/src/features/sticky-sessions/schemas.test.ts
+++ b/frontend/src/features/sticky-sessions/schemas.test.ts
@@ -39,7 +39,15 @@ describe("StickySessionsListResponseSchema", () => {
 describe("StickySessionsListParamsSchema", () => {
   it("defaults pagination parameters", () => {
     const parsed = StickySessionsListParamsSchema.parse({});
-    expect(parsed).toEqual({ staleOnly: false, offset: 0, limit: 10 });
+    expect(parsed).toEqual({
+      staleOnly: false,
+      accountQuery: "",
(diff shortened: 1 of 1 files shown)


L1st3r:
I will fix this up shortly!

Soju06:
Thank you for waiting! I’ll include everything in the new version

PR #303, opened 2026-04-02
fix(dashboard): show remaining totals in usage donuts

Daltonganger:
## Summary
- use aggregate remaining credits instead of aggregate capacity for the dashboard 5h and weekly usage donut totals
- keep the donut center value aligned with the existing per-account remaining slices and the visible “Remaining” label
- add a focused frontend regression test so the dashboard view does not fall back to capacity totals again

## Testing
- `bun run test -- src/features/dashboard/utils.test.ts src/features/dashboard/components/usage-donuts.test.tsx`
- `bun run lint -- src/features/dashboard/components/dashboard-page.tsx src/features/dashboard/utils.ts src/features/dashboard/utils.test.ts`
- `bun run typecheck`

### DIFF
--- a/frontend/src/features/dashboard/utils.test.ts
+++ b/frontend/src/features/dashboard/utils.test.ts
@@ -161,6 +161,16 @@ describe("applySecondaryConstraint", () => {
     expect(result[0].remainingPercent).toBe(90);
   });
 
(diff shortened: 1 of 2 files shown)


Soju06:
@all-contributors please add @Daltonganger for bug

PR #321, opened 2026-04-03
fix(helm): disable service links and use fully qualified image names

Soju06:
## Summary
- Add `enableServiceLinks: false` to all pod specs (deployment, migration job, helm test)
- Use fully qualified image references (`docker.io/library/...`) for cri-o compatibility

## Problem
The chart creates a `codex-lb-metrics` Service. Kubernetes automatically injects service discovery env vars into all pods in the same namespace:
```
CODEX_LB_METRICS_PORT=tcp://10.96.x.x:9090
CODEX_LB_METRICS_SERVICE_HOST=10.96.x.x
```
The app expects `CODEX_LB_METRICS_PORT` to be an integer (e.g., `9090`), causing a Pydantic validation error on startup. This affects the main deployment, migration hooks, and schema gate init containers — making the chart unusable when the metrics service exists in the same namespace.

Additionally, cri-o runtimes (used by OKE, OpenShift) reject short image names like `busybox:1.37`, requiring `docker.io/library/` prefix.

## Impact
- **Without this fix**: 100% crash rate on any cri-o cluster, and crash on any cluster when metrics service is enabled (default)
- **With this fix**: Zero-config compatibility with all major K8s distributions

## Changes
- `templates/deployment.yaml`: `enableServiceLinks: false`
- `templates/hooks/migration-job.yaml`: `enableServiceLinks: false`
- `templates/tests/test-connection.yaml`: `enableServiceLinks: false`
- Fully qualified image names where applicable

### DIFF
--- a/deploy/helm/codex-lb/templates/hooks/migration-job.yaml
+++ b/deploy/helm/codex-lb/templates/hooks/migration-job.yaml
@@ -43,33 +43,33 @@ spec:
       {{- end }}
       {{- if .Values.postgresql.enabled }}
       initContainers:
-         - name: wait-for-db
-           image: docker.io/library/postgres:16-alpine
-           imagePullPolicy: IfNotPresent
-           securityContext:
-             allowPrivilegeEscalation: false
-             readOnlyRootFilesystem: true
-             runAsNonRoot: true
-             runAsUser: 999
-             capabilities:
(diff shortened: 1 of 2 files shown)


Soju06:
@codex review

PR #329, opened 2026-04-04
fix: avoid Windows startup crash in memory monitor and add manual reg…

L1st3r:
…ression workflow

Make the resilience memory monitor portable on Windows by removing the unconditional import of the Unix-only resource module, adding a Windows-compatible RSS lookup path, and falling back safely when RSS telemetry is unavailable.

Add regression tests for Windows import behavior and the no-provider fallback, and add a manually triggered GitHub Actions workflow to smoke test the startup path on windows-latest without running on every push.


Soju06:
@all-contributors please add @L1st3r for bug

Issue #324, opened 2026-04-03
Codex CLI /models is incorrect


asmichel:
What I've tried:
    - start afresh with a bar config.toml containing only authentication

I suspect that `backend-api/codex` is the source of the issue. This is because
    `curl http://localhost:2455/v1/models`
correctly prints out models and the command
    `curl http://localhost:2455/backend-api/codex/v1/models`
from https://www.mintlify.com/Soju06/codex-lb/clients/codex-cli#error-model-not-found gives
    `{"error":{"message":"Not Found","type":"invalid_request_error","code":"not_found"}}`

Soju06:
Closing — as @huzky-v explained, this is a limitation in how the Codex CLI handles model fetches from custom providers. The CLI doesn't currently fetch `/backend-api/codex/models` without a code change on the Codex CLI side itself.

Workarounds that work today:
- Use the `requires_openai_auth = true` provider flag (same as #125) so the CLI can fetch the model list through the supported path.
- Or explicitly specify the model on the CLI, e.g. `codex --model gpt-5.5` / `codex --model gpt-5.5-mini`, instead of relying on auto-discovery.

If/when the upstream Codex CLI adds support for fetching models from non-OpenAI base URLs without the auth flag, we can revisit.

PR #331, opened 2026-04-05
fix(api): for /backend-api/codex/model, return it in codex format

huzky-v:
Address part of #324 , Updating `/backend-api/codex/models` API to a format that at least `codex-cli` accepts

Although that the `codex-cli` behavior cannot be changed this moment, and the change of `/backend-api/codex/models` does not affect the `codex-cli` `/model` because of the `codex` mechanism

It is always good to align the format with the upstream and prepare it when `codex` is compatible with that one day.

Tested on my modified `codex-cli` with success.

### DIFF
--- a/app/modules/proxy/api.py
+++ b/app/modules/proxy/api.py
@@ -325,10 +325,7 @@ async def _build_codex_models_response(api_key: ApiKeyData | None) -> Response:
         request_service_tier=None,
     )
 
-    allowed_models = set(api_key.allowed_models) if api_key and api_key.allowed_models else None
-    if api_key and api_key.enforced_model:
-        forced = {api_key.enforced_model}
-        allowed_models = forced if allowed_models is None else (allowed_models & forced)
+    allowed_models = _allowed_models_for_api_key(api_key)
 
     registry = get_model_registry()
     models = registry.get_models_with_fallback()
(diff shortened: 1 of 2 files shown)


huzky-v:
Oh it could be done by some simple shell script and aliasing.

Something like this will do
```
#!/usr/bin/env bash
set -euo pipefail

CLIENT_VERSION=$(codex --version | sed 's/^codex-cli //')

MODELS_RESPONSE=$(curl -s -H "Authorization: Bearer ${ENV_KEY}" https://your.codex-lb.location/backend-api/codex/models -o test.json)

jq -n \
  --arg fetched_at "2099-12-31T23:59:59.000000Z" \
  --arg client_version "$CLIENT_VERSION" \
  --argjson models "$(cat test.json | jq '.models')" \
  '{fetched_at: $fetched_at, client_version: $client_version, models: $models}' \
  > ~/.codex/models_cache.json

rm test.json
exec codex "$@"
```
and do an alias to codex.

Although the allowed model selection in `codex-lb` will not exactly equal to the output of `/model`, which still constrained by some kind of the `codex-cli` default list, which has 6 of them

Soju06:
@codex review

Soju06:
@codex review

Soju06:
Thank you for your contribution!

huzky-v (commented):
Just filtering out the list on `/backend-api/codex/models` for API Allowed Model is not enough for `codex-cli` as it has a default `visibility` value for various model. You need to explicitly set it to `hide` in the response so that the `/models` will not show the model 

Filtering may not align the `/v1/models` behavior to just let the app to show the models available to the API

PR #364, opened 2026-04-08
Fail closed on account_deactivated errors

### LINKED ISSUE #353: Deactivated upstream accounts can stay in the pool after usage API 401
﻿## Summary

If an upstream OpenAI account gets deactivated, `codex-lb` can keep trying to use it instead of removing it from the available pool.

In my case this caused traffic to stall until I manually disabled / removed the bad account.

## What I observed

I had an account that was deactivated on the OpenAI side. Around `2026-04-04 13:38-13:50` (`UTC+05:00`) `codex-lb` repeatedly logged this:

```text
Usage fetch failed request_id=None status=401 message=Your OpenAI account has been deactivated, please check your email for more information. If you feel this is an error, contact us through our help center at help.openai.com.
```

This kept happening over and over while normal traffic was still flowing through other websocket requests.

The practical problem was:

- the account was already dead upstream
- but `codex-lb` still kept trying to refresh usage / route through it
- and I had to manually remove that account so it would stop affecting the pool

## Why I think this happens

From reading the current code, this looks like a specific gap in the usage-refresh path.

### 1. The usage client keeps only the message, not the upstream error code

In `app/core/clients/usage.py`:

- `UsageErrorDetail` parses `message` and `error_description`
- but it does **not** parse `error.code`
- `UsageFetchError` then keeps only `status_code` and `message`

So even if upstream returns something structured like `account_deactivated`, that information appears to be dropped in this path.

### 2. The usage updater only auto-deactivates on `402` and `404`

In `app/modules/usage/updater.py`:

```py
_DEACTIVATING_USAGE_STATUS_CODES = {402, 404}
```

So a `401` coming back from the usage endpoint does **not** deactivate the account automatically, even if the message is clearly:

> Your OpenAI account has been deactivated...

### 3. Result: a permanently dead account can be treated like a non-terminal usage/auth problem

That means a deactivated account may remain in circulation until someone manually removes or pauses it.

## Expected behavior

If upstream indicates that an account is deactivated, `codex-lb` should treat that as a permanent failure and remove the account from the routable pool automatically.

At minimum, I think one of these should happen:

1. If upstream provides something like `error.code = account_deactivated`, preserve it and treat it as permanent.
2. If only the message text is available, detect the deactivation message and treat it as permanent.
3. Mark the account `DEACTIVATED` (or otherwise make it non-routable) so the balancer stops selecting it.

## Why this matters

This is more serious than a dashboard status mismatch.

A dead upstream account should fail closed and leave the pool immediately. Otherwise one bad account can keep generating retries / refresh attempts and degrade the whole LB until someone notices and removes it by hand.

## Related but different

I know there is already a sticky-session / `quota_exceeded` issue, but this seems different:

- this is not about stale quota state
- this is about an upstream-deactivated account not being treated as permanently dead in the usage path

If useful, I can also provide the exact code locations I traced locally.

SHAREN:
## Summary
- treat `account_deactivated` as a permanent failure so affected accounts become `deactivated` and leave the routing pool automatically
- preserve usage API `error.code` values and fail closed on usage-path deactivation signals, including message-only 401 fallbacks
- add regression coverage for balancer handling, usage refresh handling, and websocket handshake deactivation classification
- document the behavior change in OpenSpec

Closes #353.

## Testing
- frontend: `docker run --rm -v /home/renat/apps/codex-lb:/workspace -w /workspace/frontend oven/bun:1.3.10 sh -lc 'bun install --frozen-lockfile && bun run lint && bun run typecheck && bun run build'`
- backend: `docker run --rm -v /home/renat/apps/codex-lb:/workspace -w /workspace python:3.13-slim bash -lc 'pip install -q uv && uv sync --dev --frozen && uv run ruff check . && uv run ruff format --check . && uv run ty check'`
- targeted tests: `docker run --rm -v /home/renat/apps/codex-lb:/workspace -w /workspace python:3.13-slim bash -lc 'pip install -q uv && uv sync --dev --frozen && uv run pytest -q tests/unit/test_auth_refresh.py tests/unit/test_load_balancer.py tests/unit/test_usage_client.py tests/unit/test_usage_updater.py tests/unit/test_proxy_utils.py tests/unit/test_select_with_stickiness.py'`

### DIFF
--- a/app/core/balancer/logic.py
+++ b/app/core/balancer/logic.py
@@ -14,6 +14,7 @@ PERMANENT_FAILURE_CODES = {
     "refresh_token_expired": "Refresh token expired - re-login required",
     "refresh_token_reused": "Refresh token was reused - re-login required",
     "refresh_token_invalidated": "Refresh token was revoked - re-login required",
+    "account_deactivated": "Account has been deactivated",
     "account_suspended": "Account has been suspended",
     "account_deleted": "Account has been deleted",
 }
--- a/app/core/clients/proxy.py
+++ b/app/core/clients/proxy.py
@@ -140,6 +140,7 @@ _HOP_BY_HOP_HEADER_NAMES = frozenset(
 _AUTO_WEBSOCKET_HANDSHAKE_FALLBACK_STATUSES = frozenset({426})
 _WEBSOCKET_RESPONSE_CREATE_EXCLUDED_FIELDS = frozenset({"background", "stream"})
 _WEBSOCKET_HANDSHAKE_ERROR_HINTS = (
+    ("account_deactivated", "account has been deactivated"),
     ("usage_not_included", "usage not included"),
     ("insufficient_quota", "insufficient quota"),
     ("quota_exceeded", "quota exceeded"),
--- a/app/core/clients/usage.py
+++ b/app/core/clients/usage.py
@@ -23,6 +23,7 @@ logger = logging.getLogger(__name__)
 class UsageErrorDetail(BaseModel):
     model_config = ConfigDict(extra="ignore")
 
+    code: str | None = None
     message: str | None = None
     error_description: str | None = None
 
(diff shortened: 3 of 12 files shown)


SHAREN:
image
image

Soju06:
@codex review

Soju06:
Thanks @SHAREN! Clean fail-closed implementation with solid test coverage across all three paths (balancer, usage refresh, websocket handshake). The message-only fallback for structured-code-less 401s is a nice touch. Merged!

Soju06:
@all-contributors please add @SHAREN for code, test

PR #366, opened 2026-04-08
feat(auth): add dashboard proxy auth modes

Daltonganger:
## Summary
- add env-backed dashboard auth modes for standard, trusted-header, and disabled access, including trusted proxy header sanitization and fail-closed behavior when reverse-proxy auth is required
- update the dashboard session/settings UI to expose auth mode state, block direct access when Authelia-style headers are missing, and hide password/TOTP management when app-level auth is disabled
- document the new Docker/env configuration in README and `.env.example`, add OpenSpec change artifacts, and cover the flows with backend and frontend tests

## Validation
- uv run ruff check app/core/middleware/dashboard_auth_proxy.py app/core/middleware/__init__.py app/main.py app/core/auth/dashboard_mode.py app/modules/dashboard_auth/api.py tests/integration/test_auth_middleware.py
- uv run pytest tests/integration/test_auth_middleware.py tests/integration/test_dashboard_totp_auth.py
- bun run typecheck
- bunx vitest run --environment jsdom src/features/auth/hooks/use-auth.test.ts

PR #367, opened 2026-04-08
feat(auth): add a flag CODEX_LB_DISABLE_BOOTSTRAP_TOKEN to disable remote auth requirement

huzky-v:
Just observed that the remote needs authentication, which in some case is not necessary for remote deployment 
In my case, I have put the `codex-lb` behind a `pangolin` reverse proxy, which already handles authentication for me

So I have added `CODEX_LB_DISABLE_BOOTSTRAP_TOKEN` (default to `false`, to preserve the current behavior, needs to set password on first auth)

When `CODEX_LB_DISABLE_BOOTSTRAP_TOKEN` is set to true, there will be no authentication required, just like before.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -74,4 +74,7 @@ CODEX_LB_SHUTDOWN_DRAIN_TIMEOUT_SECONDS=30
 CODEX_LB_HTTP_CONNECTOR_LIMIT=100
 CODEX_LB_HTTP_CONNECTOR_LIMIT_PER_HOST=50
 
+# Remote bootstrap token gate (optional)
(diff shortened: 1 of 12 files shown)


huzky-v:
Seems #366 is handling the same issue with finer control. Maybe stick to that instead.

If that PR is getting merged, you may close this PR without merging it

Soju06:
Closing in favor of #366 which provides a more comprehensive solution with three dashboard auth modes (`standard`/`trusted_header`/`disabled`). The `disabled` mode covers this PR's use case. Thanks @huzky-v for the clean implementation and for flagging the overlap!

PR #373, opened 2026-04-08
Recover quota-exceeded accounts after early resets and restarts

### LINKED ISSUE #293: Accounts can stay stuck in quota_exceeded after an early upstream reset
﻿## Summary

After what looked like an unexpected early/global quota reset on the OpenAI side, some accounts in `codex-lb` stayed stuck in `quota_exceeded` even though their usage data was already fully reset.

In my case, two accounts got stuck like this:

- `sharen1994@gmail.com`
- `jeanna_curran71@outlook.com`

The dashboard/API showed all of these at the same time:

- `status = quota_exceeded`
- `primaryRemainingPercent = 100.0`
- `secondaryRemainingPercent = 100.0`
- valid auth / no deactivation reason

So the accounts looked fully reset from the usage side, but were still treated as quota-blocked from the status side.

## Why this matters

This is not just a dashboard display issue.

As far as I can tell, the balancer treats `AccountStatus.QUOTA_EXCEEDED` as a hard exclusion until its stored `reset_at` is reached, even if fresh usage rows already say the account is back to 100% remaining.

That means an early upstream reset can leave healthy accounts unavailable inside `codex-lb`.

## What I observed

From `/api/accounts`, the stuck accounts looked like this (abridged):

- `status: quota_exceeded`
- `usage.primaryRemainingPercent: 100.0`
- `usage.secondaryRemainingPercent: 100.0`
- `remainingCreditsPrimary: full`
- `remainingCreditsSecondary: full`
- `deactivationReason: null`
- auth still valid

Other accounts in the same pool were shown as `active` normally.

## Suspected cause

My guess is that this happens when:

1. an account was previously marked `quota_exceeded`
2. `codex-lb` stored a future `reset_at`
3. OpenAI resets quotas earlier than expected
4. fresh usage data now shows the account as fully reset
5. but `codex-lb` keeps the old `quota_exceeded` status because the old stored `reset_at` has not been reached yet

So usage and account status drift apart.

## Expected behavior

If fresh usage data shows the quota is fully reset again, `codex-lb` should be able to clear stale `quota_exceeded` automatically instead of waiting only for the previously stored reset timestamp.

In other words, early upstream resets should unstick accounts automatically.

## Notes

This may be rare in normal operation, but if OpenAI does an unscheduled/global reset, it seems worth handling, because otherwise some accounts stay unavailable until manually reactivated.

SHAREN:
## Summary
- persist `blocked_at` on accounts so quota-block recovery survives process restarts
- clear stale `quota_exceeded` guards when fresh governing usage proves the quota has recovered
- recover legacy stuck accounts that never recorded a block marker only when fresh healthy usage also shows a newer reset window
- cover the new behavior with unit tests and OpenSpec notes

## Why
Upstream `#289` improved early-reset handling, but recovery still depended on runtime-only markers. After a restart, or for older already-stuck accounts without a persisted block marker, `quota_exceeded` could still remain stuck even when fresh usage had clearly recovered.

This change keeps the existing mitigation, makes it durable across restarts, and keeps the legacy recovery path conservative by requiring evidence of a newer reset window before reactivating an account that has no persisted block marker.

## Testing
- `ruff check` on touched files
- `ruff format --check` on touched files
- `ty check app tests/unit/test_auth_manager.py tests/unit/test_load_balancer.py tests/unit/test_proxy_load_balancer_refresh.py tests/unit/test_usage_updater.py`
- `pytest -q tests/unit/test_auth_manager.py tests/unit/test_load_balancer.py tests/unit/test_proxy_load_balancer_refresh.py tests/unit/test_usage_updater.py`

Closes #293


Soju06:
Thanks @SHAREN — solid fix for restart-safe quota recovery. Rebased the Alembic migration onto the current head and fixed lint. Fourth PR and consistently high quality!

Issue #365, opened 2026-04-08
401 Unauthorized error after updating to v1.12 (Proxy authentication issue)


Webificio:
Hello,
I recently updated my Codex LB to version 1.12 (docker).
Upon starting, the dashboard forced me to set up the CODEX_LB_DASHBOARD_BOOTSTRAP_TOKEN and create a new password.

I completed this step, and the dashboard is now loading and working perfectly. 
However, when my client tries to communicate with Codex, I receive the following error:

**unexpected status 401 Unauthorized: Proxy authentication must be configured before remote access is allowed, url: http://127.0.0.1:2455/backend-api/codex/responses**

It seems like the dashboard authentication is fine, but the API endpoint now requires proxy authentication that wasn't required (or was configured differently) in previous versions.

Could you please guide me on how to properly configure the proxy authentication to resolve this?
Thank you!

jonathanlaniado:
Ran into the same issue. Running codex-lb in Docker on Mac, so even though the Codex desktop app and codex-lb are on the same machine, the container networking layer makes every request appear remote (e.g. `192.168.215.1` instead of `127.0.0.1`). That's what triggers the 401.

Fix was:

1. **Set `CODEX_LB_DASHBOARD_BOOTSTRAP_TOKEN`** as an env var on the container so the remote setup screen lets you set a password
2. **Enable API key auth** in the dashboard under Settings, then create a key
3. **Add the key to Codex config** (`~/.codex/config.toml`):
   ```toml
   [model_providers.codex-lb]
   env_key = "CODEX_LB_API_KEY"
   ```
4. **Set `CODEX_LB_API_KEY="sk-clb-..."`** in your environment (e.g. `.zshenv`, `.bashrc`, etc.)
5. **Restart the Codex app**

Webificio:
Thanks @jonathanlaniado

meeftah:
if you run this on local machine then you can just add `network_mode: host` to access the service using `127.0.0.1`

kSherp:
Unfortunately it didn't work for me, the environment variable is visible and configured normally, but Codex still returns 401 Unauthorized

Soju06:
Thanks — that narrows it down.

The key detail is the exact error string:

> `Proxy authentication must be configured before remote access is allowed`

[...]

PR #374, opened 2026-04-09
docs(api-keys): clarify local-only behavior when auth is disabled

aruis:
## Summary
- clarify the API key auth toggle copy to cover protected proxy requests without over-expanding the dashboard UI copy
- update README guidance to explain that disabling API key auth only allows local requests and can still reject Docker/container-networked clients as non-local
- add an OpenSpec change and sync the main `api-keys` spec wording with the implemented fail-closed behavior

## Validation
- reviewed the copy updates across the dashboard, README, and OpenSpec
- `openspec validate --specs`
- `cd frontend && bun run typecheck`

## Context
- aligns user-facing wording with the runtime behavior introduced by PR #339
- related upstream discussion: #365

### DIFF
--- a/README.md
+++ b/README.md
@@ -294,7 +294,7 @@ from openai import OpenAI
 
 client = OpenAI(
     base_url="http://127.0.0.1:2455/v1",
-    api_key="sk-clb-...",  # from dashboard when auth is enabled; local-only requests may omit it when disabled
+    api_key="sk-clb-...",  # from dashboard, or any non-empty string if auth is disabled
 )
 
 response = client.chat.completions.create(
@@ -321,7 +321,7 @@ Authorization: Bearer sk-clb-...
 
 The protected proxy routes covered by this setting are:
 
-- `/v1/*`
+- `/v1/*` (except `/v1/usage`, which always requires a valid key)
 - `/backend-api/codex/*`
 - `/backend-api/transcribe`
 
--- a/README.md
+++ b/README.md
@@ -280,7 +280,8 @@ opencode
 }
 ```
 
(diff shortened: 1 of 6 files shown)


Soju06 (commented):
## Codex Adversarial Review — 2 findings fixed

Reviewed the changes against the actual runtime implementation (`validate_proxy_api_key` vs `validate_usage_api_key` in `app/modules/proxy/api.py` and `app/core/auth/dependencies.py`). Two accuracy issues found:

### P1 (High): `/v1/usage` excluded from protected route glob

`/v1/usage` uses `validate_usage_api_key`, not the proxy auth guard. It **always** requires a valid key even when `api_key_auth_enabled=false`. The current list overstates scope.

```suggestion
- `/v1/*` (except `/v1/usage`, which always requires a valid key)
```

*(applies to line 324 of README.md)*

### P2 (Medium): Python SDK `api_key` cannot be omitted

`openai.OpenAI(api_key=...)` raises at construction time if `api_key` is None and `OPENAI_API_KEY` env var is unset — before any HTTP request. "may omit it" is misleading.

```suggestion
    api_key="sk-clb-...",  # from dashboard, or any non-empty string if auth is disabled
```

*(applies to line 297 of README.md)*

---

Everything else looks good — the local/non-local distinction is accurately documented, delta spec matches the main spec changes, and no conflicts with other active changes. Will merge once these two lines are fixed.

Soju06:
Thanks @aruis for aligning the docs with the actual fail-closed behavior — this was causing real confusion for Docker users. Pushed a small fix for the /v1/usage exception and SDK comment accuracy before merging.

Soju06:
@all-contributors please add @aruis for doc

PR #379, opened 2026-04-09
feat(ui): UI adjustments on dashboards

huzky-v:
1. Add a 12h/24h switch in setting so that the UI can show 12h/24h time, which the user chose for
Closes #369   
螢幕截圖 2026-04-09 22 17 49  

2. The pie in the donut will enlarge a slightly little bit when the legend text is hovered. 
The item will also be highlighted as box
Also there is a maximum number of legend items could be shown so that the dashboard will not get overflown
螢幕截圖 2026-04-09 22 17 25

PR #386, opened 2026-04-10
fix(ui): some append fix for #379

huzky-v:
Append #379 
1. Just found that the behavior of showing 5 legends max is not aligned with the #379 described, adjusted accordingly.
2. For the accounts, this is also scrollable now, to prevent producing overflown page

螢幕截圖 2026-04-10 20 46 49


Soju06:
Thanks @huzky-v — clean follow-up to #379. The 5-row legend and scrollable account grid makes the dashboard much more usable with many accounts.

Soju06:
@all-contributors please add @huzky-v for code, test

PR #388, opened 2026-04-11
fix(http-bridge): propagate bridged Spark model errors as HTTP 400

OverHash:
## Summary
- Reproduced an OpenCode UX mismatch for unsupported `gpt-5.3-codex-spark` on Team accounts: direct ChatGPT path surfaced the error, but the codex-lb bridge path did not.
- Compared `mitmproxy` captures and confirmed behavior difference: direct backend (to ChatGPT API) returned `HTTP 400` JSON, while codex-lb returned `HTTP 200` SSE terminal error. This causes OpenCode to not display the error correctly.
- Updated HTTP bridge error handling to preserve normalized SSE events and also propagate first terminal bridged errors as `ProxyResponseError` when `propagate_http_errors=True`, so `/v1/responses` returns `HTTP 4xx` JSON.
- Added integration coverage for the bridged upstream `type:error` payload and asserted `POST /v1/responses` returns `400` with the expected OpenAI error envelope.

## Root Cause
- In `_stream_via_http_bridge`, `propagate_http_errors` was discarded, so terminal bridged errors were always emitted as stream events under a `200` response.

Fixed and ran tests. I used red-green testing approach to fix this issue. I then ran a local build of codex-lb to make sure that OpenCode surfaces the error correctly.

### DIFF
--- a/tests/integration/test_http_responses_bridge.py
+++ b/tests/integration/test_http_responses_bridge.py
@@ -5671,7 +5671,7 @@ async def test_v1_responses_http_bridge_stream_failure_remains_valid_sse(async_c
 
 
 @pytest.mark.asyncio
-async def test_v1_responses_http_bridge_normalizes_upstream_error_event(async_client, monkeypatch):
+async def test_v1_responses_http_bridge_surfaces_upstream_error_event_as_http_400(async_client, monkeypatch):
     _install_bridge_settings(monkeypatch, enabled=True)
     account_id = await _import_account(
         async_client,
@@ -5734,10 +5734,9 @@ async def test_v1_responses_http_bridge_normalizes_upstream_error_event(async_cl
     monkeypatch.setattr(proxy_module.ProxyService, "_ensure_fresh_with_budget", fake_ensure_fresh_with_budget)
     monkeypatch.setattr(proxy_module, "connect_responses_websocket", fake_connect_responses_websocket)
 
-    events = await _collect_sse_events(
-        async_client,
+    response = await async_client.post(
         "/v1/responses",
-        json_body={
+        json={
             "model": "gpt-5.3-codex-spark",
             "instructions": "Return exactly OK.",
             "input": "hello",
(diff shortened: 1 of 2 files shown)


tests/integration/test_http_responses_bridge.py
```diff
@@ -307,6 +307,31 @@ async def send_text(self, text: str) -> None:
...
+                            "type": "invalid_request_error",
+                            "code": "invalid_request_error",
+                            "message": (
+                                "The 'gpt-5.3-codex-spark' model is not supported when using Codex "
+                                "with a ChatGPT account."
+                            ),
+                        },
+                    },
+                    separators=(",", ":"),
+                ),
```
OverHash, line 330: This is what OpenAI ChatGPT backend directly sends when you attempt to use 5.3 Codex Spark on a Team account.

Soju06 (commented):
## Codex Review — 2 findings

Tested on a clean rebase against main. The Spark model error fix (400 path) works correctly, but the same codepath runs for 429/quota errors where it strips important metadata.

### P1 (High): Rate-limit retry metadata lost in HTTP error response

**File:** `app/modules/proxy/service.py` — `_openai_error_envelope_from_response_failed_payload()`

`openai_error()` only carries `code`, `type`, `message`, `param`. When a 429 comes through the bridge path and `propagate_http_errors` fires, the HTTP JSON response drops `resets_at`, `resets_in_seconds`, and `plan_type` that the non-bridge stream path preserves. Clients like OpenCode lose retry timing info.

**Suggested fix:** Forward the full upstream `error` dict from the response payload instead of rebuilding it from scratch. Or at minimum, copy known rate-limit fields into the envelope.

### P2 (Medium): Synthesized `response.failed` drops reset metadata for balancer

**File:** `app/modules/proxy/service.py` — `_normalize_http_bridge_error_event()`

The synthetic `response.failed` event only carries `message/type/code/param`. When `_finalize_websocket_request_state()` later parses this to update account health, `resets_at`/`resets_in_seconds` are gone. The balancer falls back to generic backoff instead of the actual reset time → can bring a rate-limited account back into rotation too early.

---

The 400 path (Spark model error) is correct and well-tested. The issue is specifically with 429/quota errors going through the same reconstruction helpers. Consider preserving the full upstream error payload for non-400 cases, or pass through rate-limit fields explicitly.

Soju06:
Thanks @OverHash! Great catch on the bridge error propagation gap — the red-green testing approach paid off. Pushed a follow-up commit preserving rate-limit metadata (plan_type/resets_at/resets_in_seconds) for 429 bridge errors so the balancer and clients get proper retry timing.

Soju06:
@all-contributors please add @OverHash for code, test

PR #390, opened 2026-04-11
Guard oversized response.create before upstream websocket send

SHAREN:
﻿## Summary
- guard serialized upstream `response.create` payloads before attempting the upstream websocket send
- slim historical inline images and oversized historical tool outputs before the most recent user turn when that is enough to fit under budget
- return deterministic local `payload_too_large` errors for HTTP and websocket Responses routes when the request still does not fit
- persist oversized payload dumps and add OpenSpec + regression coverage for the guarded/slimmed paths

## Why
Threads with accumulated screenshots, inline images, and large tool outputs could build a `response.create` payload large enough to trigger upstream websocket `1009 (message too big)`. That surfaced as opaque `stream_incomplete` failures and reconnect loops. This change turns that into a local preflight guard and recovery path.

## Failure mode
Large threads could accumulate historical inline images, Playwright screenshots, and oversized tool outputs inside the serialized `response.create` payload. Even a small follow-up message could then fail because the proxy still had to resend that historical payload upstream.

In production this surfaced as:

`Upstream websocket closed before response.completed: received 1009 (message too big); then sent 1009 (message too big)`

This is a websocket message-size problem in bytes, not just a token-window overflow.

## OpenSpec
- added `openspec/changes/guard-oversized-response-create`

## Testing
- `ruff check app/modules/proxy/service.py tests/integration/test_http_responses_bridge.py tests/integration/test_proxy_websocket_responses.py tests/unit/test_proxy_utils.py`
- `python -m py_compile app/modules/proxy/service.py`
- `pytest -q tests/unit/test_proxy_utils.py -k "payload_too_large or slim_response_create or prepare_websocket_response_create_request"`
- `pytest -q tests/integration/test_http_responses_bridge.py tests/integration/test_proxy_websocket_responses.py`
- `npx -y @fission-ai/openspec validate --specs`

### DIFF
--- a/app/modules/proxy/service.py
+++ b/app/modules/proxy/service.py
@@ -5849,10 +5849,11 @@ def _openai_error_envelope_from_response_failed_payload(
     param_value = error_payload.get("param")
     if isinstance(param_value, str) and param_value.strip():
         envelope["error"]["param"] = param_value.strip()
+    error_detail = envelope["error"]
     for key in ("plan_type", "resets_at", "resets_in_seconds"):
         value = error_payload.get(key)
         if value is not None:
-            envelope["error"][key] = value  # type: ignore[literal-required]
+            cast(dict[str, object], error_detail)[key] = value
     return envelope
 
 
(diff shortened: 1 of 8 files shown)


Soju06:
Thanks @SHAREN — great production fix for the 1009 message-too-big issue. Pushed review fixes: removed silent progressive drop (now fails with payload_too_large per spec), scoped 1009 dumps to the offending request, preserved tool-continuation inputs, and added object-valued content slimming. Third PR from you and all solid work!

Soju06:
@all-contributors please add @SHAREN for code, test

PR #396, opened 2026-04-13
fix(proxy): websocket connect-phase failover + deterministic failover integration tests

### LINKED ISSUE #385: Don't surface usage_limit_reached to Codex Desktop when other accounts are still available
## Summary

I use `codex-lb` as the provider for Codex Desktop, with multiple OpenAI accounts behind it.

When one account runs out of usage, `codex-lb` immediately returns `usage_limit_reached` to Codex Desktop.

Codex Desktop treats this as a terminal failure and stops the current request, even though other accounts in the pool may still be available.

## Why this matters

From the Codex Desktop side, it looks like the whole provider is out of quota.

But in reality, only one account is exhausted. Other accounts may still be healthy.

Because of this, the current turn dies too early even though the pool could still continue working.

## Expected behavior

If one account is out of quota but other accounts are still available, `codex-lb` should not immediately send the raw `usage_limit_reached` error back to Codex Desktop.

Instead, it should do one of these:

1. Switch to another available account internally.
2. Return a different retryable proxy-level error so Codex Desktop retries the request, and the retry can be routed to another healthy account.

## Goal

A single exhausted account should not kill the current Codex Desktop request if there are still healthy accounts left in the pool.

## Desired flow

- account A hits `usage_limit_reached`
- `codex-lb` does not stop the request immediately
- Codex Desktop retries
- `codex-lb` routes the retry to account B
- work continues

## Remaining gap after #396

PR #396 fixed websocket connect-phase failover, but on `codex-lb` v1.13.1 Codex Desktop can still receive raw `usage_limit_reached` on the HTTP responses bridge path.

That means the original problem is still reproducible in this case:

- account A is selected for the current turn
- account A has just exhausted quota
- other accounts in the pool still have quota
- Codex Desktop still gets a terminal quota error and stops

## What is still needed

For the HTTP bridge/session creation path, if the selected account returns `usage_limit_reached` (or a similar quota/rate-limit error) and there are still spare healthy accounts, `codex-lb` should not return that raw quota error to Codex Desktop.

Instead, it should return a retryable temporary proxy-level error so Codex Desktop retries automatically, and the retry can be routed to another account.

In practice, this follow-up uses the existing `502 upstream_unavailable` contract as a minimal fallback.

## User-facing scenario

A typical Codex Desktop flow looks like this:

- I type a message and send it
- the currently selected account has just run out of quota
- another account in the pool still has quota
- today, the message fails immediately and I have to resend it manually
- desired behavior: Codex Desktop gets a retryable temporary error, retries automatically, `codex-lb` selects another account, and the turn continues without manual resend

## Desired fallback flow

- account A hits `usage_limit_reached`
- `codex-lb` marks account A as exhausted for routing
- if another account is still available, `codex-lb` returns a retryable temporary proxy error instead of raw quota
- Codex Desktop retries automatically
- the retry lands on account B
- work continues without the user having to resend the message

Kazet111:
Closes #385

  ## Summary
  This PR implements and hardens deterministic failover behavior across transports, with a focus on websocket connect-phase handling and missing integration coverage.

  ## What changed
  - Added websocket connect-phase failover logic (before first downstream frame) in `ProxyService`.
  - Refactored `_connect_proxy_websocket` into smaller internal helpers to reduce complexity while preserving behavior.
  - Replaced hardcoded retry attempt values with named constants:
    - `_WEBSOCKET_MAX_ACCOUNT_ATTEMPTS`
    - `_STREAM_MAX_ACCOUNT_ATTEMPTS`
  - Added unit coverage for websocket handshake failover on `429 usage_limit_reached`.
  - Added integration coverage for:
    - streaming connect-phase `429` -> transparent failover to account B (T21)
    - compact `quota_exceeded` -> transparent failover to account B (T22)
    - mid-stream error -> surfaced to client, no failover (T23)
    - all accounts `DRAINING` -> best `DRAINING` account still selected (T24)
  - Updated OpenSpec task checklist for completed items (T19, T21–T24).

  ## Notes
  - No behavioral changes were introduced for post-first-frame websocket errors (they are still surfaced, not failed over).

  ## Validation
  - `ruff check` on modified files
  - `py_compile` on modified files
  - `pytest tests/unit/test_proxy_utils.py -k "connect_proxy_websocket or stream_with_retry" -q --timeout=120`

### DIFF
--- a/app/modules/proxy/service.py
+++ b/app/modules/proxy/service.py
@@ -214,6 +214,8 @@ _ACCOUNT_RECOVERY_RETRY_CODES = frozenset(
 _TRANSIENT_RETRY_CODES = frozenset({"server_error"})
 _MAX_TRANSIENT_SAME_ACCOUNT_RETRIES = 3
 _COMPACT_MAX_ACCOUNT_ATTEMPTS = 2
+_STREAM_MAX_ACCOUNT_ATTEMPTS = 3
+_WEBSOCKET_MAX_ACCOUNT_ATTEMPTS = 3
 
 
 @dataclass(frozen=True, slots=True)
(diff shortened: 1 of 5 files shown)


Soju06:
Thank you for the thorough failover work @Kazet111! 🎉

The websocket connect-phase failover, constant extraction, and comprehensive T21-T24 integration tests are a great addition. The Codex adversarial review caught a few edge cases (failover error surface, sticky mapping preservation) which were resolved together.

@all-contributors please add @Kazet111 for code and test

PR #397, opened 2026-04-13
fix: prevent context blowup by preserving previous_response_id on bridge recovery - IMPORTANT FIX FOR CONTEXT BLOWING UP

balakumardev:
## Problem

When an upstream websocket session is lost (session.closed) or a request fails mid-flight, the HTTP bridge recovery paths introduced in v1.12.0 return 400 previous_response_not_found errors. This tells the Codex CLI that the server-side conversation state is gone, causing it to drop previous_response_id and resend the entire conversation history as input.

The result is catastrophic for context window usage:

| Metric | Before (v1.11.x) | After (v1.12.0) |
|--------|------------------|-----------------|
| Per-turn growth | ~2-3K tokens | ~70K tokens |
| Turns to fill 876K window | 200+ | 17 |
| Compactions per session | 0 | 2+ |

## Root Cause

Three recovery paths in the bridge return 400 previous_response_not_found instead of 502 upstream_unavailable:

1. **Mid-request websocket failure** - when the upstream websocket dies during send() and previous_response_id is present, the error handler returns 400 previous_response_not_found instead of 502 upstream_unavailable.

2. **_retry_http_bridge_request_on_fresh_upstream** - when called with send_request=True and previous_response_id is present, immediately marks the request as previous_response_not_found without attempting recovery.

3. **_retry_http_bridge_precreated_request** - same pattern.

When the CLI receives 400 previous_response_not_found, it interprets this as "the server doesn't have this conversation anymore" and resends the full conversation history without previous_response_id, inflating per-turn context by ~20x.

## Fix

- When session.closed and previous_response_id is present, raise 502 immediately (matching pre-v1.12.0 behaviour) so the client retries with previous_response_id intact.
- In error handler, always raise 502 instead of 400 previous_response_not_found.
- In retry helpers, return False without marking previous_response_not_found so callers raise a retriable 502.

## Test Results

7 regression tests added in test_bridge_context_blowup.py. Ran against both broken and fixed code:

| Test | Broken (pre-fix) | Fixed |
|------|:---:|:---:|
| test_mid_request_failure_with_previous_response_id_raises_502 | FAIL (returns 400) | PASS (returns 502) |
| test_retry_with_previous_response_id_returns_false_without_marking_error | FAIL (sets previous_response_not_found) | PASS |
| test_closed_session_with_previous_response_id_raises_502 | PASS | PASS |
| test_closed_session_without_previous_response_id_attempts_recovery | PASS | PASS |
| test_healthy_growth_rate_stays_within_budget | PASS | PASS |
| test_broken_growth_rate_fills_window_fast | PASS | PASS |
| test_error_code_determines_growth_rate | PASS | PASS |

All 44 existing test_proxy_http_bridge.py tests pass with zero regressions.

## Live Verification

Tested on a single-instance deployment with gpt-5.4 (876K effective context window):

- **v1.12.0 (broken)**: 4 user messages -> 853K tokens in 17 turns -> compaction triggered, avg 70K tok/turn
- **v1.11.x (reverted)**: 4 user messages -> ~80K tokens in 17 turns -> no compaction, avg 518 tok/turn

### DIFF
--- a/app/modules/proxy/service.py
+++ b/app/modules/proxy/service.py
@@ -3373,7 +3373,7 @@ class ProxyService:
             # history, inflating per-turn context by ~20x.
             raise ProxyResponseError(
                 502,
-                openai_error("upstream_unavailable", str(exc) or "Upstream websocket closed"),
+                openai_error("bridge_owner_unreachable", str(exc) or "Upstream websocket closed"),
             ) from exc
 
     async def _maybe_prewarm_http_bridge_session(
--- a/app/modules/proxy/service.py
+++ b/app/modules/proxy/service.py
@@ -3579,7 +3579,8 @@ class ProxyService:
         text_data: str,
         send_request: bool = True,
     ) -> bool:
-        effective_send = send_request and request_state.previous_response_id is None
+        if request_state.previous_response_id is not None and send_request:
+            return False
         if request_state.replay_count >= 1:
             return False
         request_state.replay_count += 1
@@ -3598,7 +3599,7 @@ class ProxyService:
                 request_state=request_state,
                 restart_reader=True,
             )
-            if effective_send:
+            if send_request:
                 await session.upstream.send_text(text_data)
             session.last_used_at = time.monotonic()
             return True
--- a/app/modules/proxy/service.py
+++ b/app/modules/proxy/service.py
@@ -3246,25 +3246,14 @@ class ProxyService:
         queue_limit: int,
     ) -> None:
         if session.closed:
-            # When previous_response_id is present, reconnecting the upstream
-            # websocket creates a fresh server-side session that won't recognise
-            # the old response id.  Rather than silently falling back to a full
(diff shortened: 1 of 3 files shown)


Soju06:
Thank you for this important fix @balakumardev! 🎉

The context blowup issue is a critical regression — great catch with the session log analysis showing 70K vs 2.3K tokens/turn.

I've rebased your changes on main and applied Codex adversarial review fixes (preserving the reconnect recovery path for between-turn disconnects). The reviewed version is in #401.

Closing this in favor of #401 which includes your original fix + review improvements.

Soju06:
Thank you for catching this critical context blowup regression @balakumardev! 🎉

The session log analysis (2.3K vs 70K tokens/turn) was incredibly helpful in diagnosing the root cause. The Codex adversarial review identified some edge cases around multi-replica recovery and reconnect paths which were resolved together.

@all-contributors please add @balakumardev for code and test

PR #399, opened 2026-04-14
fix(auth): allow explicit unauthenticated proxy client CIDRs

ihazgithub:
## Summary

- add an explicit `CODEX_LB_PROXY_UNAUTHENTICATED_CLIENT_CIDRS` escape hatch for protected proxy routes when API key auth is disabled
- apply that allowlist only to the raw socket peer IP seen by the container
- keep the fix intentionally narrow and proxy-only
- leave global `is_local_request()` behavior unchanged
- keep dashboard/bootstrap auth protected
- add regression coverage for settings parsing, HTTP proxy access, websocket proxy access, and dashboard isolation

## Why

This was prompted by a real regression I hit after updating `codex-lb` from `1.11.0` to `1.12.0`.

My Codex desktop app workflow on macOS stopped working through Docker Desktop after that update: `/backend-api/codex/responses` started failing with `401 invalid_api_key`, while the same setup still worked on `1.11.0`.

That regression is why I investigated this path and contributed this fix.

## What This Changes

When API key auth is disabled, `codex-lb` still keeps the default fail-closed behavior for non-local proxy requests.

This change only adds a narrow opt-in allowlist for specific raw socket peer IPs via `CODEX_LB_PROXY_UNAUTHENTICATED_CLIENT_CIDRS`, so operators can restore local Docker Desktop-style workflows without broadly weakening security.

Forwarded headers are not used for this exception. The match is based only on the observed raw socket peer IP.

Dashboard/bootstrap auth remains protected and does not use this new allowlist.

## Validation

- added targeted regression tests for:
  - settings parsing and CIDR validation
  - HTTP proxy access with API key auth disabled
  - websocket proxy access with API key auth disabled
  - dashboard isolation, to confirm dashboard/bootstrap auth is still protected
- manually validated on Docker Desktop for Mac and with the Codex desktop app

One important detail from manual validation: on my machine, the actual socket peer IP seen by the container was `172.64.155.209`, so the working value for `CODEX_LB_PROXY_UNAUTHENTICATED_CLIENT_CIDRS` during validation was `172.64.155.209/32`, not `192.168.65.1/32`.

The exact CIDR is environment-specific. The key point is that the configured CIDR must match the observed raw socket peer IP that the container actually sees.

## Result

This restores the Codex desktop app workflow for the Docker Desktop case I observed, without changing the shared locality rules and without bypassing dashboard/bootstrap auth.

### DIFF
--- a/app/core/auth/dependencies.py
+++ b/app/core/auth/dependencies.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import hashlib
 import logging
+from ipaddress import ip_address, ip_network
 from typing import cast
 
 from fastapi import Request, Security
@@ -11,6 +12,7 @@ from starlette.requests import HTTPConnection
 from app.core.auth.api_key_cache import get_api_key_cache
 from app.core.auth.dashboard_mode import DashboardAuthMode, get_dashboard_request_auth
 from app.core.clients.usage import UsageFetchError, fetch_usage
+from app.core.config.settings import get_settings
 from app.core.config.settings_cache import get_settings_cache
 from app.core.exceptions import DashboardAuthError, ProxyAuthError, ProxyUpstreamError
 from app.core.request_locality import is_local_request
@@ -56,7 +58,8 @@ async def validate_proxy_api_key_authorization(
     settings = await get_settings_cache().get()
     if not settings.api_key_auth_enabled:
         if request is not None and not is_local_request(request):
-            raise ProxyAuthError("Proxy authentication must be configured before remote access is allowed")
+            if not _is_proxy_unauthenticated_socket_peer_allowed(request):
+                raise ProxyAuthError("Proxy authentication must be configured before remote access is allowed")
         return None
 
     token = _extract_bearer_token(authorization)
@@ -150,6 +153,20 @@ def get_dashboard_request_auth_mode() -> DashboardAuthMode:
     from app.core.config.settings import get_settings
 
     return get_settings().dashboard_auth_mode
+
+
+def _is_proxy_unauthenticated_socket_peer_allowed(request: HTTPConnection) -> bool:
+    socket_host = request.client.host if request.client else None
+    if socket_host is None:
+        return False
+
+    try:
+        socket_ip = ip_address(socket_host)
+    except ValueError:
+        return False
+
+    configured_cidrs = get_settings().proxy_unauthenticated_client_cidrs
+    return any(socket_ip in ip_network(cidr, strict=False) for cidr in configured_cidrs)
 
 
 # --- Codex usage caller identity auth ---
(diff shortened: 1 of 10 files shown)


ihazgithub:
## FAQ / reviewer appendix

### What problem does this PR address?
After updating `codex-lb` from `1.11.0` to `1.12.0`, my Codex desktop app workflow on macOS stopped working through Docker Desktop. The main symptom was that `/backend-api/codex/responses` started failing with `401 invalid_api_key`, while the same setup still worked on `1.11.0`.

### Why did I work on this?
This was not theoretical. It broke a real workflow I use, which is why I investigated it and contributed the fix.

### What was actually happening?
In newer versions, unauthenticated proxy traffic was handled more strictly. On Docker Desktop for Mac, host-to-container traffic that feels local from the Mac side can appear inside the container as a non-loopback socket peer IP. That meant the Codex desktop app path was being classified as remote rather than local.

### What does this fix do?
This adds a new `CODEX_LB_PROXY_UNAUTHENTICATED_CLIENT_CIDRS` setting. In plain English, it is a narrow opt-in allowlist for specific raw socket peer IPs or CIDRs for unauthenticated proxy traffic when proxy API-key auth is disabled.

### Why is the fix narrow?
It does not change global `is_local_request()` behavior. It does not weaken dashboard/bootstrap auth. It does not trust forwarded headers for this exception. It applies only to proxy auth, and the matching is based on the raw socket peer IP seen by the container.

### Does this bypass dashboard auth?
No. Dashboard/bootstrap auth remains protected, and that was explicitly tested during validation.

### Why is the CIDR example environment-specific?
During manual validation on my machine, the container saw the socket peer as `172.64.155.209`, so the working value was `172.64.155.209/32`. That exact value is environment-specific. The important part is that the configured CIDR must match the raw socket peer IP the container actually sees.

### How was this tested?
- regression confirmed between `1.11.0` and `1.12.0`
- direct proxy validation with `curl`
- dashboard isolation validation
- Codex desktop app validation on macOS via Docker Desktop
- websocket acceptance on `/backend-api/codex/responses`

### Short takeaway
This patch restores the broken Docker Desktop plus Codex desktop workflow I observed after `1.12.0`, while keeping the fix narrow and preserving dashboard/bootstrap protection.

Soju06:
Thank you for the detailed regression investigation @ihazgithub! 🎉

The Docker Desktop 1.11→1.12 regression was well-diagnosed, and the narrow `CODEX_LB_PROXY_UNAUTHENTICATED_CLIENT_CIDRS` escape hatch is a clean solution that keeps the fail-closed default while unblocking legitimate local Docker workflows.

The Codex adversarial review came back clean on the first pass — solid work.

@all-contributors please add @ihazgithub for code and test

ihazgithub:
@Soju06 Great! Happy to help! Cheers!

PR #401, opened 2026-04-14
fix(proxy): preserve previous_response_id on bridge recovery to prevent context blowup

Soju06:
## Summary

Closes #397 — rebased on main with Codex review fixes applied.

**Original author**: @balakumardev

### Problem
v1.12.0 bridge recovery paths return `400 previous_response_not_found` instead of `502 upstream_unavailable`, causing Codex CLI to drop `previous_response_id` and resend full conversation history (~70K tokens/turn instead of ~2-3K).

### Fix
- Closed-session path: attempt reconnect (`send_request=False`) before raising 502, preserving between-turn recovery
- Mid-request failure: always raise 502 (not 400) so client retries with `previous_response_id` intact
- `_retry_http_bridge_request_on_fresh_upstream`: reconnect but skip send when `previous_response_id` present (`effective_send` flag)
- `_retry_http_bridge_precreated_request`: return False for `previous_response_id` requests (caller raises 502)

### Codex Review Findings (fixed)
- **P1**: Preserved closed-session reconnect path (was bypassed by early 502)
- **P2**: Preserved local rebind signal via `effective_send` pattern

### Tests
- 7 new unit tests (`test_bridge_context_blowup.py`)
- 2 integration tests updated (400→502 expectation)
- All 76 bridge tests pass

### DIFF
--- a/app/modules/proxy/service.py
+++ b/app/modules/proxy/service.py
@@ -3246,25 +3246,14 @@ class ProxyService:
         queue_limit: int,
     ) -> None:
         if session.closed:
-            # When previous_response_id is present, reconnecting the upstream
-            # websocket creates a fresh server-side session that won't recognise
-            # the old response id.  Rather than silently falling back to a full
-            # conversation replay (which inflates per-turn context by ~20x), raise
-            # a 502 so the client can retry the original request on a new session
-            # with previous_response_id intact.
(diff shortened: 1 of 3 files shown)


Soju06:
Closing in favor of merging the original #397 directly.

PR #416, opened 2026-04-15
fix(proxy): retry active bridged previous_response_id continuations

Daltonganger:
## Summary
- retry active HTTP bridge continuations with `previous_response_id` once on a fresh upstream websocket instead of forcing a 300 second idle timeout
- keep `previous_response_not_found` fail-closed only for requests that never had a matching live bridged session before submission
- document the behavior change in OpenSpec and keep the regression coverage aligned with the faster upstream-unavailable failure path

## Testing
- `./.venv/bin/python -m pytest tests/integration/test_http_responses_bridge.py -k 'previous_response_id_recovery or send_failure_returns_upstream_unavailable or precreated_disconnect_returns_upstream_unavailable'`
- `./.venv/bin/python -m pytest tests/integration/test_http_responses_bridge.py -k 'opens_fresh_session_for_previous_response_id_recovery or retry_http_bridge_precreated_request'`


Daltonganger:
All fixed now! It works on my website again, and it does a retry! Thank you. Please check it all and please push it in 0.13.1! Thx

Soju06 (changes requested):
Codex review loop ran 9 rounds — core retry guard removal has a fundamental safety gap.

### What was fixed (owner-pushed)

- **Removed** `_http_bridge_should_prefer_local_previous_response_recovery` helper entirely — it bypassed owner-forward on the initial request path. Every approach (local session check, instance-id match, active-owner check, stale-owner check) broke a different multi-replica scenario. The reattach path (`_http_bridge_should_attempt_local_previous_response_recovery`, post-forward-failure) already handles local recovery correctly.
- **Added** turn-state anchor gate on both retry functions — continuation retries now require `_preferred_http_bridge_reconnect_turn_state(session) is not None` before resending, preventing retries on sessions with no continuity metadata.

### Remaining P1 — no upstream idempotency

The bridge has no request-level dedup mechanism. Both retry paths (`send_failure` + `precreated`) can resend a `response.create` that the upstream already accepted:

```
1. LB → Upstream: send_text(response.create, previous_response_id=X)
2. Upstream accepts, starts processing
3. WebSocket drops before response.created arrives
4. LB retries on fresh socket: send_text(same request)
5. Two child responses created → continuity anchor corrupted
```

The turn-state anchor check reduces the surface (no retry without reconnect metadata) but doesn't eliminate the race — a session with valid turn-state can still hit the ambiguous-delivery window.

### Recommendation

Either:
1. **Add a request-level idempotency key** to `response.create` payloads so upstream can dedup replayed continuations, or
2. **Keep the `previous_response_id` guards** for `send_request=True` and `precreated` paths (only allow `send_request=False` reconnect-without-resend, which was already permitted before this PR)

Option 2 is minimal — the PR's value narrows to the initial routing + OpenSpec artifacts, but the retry behavior stays safe.

Daltonganger:
Addressed in `4c72cde` (`fix(proxy): fail fast on unsafe continuation retries`).

I reverted the unsafe part of the retry behavior for `previous_response_id` continuations:

- `send_failure` no longer replays a continuation request on a fresh upstream websocket
- `precreated` retry no longer replays a continuation request either
- reconnect-without-resend is still allowed for the closed-session recovery path
- failures still surface quickly as retriable `502`-style upstream availability errors instead of hanging or returning `previous_response_not_found`

The reason is the same safety concern from this review: after an ambiguous websocket drop, we cannot prove whether upstream already accepted the original `response.create`. Replaying the same continuation could create duplicate child responses and corrupt continuity.

This keeps the anti-hang behavior that Opencode needs, while restoring the v1.12-style safety invariant for `previous_response_id` continuations.

I also updated the regression coverage to assert that these continuation paths fail fast without reconnect+resend.

PR #417, opened 2026-04-15
fix(auth): accept API keys on /api/codex/usage

Daltonganger:
## Summary
- allow `/api/codex/usage` to accept LB API-key callers when they do not send `chatgpt-account-id`
- return a `RateLimitStatusPayload` compatibility response for API-key callers using existing self-usage and credit-limit data
- preserve the existing ChatGPT identity validation path for callers that do send `chatgpt-account-id`

## Testing
- `./.venv/bin/python -m pytest tests/integration/test_codex_usage_api.py`
- `./.venv/bin/python -m pytest tests/integration/test_v1_usage.py -k 'v1_usage_returns_aggregate_credit_limits_when_upstream_usage_exists or v1_usage_overrides_aggregate_credit_windows_with_api_key_credit_limits'`

### DIFF
--- a/app/core/auth/dependencies.py
+++ b/app/core/auth/dependencies.py
@@ -172,7 +172,7 @@ def _is_proxy_unauthenticated_socket_peer_allowed(request: HTTPConnection) -> bo
 # --- Codex usage caller identity auth ---
 
 
-async def validate_codex_usage_identity(request: Request) -> None:
+async def validate_codex_usage_identity(request: Request) -> ApiKeyData | None:
     token = _extract_bearer_token(request.headers.get("Authorization"))
     if not token:
         raise ProxyAuthError("Missing ChatGPT token in Authorization header")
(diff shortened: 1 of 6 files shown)


Soju06 (approved):
LGTM — approved after codex review loop (9 rounds, 0 findings on final pass).

### Owner-pushed commits (7)

| Commit | Fix |
|---|---|
| `dfa7873` | **P1**: remove aggregate workspace limits from API-key codex usage path — use key's own `raw_limits` directly to avoid reporting workspace-wide `current_value` instead of key-specific counters. Regression test added (seeds aggregate data, asserts key-specific %). |
| `b76bdd5` | **P2**: restrict `_select_codex_usage_limit` to `credits` type only — prevents non-credit limits (token/cost) leaking into codex quota windows. Floor `used_percent` instead of rounding to avoid premature 100% / `limit_reached` near boundary. |
| `d6c67c7` | **P2**: add `daily`→primary, `weekly`→secondary window fallback for keys without 5h/7d credit limits. |
| `04decd5` | Include `monthly` in secondary window fallback chain. |
| `9559c9d` → `448d34c` | Tried aggregate fallback for no-limit keys, reverted (leaks workspace quotas to individual callers). `has_credits` now reflects actual `remaining_value > 0`. |
| `23b53e0` | Restore secondary (longer) window preference for `credits.balance` — primary window is the blocking limit shown in `rate_limit`, secondary is the total balance. |

All CI green. Codex review clean.

PR #424, opened 2026-04-16
fix(chat-completions): normalize provider thinking aliases

### LINKED ISSUE #205: Support `thinking` / `enable_thinking` params in Chat Completions API for multi-provider compatibility
Hi all, thanks for the great work, however i encountered an issue when migrating existing project to codex-lb proxy 

## Problem

The Chat Completions endpoint uses `extra="allow"` on `ChatCompletionsRequest`, which transparently passes unknown parameters through to upstream ChatGPT. 

When clients send `thinking` (Claude format) or `enable_thinking` (Qwen/DeepSeek format), these are forwarded as-is to ChatGPT's backend-api, which rejects them with:

```
proxy_error_response request_id=... method=POST path=/v1/chat/completions status=502
code=None message=Unsupported parameter: thinking
```

Many projects are built with SDKs targeting one of these providers. When migrating to use codex-lb as a proxy to ChatGPT, users must rewrite their thinking/reasoning parameter handling, which creates friction for adoption.

For example, a project using the OpenAI Python SDK with Qwen:

```python
client = OpenAI(base_url="https://dashscope.aliyuncs.com/compatible-mode/v1", api_key="...")

client.chat.completions.create(
    model="qwen-plus",
    messages=[{"role": "user", "content": "hello"}],
    extra_body={"enable_thinking": True},
    stream=True,
)
```

Switching `base_url` to codex-lb causes a 502 because `enable_thinking` is passed through to ChatGPT unchanged.

### Comparison with ChatMock

[ChatMock](https://github.com/RayBytes/ChatMock) handles this correctly because `start_upstream_request()` uses a **whitelist approach** — it manually constructs the upstream payload with only known fields. Unknown client parameters like `thinking` are naturally dropped.

codex-lb uses a **passthrough approach** (`extra="allow"`) for forward compatibility with new OpenAI API features, but this causes incompatible third-party parameters to leak through.

## Questions 

Different providers (Claude, Qwen, DeepSeek) all use different parameter names for the same concept. Many users are migrating existing projects that use these SDKs. Are there any plans to normalize these third-party thinking parameters (`thinking`, `enable_thinking`) to ChatGPT's `reasoning` format, so that switching `base_url` to codex-lb "just works" without client code changes?

Thanksss

stemirkhan:
## Summary
- normalize provider-specific `thinking` / `enable_thinking` aliases into the canonical `reasoning` payload
- apply the normalization in Chat Completions request mapping and in the shared OpenAI-compatible Responses sanitation path
- add unit + integration coverage and an OpenSpec change for the compatibility behavior

## Why
Fixes #205.

Some OpenAI-compatible clients reuse provider-specific reasoning controls when switching their `base_url` to `codex-lb`. Those aliases were leaking through to the upstream ChatGPT backend unchanged, which caused avoidable upstream validation failures instead of a smooth compatibility path.

This keeps the fix intentionally small:
- `enable_thinking: true` maps to `reasoning.effort = "medium"`
- Anthropic-style `thinking` payloads are normalized to the same canonical shape
- explicit `reasoning` still wins over provider aliases

## Testing
- `/home/temirkhan/codex-lb-my/codex-lb-upstream/.venv/bin/python -m pytest tests/unit/test_chat_request_mapping.py tests/unit/test_openai_requests.py tests/integration/test_openai_compat_features.py -q`
- `openspec validate --specs`

## Notes
- scoped to OpenAI-compatible request compatibility only
- no API surface expansion and no dashboard changes

### DIFF
--- a/app/core/openai/chat_requests.py
+++ b/app/core/openai/chat_requests.py
@@ -11,6 +11,7 @@ from app.core.openai.requests import (
     ResponsesRequest,
     ResponsesTextControls,
     ResponsesTextFormat,
+    normalize_reasoning_aliases,
     normalize_tool_type,
     validate_tool_types,
 )
@@ -133,6 +134,7 @@ class ChatCompletionsRequest(BaseModel):
         reasoning_effort = data.pop("reasoning_effort", None)
         if reasoning_effort is not None and "reasoning" not in data:
             data["reasoning"] = {"effort": reasoning_effort}
+        normalize_reasoning_aliases(data)
         if response_format is not None:
             _apply_response_format(data, response_format)
         if isinstance(stream_options, Mapping):
--- a/app/core/openai/requests.py
+++ b/app/core/openai/requests.py
@@ -522,18 +522,11 @@ def _sanitize_interleaved_reasoning_input(payload: MutableJsonObject) -> None:
     payload["input"] = _sanitize_input_items(input_items)
 
 
-def _normalize_openai_compatible_aliases(payload: MutableJsonObject) -> None:
+def normalize_reasoning_aliases(payload: MutableJsonObject) -> None:
     reasoning_effort = payload.pop("reasoningEffort", None)
     reasoning_summary = payload.pop("reasoningSummary", None)
-    text_verbosity = payload.pop("textVerbosity", None)
-    top_level_verbosity = payload.pop("verbosity", None)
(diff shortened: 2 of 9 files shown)


Soju06:
Re-running CI and triggering codex review on the latest commit. Will follow up after both come back.

@codex review

Soju06:
@all-contributors please add @stemirkhan for bug

Issue #210, opened 2026-03-15
Persist previous_response_id chains, require neon postgresql runtime, and normalize API key expirations


xirothedev:
## Summary

This branch bundles three compatibility/runtime changes for `codex-lb`:

- persist `previous_response_id` response chains so `/v1/responses` can continue prior conversations after restart across HTTP and WebSocket transports
- require Neon/PostgreSQL for runtime and migrations, with a dedicated migration DSN and no SQLite runtime fallback on the main path
- normalize timezone-aware API key expiration datetimes before persisting them to PostgreSQL

### 1. `previous_response_id` is not durable today

`codex-lb` currently rejects or loses `previous_response_id` continuity because the proxy does not persist enough local response state to rebuild prior turns after restart. That leaves `/v1/responses` incompatible with clients that continue conversations by referencing the previous response id instead of resending the full transcript.

### 2. Runtime database defaults do not match the intended deployment contract

The repo still carries SQLite-oriented runtime assumptions in config, startup, docs, and tests even though the intended deployment model is Neon/PostgreSQL. That mismatch makes runtime behavior and setup guidance drift from the production path.

### 3. API key expiration writes fail with timezone-aware input

The dashboard submits ISO 8601 `expiresAt` values with timezone information, but PostgreSQL writes fail because the backend persists them into a `timestamp without time zone` column without normalizing them first.

## Proposed Changes

- store durable response snapshots keyed by `response_id`
- resolve `previous_response_id` into replayable input history before forwarding upstream
- prefer the account that served the referenced response when it is still eligible, and fall back cleanly otherwise
- support the same continuity behavior for collected HTTP responses, streaming HTTP responses, and WebSocket Responses traffic
- require `CODEX_LB_DATABASE_URL` and `CODEX_LB_DATABASE_MIGRATION_URL` for the Neon/PostgreSQL runtime path
- remove SQLite-specific runtime fallback behavior from startup expectations and local setup docs
- normalize timezone-aware API key expiration datetimes to UTC-naive before persistence on create and update flows

## Scope

OpenSpec artifacts already included on this branch:

- `support-previous-response-id-persistence`
- `neon-postgresql-runtime`
- `fix-api-key-expiration-timezone`

Primary implementation areas:

- proxy request/response flow and account selection
- database session and migration configuration
- API key service persistence normalization
- integration and unit regression coverage

## Validation Notes

- `.venv/bin/python -m compileall app tests` passes in the current environment
[...]

PR #421, opened 2026-04-16
fix(proxy): prefer budget-safe routing and support image-generation compatibility ("code":"invalid_request_error","param":"tools")

mhughdo:
## Summary
- reallocate durable backend `codex_session` affinity once the pinned account crosses the configured budget threshold and a healthier candidate exists
- prefer budget-safe Responses routing candidates over already-pressured candidates whenever any safe candidate exists
- allow built-in Responses tools on full `/backend-api/codex/responses` and `/v1/responses` routes while stripping tool-only fields from compact requests
- raise the default upstream Responses event/message limit to 16 MiB and make auto transport prefer upstream HTTP for `image_generation` requests

## Why
Repeated `usage_limit_reached` failures were still occurring even while other accounts had headroom.

Live DB inspection showed the failures were being routed to accounts whose local primary usage at request time was already `97-100%`. The proxy was still allowing those near-exhausted `ACTIVE` accounts to win fresh selection, and durable backend `codex_session` affinity only reallocated under narrower conditions.

Separately, recent Codex Desktop builds now send newer built-in Responses tools such as `image_generation`. That exposed proxy-side compatibility issues including this concrete failure on full Responses requests:

```json
{"type":"error","status":400,"error":{"message":"Invalid request payload","type":"invalid_request_error","code":"invalid_request_error","param":"tools"}}
```

The underlying issues were:
- full Responses and compact requests did not sanitize tool payloads correctly for the different upstream contracts
- websocket-preferred models such as `gpt-5.4` could still carry large inline `image_generation` outputs over upstream websocket, which made the proxy hit a local frame ceiling and fail before `response.completed`

## Verification
- `uv run pytest tests/integration/test_proxy_sticky_sessions.py -q -k "test_proxy_stream_sticky_threads_reallocate_by_prompt_cache_key or test_proxy_codex_session_id_pins_responses_and_compact_without_sticky_threads or test_proxy_codex_session_id_reallocates_when_pinned_budget_exhausted or test_proxy_codex_session_id_compact_first_pins_followup_stream_without_sticky_threads or test_proxy_sticky_switches_when_pinned_rate_limited"`
- `uv run pytest tests/unit/test_proxy_load_balancer_refresh.py -q -k "prefers_budget_safe_account_when_any_exist"`
[...]

PR #448, opened 2026-04-21
fix(proxy): prevent context blowup by trimming input on client-supplied previous_response_id

balakumardev:
## Problem

Long Codex CLI sessions experience monotonic input token growth — from ~49K on the first turn to 800K+ after several turns — because the proxy fails to trim redundant input items when the **client** (not the proxy) supplies `previous_response_id`.

Related issues: #395, #423, #210.

## Root cause

The existing trimming logic at line 648 only checked `proxy_injected_previous_response_id`, which is `True` only when the proxy itself injects a `previous_response_id` during fresh reattach (e.g., after restart). However, Codex CLI sends its own `previous_response_id` on every turn after the first. When the client provides it, the proxy's flag stayed `False`, so trimming never fired — and the full input history was resent every single turn.

## Measured impact (before fix)

Data from Codex CLI's local logs:

| Session | Token growth per turn | Pattern |
|---------|----------------------|---------|
| Thread 1 (morning) | 249K → 250K → 255K → 449K → 651K → 854K | Unbounded growth |
| Thread 2 (afternoon) | 94K → 121K → 147K → 174K → 201K → 228K → 256K → 283K → 311K → 339K | ~27K per turn |
| Thread 3 (after fix) | 49K → 49K → 49K → 49K → 49K → 49K | Flat (stable) |

The ~49K baseline is the Codex CLI system prompt (instructions field), which must be resent every turn and is prompt-cached by OpenAI.

### 1. Coerce store=true to false instead of rejecting

The validators no longer raise a ValueError when a client sends store=true. Instead, they silently coerce to false. This unblocks Codex CLI which sends store: true by default.

### 2. Trim input when either proxy or client supplies previous_response_id

The trimming condition now fires when the client supplies previous_response_id, not just when the proxy injects it during reattach. The upstream conversation state is valid regardless of who provided the ID.

**Before:** Only trimmed on proxy reattach (proxy_injected_previous_response_id=True)
**After:** Trims whenever any previous_response_id is present (proxy-injected OR client-supplied)

### 3. Preserve full input count after trimming

The bridge tracks the full logical input length after a trimmed request completes, preventing the next turn from forgetting older items.

## Impact

| Metric | Before | After |
|--------|--------|-------|
| Codex CLI store: true | 400 error | Accepted (coerced to false) |
| Input tokens (turn N+1) | ~250K+ and growing | ~49K flat (system prompt only, prompt-cached) |
| Token growth pattern | Monotonic (unbounded) | Flat (stable) |
| cached_input_tokens on follow-ups | Low | ~49K (system prompt cached by OpenAI) |

## Validation

- Verified trimming fires on follow-up turns via debug logging
- Confirmed total_usage_tokens stays flat at ~49K across 12+ consecutive turns
- Previous sessions showed growth from 249K to 854K (same thread, same day)
- uvx ruff check . — clean
- uvx ruff format --check . — clean
- pytest on touched test files — 150 passed

## Related issues

- Fixes context blowup on long sessions (#423, #210)
- Fixes the Codex CLI 400 rejection (#395)


Soju06:
@codex review

Soju06 (changes requested):
## Review — REQUEST CHANGES

Great root-cause analysis on `store=false` — the fix direction (Codex-bridge-scoped `store=true` coercion + durable `previous_response_id` reattach) is solid and I want to get this merged. Two items to resolve first.

---

### 🚨 Blocking: silent correctness regression on edited history prefixes

@chatgpt-codex-connector caught this at `app/modules/proxy/service.py:661` and I reproduced the scenario:

```
last_completed_input_count = 5                            # prior turn ended here
client sends input = [i1, i2, i2_edited, i3, i4, i5, i6]  # 7 items, mid-prefix edit
trim logic: input[5:] = [i5, i6]                          # prefix silently dropped
upstream: uses stored previous_response_id (still has original i3)
result: edit is silently lost, OpenAI answers as if the original prefix still held
```

The trim currently assumes the new `input` is a strict append to whatever the session saw last. That holds for Codex CLI's normal loop, but it breaks:

- `/compact` / `/edit` flows that rewrite an earlier item
- Clients that re-send a normalised full history every turn with any in-place correction
- Any future client behavior that does not preserve prefix equality

Because there is no equality check, the proxy will happily drop an edited user message and the user never sees why their correction was ignored. That is the worst failure mode here (no error, wrong answer).

#### Suggested fix

Gate the slice on a prefix-equality check against what the session has already forwarded. Rough shape:

```python
if (
    proxy_injected_previous_response_id
    and session.last_completed_input_count > 0
    and isinstance(effective_payload.input, list)
    and len(effective_payload.input) > session.last_completed_input_count
    and _input_prefix_matches_stored(
        effective_payload.input[: session.last_completed_input_count],
        session.last_completed_input_items,  # you'll need to keep this too
    )
):
    ...
```

If the prefix diverges, fall back to sending the full `input` (slower, but correct). Add a warning log when the mismatch triggers so we can observe how often it happens in practice.

An alternative is to store and compare a lightweight fingerprint (e.g. hash) of each completed item rather than the items themselves, to avoid blowing up session memory.

---

### ⚠️ Non-blocking: PR description vs actual code

The current description reads as if public validators are coerced to `store=true`, which is not what the diff does:

- The three `_ensure_store_false` validators still always return `False`. They only changed from `raise ValueError` to silent coerce.
- The actual `store=true` upstream override lives only in `_prepare_response_bridge_request_state` and is gated on `force_store=codex_session_affinity`.

That is the correct design (narrow scope, no public API surface change), but please update the PR body so reviewers and release notes reflect what actually ships. Current wording will also confuse anyone grepping for `store=true` in the code.

---

### Not blocking, just noting

- 18/18 CI green ✅
- Text-clean rebase on top of #421 (e632d94) and #425 (dbf4775); no semantic collision with the `allow_builtin_tools` / `_strip_compact_unsupported_fields` changes from #421
- New `_WebSocketRequestState.input_item_count` and `_HTTPBridgeSession.last_completed_input_count` fields look well scoped; logging at INFO level for `store_context_input_trimmed` is useful

### TL;DR

Fix the prefix-equality guard + refresh PR description and this is merge-ready. Happy to re-review on update.

Soju06:
Applied the prefix-fingerprint guard directly (pushed as `29e8b74` to your branch — `maintainerCanModify` was on, so I went ahead rather than blocking on a round-trip).

## What changed

- New helper `_fingerprint_input_items()` (SHA-256 of canonical JSON, sorted keys) in `service.py`
- `_WebSocketRequestState.input_full_fingerprint` captures the fingerprint of each request's full input list
- On `response.completed`, the fingerprint is promoted onto the session alongside `last_completed_input_count`
- Trim decision now checks `sha256(incoming_input[:stored_count]) == session.last_completed_input_prefix_fingerprint` before slicing; mismatch falls back to sending the full input + emits `store_context_input_trim_skipped_prefix_mismatch` at WARNING

## New test

`test_stream_via_http_bridge_skips_trim_when_stored_prefix_was_edited` — constructs a session with a fingerprint of the unedited prefix, then sends a request where item 2 is edited. Asserts that `_prepare_http_bridge_request` is only called **once** with the full 3-item input, never a second time with a 1-item trimmed slice.

## Ran locally

- `uvx ruff check .` — clean
- `uvx ruff format --check .` — clean
- `uv run pytest tests/unit/test_proxy_http_bridge.py tests/unit/test_openai_requests.py tests/integration/test_openai_compat_features.py tests/integration/test_proxy_websocket_responses.py -q` — **173 passed**

Please take a look at the fingerprint approach — if you prefer per-item fingerprints (trades a constant-memory prefix hash for granular edit localization), I'm happy to switch, but the full-prefix hash felt right for the "append vs edited" decision we actually care about here.

Also noted in the earlier review: please refresh the PR description before we merge — the current wording reads as if `store=true` is now public, but the final diff keeps `_ensure_store_false` coercing to `False` and scopes the real upstream override narrowly. Reviewers and release notes will want the accurate story.

Soju06 (approved):
Re-reviewing after pushing `29e8b74` that addresses the prefix-equality guard I requested. 

- Edited-prefix silent regression: ✅ fixed with fingerprint check + fallback + warning log
- Test coverage: ✅ new negative test + existing trim test adapted
- Lint / format / tests all green locally (173 passed)

Still non-blocking: PR description needs a refresh to match the final design (public validators still coerce to False; upstream `store=true` override is gone after your `020dd6d`; real fix is the Codex-bridge trim path on client-supplied `previous_response_id`).

Approving on the assumption you update the body before merge. This is the #395 / #423 blocker fix 30+ users are waiting on.

Soju06:
@codex review

Soju06:
@chatgpt-codex-connector nailed another one — the fingerprint I was recording on the trimmed-suffix `request_state` would have been promoted onto the session on `response.completed` and broken trimming on every subsequent turn.

Fixed in `3eb1f27`:

```python
# after restoring input_item_count, force the full-input fingerprint back
request_state.input_full_fingerprint = _fingerprint_input_items(incoming_input_list)
```

Also added an assertion to the existing trim preservation test that the `_WebSocketRequestState.input_full_fingerprint` equals the hash of the original 3-item input (not the 1-item trimmed suffix), so any future regression here fails loudly.

@codex review — please take another look.

Ran locally:
- `uvx ruff check .` — clean
- `uvx ruff format --check .` — clean
- `uvx ty check app/modules/proxy/service.py` — clean
- `uv run pytest tests/unit/test_proxy_http_bridge.py -k 'trim or prefix or preserves'` — 4 passed

PR #456, opened 2026-04-21
fix(proxy): inject session-level previous_response_id to enable input trimming for all clients

balakumardev:
## Problem

Codex CLI and other clients that don't send `previous_response_id` in their request payload bypass the input trimming logic entirely, causing input tokens to grow monotonically within a bridge session.

**Observed:** 31K → 201K input tokens across 15 requests in a single session, with zero trimming despite the upstream caching 99% of the input (same WebSocket connection).

The existing trimming (from #448) only fires when `has_previous_response_id` is true — i.e., when the client sends `previous_response_id` or the durable lookup injects one. But:
1. **Codex CLI never sends `previous_response_id`** in its request payload
2. **Durable lookup injection** only fires when a canonical session key exists from a prior lookup, which doesn't always happen (e.g., session key transitions from `session_header` to `turn_state_header` after idle timeout)

This means the common case of a Codex CLI session staying on the same bridge WebSocket never gets trimming.

## Fix

After a `response.completed` event, record the response ID on the bridge session (`last_completed_response_id`). On the next request through the same session, if no `previous_response_id` is available from the client or durable lookup, inject the session's last completed response ID.

### Why this is reliable

- **Same WebSocket connection** → the upstream conversation state is guaranteed to contain this response (it was produced on the very connection we're about to use)
- **WebSocket reconnects** → new session is created → `last_completed_response_id = None` → no stale injection
- **Upstream lost the response** → existing error recovery at the `_HTTPBridgeOwnerForward` failure path handles the "unknown previous_response_id" error

### Changes

1. `_HTTPBridgeSession.last_completed_response_id` — new field to track the most recent successful response ID on this session
2. `_process_http_bridge_upstream_text` — on `response.completed`, store the response ID alongside the existing `last_completed_input_count` and fingerprint
3. `_stream_via_http_bridge` — before the trimming block, inject `session.last_completed_response_id` as `previous_response_id` when neither the client nor durable lookup provided one

### Expected behavior

| Request # | Before (no fix) | After (with fix) |
|-----------|----------------|-----------------|
| 1 | 31K tokens (full) | 31K tokens (full) |
| 2 | 71K tokens (full) | ~1K tokens (trimmed) |
| 3 | 99K tokens (full) | ~1K tokens (trimmed) |
| ... | grows monotonically | stays flat |
| 15 | 201K tokens | ~1K tokens |

### DIFF
--- a/app/modules/proxy/service.py
+++ b/app/modules/proxy/service.py
@@ -579,10 +579,15 @@ class ProxyService:
         if proxy_injected_previous_response_id:
             request_state.proxy_injected_previous_response_id = True
             request_state.fresh_upstream_request_text = fresh_upstream_request_text or text_data
-            # Durable-anchor injection only fires for trim-safe full-resend
-            # payloads, so the unanchored request text is a valid fresh-turn
-            # replay target.
-            request_state.fresh_upstream_request_is_retry_safe = True
+            # Durable-anchor injection actually runs when the incoming
+            # payload is *not* a full resend (see the
+            # ``not _http_bridge_payload_looks_like_full_resend(payload)``
+            # guard above), so the captured unanchored text is typically
+            # just a short follow-up. Replaying it as a fresh turn would
+            # drop the conversational context the anchor was pointing at.
+            # Only the trim branch below (which verifies the stored prefix
+            # fingerprint) is allowed to flip this flag to ``True``.
+            request_state.fresh_upstream_request_is_retry_safe = False
         session_or_forward = await self._get_or_create_http_bridge_session(
             bridge_session_key,
             headers=dict(headers),
--- a/app/modules/proxy/service.py
+++ b/app/modules/proxy/service.py
@@ -734,16 +734,15 @@ class ProxyService:
         session = session_or_forward
         # --- Session-level previous_response_id injection ---
         # If the client didn't send previous_response_id and the durable
-        # lookup didn't inject one, but this bridge session has already
-        # completed a request on the *same* WebSocket connection, inject the
-        # session's last completed response ID.  This is reliable because the
-        # upstream conversation state is guaranteed to contain this response
-        # — it was produced on the very connection we are about to use.
-        # Without this, clients like Codex CLI that never send
-        # previous_response_id would bypass the trimming logic below and
-        # cause input tokens to grow monotonically for the whole session.
+        # lookup didn't inject one, but this bridge session is carrying
+        # Codex-style conversational continuity and has already completed a
+        # request on this logical conversation, inject the session's last
+        # completed response ID. Soft affinity reuse (for example prompt
+        # cache / sticky-thread sharing) must stay self-contained, so it
+        # must never inherit a previous_response_id from another request.
         if (
-            not proxy_injected_previous_response_id
+            session.codex_session
+            and not proxy_injected_previous_response_id
             and effective_payload.previous_response_id is None
             and session.last_completed_response_id is not None
         ):
--- a/app/modules/proxy/service.py
+++ b/app/modules/proxy/service.py
@@ -579,6 +579,10 @@ class ProxyService:
         if proxy_injected_previous_response_id:
             request_state.proxy_injected_previous_response_id = True
             request_state.fresh_upstream_request_text = fresh_upstream_request_text or text_data
+            # Durable-anchor injection only fires for trim-safe full-resend
+            # payloads, so the unanchored request text is a valid fresh-turn
+            # replay target.
+            request_state.fresh_upstream_request_is_retry_safe = True
         session_or_forward = await self._get_or_create_http_bridge_session(
             bridge_session_key,
             headers=dict(headers),
@@ -801,6 +805,14 @@ class ProxyService:
             request_state.preferred_account_id = durable_lookup.account_id if durable_lookup is not None else None
             request_state.proxy_injected_previous_response_id = True
             request_state.fresh_upstream_request_text = fresh_upstream_request_text
+            # Session-level anchor injection may be attached to a payload
+            # that relied on the anchor for context (for example a
+            # single-item follow-up turn whose prior history is only
+            # represented by ``previous_response_id``). Replaying without
+            # the anchor would silently turn it into a fresh turn and drop
+            # conversational context, so opt this path out of fresh-upstream
+            # fresh-turn replay.
+            request_state.fresh_upstream_request_is_retry_safe = False
             logger.info(
                 "session_anchor_injected request_id=%s response_id=%s",
                 request_id,
@@ -848,6 +860,13 @@ class ProxyService:
                 if proxy_injected_previous_response_id:
                     request_state.proxy_injected_previous_response_id = True
                     request_state.fresh_upstream_request_text = fresh_upstream_request_text
+                    # The trim branch only fires when the untrimmed payload
+                    # is a true full resend whose prefix exactly matches the
+                    # already-stored context, so the unanchored request text
+                    # is a safe fresh-turn replay target regardless of
+                    # whether the anchor came from the durable or
+                    # session-level injection path.
+                    request_state.fresh_upstream_request_is_retry_safe = True
                 logger.info(
                     "store_context_input_trimmed request_id=%s original_items=%s trimmed_to=%s previous_response_id=%s",
                     request_id,
@@ -4738,7 +4757,16 @@ class ProxyService:
             # upstream already accepted the continuation. Re-sending the same
             # previous_response_id request can fork continuity with duplicate
             # child responses, so only reconnect-without-resend is allowed.
-            if not request_state.proxy_injected_previous_response_id or not request_state.fresh_upstream_request_text:
+            # The single exception is proxy-injected anchors on trim-safe
+            # full-resend payloads: dropping the anchor and replaying the
+            # original unanchored request is equivalent to the client's own
+            # retry. Session-level injections do not opt in because their
+            # payload may depend on the anchor for context preservation.
+            if (
+                not request_state.proxy_injected_previous_response_id
+                or not request_state.fresh_upstream_request_text
+                or not request_state.fresh_upstream_request_is_retry_safe
+            ):
                 return False
             retry_text_data = request_state.fresh_upstream_request_text
         if request_state.replay_count >= 1:
@@ -7749,6 +7777,16 @@ class _WebSocketRequestState:
     session_id: str | None = None
     proxy_injected_previous_response_id: bool = False
     fresh_upstream_request_text: str | None = None
+    # True only when ``fresh_upstream_request_text`` contains a *safe* pre-
+    # injection form of this request that can be replayed as a fresh turn.
+    # Durable-anchor injection captures the original unanchored full-resend
+    # payload, so dropping the anchor and replaying is equivalent to the
+    # client's own retry. Session-level anchor injection does **not** set
+    # this: the original payload may have omitted history the conversation
+    # depended on (for example a single-item follow-up whose context came
+    # entirely from the injected anchor), and dropping the anchor there
+    # would silently turn a continuation into a context-free fresh turn.
+    fresh_upstream_request_is_retry_safe: bool = False
     request_stage: str = "first_turn"
     preferred_account_id: str | None = None
     error_code_override: str | None = None
--- a/app/modules/proxy/service.py
+++ b/app/modules/proxy/service.py
@@ -489,6 +489,7 @@ class ProxyService:
             durable_lookup = None
         effective_payload = payload
         proxy_injected_previous_response_id = False
+        fresh_upstream_request_text: str | None = None
         if durable_lookup is not None:
             bridge_session_key = _HTTPBridgeSessionKey(
                 durable_lookup.canonical_kind,
@@ -513,6 +514,14 @@ class ProxyService:
                     update={"previous_response_id": durable_lookup.latest_response_id}
                 )
                 proxy_injected_previous_response_id = True
+                _fresh_request_state, fresh_upstream_request_text = self._prepare_http_bridge_request(
+                    payload,
+                    headers,
+                    api_key=api_key,
+                    api_key_reservation=api_key_reservation,
+                    request_id=request_id,
+                )
+                del _fresh_request_state
                 _log_http_bridge_event(
                     "fresh_reattach_anchor_injected",
                     bridge_session_key,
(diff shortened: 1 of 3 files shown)


app/modules/proxy/service.py
```diff
@@ -653,6 +653,44 @@ async def _stream_via_http_bridge(
...
+        # lookup didn't inject one, but this bridge session has already
+        # completed a request on the *same* WebSocket connection, inject the
+        # session's last completed response ID.  This is reliable because the
+        # upstream conversation state is guaranteed to contain this response
+        # — it was produced on the very connection we are about to use.
+        # Without this, clients like Codex CLI that never send
+        # previous_response_id would bypass the trimming logic below and
+        # cause input tokens to grow monotonically for the whole session.
+        if (
+            not proxy_injected_previous_response_id
```
balakumardev: This now fires for every reused `_HTTPBridgeSession`, not just hard Codex/session-header continuity. Soft keys like `prompt_cache` and `sticky_thread` are also reusable (`_make_http_bridge_session_key` marks only `turn_state_header`/`session_header` as hard), so the next unrelated request on the same shared bridge socket will inherit `previous_response_id` from the prior caller and get treated as a continuation. That is a correctness/data-isolation regression for cache-affinity traffic. Can we gate this to hard conversational sessions only (for example `session.codex_session` / hard keys) and add a regression test for the soft-key case?

balakumardev:
Left one blocking inline comment on the new session-level  injection. In its current form it also applies to soft bridge keys ( / ), which can accidentally chain unrelated requests together instead of keeping them self-contained.

balakumardev:
Left one blocking inline comment on the new session-level `previous_response_id` injection. In its current form it also applies to soft bridge keys (`prompt_cache` / `sticky_thread`), which can accidentally chain unrelated requests together instead of keeping them self-contained.

Soju06:
@codex review

Soju06 (changes requested):
Concurring with @balakumardev's own self-review — this reproduces on soft keys and is a blocking correctness issue before merge.

## The regression path, concretely

Let me re-verify from the code. `_HARD_HTTP_BRIDGE_AFFINITY_KINDS = frozenset({"turn_state_header", "session_header"})` — so the bridge session key strength for `prompt_cache` or `sticky_thread` is `soft`. A soft bridge session is a **shared pipe for unrelated requests that happen to share a cache key or UI thread**, not a conversational continuity channel.

With the current diff, the new injection fires on **any** reused `_HTTPBridgeSession` that has completed at least one response:

```
T0: request A (prompt_cache=X, no previous_response_id) completes → session.last_completed_response_id = respA
T1: unrelated request B (prompt_cache=X, no previous_response_id)
    → proxy injects previous_response_id=respA on B's payload
    → B is now treated as a continuation of A
```

That silently merges the two requests' contexts from upstream's perspective, and the trimming path will then happily drop B's input prefix if the fingerprint happens to match. Worst case: cross-user or cross-request leakage on a shared cache-affinity pool.

## Fix suggestion

Gate the session-level injection to **hard** keys only — which is precisely what "conversational continuity" is supposed to mean in this codebase. Two options:

1. `session.codex_session` boolean that's already on `_HTTPBridgeSession` (set when Codex session headers are present). Narrowest scope.
2. `session.key.strength == "hard"` which generalises to future hard kinds as well.

I'd prefer **(2)** because it keeps the invariant ("only chain on hard conversational sessions") independent of Codex-specific flags.

## Regression test

Please add one covering the soft-key case — something like `test_stream_via_http_bridge_does_not_inject_session_anchor_for_soft_bridge`. Set up a session with `affinity_kind=prompt_cache`, pre-populate `last_completed_response_id = "resp_other"`, then send a request with no `previous_response_id` and assert that `prepared_payload.previous_response_id` stays `None`.

## Not blocking, but nice-to-have

- Log INFO only on hard-key injection; soft-key skips shouldn't spam the session.
- The new field `last_completed_response_id` on `_HTTPBridgeSession` is written inside `_process_http_bridge_upstream_text` — worth a short comment there too explaining that it's **only meaningful for hard keys** and is a no-op on soft paths.

Also triggered a `@codex review` for cross-check. Happy to re-approve once the gate + regression test land.

app/modules/proxy/service.py
balakumardev: Fixed on the PR branch.

Changes:
- gated session-level anchor injection to true Codex continuity sessions (`session.codex_session`) so soft prompt-cache / sticky-thread reuse stays self-contained
- added a regression test covering the soft-reuse case

Verification:
- `uvx ruff check app/modules/proxy/service.py tests/unit/test_proxy_http_bridge.py tests/unit/test_bridge_context_blowup.py`
- `uv run pytest tests/unit/test_proxy_http_bridge.py tests/unit/test_bridge_context_blowup.py -k "injects_durable_previous_response_anchor or does_not_inject_session_anchor_for_soft_reuse or preserves_full_input_count_after_trimming or skips_trim_when_stored_prefix_was_edited or proxy_injected_previous_response or previous_response_id"`

Commits:
- `919e501` fix(proxy): scope session anchor injection to codex continuity
- `7b55c80` fix(proxy): preserve fresh-upstream retries for injected anchors

app/modules/proxy/service.py
```diff
@@ -653,6 +653,44 @@ async def _stream_via_http_bridge(
...
+            and effective_payload.previous_response_id is None
+            and session.last_completed_response_id is not None
+        ):
+            effective_payload = effective_payload.model_copy(
+                update={"previous_response_id": session.last_completed_response_id}
```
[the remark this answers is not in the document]
balakumardev: Fixed on the PR branch in `7b55c80`.

I kept the injected-anchor path, but preserved fresh-upstream retries by storing the original unanchored request text on proxy-injected anchors and replaying that full request after a reconnect. That keeps the websocket-blip recovery path for Codex-style clients instead of turning the turn into a hard 502.

Coverage added:
- `tests/unit/test_bridge_context_blowup.py::TestRetryHelperPreservesPreviousResponseId::test_retry_with_proxy_injected_previous_response_replays_full_request`
- reran the focused bridge slice covering both anchor injection and retry preservation

Soju06:
Correction to my previous comment — rebasing, CI approval, and `@codex review` triggering are maintainer-side actions, not something the contributor is expected to do manually.

What I meant was simply:

- the logic issues I blocked on look resolved in the current branch
- the PR branch is currently `CONFLICTING / DIRTY` against `main`
- once a maintainer rebases or otherwise refreshes the branch onto current `main`, **we** can re-run CI / re-trigger Codex review from the maintainer side and continue from there

Sorry for the sloppy wording there.

Soju06:
GPT5.4 can't understand what I'm saying lol

Soju06:
@codex review

Soju06:
Pushed fixes for the two issues @chatgpt-codex-connector flagged on the latest pass (`58777e8`). Both were real — thanks for catching them before merge.

## P1 — Guard session anchor injection when full-resend trim would not apply

The injection previously fired for every codex-session follow-up as soon as `session.last_completed_response_id` existed, without mirroring the durable path's full-resend guard. If the incoming payload could not be trimmed (non-list input, shorter history, or prefix fingerprint mismatch), the upstream call received both the full history **and** an injected `previous_response_id` — duplicating context and distorting output/cost.

Fix: gate the injection on the same trimmability check the trim branch performs, so the anchor is only attached when the prefix would actually be stripped.

```python
session_anchor_trimmable = (
    stored_count_preview > 0
    and stored_fingerprint_preview is not None
    and isinstance(incoming_input_preview, list)
    and len(incoming_input_preview) > stored_count_preview
    and _fingerprint_input_items(
        cast(list[JsonValue], incoming_input_preview)[:stored_count_preview]
    )
    == stored_fingerprint_preview
)
if (
    session.codex_session
    and not proxy_injected_previous_response_id
    and effective_payload.previous_response_id is None
    and session.last_completed_response_id is not None
    and session_anchor_trimmable
):
    ...
```

Added `test_stream_via_http_bridge_skips_session_anchor_injection_when_trim_would_not_apply` to lock this in — constructs a codex continuity session with a completed response and a non-list (string) input, then asserts `_prepare_http_bridge_request` sees `previous_response_id=None`.

## P2 — Update `last_completed_response_id` for non-list completed requests

`session.last_completed_response_id` was scoped inside the `input_item_count > 0` branch, which only populates for list inputs. Mixed-input sessions therefore reused a stale response id (or none) on later turns, weakening continuity and trimming.

Fix: move the response id update out of the input-count branch so every completed turn refreshes the session anchor. Keep the input count / fingerprint update scoped to list inputs since prefix trimming is only meaningful there.

```python
if event_type == "response.completed" and terminal_request_state is not None:
    if response_id is not None:
        session.last_completed_response_id = response_id
    if terminal_request_state.input_item_count > 0:
        session.last_completed_input_count = terminal_request_state.input_item_count
        session.last_completed_input_prefix_fingerprint = terminal_request_state.input_full_fingerprint
```

## Ran locally

- `uvx ruff check .` — clean
- `uvx ruff format --check .` — clean
- `uvx ty check app/modules/proxy/service.py` — clean
- `uv run pytest tests/unit/test_proxy_http_bridge.py tests/unit/test_bridge_context_blowup.py -q` — **89 passed**

@codex review — please take another look.

Soju06:
Pushed a fix for the P1 @chatgpt-codex-connector flagged on `_retry_http_bridge_request_on_fresh_upstream` (`15922ae`). Real issue — good catch.

## Problem

The retry helper was dropping *any* proxy-injected `previous_response_id` whose `fresh_upstream_request_text` existed and replaying that as a fresh turn on send failure. That is safe for durable-anchor injections (which capture the original full-resend payload before injecting the anchor), but unsafe for the new session-level anchor injection in this PR: its unanchored payload may have depended on the anchor for context preservation (for example a single-item follow-up whose history lives only in the stored conversation). Replaying without the anchor silently converts a continuation into a context-free fresh turn.

## Fix

Added `_WebSocketRequestState.fresh_upstream_request_is_retry_safe` and only opt injections in when the captured unanchored payload is a true standalone replay target:

| Injection path                                                    | `is_retry_safe` |
| ----------------------------------------------------------------- | --------------- |
| Durable-anchor injection on reattach (captures full resend)       | `True`          |
| Trim branch re-preparation (trim only fires on full resend)       | `True`          |
| Session-level anchor injection on codex continuity session        | `False`         |

`_retry_http_bridge_request_on_fresh_upstream` now gates the fresh-turn replay on this flag in addition to the existing `proxy_injected_previous_response_id + fresh_upstream_request_text` check:

```python
if request_state.previous_response_id is not None and send_request:
    if (
        not request_state.proxy_injected_previous_response_id
        or not request_state.fresh_upstream_request_text
        or not request_state.fresh_upstream_request_is_retry_safe
    ):
        return False
    retry_text_data = request_state.fresh_upstream_request_text
```

So a session-level injection now surfaces the original send failure as a retriable error instead of running as a fresh turn.

## New regression tests

- `test_retry_http_bridge_request_on_fresh_upstream_replays_retry_safe_injection_without_anchor` — `is_retry_safe=True` replay succeeds and drops the anchor, matching the durable path.
- `test_retry_http_bridge_request_on_fresh_upstream_refuses_session_level_injection_replay` — `is_retry_safe=False` replay refused, reconnect/send not invoked.

Also updated `test_retry_with_proxy_injected_previous_response_replays_full_request` in `tests/unit/test_bridge_context_blowup.py` to set the flag explicitly for the durable-anchor path.

## Ran locally

- `uvx ruff check .` — clean
- `uvx ruff format --check .` — clean
- `uvx ty check app/modules/proxy/service.py` — clean
- `uv run pytest tests/unit/test_proxy_http_bridge.py tests/unit/test_bridge_context_blowup.py -q` — **91 passed**

@codex review — please take another look.

Soju06:
Nice — I had the polarity backwards. Pushed `c9a3ba0` to narrow `fresh_upstream_request_is_retry_safe=True` to only the trim-verified path.

## What I had wrong

I set durable-anchor injection to `True` under the assumption that it captured a full-resend payload. Looking again at the reattach block:

```python
if (
    ...
    and not _http_bridge_payload_looks_like_full_resend(payload)
):
    ...
    proxy_injected_previous_response_id = True
    _fresh_request_state, fresh_upstream_request_text = self._prepare_http_bridge_request(
        payload, ...
    )
```

The `not _..._full_resend(payload)` guard means durable injection actually runs when the incoming request is **not** a full resend. The captured unanchored text is usually a short single-item follow-up whose context lived only in the anchor, so replaying it as a fresh turn would silently strip conversation context.

## New policy

`fresh_upstream_request_is_retry_safe` is `True` only when the captured text has been *proven* standalone:

| Injection path                                                           | `is_retry_safe` |
| ------------------------------------------------------------------------ | --------------- |
| Durable-anchor injection on reattach (non-full-resend payload)           | `False`         |
| Session-level anchor injection on codex continuity                       | `False`         |
| Trim branch re-preparation (stored prefix fingerprint verified)          | `True`          |

The trim branch is the only one that verifies the incoming payload actually contains the already-stored prefix, so it's the only one where dropping the anchor and replaying is equivalent to the client's own retry.

Retry semantics from the caller's perspective are unchanged: unsafe injections still surface the original send failure as a retriable error instead of executing as a fresh turn.

## Ran locally

- `uvx ruff check .` — clean
- `uvx ruff format --check .` — clean
- `uvx ty check app/modules/proxy/service.py` — clean
- `uv run pytest tests/unit/test_proxy_http_bridge.py tests/unit/test_bridge_context_blowup.py -q` — **91 passed**

@codex review — please take another look.

Soju06 (approved):
All blocking reviews addressed:

- soft-key cross-chain: gated via `session.codex_session`
- prefix-mismatch silent regression: fingerprint + fallback warning
- non-trimmable anchor injection: trimmability precheck (`session_anchor_trimmable`)
- non-list completed id update: hoisted out of input-count branch
- unsafe fresh-upstream replay: `fresh_upstream_request_is_retry_safe`, True only for trim-verified path

CI 18/18 green, `@codex review` clean on `c9a3ba0`, 91 unit tests passed locally. Closes #395 / #423 / #463 / #464 pending next release.

Soju06:
Huge thanks @balakumardev — this was a non-trivial one. You not only shipped the core architecture (session-level anchor injection + fingerprint-guarded trim extension from #448) but also engaged across multiple review rounds, responded to every blocker from the Codex review bot within hours, and your own self-review caught the initial soft-key cross-chain issue before anyone else did. Directly closes #395 / #423 / #463 / #464 once this ships.

@all-contributors please add @balakumardev for ideas

PR #475, opened 2026-04-23
fix(api-limit): Add fallback for api limit reset

huzky-v:
螢幕截圖 2026-04-23 21 00 08

Added a fallback to reset expired api limit per hour as the old logic relies on lazy reset, which will make stale UI limit still "visually exist" before shooting the request (the lazy reset)
But still preserve the lazy reset for accurate timing

And also a very minor UI adjustment for cleaner time setting, removed the icon and the example
螢幕截圖 2026-04-24 02 07 12


huzky-v:
@codex review

PR #476, opened 2026-04-23
feat(proxy): make response.create payload limit configurable

xCatalitY:
## Summary

The 15 MiB `response.create` websocket payload limit is hardcoded in both `app/modules/proxy/service.py` and `app/core/clients/proxy.py`. This makes it impossible to raise the ceiling without patching source and restarting.

This adds `CODEX_LB_UPSTREAM_RESPONSE_CREATE_MAX_BYTES` to `Settings` so operators can tune the limit via env var or `.env` file. The warn threshold stays proportional at 80% of max. Default is unchanged (15 MiB).

## Changes

- Add `upstream_response_create_max_bytes` field to `Settings` (pydantic, `gt=0`, default `15 * 1024 * 1024`)
- Both modules read the value from settings at import time instead of hardcoding
- Warn threshold derived as `int(max * 0.8)` — no separate config needed
- Existing tests unaffected (they monkeypatch the module-level constant directly)

## Test plan

- [x] Verified default value matches original (15728640 bytes)
- [x] Verified env var override: `CODEX_LB_UPSTREAM_RESPONSE_CREATE_MAX_BYTES=33554432` → 32 MiB
- [x] Ran `test_proxy_utils.py` — all 207 tests pass

### DIFF
--- a/app/core/clients/proxy.py
+++ b/app/core/clients/proxy.py
@@ -89,8 +89,8 @@ _IMAGE_INLINE_MAX_BYTES = 8 * 1024 * 1024
 _IMAGE_INLINE_CHUNK_SIZE = 64 * 1024
 _IMAGE_INLINE_TIMEOUT_SECONDS = 8.0
 _BLOCKED_LITERAL_HOSTS = {"localhost", "localhost.localdomain"}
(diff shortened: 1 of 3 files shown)


Soju06:
@codex review

Soju06:
@all-contributors please add @xCatalitY for code

PR #481, opened 2026-04-24
fix(oauth): make manual callback idempotent

rio-jeong:
## Summary
- Make `/api/oauth/manual-callback` idempotent after the current OAuth attempt has already reached `success`.
- Avoid re-exchanging an already-consumed authorization code when the automatic localhost callback persisted the account first.
- Preserve the successful OAuth/account-add state instead of overwriting it with a later manual-callback token exchange error.

## Observed Bug Evidence
I hit this in a real local Docker dashboard OAuth flow, not only from code inspection.

Environment:
- Dashboard: `http://localhost:2455`
- OAuth callback port exposed as `1455:1455`
- Redirect URI: `http://localhost:1455/auth/callback`

Actual behavior observed:
- The account was added successfully by the automatic callback path.(But I didn't know that.)
- So I pasted the same callback URL into the manual callback field, the dashboard showed an OAuth/account-add failure message.
- The account list still showed the newly added account, so the final UI state contradicted the persisted account state.

Most important evidence is the timing: the account existed before the manual callback request that reported the token error.

```text
account created_at: 2026-04-24 05:30:33 UTC
manual callback:    2026-04-24 05:30:40 UTC
```

Relevant logs from the same flow:

```text
2026-04-24T05:30:17Z POST /api/oauth/start HTTP/1.1 200 OK
2026-04-24T05:30:40Z OAuth token request failed request_id=60915954-d78a-4953-92e3-cf0a0b6316d8 status=400
2026-04-24T05:30:40Z POST /api/oauth/manual-callback HTTP/1.1 200 OK
```

So the account was created about 7 seconds before the manual callback tried to exchange the same already-consumed authorization code.

## Root Cause
The browser OAuth flow has two completion paths: the local callback server at `localhost:1455` and the dashboard manual callback endpoint. In local Docker setups where `1455:1455` is exposed, the automatic callback can complete first and create the account. If the user then submits the same callback URL manually, the previous implementation tried to exchange the same authorization code again. Upstream rejects the consumed code with HTTP 400, and `manual_callback()` converted that into `_set_error()`, causing the dashboard to show an authorization failure even though the account was already added.

## Fix
`manual_callback()` now checks the current OAuth state before exchanging the authorization code. If the OAuth attempt is already `success`, the manual callback returns success without calling `exchange_authorization_code()` again.

## Test Plan
- [x] `uv run --frozen --python $env:APPDATA\uv\python\cpython-3.13.12-windows-x86_64-none\python.exe python -m pytest tests\integration\test_oauth_flow.py -q`

### DIFF
--- a/app/modules/oauth/service.py
+++ b/app/modules/oauth/service.py
@@ -232,14 +232,18 @@ class OauthService:
         code = params.get("code", [None])[0]
         state = params.get("state", [None])[0]
 
+        async with self._store.lock:
+            current_status = self._store.state.status
+            expected_state = self._store.state.state_token
+            verifier = self._store.state.code_verifier
+
+        if current_status == "success":
+            return ManualCallbackResponse(status="success")
+
         if error:
             message = f"OAuth error: {error}"
             await self._set_error(message)
             return ManualCallbackResponse(status="error", error_message=message)
-
-        async with self._store.lock:
-            expected_state = self._store.state.state_token
-            verifier = self._store.state.code_verifier
 
         if not code or not state or state != expected_state or not verifier:
             message = "Invalid OAuth callback: state mismatch or missing code."
--- a/app/modules/oauth/service.py
+++ b/app/modules/oauth/service.py
@@ -237,7 +237,17 @@ class OauthService:
             expected_state = self._store.state.state_token
             verifier = self._store.state.code_verifier
 
-        if current_status == "success":
+        # Idempotent return only when this manual-callback corresponds to the
+        # same OAuth attempt that already succeeded (state token matches the
+        # current attempt). This avoids reporting success for stale callback
+        # URLs from a different/previous attempt, which would skip state/code
+        # validation and token persistence.
+        if (
+            current_status == "success"
+            and state
+            and expected_state
+            and state == expected_state
+        ):
             return ManualCallbackResponse(status="success")
 
         if error:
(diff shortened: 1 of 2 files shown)


Soju06:
@codex review

Soju06:
Pushed a follow-up commit (b0dfa8d) on top of this branch (`maintainerCanModify=true`) addressing the codex review **P2** finding ("Limit idempotent success to the active OAuth attempt").

### What changed

In `app/modules/oauth/service.py::manual_callback`, the idempotent early return now requires the incoming `state` to match the currently-stored attempt's `state_token` in addition to `status == "success"`. Stale callback URLs from a different / previous OAuth attempt no longer bypass state/code validation.

```python
if (
    current_status == "success"
    and state
    and expected_state
    and state == expected_state
):
    return ManualCallbackResponse(status="success")
```

### Regression coverage

Added two integration tests in `tests/integration/test_oauth_flow.py`:

- `test_manual_callback_is_idempotent_for_same_attempt` — same callback URL submitted twice for the same attempt: the second call returns success without re-calling `exchange_authorization_code`.
- `test_manual_callback_after_success_rejects_stale_callback` — a stale URL (different `state`) arriving after success is rejected with the existing `Invalid OAuth callback: state mismatch or missing code.` error rather than masked as success.

`tests/integration/test_oauth_flow.py`: 9 passed locally.

@rio-jeong if you'd prefer a different shape for the gate (e.g. comparing against a one-shot token consumed on first manual-callback success) feel free to amend / rework — happy to defer to your direction. Otherwise this should resolve the codex P2 and keep the original bug you reported fixed.

@codex review

Soju06:
@codex review

Soju06:
@all-contributors please add @rio-jeong for code, bug, test

Soju06:
@all-contributors please add @rio-jeong for bug, code, test

PR #485, opened 2026-04-24
fix(proxy): load balancer filter

### LINKED ISSUE #478: [bug] model sticks to account, even when it's disabled
### Background
I'm having a business/team and free/private account on the same address registered in _codex-lb_. `free` doesn't seem to support `gpt-5.5` now, while business/team does.

### Bug
When disabling business/team and re-run a prompt on a free account with the same model, it gets run against disabled/paused business/team again. I think _codex-lb_ gets confused, since it has the same account name (but different ID !).

### Steps to reproduce
command:
```
curl -s http://127.0.0.1:2455/v1/responses -H "Content-Type: application/json" -d '{"model": "gpt-5.5", "input": "Say hi."}'
```

1. enable business/team account & run -> query gets executed on business/team
2. pause business/team account (free is active only ) & run ->  query STILL executed on business/team (although it's paused)

### Expected
Run the prompt on the available/unpaused accounts only and throw an error when the model isn't available on the plan.

### environment
- codex-lb Version: 1.15.0

tobwen:
### Summary
Fixes #478 by applying proposed fix by @Soju06: filter out `PAUSED` and `DEACTIVATED` accounts at the `_load_selection_inputs()` level so disabled accounts never enter sticky, budget, or fallback selection.

### Problem
`_load_selection_inputs()` passes **all** accounts (including paused/deactivated) into the sticky/budget/fallback selection pool. This causes paused accounts to remain eligible for sticky session reuse, budget-weighted selection, and fallback routing.

### Fix
Added `_selectable_accounts()` and called it at the top of `_load_selection_inputs()`. That means paused/deactivated accounts are removed before:

- sticky pin evaluation
- budget pressure decisions
- fallback routing

The unfiltered `all_accounts` list is still preserved for the no-accounts / model-support error paths, as requested by @Soju06:

### Changes
| File | Change |
|------|--------|
| `app/modules/proxy/load_balancer.py` | Add `_selectable_accounts()`, filter in `_load_selection_inputs()` |
| `tests/unit/test_proxy_load_balancer_refresh.py` | Regression test: paused account excluded from sticky pool and no sticky upsert for paused ID |

### Coverage
@Soju06 asked to cover some points...

1. Keep `all_accounts` for the no-accounts error path.
   - Kept. Only the working `accounts` pool is filtered.

2. Double-check sticky cleanup semantics.
   - Covered by test. A paused pinned account is no longer reachable in the filtered pool.
   - The stale sticky row is deleted when the pinned ID is no longer present.
   - The regression test asserts that no sticky row is ever upserted for the paused account ID.

3. Add regression coverage in `tests/unit/test_proxy_load_balancer_refresh.py`.
   - Added exactly there.
   - Verified red on `main`, green on this branch.

### Why I kept the sticky behavior minimal
I intentionally did **not** introduce a broader sticky-behavior redesign in this patch.

@Soju06 explicitly asked for the selection-pool filter plus a regression test. The sticky-path note (`"should be actively deleted or at least not re-upserted onto a different account"`) was phrased as a double-check, not as a hard requirement to change rebinding semantics.

So this patch stays minimal:

- it prevents paused accounts from being selected
- it ensures the paused account ID is not re-upserted
- it avoids adding new sticky migration behavior beyond what is needed for the bugfix

### Testing
```
35 passed, 3 skipped (load balancer tests)
1281 passed, 39 skipped (full unit suite)
```

Targeted regression proof:
- `main` + new regression test: **fails**
- this branch + same test: **passes**

Manual verification against running instance with real accounts:

1. Account 1 active, Account 2 paused → request → **200** → pass
2. Account 1 paused, Account 2 paused → request → **503 no_accounts** → pass
3. Account 1 paused, Account 2 active →  request → **200** → pass
4. Dashboard was checked to verify there wasn't stickyness on disabled accounts.

### DIFF
--- a/app/modules/proxy/load_balancer.py
+++ b/app/modules/proxy/load_balancer.py
@@ -411,7 +411,7 @@ class LoadBalancer:
         async with self._repo_factory() as repos:
             all_accounts = await repos.accounts.list_accounts()
             effective_limit_name = additional_limit_name or _gated_limit_name_for_model(model)
-            accounts = all_accounts
+            accounts = _selectable_accounts(all_accounts)
             if account_ids is not None:
                 allowed_account_ids = set(account_ids)
                 accounts = [account for account in accounts if account.id in allowed_account_ids]
@@ -1165,6 +1165,10 @@ def _filter_accounts_for_model(accounts: list[Account], model: str) -> list[Acco
     if allowed_plans is None:
         return accounts
     return [a for a in accounts if a.plan_type in allowed_plans]
+
+
+def _selectable_accounts(accounts: list[Account]) -> list[Account]:
+    return [account for account in accounts if account.status not in (AccountStatus.DEACTIVATED, AccountStatus.PAUSED)]
 
 
 def _gated_limit_name_for_model(model: str | None) -> str | None:
--- a/tests/unit/test_proxy_load_balancer_refresh.py
+++ b/tests/unit/test_proxy_load_balancer_refresh.py
@@ -146,7 +146,8 @@ class StubUsageRepository(UsageRepository):
 
 class StubStickySessionsRepository(StickySessionsRepository):
     def __init__(self) -> None:
-        pass
+        self.upserts: list[StickySession] = []
+        self.deletes: list[tuple[str, StickySessionKind | None]] = []
 
     async def get_account_id(
         self,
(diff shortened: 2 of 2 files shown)


Soju06:
@codex review

Soju06:
@codex review

Soju06:
@all-contributors please add @tobwen for code, test, bug

Soju06:
@all-contributors please add @tobwen for code, test, bug

PR #487, opened 2026-04-25
feat(cli): add Codex session retag command

nicefellow1234:
## Summary

This PR adds a lower-surface Codex session provider migration command:

```bash
codex-lb codex-sessions retag --from openai --to codex-lb [--dry-run] [--codex-home PATH] [--yes]
```

It replaces the manual README `find ... sed` + `sqlite3 UPDATE` migration snippet with a backup-first CLI flow that can retag Codex sessions between `openai` and `codex-lb`.

## What Changed

- Added `codex-lb codex-sessions retag` for one-time provider retagging.
- Supports both Codex storage formats:
  - JSONL session files under `sessions/**/*.jsonl`.
  - `state_*.sqlite` thread rows used by newer Codex CLI versions.
- Adds `--dry-run` for read-only previews before writing.
- Adds `--codex-home PATH` so users can target native, WSL, or mounted Docker paths explicitly.
- Adds `--yes` for non-interactive writes after users confirm Codex/Codex CLI is closed.
- Creates backups before mutation under `<codex-home>/backups/provider-retag/<timestamp>/`.
- Uses Python's built-in `sqlite3`, so no external `sqlite3` CLI is required.
- Restricts WSL Codex-home autodetect to the current Windows `USERPROFILE`; other Windows profiles require explicit `--codex-home`.
- Opens SQLite dry-run/count reads in immutable read-only mode, which lets read-only Docker bind mounts work cleanly.
- Uses `sqlite3.Connection.backup()` for SQLite backups and copy-write fallback snapshots so WAL-backed committed rows are included.
- Consolidates fallback temp DBs before copying them back and removes stale `-wal` / `-shm` sidecars after replacement.
- Adds a runtime Docker `codex-lb` launcher so the CLI command is available inside the built image even though the project itself is not installed into the venv during image build.
- Keeps Docker usage opt-in through a one-off bind mount instead of adding a permanent Compose mount.
- Keeps container/cgroup runtime detection scoped to this command instead of changing shared Settings behavior.
- Keeps default Docker/Compose ports unchanged.
- Keeps server-only imports lazy so the retag CLI path does not import uvicorn/FastAPI just to show help or run migration commands.

## Behavioral Concerns Addressed

- **SQLite while Codex CLI is running:** real writes print a warning to close Codex/Codex CLI first; non-interactive writes require `--yes`; locked SQLite failures are surfaced with a close-and-retry instruction.
- **Runtime detection side effects:** cgroup/container checks live only in the retag command's Codex-home discovery, not in core Settings or OAuth/server behavior.
- **WSL profile safety:** autodetect maps only the current Windows `USERPROFILE` to `/mnt/<drive>/.../.codex`; cross-profile retagging is opt-in via `--codex-home`.
- **Docker mounts:** Docker does not get a default `~/.codex` mount. Users mount Codex data only for the one-off command, e.g. `-v ~/.codex:/codex-home` and `--codex-home /codex-home`.
- **Read-only Docker dry-runs:** SQLite scans use `mode=ro&immutable=1`, so dry-run works with `:ro` bind mounts.
- **WAL safety:** backup and copy-write fallback paths use SQLite backup/checkpoint primitives instead of raw-copying only the main `.sqlite` file.

## Usage

Native / WSL / macOS / Linux dry-run:

```bash
codex-lb codex-sessions retag --from openai --to codex-lb --dry-run
```

Apply after closing Codex/Codex CLI:

```bash
codex-lb codex-sessions retag --from openai --to codex-lb --yes
```

Docker dry-run with an explicit one-off mount:

```bash
docker run --rm \
  -v ~/.codex:/codex-home \
  ghcr.io/soju06/codex-lb:latest \
  codex-lb codex-sessions retag --from openai --to codex-lb \
    --codex-home /codex-home --dry-run
```

Docker apply after closing Codex/Codex CLI:

```bash
docker run --rm \
  -v ~/.codex:/codex-home \
  ghcr.io/soju06/codex-lb:latest \
  codex-lb codex-sessions retag --from openai --to codex-lb \
    --codex-home /codex-home --yes
```

Reverse direction:

```bash
codex-lb codex-sessions retag --from codex-lb --to openai --dry-run
codex-lb codex-sessions retag --from codex-lb --to openai --yes
```

### Docker dry-run with host Codex home mounted read-only

Docker dry-run

### Docker apply against a mounted Codex home copy

Docker apply

### WSL dry-run against Windows Codex home

WSL dry-run

### WSL apply against a Codex home copy

WSL apply

## Validation

- `uv run ruff check app/cli.py app/codex_sessions_retag.py tests/unit/test_cli.py tests/unit/test_codex_sessions_retag.py`
- `uv run ruff format --check app/cli.py app/codex_sessions_retag.py tests/unit/test_cli.py tests/unit/test_codex_sessions_retag.py`
- `uv run ty check`
- `TMPDIR=/tmp TMP=/tmp TEMP=/tmp uv run pytest -q tests/unit/test_cli.py tests/unit/test_codex_sessions_retag.py`
- Unit coverage includes WSL current-user detection, immutable read-only SQLite scans, WAL-consolidated SQLite backups, and WAL-safe copy-write fallback replacement.
- Built separate local image: `docker build -t codex-lb:provider-session-retag-test .`
- Docker dry-run tested with real Codex home mounted read-only: `/mnt/c/Users/nicef/.codex:/codex-home:ro`
- Docker apply tested against a synthetic mounted Codex home copy; verified both JSONL and SQLite changed to `codex-lb` and backup was created.
- WSL dry-run tested against real Windows Codex home path: `/mnt/c/Users/nicef/.codex`
- WSL apply tested against a synthetic Codex home copy; verified both JSONL and SQLite changed to `codex-lb` and backup was created.

## Notes

This branch intentionally removes the earlier dashboard/API/SSE/frontend/Compose-default-mount approach. The migration stays as a CLI-only utility for a one-time workflow, while preserving the same backup-first JSONL + SQLite behavior.

Commit used for screenshots: `78f4dbc6e6126ce6685797fec6bfc15f7d5e2b2f`


nicefellow1234:
@Codex review

Soju06:
Hey @nicefellow1234 — first off, thanks for the very thorough work on this. The implementation is genuinely careful: backups before mutation, JSONL + SQLite parity, SSE progress, runtime detection across Docker/WSL/Windows/macOS/Linux, and an explicit Docker-mount story. The screenshots and OpenSpec change notes are great too.

Before going further on review, I want to flag a scope concern up front so we don't spend a lot of round-trip cycles on something I might end up not merging in this shape.

## Concern: scope vs. lifetime use of the feature

Provider re-tagging is essentially a **one-time migration** per user — typically run once when switching from direct OpenAI to codex-lb (or back), and rarely again. The current README path for that is two shell lines (`find ... sed` + `sqlite3 UPDATE threads`).

This PR adds, to support that one-time migration:

- a new backend module (`app/modules/codex_sessions/`, ~820 lines: api + service + schemas)
- a frontend Settings panel + SSE client (~600 lines incl. tests)
- runtime detection (`_in_container` cgroup hardening, WSL `/mnt/c/Users/*/.codex` discovery, `detect_runtime_environment`) added to core `Settings`
- a new default Docker-Compose bind mount (`${CODEX_LB_CODEX_HOST_DIR:-~/.codex}:/codex-home`) and `CODEX_LB_CODEX_HOME=/codex-home`
- new env vars and README sections covering Docker / WSL / native overrides

That's a meaningful amount of permanent surface area + ongoing maintenance for a workflow most users will run once. The Docker-Compose default-mount change in particular affects every existing Compose user, even those who will never use the switcher.

I also have two smaller behavioral concerns I'd want to address regardless of shape:

1. **SQLite write while codex-cli is running**. `_update_sqlite_provider` issues `UPDATE threads ...` directly against `state_*.sqlite`. If the user's codex-cli is alive at the same time, this can race / lock. There's `PRAGMA busy_timeout=5000`, but no UI guard telling the user to close codex-cli first.
2. **`_in_container()` cgroup hardening side-effects**. The cgroup-marker check now also influences any other code path that calls `_in_container()` (e.g. OAuth callback host defaulting). On hosts where cgroups happen to contain `docker`/`kubepods`/`containerd` substrings, this can flip codex-lb into container mode even when running natively.

## Lower-surface alternative I'd prefer

I think this maps better onto a **CLI command** rather than a dashboard module:

```
codex-lb codex-sessions retag --from openai --to codex-lb [--dry-run] [--codex-home PATH]
```

With:

- the same backup-first / JSONL + SQLite parity / clear progress that this PR has
- no new HTTP endpoints, no SSE job, no frontend panel, no permanent Compose bind mount
- README updated to point to the command instead of the current `find`/`sqlite3` snippet
- runtime detection scoped to the command, not pushed into core `Settings`

For Docker users, mounting `~/.codex` becomes opt-in (`docker run --rm -v ~/.codex:/codex-home codex-lb codex-sessions retag ...`) instead of an always-on default mount on every Compose deployment.

That gives users the same end result (no manual sed / sqlite3, with backups) at maybe ~1/4 the permanent surface area.

## What I'd like to do

I haven't closed this PR. Two reasonable paths from here:

1. **You're up for re-scoping**: I close this PR and you (or anyone) opens a follow-up PR adding just the CLI command + README pointer. The retagging logic in `service.py` is mostly reusable — it's the API/SSE/frontend/Compose-mount layer I'd drop.
2. **You'd rather not re-scope**: I'd close this as `wontfix` / `out-of-scope` and just keep the README shell snippet. No hard feelings — the work is solid, the disagreement is purely about whether this belongs as a permanent dashboard surface.

Either way, I really appreciate the effort and the level of polish. Let me know which direction you'd prefer, and I'll act on it. 🙏

nicefellow1234:
I reworked and pushed the PR as the CLI-only implementation requested in review.

Highlights:
- Removed the dashboard/API/SSE/frontend surface and default Compose Codex-home mount.
- Added `codex-lb codex-sessions retag` with dry-run, explicit `--codex-home`, backup-first JSONL + SQLite updates, and `--yes` for non-interactive writes.
- Addressed SQLite behavior by warning users to close Codex/Codex CLI before writes and by using immutable read-only SQLite scans for Docker read-only dry-runs.
- Kept container/cgroup detection scoped to the command, not core Settings.
- Added Docker runtime `codex-lb` launcher support and opt-in Docker mount docs.
- Added screenshots for Docker dry-run/apply and WSL dry-run/apply flows.

Validation is listed in the PR body, including Docker and WSL test evidence. Current pushed commit: `a83c8ed661d3403cabe47471b7049a636fdaa1c7`

Soju06:
Re-scope LGTM at a glance — thanks for trimming this back so cleanly. Approving the CI run and triggering codex review now; will follow up after both come back.

@codex review

Soju06:
Thanks for the very clean re-scope @nicefellow1234 — the CLI shape (`codex-lb codex-sessions retag`), the `--dry-run` / `--yes` ergonomics, the read-only SQLite scan via `mode=ro&immutable=1`, the in-place + copy-write fallback, the per-command runtime detection, and the screenshots covering the Docker dry-run/apply and WSL dry-run/apply paths are all spot on. This is a much more proportionate way to ship the workflow.

A few items to address before merge — most are codex-review findings that I think are real, plus a CI fix.

## P1 — only one is purely from me; the rest were flagged by codex review on the current commit

### 1. Restrict WSL Codex-home autodetect to the current Windows user (codex review P1)

`_discover_wsl_windows_codex_home()` currently scans every `/mnt/c/Users/*/.codex` directory and returns the highest-scoring one. On a machine with multiple Windows profiles, this can pick another user's Codex data and rewrite it in place. Because this command does mutating writes, that's a data-integrity risk for the autodetect path.

Suggested fix:

- Prefer `$USERPROFILE` mapped through `/mnt/c/Users/<that-user>/.codex` (matching the current Windows account) when it exists.
- If `USERPROFILE` is unset or the mapped path doesn't exist, fall back to a single, deterministic candidate (e.g. the WSL user's `~/.codex`) rather than scoring across all `/mnt/c/Users/*` profiles.
- Document `--codex-home` as the explicit override for any other Windows profile.

This trims the WSL autodetect to "current Windows account, deterministic" and keeps the cross-profile case opt-in via the existing `--codex-home`.

### 2. Include SQLite WAL sidecars in the pre-mutation backup (codex review P1)

`_create_backup()` copies each `state_*.sqlite` directly with `shutil.copy2`, but doesn't account for `state_*.sqlite-wal` / `state_*.sqlite-shm` if the database is in WAL journal mode (which Codex's `state_*.sqlite` typically is). If the user runs the retag right after an unclean Codex shutdown, the backup can be missing committed-but-not-checkpointed data, and the rollback claim would be incomplete in exactly the scenario where rollback matters.

Suggested fix:

- Either: copy `state_*.sqlite`, `state_*.sqlite-wal`, and `state_*.sqlite-shm` together when they exist next to the main DB.
- Or (simpler and more correct): use `sqlite3` Python `Connection.backup()` against a freshly-opened connection so SQLite itself produces a consolidated, fully-checkpointed copy on disk. That sidesteps the WAL-sidecar question entirely.

I'd lean toward the `Connection.backup()` approach since it's a one-call, journal-mode-agnostic primitive and matches the safety guarantee in the README.

### 3. Preserve WAL state in the SQLite copy-write fallback (codex review P1)

`_update_sqlite_provider_via_copy()` copies the live `.sqlite` file with `shutil.copy2`, updates the temp copy, then `shutil.copy2`s it back over the original. If the source DB is in WAL mode and there are uncheckpointed commits in the live `-wal`, the snapshot used for retagging won't reflect them, and the write-back can also overwrite the main file with a stale snapshot — silently dropping WAL-backed thread rows in the same Docker bind-mount path this fallback was built for.

Suggested fixes (any one is sufficient):

- Before the fallback, open the live DB and run `PRAGMA wal_checkpoint(TRUNCATE)` (or `Connection.backup()` the live DB into the temp copy) so the temp file reflects all committed data.
- Or refuse the in-place write if `-wal` exists and is non-empty, instructing the user to close Codex first (matching the existing "close Codex" guidance) rather than silently degrading.

The first is friendlier; the second is the conservative safe-by-default option.

## CI: Type check (ty) failing

`uv run ty check` is red on `app/cli.py:92`:

```
def _build_log_config() -> dict[str, object]:
    from app.core.runtime_logging import build_log_config

    return build_log_config()
```

`build_log_config()` returns `LogConfig = dict[str, LogConfigValue]` (where `LogConfigValue = str | bool | None | dict[str, LogConfigValue]`), not `dict[str, object]`, so the return-type annotation should be `LogConfig`:

```python
from app.core.runtime_logging import LogConfig, build_log_config

def _build_log_config() -> LogConfig:
    return build_log_config()
```

(There's also an `unused-type-ignore` warning on `tests/unit/test_proxy_utils.py:179` in the same job, but that file isn't touched by this PR — it looks like an unrelated stale suppression on `main`. Please ignore that one; I'll clean it up separately.)

## Codex review noise to ignore

Codex review left three additional comments under `app/modules/codex_sessions/api.py` and `app/core/config/settings.py` (P1 about `_discover_wsl_windows_codex_home` in `settings.py`, P2 about `_switch_jobs` retention, P2 about `_switch_event_stream` broadcast). Those all target the **previous** dashboard/SSE shape that you removed in this re-scope and don't apply to the current diff — the re-review ended up scoring against a snapshot that no longer exists. Safe to ignore.

- The Dockerfile launcher (`/usr/local/bin/codex-lb` shim → `python -m app.cli`) is a nice touch for a CLI that lives in a venv-less image. ✅
- `_write_text_atomically` using `NamedTemporaryFile` + `Path.replace` is the correct idiom for the JSONL writes. ✅
- The READMEs `--dry-run` first → `--yes` apply flow is great UX guidance.

Once items 1–3 plus the `cli.py` ty fix are addressed, this should be ready. Thanks again — really appreciate how willingly you re-scoped this. 🙏

nicefellow1234:
Pushed a follow-up fix for `uv run ty check`.

- `_build_log_config()` now uses the existing `LogConfig` type alias through a `TYPE_CHECKING` import, keeping server-only imports lazy while satisfying ty.
- Re-ran `uv run ty check`: passed.
- Re-ran focused lint/format checks and the focused unit suite: passed (`13 passed`).

Current pushed commit: `4c3a3a697dee9aae2b36f67f40bf5f946930773d`

nicefellow1234:
Addressed the latest review items and pushed `78f4dbc6e6126ce6685797fec6bfc15f7d5e2b2f`.

- **WSL autodetect:** `_discover_wsl_windows_codex_home()` no longer scans all `/mnt/c/Users/*/.codex` profiles. It maps only the current Windows `USERPROFILE` to the corresponding WSL path and otherwise falls back to the deterministic normal default path. README now calls out `--codex-home /mnt/c/Users/<name>/.codex` for other profiles.
- **SQLite backup WAL safety:** pre-mutation backups now use `sqlite3.Connection.backup()` instead of `shutil.copy2` on the main DB file, then checkpoint/switch the backup copy out of WAL mode so the backup is a consolidated SQLite file.
- **Copy-write fallback WAL safety:** the fallback temp DB is also created through `Connection.backup()`, then consolidated before copy-back; stale destination `-wal` / `-shm` sidecars are removed after replacement.
- **CI type check:** the `LogConfig` return annotation fix is still in place and `uv run ty check` passes.

Validation rerun:
- `uv run ruff check app/cli.py app/codex_sessions_retag.py tests/unit/test_cli.py tests/unit/test_codex_sessions_retag.py`
- `uv run ruff format --check app/cli.py app/codex_sessions_retag.py tests/unit/test_cli.py tests/unit/test_codex_sessions_retag.py`
- `uv run ty check`
- `TMPDIR=/tmp TMP=/tmp TEMP=/tmp uv run pytest -q tests/unit/test_cli.py tests/unit/test_codex_sessions_retag.py` -> `18 passed`

Komzpa:
@codex review

nicefellow1234:
Addressed the latest Codex review P1 about unreadable JSONL files and pushed `e7a8442`.

- `_read_jsonl_records()` now propagates file read failures instead of treating unreadable files as having no matching provider.
- The CLI now reports filesystem access failures as a clear `Unable to read or write Codex session files: ...` error instead of a traceback.
- Added regression coverage for unreadable JSONL scan behavior and CLI reporting.
- `uv.lock` refreshed the editable package version to match `pyproject.toml` (`1.17.0`).

Validation:
- `uv run ruff check app/cli.py app/codex_sessions_retag.py tests/unit/test_cli.py tests/unit/test_codex_sessions_retag.py`
- `uv run ruff format --check app/cli.py app/codex_sessions_retag.py tests/unit/test_cli.py tests/unit/test_codex_sessions_retag.py`
- `uv run ty check`
- `TMPDIR=/tmp TMP=/tmp TEMP=/tmp uv run pytest -q tests/unit/test_cli.py tests/unit/test_codex_sessions_retag.py` -> `20 passed`

Komzpa:
@codex review

nicefellow1234:
Addressed the latest Codex review P2 about `PORT` parsing and pushed `7dc27ac`.

- `codex-sessions retag` no longer eagerly converts the server-only `PORT` default during argument parsing.
- Server startup still validates `--port` / `PORT` and now reports a clean CLI error for non-integer values.
- Added regression coverage for both paths: bad `PORT` is ignored by retag, but rejected on server startup.

Validation:
- `uv run ruff check app/cli.py app/codex_sessions_retag.py tests/unit/test_cli.py tests/unit/test_codex_sessions_retag.py`
- `uv run ruff format --check app/cli.py app/codex_sessions_retag.py tests/unit/test_cli.py tests/unit/test_codex_sessions_retag.py`
- `uv run ty check`
- `TMPDIR=/tmp TMP=/tmp TEMP=/tmp uv run pytest -q tests/unit/test_cli.py tests/unit/test_codex_sessions_retag.py` -> `22 passed`

Komzpa:
@codex review

Soju06:

Komzpa:
Revived on a maintainer-owned branch as #763 because this PR head has maintainerCanModify=false and cannot be updated here. Keeping credit to the original implementation in the replacement PR.

PR #495, opened 2026-04-25
fix(proxy): pre-validate strict JSON schemas to surface invalid_json_schema (#491)

### LINKED ISSUE #491: [bug] /v1/responses surfaces stream_incomplete instead of upstream invalid_json_schema (and retries permanent schema errors)
## Summary

When `/v1/responses` receives an invalid JSON schema (strict-mode violation), the codex backend returns a detailed `invalid_json_schema` error via a `response.failed` event. However, codex-lb's websocket bridge fails to surface this and replaces it with a generic `502 stream_incomplete`. The same schema sent to `/v1/chat/completions` preserves the original error message thanks to the conversion path.

Additionally, despite this being a deterministic (permanent) error, the bridge treats it like a transient failure and runs a reattach/retry, sending the same invalid request to the same account a second time.

## Reproduction

On the dev instance (`http://127.0.0.1:2456`):

### Same schema, two endpoints

```bash
KEY=...

# A. /v1/chat/completions  → preserves the friendly invalid_json_schema message
curl -s -X POST http://127.0.0.1:2456/v1/chat/completions \
  -H "Authorization: Bearer $KEY" -H "Content-Type: application/json" \
  -d '{
    "model":"gpt-5.5",
    "messages":[{"role":"user","content":"hi"}],
    "response_format":{
      "type":"json_schema",
      "json_schema":{
        "name":"person","strict":true,
        "schema":{"type":"object","properties":{"name":{"type":"string"}},"required":["name"]}
      }
    }
  }'
```

Response (HTTP 502):
```json
{"error":{
  "message":"Invalid schema for response_format 'person': In context=(), 'additionalProperties' is required to be supplied and to be false.",
  "type":"invalid_request_error",
  "code":"invalid_json_schema",
  "param":"text.format.schema"
}}
```

```bash
# B. /v1/responses (identical schema) → message is lost as stream_incomplete
curl -s -X POST http://127.0.0.1:2456/v1/responses \
  -H "Authorization: Bearer $KEY" -H "Content-Type: application/json" \
  -d '{
    "model":"gpt-5.5",
    "input":"hi",
    "text":{"format":{
      "type":"json_schema","name":"person","strict":true,
      "schema":{"type":"object","properties":{"name":{"type":"string"}},"required":["name"]}
    }}
  }'
```

Response (HTTP 502):
```json
{"error":{
  "message":"Upstream websocket closed before response.completed (close_code=1000)",
  "type":"server_error",
  "code":"stream_incomplete"
}}
```

### Bridge retry log (case B)

```
proxy_error_response request_id=10364f0a-... path=/v1/responses status=502 code=stream_incomplete ...
http_bridge_event event=retry_precreated bridge_kind=prompt_cache ...
http_bridge_event event=reconnect request_stage=reattach ... selected_account_id=<same as before>
proxy_error_response request_id=bc495a61-... path=/v1/responses status=502 code=stream_incomplete ...
```

→ Permanent error, but the bridge runs another reattach against the same account.

## Comparison (reference)

The same schema sent directly to OpenAI's official `/v1/responses` API:

```json
HTTP 400
{"error":{
  "message":"Invalid schema for response_format 'person': In context=(), 'additionalProperties' is required to be supplied and to be false.",
  "type":"invalid_request_error",
  "param":"text.format.schema",
  "code":"invalid_json_schema"
}}
```

→ The codex backend returns the equivalent error via a `response.failed` event, but the websocket close arrives with `close_code=1000` (clean shutdown), and codex-lb does not surface the final `response.failed` payload before mapping it to `stream_incomplete`.

## Impact

1. Clients (Graphiti, custom SDK integrations, etc.) cannot debug schema problems — the message does not indicate which node violates strict mode, only "upstream closed".
2. Invalid requests still incur retry/reconnect cost (extra hit on the same account, prompt-cache key recomputation, etc.).
3. Asymmetry between endpoints: `/v1/chat/completions` preserves the message, `/v1/responses` does not.

## Suggested fix

1. **Surface the upstream `response.failed` payload**: capture the final `response.failed` event's `error.code` / `error.message` before the websocket close and include it in the 502 body. Five close paths in `app/modules/proxy/service.py` (around lines 947-993, 1932-1933, 2279-2312, 4516-4517, 4721-4744) all overwrite the upstream error with a generic `stream_incomplete`.
2. **Skip retry on deterministic errors**: when `error.code` is something like `invalid_json_schema` / `invalid_request_error`, skip the reattach/retry.

## Environment

- codex-lb: dev build (recent main)
- Model: gpt-5.5
- Calls: direct `/v1/responses` and `/v1/chat/completions` conversion path

Related: structured outputs / strict json_schema (mirrors OpenAI's official policy — the codex backend is not stricter, codex-lb just fails to surface the error).

Soju06:
Closes #491.

## Problem

When a client sends a strict-mode JSON schema (`response_format` /
`text.format` with `strict: true`) that violates OpenAI's structured-outputs
rules — for example `additionalProperties:false` missing on an `object` node,
a missing `type` key, or `additionalProperties:true` on a nested object —
the Codex backend rejects it and closes the websocket session with
`close_code=1000`, sending the original `invalid_json_schema` detail in a
`response.failed` event. The proxy currently overwrites that payload with
a generic `502 stream_incomplete`, so callers (Graphiti, raw SDK clients,
etc.) lose the actionable schema-path diagnostic. Worse, the bridge treats
this as a transient close and runs a `retry_precreated` / `reconnect` cycle
against the same account, wasting another upstream hit on a permanently
invalid request.

`/v1/chat/completions` already preserved the message because the conversion
path captures the upstream error envelope, but `/v1/responses` did not — an
asymmetry between the two endpoints.

## Fix

Validate strict-mode schemas locally before any upstream connection is
opened, mirroring OpenAI's published rules:

* every `object` schema must have `additionalProperties: false`;
* every schema node must declare a `type` (no empty `{}`);
* `anyOf` / `oneOf` / `allOf` branches and `$defs` are walked recursively.

Schemas with `strict` omitted or `strict: false` are passed through
unchanged so the upstream API keeps owning that policy.

The check is done after `ResponsesRequest` is constructed (in
`normalize_responses_request_payload`, plus a callable
`enforce_strict_text_format` used by `/v1/chat/completions` and
`/v1/responses`). On violation, `ClientPayloadError` is raised with
`code=invalid_json_schema`, `type=invalid_request_error`,
`param=text.format.schema`, and is rendered into the OpenAI-style error
envelope by the new `openai_client_payload_error` helper. The result is a
deterministic 400 that matches the message reporters see when they call
`api.openai.com` directly.

Because the request never reaches the upstream websocket, the bridge no
longer triggers a `retry_precreated` cycle for permanently invalid
schemas — the second part of #491 — and the two endpoints become
consistent.

## Verification

Empirical comparison of the proxy (against the dev instance) and OpenAI's
official `/v1/responses` API for the same payloads:

| Schema (strict=true)                                                | OpenAI official     | codex-lb before  | codex-lb after  |
| ------------------------------------------------------------------- | ------------------- | ---------------- | --------------- |
| missing `additionalProperties` at root                              | 400 invalid_json_schema | 502 stream_incomplete | 400 invalid_json_schema |
| nested `additionalProperties: true` (`dict[str, Any]` field)        | 400 invalid_json_schema | 502 stream_incomplete | 400 invalid_json_schema |
| empty `{}` schema node (`Any` field)                                | 400 invalid_json_schema | 502 stream_incomplete | 400 invalid_json_schema |
| valid `additionalProperties:false` + all required                   | 200                 | 200              | 200             |
| `strict: false` (no `additionalProperties`)                         | 200                 | 200              | 200             |

The `400` body now carries the same envelope as upstream:

```json
{
  "error": {
    "message": "Invalid schema for response_format 'graph': In context=('properties', 'nodes', 'items'), 'additionalProperties' is required to be supplied and to be false.",
    "type": "invalid_request_error",
    "code": "invalid_json_schema",
    "param": "text.format.schema"
  }
}
```

Bridge logs no longer show `retry_precreated` / `reconnect` events for
these requests because the connection is never opened.

## Tests

* New unit suite `tests/unit/test_strict_schema_validation.py` covering
  the validator (root violation, nested violation, `dict[str, Any]`,
  `Any` field, valid pass-through, combinator recursion) and the
  pipeline integration through both `normalize_responses_request_payload`
  and `enforce_strict_text_format` (chat-completions path).
* Updated existing
  `tests/integration/test_openai_compat_features.py::test_v1_chat_completions_maps_response_format`
  fixture to be strict-compliant, plus a new
  `test_v1_chat_completions_rejects_strict_schema_violation` integration
  test asserting the 400 envelope shape.
* Full unit suite passes (1320 tests).

## Notes

* Only strict-mode constraints are enforced. The proxy does not rewrite
  schemas to make them compliant — that would silently change the
  client's intent. Returning a deterministic 400 lets the client decide
  whether to drop `strict`, add `additionalProperties: false`, or fix
  optional-field handling.
* The validator handles `$ref` / combinator nodes by recursing without
  requiring a top-level `type`, since strict mode allows those shapes.
* The `openai_client_payload_error` helper preserves backward
  compatibility with existing `ClientPayloadError` callsites that don't
  set `code` / `error_type` — they still get the legacy
  `invalid_request_error` payload.

### DIFF
--- a/app/core/openai/exceptions.py
+++ b/app/core/openai/exceptions.py
@@ -2,6 +2,23 @@ from __future__ import annotations
 
 
 class ClientPayloadError(ValueError):
-    def __init__(self, message: str, *, param: str | None = None) -> None:
+    """Validation error that should be surfaced to the client as a 400.
+
+    ``code`` and ``error_type`` allow callers to mirror specific OpenAI
+    error shapes (for example ``invalid_json_schema`` /
+    ``invalid_request_error`` for strict-mode schema violations) instead
+    of the generic ``invalid_request_error`` payload used by default.
+    """
+
+    def __init__(
+        self,
+        message: str,
+        *,
+        param: str | None = None,
+        code: str | None = None,
+        error_type: str | None = None,
+    ) -> None:
         super().__init__(message)
         self.param = param
+        self.code = code
+        self.error_type = error_type
--- a/app/core/openai/strict_schema.py
+++ b/app/core/openai/strict_schema.py
@@ -100,6 +100,18 @@ def _find_violation(
 
         properties = node.get("properties")
         if is_json_mapping(properties):
+            required_raw = node.get("required")
+            required_set: set[str] = set()
+            if is_json_list(required_raw):
+                required_set = {item for item in required_raw if isinstance(item, str)}
+            missing_required = [str(prop_name) for prop_name in properties.keys() if str(prop_name) not in required_set]
+            if missing_required:
+                missing_list = ", ".join(f"'{name}'" for name in missing_required)
+                return (
+                    path,
+                    "'required' is required to be supplied and to be an array including every key in properties. "
+                    f"Missing {missing_list}",
+                )
             for prop_name, prop_schema in properties.items():
                 violation = _find_violation(
                     prop_schema,
--- a/app/core/openai/strict_schema.py
+++ b/app/core/openai/strict_schema.py
@@ -1,0 +1,146 @@
+"""Validation for OpenAI strict JSON schema mode.
+
+When ``strict: true`` is set on a ``response_format`` (or ``text.format``)
+of type ``json_schema``, the upstream Codex/Responses API enforces several
+constraints on the schema (mirroring the public OpenAI Structured Outputs
+policy). The most common violations are:
+
+* every ``object`` schema must have ``additionalProperties: false``;
+* every property defined under ``properties`` must be listed in
+  ``required``;
+* schema nodes must have a ``type`` key (no empty ``{}``).
+
+When these violations reach the Codex backend over a websocket, the
+session is terminated with ``close_code=1000`` and the proxy currently
+surfaces a generic ``stream_incomplete`` 502 instead of the upstream
+``invalid_json_schema`` detail. We detect these violations locally so the
+client receives a 400 with the OpenAI-style error message before any
+upstream request is opened.
+
+Only the strict-mode constraints are enforced here. Schemas with
+``strict`` unset or ``strict: false`` pass through unchanged.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+
+from app.core.types import JsonValue
+from app.core.utils.json_guards import is_json_list, is_json_mapping
+
+
+@dataclass(frozen=True)
+class StrictSchemaError:
+    """A single strict-mode schema violation, formatted for the API."""
+
+    code: str
+    message: str
+    param: str
+
+
+def validate_strict_json_schema(
+    schema: JsonValue,
+    *,
+    name: str | None,
+    param: str,
+) -> StrictSchemaError | None:
+    """Return the first strict-mode violation in ``schema`` or ``None``.
+
+    The returned message mirrors the upstream OpenAI Responses API format:
+
+        Invalid schema for response_format '<name>':
+        In context=(<path>), <reason>.
+    """
+
+    label = name or "response_format"
+    violation = _find_violation(schema, ())
+    if violation is None:
+        return None
+    context_path, reason = violation
+    rendered_context = _render_context(context_path)
+    message = f"Invalid schema for response_format '{label}': In context={rendered_context}, {reason}."
+    return StrictSchemaError(
+        code="invalid_json_schema",
+        message=message,
+        param=param,
+    )
+
+
+def _render_context(path: tuple[str, ...]) -> str:
+    if not path:
+        return "()"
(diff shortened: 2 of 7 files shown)


Soju06:
Thanks for the catch — addressed in a28db5e.

Strict mode now rejects schemas where any key in `properties` is missing from `required`, including the empty-`required` case. The diagnostic mirrors the upstream OpenAI API message verbatim, verified against `api.openai.com` so callers see the same body regardless of which side rejects the request.

Added two tests covering missing keys and empty/missing `required` arrays.

@codex review

PR #496, opened 2026-04-25
fix: isolate concurrent browser OAuth flows

### LINKED ISSUE #472: Concurrent browser OAuth sessions: second login fails with "Invalid OAuth callback: state mismatch or   missing code"
## Summary
  Starting two browser-based OAuth "Add account" flows concurrently (two different accounts) causes one of them
  to fail with `Invalid OAuth callback: state mismatch or missing code.` The single shared `/auth/callback`
  server plus reliance on the system default browser makes the two flows interfere with each other.

  ## Environment
  - codex-lb image: `ghcr.io/soju06/codex-lb:1.14.1` (digest
  `sha256:1047c2b6e55595ab09b6710a31a5f31e007d29878040df7d1f6c4f6de0630ebf`)
  - Deployment: Docker (single container)
  - Host OS: Linux (Ubuntu-based, kernel 6.8)
  - Client(s) affected: Web dashboard users adding accounts via the browser OAuth flow from two sessions/tabs at
  the same time

  ## Steps to Reproduce
  1. Open the codex-lb dashboard in two separate browser sessions/tabs (or two operators).
  2. In tab A, click "Add account" and start the browser OAuth flow for account A.
  3. Before completing A, in tab B, click "Add account" and start the browser OAuth flow for account B.
  4. Complete the Chat/Codex login in tab A, then in tab B (or the opposite order).

  ## Expected Behavior
  Both sessions should complete independently. The shared callback server should route each callback to the
  correct pending session via the `state` parameter, and both accounts should be added.

  ## Actual Behavior
  One of the two sessions finishes with:
  Invalid OAuth callback: state mismatch or missing code.
  and the account is not added. The other session usually succeeds. Reproduces consistently when both flows are
  in-flight at the same time.

  ## Logs
  oauth_callback received state= code=
  oauth_callback received state= code=
  oauth_callback terminalized session= status=error msg="Invalid OAuth callback: state mismatch or missing code."

  (Anonymized; no API keys or account IDs included.)
  ## Impact
  Operators who manage multiple accounts cannot add them in parallel — they must serialize OAuth flows one at a
  time. For teams onboarding many accounts this significantly slows account import. Workaround: complete each
  browser OAuth flow fully before starting the next.

  ## Suspected Cause (guess)
  Two areas in `app/modules/oauth/service.py` look relevant — this is speculation, not confirmed:

  1. The browser flow uses a single shared callback server (`BrowserCallbackCoordinator` + `OAuthCallbackServer`
  on port `1455`) and a shared `_OAUTH_STORE`. While `_state_token_index` is keyed by `state_token` and should in
   principle support multiple concurrent pending sessions, if the OS-level default browser re-uses an existing
  tab/session, the PKCE `code_verifier` recorded on one session may be used against the wrong callback. Worth
  verifying whether `_handle_callback` is always reached once per started flow when two flows are active.
  2. Branch at `service.py:460-462` emits `"Invalid OAuth callback state."` when `code` or `verifier` is missing
  — but the user-facing message returned from `manual_callback` at lines 409-413 is `"Invalid OAuth callback:
  state mismatch or missing code."` for the same underlying condition (missing `code` or `verifier`), which masks
   the real cause in logs/UI. A narrower error message per condition would help triage.

  ## Additional Context
  - Observed on a remote deployment behind a public HTTPS domain, with OAuth redirect configured accordingly.
  - Device flow (`_start_device_flow`) is not affected; only the browser/manual callback path.
  - Both accounts being added are **different** emails, so this is not a same-email merge conflict.

stemirkhan:
## Summary

This fixes #472 by isolating dashboard OAuth state per flow instead of
storing browser PKCE state in a single global slot.

Previously, if two browser OAuth sessions were started at the same time
from different tabs or operators, the second flow would overwrite the
first flow's `state` and `code_verifier`. When the first callback later
arrived, it failed with:

`Invalid OAuth callback: state mismatch or missing code.`

## What changed

- Added per-flow OAuth state tracking in the backend
- Added a stable `flowId` to dashboard OAuth payloads
- Matched browser callbacks against the correct in-flight flow by `state`
- Kept the shared callback server alive until all pending browser flows finish
- Threaded `flowId` through the frontend OAuth hook and API client
- Added regression coverage for overlapping browser OAuth flows
- Added an OpenSpec change for concurrent browser OAuth flow isolation

## Root cause

The dashboard OAuth service stored browser flow state in one global slot.
Starting a second browser flow overwrote the first flow's PKCE verifier
and callback state, so the first callback could no longer be validated.

## Testing

Passed locally:
- `pytest tests/integration/test_oauth_flow.py -q`
- `ruff check app/modules/oauth/api.py app/modules/oauth/schemas.py app/modules/oauth/service.py tests/integration/test_oauth_flow.py`
- `ruff format --check app/modules/oauth/api.py app/modules/oauth/schemas.py app/modules/oauth/service.py tests/integration/test_oauth_flow.py`
- `npm exec eslint -- src/features/accounts/api.ts src/features/accounts/schemas.ts src/features/accounts/hooks/use-oauth.ts src/features/accounts/hooks/use-oauth.test.ts`

Blocked in local environment:
- `npm run typecheck` fails on a pre-existing error in `frontend/src/components/donut-chart.tsx`
- `npm run test -- src/features/accounts/hooks/use-oauth.test.ts` fails with a local `ERR_REQUIRE_ESM` / jsdom worker issue
- `openspec validate --specs` could not be run because the `openspec` CLI is not installed in this environment

Closes #472


Soju06:
Thanks @stemirkhan — really thorough fix and the per-flow `OAuthState` + `state_token` index design is exactly the right shape for this. The frontend `flowId` threading and the regression test against two concurrent browser flows are nice. CI is fully green and the diff is mergeable.

One blocker before I'd merge, plus a few smaller things to consider.

## Blocker — restore the manual-callback regression coverage from #481

This PR (intentionally or not) drops two integration tests that just landed in #481:

- `test_manual_callback_is_idempotent_for_same_attempt`
- `test_manual_callback_after_success_rejects_stale_callback`

The behavior they covered is still preserved on the code side — your new check

```python
if flow is not None and flow.status == "success" and state == flow.state_token:
    return ManualCallbackResponse(status="success")
```

is the per-flow equivalent of the global gate I added in #481, and the validation path further down (`if not code or not state or flow is None or not verifier:`) still rejects stale URLs. So the runtime behavior is fine.

But losing the two regression tests means the next refactor of `manual_callback` can quietly break either property without anyone noticing. Could you re-add them in the new shape, e.g.:

- `test_manual_callback_is_idempotent_for_same_attempt`: start a browser flow → submit the same callback URL twice with the same `flowId` → both return `success` and `exchange_authorization_code` is called exactly once.
- `test_manual_callback_after_success_rejects_stale_callback`: start a browser flow → succeed → submit a callback URL with a fabricated `state` (no flow has that state token) → returns the existing `Invalid OAuth callback: state mismatch or missing code.` error.

Both can be rewritten against your new model in maybe 25 lines each.

## Smaller items, mostly questions

1. **State-token uniqueness vs. the `state == flow.state_token` recheck (idempotent path).** Once you look the flow up via `get_flow_by_state_token_locked(state)`, the equality recheck is technically redundant — the flow is keyed on that state token. Not worth changing, but worth pointing out so a future reader doesn't reason about "two flows with the same state".

2. **`_state_token_index` cleanup after error/success.** I don't see anywhere that removes a flow from `_flows` / `_state_token_index` after it terminalizes, except in `OAuthStateStore.reset()`. In a long-lived process where users add many accounts over time, both maps grow without bound (small per-flow, but technically unbounded). Worth either:
   - removing the entry from both maps on terminal `_set_success` / `_set_error`, while still keeping the latest flow as `self._state` for `get_oauth_status` without `flow_id`, or
   - capping with a small TTL / LRU. Not a hard blocker since each entry is small, but noting it for cleanliness.

3. **Callback-server lifecycle on `OSError`.** The new flow stores the candidate `OAuthCallbackServer` into `self._store._callback_server` *before* `start()`. If `start()` raises `OSError`, we clear it under the lock — good. But the flow itself remains in `_flows` as `pending` even though no callback server is listening. That's the same behavior as before, so not a regression — just calling it out, since it can produce flows that never get a real callback if a port collision happens. The device-flow fallback still kicks in via `_start_browser_flow_or_device`, so end-user impact is minimal.

4. **`oauth_status(flow_id)` falls back to `self._store.state` when the flow is missing.** That's the right call for backwards compat with frontends that don't yet send `flowId`, but in the post-fix world where every browser flow has a `flowId`, sending an unknown `flowId` returning the latest flow's status could mislead the frontend. Maybe consider returning `pending` (or an explicit `flow_not_found`) for unknown `flow_id` once the schema change has had a release or two to settle. Not a blocker — fine to leave the compat behavior in this PR.

## Local CI notes you mentioned

- `npm run typecheck` red on `frontend/src/components/donut-chart.tsx` — that's a pre-existing main breakage, not from this PR. Leaving it for a separate fix.
- `openspec validate --specs` not runnable locally — fine, your `openspec/changes/...` files look well-formed; CI's spec validation passed.

- The `_stop_callback_server_if_idle()` rename + `has_pending_browser_flows_locked()` guard is exactly the right idea.
- Threading `flowId` through `useOauth` and the API client cleanly is appreciated.

Just re-add the two regression tests from #481 in the new per-flow shape and I'll merge. Thanks again 🙏

Soju06:
@codex review

Soju06:
Quick follow-up — codex review just landed and flagged two things on top of the test-coverage point I raised earlier. Both are real and worth folding into the same revision:

## P1 — `/api/oauth/status` doesn't accept `flowId` due to missing alias

The frontend (`frontend/src/features/accounts/api.ts`) sends the camelCase `flowId` query param, but `oauth_status()` declares `flow_id: str | None = None` without a query alias:

```python
# app/modules/oauth/api.py
async def oauth_status(
    flow_id: str | None = None,
    context: OauthContext = Depends(get_oauth_context),
) -> OauthStatusResponse | JSONResponse:
```

FastAPI won't bind `flowId` to `flow_id` without an explicit `Query(default=None, alias="flowId")`, so the parameter silently stays `None` and the service falls back to `self._store.state` (the latest flow). Net effect: status polling for two concurrent flows shows the same status — exactly the isolation this PR is designed to provide.

Suggested fix:

```python
from fastapi import Query

async def oauth_status(
    flow_id: str | None = Query(default=None, alias="flowId"),
    context: OauthContext = Depends(get_oauth_context),
) -> OauthStatusResponse | JSONResponse:
```

It would be worth adding a regression test that polls `/api/oauth/status?flowId=...` for the *first* of two concurrent flows after the *second* terminalizes, asserting the first still returns `pending` (not the second's terminal state).

## P2 — manual-callback idempotent fast-path ignores caller's `flow_id`

In `manual_callback`, the idempotent success path matches purely by `state` token:

```python
flow = self._store.get_flow_by_state_token_locked(state)
...
if flow is not None and flow.status == "success" and state == flow.state_token:
    return ManualCallbackResponse(status="success")
```

If a user finishes flow A, then starts flow B and accidentally submits flow A's old callback URL into flow B's manual-callback dialog, this returns success for flow B even though flow B's token exchange never happens. The caller-provided `flow_id` should be honored when present:

```python
if (
    flow is not None
    and flow.status == "success"
    and state == flow.state_token
    and (flow_id is None or flow_id == flow.flow_id)
):
    return ManualCallbackResponse(status="success")
```

When `flow_id` doesn't match, fall through to the regular validation path (which will produce `Invalid OAuth callback: state mismatch or missing code.` because flow B's verifier won't match). That's the same shape as the stale-callback regression I asked you to re-add — a test for it would land on this naturally.

## Recap

Combining with my earlier comment, the requested changes for this revision are:

1. **Re-add** the manual-callback regression tests dropped from #481 in the new per-flow shape (idempotent same-attempt + stale-callback rejection).
2. **P1 (codex)**: add the `alias="flowId"` () on `oauth_status` and ideally a regression test for status isolation.
3. **P2 (codex)**: tighten the idempotent fast-path to also respect the caller's `flow_id` when provided.

Smaller items I mentioned (`_state_token_index` cleanup, `oauth_status` fallback semantics for unknown flow_id) are nice-to-have, not blockers.

Thanks again for the thorough fix — close to ready 🙏

Soju06:
Hi @stemirkhan -- friendly reminder that this PR is ready to land except for two findings codex review left on `a6a0bc8` back on April 27. Re-pinging in case the notifications got buried. The fix shape itself is correct and the OpenSpec change is good; we just want both flow-isolation edge cases closed before the gate goes in.

## P1 -- `/api/oauth/status` does not bind `flowId`

The frontend hook now sends `GET /api/oauth/status?flowId=<id>`, but `app/modules/oauth/api.py` declares the parameter as `flow_id` without a query alias. FastAPI matches query parameters by exact name, so `flowId` is dropped and `oauth_status` falls back to whatever `flow_id=None` resolves to -- in practice, the last-written global state.

That defeats the entire isolation goal of this PR: two concurrent tabs end up polling the same status surface again, just via the new endpoint instead of the old global slot. The frontend will report a misleading status for one of the flows.

Minimal fix:

```python
# app/modules/oauth/api.py
from fastapi import Query

async def oauth_status(
    flow_id: str | None = Query(default=None, alias="flowId"),
    ...
):
    ...
```

Or, alternatively, change the frontend caller to use `flow_id` snake_case (less idiomatic for JS but consistent with the rest of FastAPI's query surface here). I'd lean toward the alias on the backend side -- it keeps the public JS-style camelCase that the schema already uses for `flowId`.

Test (existing `test_oauth_flow.py` covers the body case; we want one for the query path too):

```python
async def test_oauth_status_resolves_camelcase_flow_id_query(client):
    flow_id = await _start_browser_flow(client)
    resp = await client.get(f"/api/oauth/status?flowId={flow_id}")
    assert resp.status_code == 200
    assert resp.json()["flow_id"] == flow_id
```

## P2 -- idempotent success fast-path ignores the requested flow

`app/modules/oauth/service.py` (around line 319) has an idempotent fast-path that returns success for any callback whose `state` matches an already-successful flow, regardless of `flow_id`. So this sequence misreports success:

1. tab A starts flow F_A, completes it, success persisted with `state=S_A`
2. tab B starts a new flow F_B, fresh `state=S_B`
3. tab B's user accidentally submits the **old** callback URL for F_A (e.g. they pasted it from history)
4. fast-path sees `state=S_A` matches a persisted success and returns success for F_B without doing the token exchange

The user sees "OAuth completed" in tab B, but no new token was actually exchanged for F_B -- the account row F_B was attached to is still missing credentials. We've seen the symptoms of this class of bug before (#481-adjacent territory) where a UI-level success badge doesn't match what the proxy actually stored.

Minimal fix: gate the fast-path on `flow_id` matching too. Something like:

```python
def _fast_path_already_successful(state: str, requested_flow_id: str | None) -> bool:
    persisted = self._completed_by_state.get(state)
    if persisted is None:
        return False
    if requested_flow_id is None:
        return True  # legacy clients without flowId still pass through
    return persisted.flow_id == requested_flow_id
```

Test:

```python
async def test_oauth_callback_idempotent_fast_path_rejects_mismatched_flow(client):
    # flow A completes
    flow_a = await _complete_flow(client, ...)
    # flow B starts but submits flow A's callback
    flow_b = await _start_browser_flow(client)
    resp = await client.post("/api/oauth/manual-callback", json={
        "flow_id": flow_b,
        "state": flow_a.state,
        "code": flow_a.code,
    })
    assert resp.status_code == 400
    assert "flow" in resp.json()["detail"].lower()
```

## Nothing else blocking

The rest of the patch is in good shape -- per-flow state tracking, the shared callback-server stay-alive, the frontend hook thread-through, and the OpenSpec change all read cleanly. After these two are in I'll re-run `@codex review` and we can land it.

If you can't get back to this in the next few days, ping me here and I'll prepare a follow-up commit on a branch and link it as an alternative, just so the fix doesn't sit indefinitely (this is a real concurrent-OAuth bug and worth getting in before the next release).

Komzpa:
@codex review

Soju06:

Komzpa:
Revived on a maintainer-owned branch as #753 because this PR head has maintainerCanModify=false and cannot be updated here. Keeping credit to the original implementation in the replacement PR.

PR #508, opened 2026-04-28
feat(frontend): add GitHub link to status bar

aruis:
## Summary
- add a right-aligned GitHub icon link to the dashboard status bar
- cover the status bar link target and external-link attributes
- make frontend test setup tolerate environments where localStorage lacks standard Storage methods

## Tests
- bun run lint
- bun run typecheck
- bun run test:coverage
- bun run build

### DIFF
--- a/frontend/src/components/layout/status-bar.test.tsx
+++ b/frontend/src/components/layout/status-bar.test.tsx
@@ -1,0 +1,33 @@
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import { render, screen } from "@testing-library/react";
+import { describe, expect, it } from "vitest";
+
+import { StatusBar } from "@/components/layout/status-bar";
+
+function renderStatusBar() {
+  const queryClient = new QueryClient({
+    defaultOptions: {
+      queries: {
+        retry: false,
+      },
+    },
+  });
+
+  return render(
+    <QueryClientProvider client={queryClient}>
+      <StatusBar />
+    </QueryClientProvider>,
+  );
+}
+
+describe("StatusBar", () => {
+  it("links to the official GitHub repository", () => {
+    renderStatusBar();
+
+    const link = screen.getByRole("link", { name: "Open official GitHub repository" });
+
+    expect(link).toHaveAttribute("href", "https://github.com/soju06/codex-lb");
+    expect(link).toHaveAttribute("target", "_blank");
+    expect(link).toHaveAttribute("rel", "noreferrer");
+  });
+});
--- a/frontend/src/components/layout/status-bar.tsx
+++ b/frontend/src/components/layout/status-bar.tsx
@@ -1,11 +1,13 @@
 import { useEffect, useState } from "react";
-import { Activity, ArrowRightLeft, Tag } from "lucide-react";
+import { Activity, ArrowRightLeft, Github, Tag } from "lucide-react";
 import { useQuery } from "@tanstack/react-query";
 
 import { getDashboardOverview } from "@/features/dashboard/api";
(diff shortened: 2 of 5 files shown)


Soju06:
@codex review

Soju06:
Thanks @aruis -- the GitHub status-bar link is small, clean, and the test asserts the right three attributes (href / target / rel). Codex review flagged one P1 in the new test-setup shim that's worth fixing before merging.

## P1 -- guard localStorage access in the shim

The new feature-detection in `frontend/src/test/setup.ts` dereferences `window.localStorage.clear` directly:

```ts
// frontend/src/test/setup.ts:~53
if (typeof window.localStorage.clear !== "function") {
  // install shim
}
```

This breaks in two real environments the shim is supposed to handle:

1. `window.localStorage` is `undefined` (some sandboxed runners) -> `Cannot read properties of undefined (reading 'clear')`.
2. The `localStorage` getter throws `SecurityError` (browser private mode, sandboxed iframes, certain CI runners) -> the access itself throws.

In both cases the test setup raises at import time and the entire suite aborts instead of falling back to the mock the shim is meant to install.

## Suggested fix

Either optional-chain + nullish check, or wrap in `try/catch` so the shim install path always runs in degraded environments:

```ts
function ensureLocalStorageShim() {
  let storage: Storage | undefined;
  try {
    storage = window.localStorage;
  } catch {
    storage = undefined;
  }

  if (!storage || typeof storage.clear !== "function") {
    // existing shim install: Object.defineProperty(window, "localStorage", { value: shim, ... })
    installShim();
  }
}
```

A tiny vitest regression to pin the behavior:

```ts
import { describe, expect, it, vi } from "vitest";

describe("localStorage shim", () => {
  it("installs even when window.localStorage getter throws", () => {
    const original = Object.getOwnPropertyDescriptor(window, "localStorage");
    Object.defineProperty(window, "localStorage", {
      configurable: true,
      get() {
        throw new DOMException("sandboxed", "SecurityError");
      },
    });
    try {
      expect(() => ensureLocalStorageShim()).not.toThrow();
      expect(typeof window.localStorage.clear).toBe("function");
    } finally {
      if (original) Object.defineProperty(window, "localStorage", original);
    }
  });

  it("installs when window.localStorage is undefined", () => {
    const original = Object.getOwnPropertyDescriptor(window, "localStorage");
    Object.defineProperty(window, "localStorage", { configurable: true, value: undefined });
    try {
      expect(() => ensureLocalStorageShim()).not.toThrow();
      expect(typeof window.localStorage.clear).toBe("function");
    } finally {
      if (original) Object.defineProperty(window, "localStorage", original);
    }
  });
});
```

## Not blocking otherwise

The status-bar change is a clean, well-tested UI addition (right-aligned GitHub icon, `target="_blank"` + `rel="noreferrer"`, accessible link name).

Once the shim guard lands I'll re-run `@codex review` and queue for merge.

aruis:
Thanks for the review. I addressed the P1 by moving the localStorage fallback into an exported ensureLocalStorageShim() helper that safely reads window.localStorage inside try/catch and installs the mock when storage is missing or inaccessible.

I also added regression coverage for both cases:
- window.localStorage is undefined
- reading window.localStorage throws SecurityError

Local checks run:
- bun run test src/test/local-storage-shim.test.ts src/hooks/use-time-format.test.ts src/features/settings/components/appearance-settings.test.tsx src/components/layout/status-bar.test.tsx
- bun run typecheck
- bun run lint

aruis:
Hi @Soju06, just checking in. The localStorage shim feedback has been addressed and CI is green. Let me know if anything else is needed.

Komzpa:
@codex review

PR #465, opened 2026-04-22
feat(auth): make dashboard session lifetime configurable

mikabytes:
## Summary

  Make dashboard password session lifetime configurable instead of hard-coding it to 12 hours.

  This change:
  - adds persisted `dashboard_session_ttl_seconds` to dashboard settings
  - uses the configured TTL for both encrypted dashboard session expiry and cookie `Max-Age`
  - adds a Settings UI control for dashboard session lifetime
  - warns in the UI when the configured lifetime exceeds 30 days
  - allows longer lifetimes when operators intentionally choose them
  - adds/update OpenSpec change artifacts for the behavior

  
image

  ## Behavior Notes

  - Existing sessions keep their original embedded expiry.
  - Newly issued password-backed dashboard sessions use the configured lifetime.
  - The UI warns when the configured lifetime exceeds 30 days, but does not block it.
  
image

  ## Validation

  - Inspected the cookie in the browser and verified that it expires as expected after logout/login

  Ran:
  - `uv run pytest -q tests/unit/test_dashboard_auth_api.py tests/integration/test_settings_api.py tests/unit/
  test_dashboard_auth_password_service.py`
  - `uv run pytest -q tests/integration/test_migrations.py`
  - `cd frontend && bun run test src/features/settings/schemas.test.ts src/features/settings/hooks/use-settings.test.ts src/features/
  settings/components/totp-settings.test.tsx src/features/settings/components/session-settings.test.tsx src/features/settings/components/
  routing-settings.test.tsx`
  - `cd frontend && bun run build`
  - `openspec validate --specs`

  ## OpenSpec

  Change:
  - `openspec/changes/configure-dashboard-session-lifetime/`

PR #511, opened 2026-04-29
fix(auth): preserve existing session expiry through TOTP and tighten hour input

Soju06:
## Summary

Follow-up to #465. Addresses two codex review findings on the dashboard session lifetime feature.

## verify_totp now inherits the existing password session's expiry

`verify_totp` was loading the current `dashboard_session_ttl_seconds` setting and minting a brand-new session with that TTL. If an operator changed the setting between password login and TOTP submission, the follow-up TOTP-verified session would silently use the new TTL, contradicting the documented promise that "existing sessions keep their original embedded expiry".

This change makes `verify_totp` look up the live state of the incoming password session, compute its remaining lifetime against the current clock, and use that for the new TOTP-verified session instead of the caller-provided `ttl_seconds`. The supplied `ttl_seconds` is only used as a fallback when no live session state is decodable. The API layer now also sets the session cookie's `Max-Age` to the actually-applied TTL so the cookie expiry never advertises a longer lifetime than the encrypted session payload itself.

`verify_totp` now returns `tuple[str, int]` (session id + applied TTL) to make the applied lifetime explicit at the API boundary. The single call site in `app/modules/dashboard_auth/api.py` is updated.

## Hour input no longer silently floors decimals

The Settings UI parsed the dashboard session lifetime input with `Number.parseInt("1.5", 10)` which silently truncated to `1`. Combined with the backend accepting any `>= 3600` second TTL (e.g. 5400s = 1.5h), this could:

- accept a decimal entry and save it as a smaller integer hour value
- mis-render an existing non-hour TTL by rounding down on first paint and immediately treating the field as "changed"

The control now:

- requires a strict integer-hours pattern (`/^\d+$/`) before treating input as valid
- shows a non-blocking inline error for non-integer entries
- formats stored sub-hour TTLs as decimals (`5400s -> "1.50"`) and disables save until the operator types a whole-hour value, so we never overwrite a non-aligned setting just by clicking save

## Tests

- `tests/unit/test_dashboard_auth_password_service.py` — two new tests covering inherited expiry on success and the fallback path when no live session state is available
- `frontend/src/features/settings/components/session-settings.test.tsx` — decimal-rejection test and an existing non-hour TTL render test

## Drive-by

Removed an unused `# type: ignore[attr-defined]` in `tests/unit/test_proxy_utils.py` that ty has been flagging in CI as `unused-type-ignore-comment` since the symbol now resolves cleanly.

## Local verification

- `uv run ruff check app/modules/dashboard_auth/ tests/unit/test_dashboard_auth_password_service.py` — clean
- `uv run ruff format --check app/modules/dashboard_auth/ tests/unit/test_dashboard_auth_password_service.py` — clean
- `uv run ty check` — `All checks passed!`
- `uv run pytest tests/unit/test_dashboard_auth_password_service.py tests/unit/test_dashboard_session_store.py tests/unit/test_dashboard_auth_api.py tests/unit/test_proxy_utils.py -q` — `223 passed`
- Frontend `vitest` was not run locally due to a pre-existing `ERR_REQUIRE_ESM` issue noted on #465; CI will exercise it.

### DIFF
--- a/app/modules/dashboard_auth/api.py
+++ b/app/modules/dashboard_auth/api.py
@@ -412,7 +412,7 @@ async def verify_totp(
         ) from exc
     try:
         session_ttl_seconds = (await get_settings_cache().get()).dashboard_session_ttl_seconds
-        session_id = await context.service.verify_totp(
+        session_id, applied_ttl_seconds = await context.service.verify_totp(
             session_id=current_session_id,
             code=payload.code,
             ttl_seconds=session_ttl_seconds,
@@ -432,7 +432,7 @@ async def verify_totp(
         password_session_id=session_id,
     )
     json_response = JSONResponse(status_code=200, content=response.model_dump(by_alias=True))
-    _set_session_cookie(json_response, session_id, request, max_age_seconds=session_ttl_seconds)
+    _set_session_cookie(json_response, session_id, request, max_age_seconds=applied_ttl_seconds)
     return json_response
 
 
--- a/app/modules/dashboard_auth/service.py
+++ b/app/modules/dashboard_auth/service.py
@@ -194,13 +194,19 @@ class DashboardAuthService:
         await self._repository.clear_password_and_totp()
 
     async def _require_active_password_session(self, session_id: str | None) -> DashboardAuthSettingsProtocol:
+        settings, _state = await self._require_active_password_session_with_state(session_id)
+        return settings
+
+    async def _require_active_password_session_with_state(
+        self, session_id: str | None
+    ) -> tuple[DashboardAuthSettingsProtocol, DashboardSessionState]:
         settings = await self._repository.get_settings()
         if settings.password_hash is None:
             raise PasswordSessionRequiredError("Password-authenticated session is required")
         session = self._session_store.get(session_id)
         if session is None or not session.password_verified:
             raise PasswordSessionRequiredError("Password-authenticated session is required")
-        return settings
+        return settings, session
 
     async def _require_totp_verified_session(self, session_id: str | None) -> DashboardAuthSettingsProtocol:
         settings = await self._require_active_password_session(session_id)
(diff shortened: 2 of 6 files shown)


Soju06:
@codex review

Soju06:
Pushed `6f0167b` to address codex P2 (`Reuse verified session state when deriving inherited TOTP TTL`).

The race was real: `verify_totp` called `session_store.get(session_id)` twice — once inside `_require_active_password_session` and once in the inherited-TTL branch. If the original password session expired between those two calls (or if any caller-injected `get` returned a different result), the inherited-TTL branch fell through to the `ttl_seconds` fallback and minted a fresh full-length session.

## Fix

- Add `_require_active_password_session_with_state` returning `(settings, DashboardSessionState)` from a single `store.get()`.
- `verify_totp` uses that snapshot to compute `inherited_ttl`, removing the second `get()` entirely. The `ttl_seconds` parameter no longer takes the inherited path; it is now an upper bound the API layer can pass for symmetry, but the actual applied TTL is always the snapshot's remaining lifetime.
- The fallback branch is gone — there is no way for the state to be missing after the active-session check, by construction.

## Tests

Replaced the `fallback` test with `test_verify_totp_does_not_call_session_store_get_twice` which pins the single-lookup contract: `store.get` is called exactly once with the incoming `session_id`, and the applied TTL is the snapshot's remaining lifetime.

## Local verification

- `uv run ruff format --check app/modules/dashboard_auth/ tests/unit/test_dashboard_auth_password_service.py` — clean
- `uv run ruff check app/modules/dashboard_auth/ tests/unit/test_dashboard_auth_password_service.py` — clean
- `uv run ty check` — `All checks passed!`
- `uv run pytest tests/unit/test_dashboard_auth_password_service.py tests/unit/test_dashboard_session_store.py tests/unit/test_dashboard_auth_api.py -q` — `12 passed`

@codex review

Soju06:
@all-contributors please add @Soju06 for code, test

PR #514, opened 2026-04-29
fix(proxy): include STICKY_THREAD in sticky budget-pressure guard

jgorostegui:
## Summary

This fixes a sticky-thread routing edge case where a thread can stay pinned to an `ACTIVE` account that is already over the sticky reallocation budget threshold, even when another account still has capacity.

The issue is that the budget-pressure guard covered `PROMPT_CACHE` and `CODEX_SESSION`, but not `STICKY_THREAD`. Adding `STICKY_THREAD` lets dashboard sticky threads use the existing rebind path instead of waiting for the pinned account to fail or be paused manually.

## Changes

- Add `StickySessionKind.STICKY_THREAD` to the sticky budget-pressure guard.
- Keep the existing no-thrash behavior when every candidate is also above the threshold.
- Add OpenSpec coverage for sticky-thread budget-pressure behavior.
- Add focused unit tests for rebind, all-candidates-pressured, and below-threshold cases.

## Tested

- `uv run pytest tests/unit/test_select_with_stickiness.py -q` (`33 passed`)
- `git diff --cached --check`

Not run locally: `openspec validate --specs`, because the `openspec` CLI is not installed in this environment.


Komzpa:
@codex review

Soju06:
@all-contributors please add @jgorostegui for code, test

PR #515, opened 2026-04-29
feat(proxy): add /backend-api/files upload protocol

### LINKED ISSUE #510: Payload too large: 413 error
The thread included multiple screenshots/images and ongoing coding work. The payload reached 19,625,039 bytes, exceeding the websocket limit of 15,728,640 bytes.

Image

Soju06:
## Summary

Implements the upstream Codex CLI three-step file upload protocol so clients can attach large files (e.g. images, PDFs) without hitting the 16 MiB websocket message ceiling on `/responses`.

Fixes #510.

## What

- **`POST /backend-api/files`** — registers a file with `{file_name, file_size, use_case}` and returns the upstream `{file_id, upload_url}`. The `upload_url` is an Azure Blob SAS link; clients PUT the bytes directly to it (the 512 MiB upstream cap is enforced at the codex-lb edge by `FileCreateRequest`).
- **`POST /backend-api/files/{file_id}/uploaded`** — finalizes; the upstream client polls while `status == "retry"`. The codex-lb client mirrors that loop server-side with a 30 s budget / 250 ms retry delay (matching `codex-rs/codex-api/src/files.rs::upload_local_file`).
- Files registered this way can be referenced from `/responses` prompts as `{"type": "input_file", "file_id": "..."}` instead of being inlined as base64. This bypasses the per-message 16 MiB ceiling.
- **Removed the existing `input_file.file_id is not supported` validation** in `ResponsesRequest._validate_input_type` and `ResponsesCompactRequest._validate_input_type` (4 sites). `_has_input_file_id` / `_is_input_file_with_id` helpers are kept for diagnostics.

## Why

Issue #510 reports `Payload too large: 413` after a long conversation accumulates several screenshots inline. I verified empirically against my dev codex-lb (with `CODEX_LB_UPSTREAM_RESPONSE_CREATE_MAX_BYTES=33554432`) that:

- 14 MiB / 15.5 MiB payloads → upstream accepts and processes
- **17 MiB+ payloads → upstream WebSocket closes with 1009 (message too big)**

So the 16 MiB ceiling is upstream-side; raising codex-lb's local cap any further is wasted. The actual fix is to give clients a separate path for large attachments, which is the protocol upstream Codex CLI already implements.

## Scope

- Backend client + service + API (3 endpoints, account-selection / freshness / 401-retry plumbing reused from `transcribe`).
- Pydantic edge validation (`file_size > 0` and `<= 512 * 1024 * 1024`).
- Logging into request-logs as synthetic models `files-create` and `files-finalize` for dashboard filtering.
- Tests (9 unit + 11 integration + 3 regression unit + 1 converted integration = 24 changes, all green).
- OpenSpec change notes under `openspec/changes/add-backend-api-files-protocol/`.

## Out of scope

- Dashboard "files" view UI.
- Image-resize-before-upload (codex CLI handles client-side via `load_for_prompt_bytes`).
- Persisted file metadata in codex-lb's DB (the upstream contract is stateless from our perspective; we just forward).

## Verification

- `uv run ruff check app tests` — clean
- `uv run ruff format --check app tests` — 426 files already formatted
- `uv run ty check app/...` (all changed modules) — `All checks passed!`
- `pytest tests/unit -q` — **1472 passed, 3 skipped**
- `pytest tests/integration/test_proxy_*.py tests/integration/test_openai_compat_features.py -q` — **148 passed**

## Notes

- `FileCreateRequest` uses `extra="allow"` so unknown upstream-side fields (if the contract grows) are forwarded verbatim.
- The finalize endpoint surfaces `status: "retry"` to the client when our 30 s budget is exhausted. Direct Codex CLI callers will keep polling on their end (they have their own loop), so we don't synthesize a 5xx — same shape as `codex-rs/codex-api/src/files.rs`.


Soju06:
@codex review

Soju06:
@codex review

Soju06:
Pushed `5f0be47` addressing both codex review findings:

**P1 — finalize polling overshoot**

`finalize_file` previously only checked the deadline *before* sleeping, so the inter-poll sleep crossing the budget boundary still let one more `POST /files/{id}/uploaded` go out — and that extra call could block on its own request timeout well past the documented 30 s finalize budget.

Re-check the deadline immediately after the sleep so we never overshoot:

```python
await asyncio.sleep(_FILE_FINALIZE_POLL_DELAY_SECONDS)
if time.monotonic() >= deadline:
    return parsed
```

Updated regression: `test_finalize_file_returns_last_retry_after_budget_exhaustion` now asserts exactly one upstream call (the initial poll), not two.

**P2 — file upstream calls now respect remaining request budget**

Added `push_files_timeout_overrides` / `pop_files_timeout_overrides` (mirroring the transcribe path's contextvars) and `_proxy_files_call` now pushes the remaining per-request budget into them around the upstream invocation:

```python
timeout_tokens = push_files_timeout_overrides(
    connect_timeout_seconds=remaining_budget,
    total_timeout_seconds=remaining_budget,
)
```

`finalize_file` also clamps its 30 s poll budget to the effective per-poll total so the poll loop cannot outlive the parent request when the caller has less than 30 s left.

**Verification**

- `uv run ruff check / format --check` clean on touched files
- `uv run ty check` clean on touched files
- `pytest tests/unit/test_files_client.py tests/integration/test_proxy_files.py` — 20 passed

@codex review

Soju06:
Pushed `49c0b62` for the new P1 codex flagged on `5f0be47` (`Derive finalize poll timeout from remaining deadline`).

The previous fix bounded the *budget* but the per-poll `aiohttp.ClientTimeout` was still computed once before the loop and reused for every retry. A late retry could therefore start with the full original timeout while only a few hundred ms remained until the 30 s finalize deadline, blowing past both budgets on slow networks.

Now the per-poll timeout is recomputed from `deadline - time.monotonic()` at the top of each iteration:

```python
remaining = deadline - time.monotonic()
if remaining <= 0:
    return parsed
per_poll_total = min(effective_per_poll_total, remaining)
timeout = aiohttp.ClientTimeout(
    total=per_poll_total,
    sock_connect=min(effective_connect, per_poll_total),
)
```

`parsed` is seeded to `{"status": "retry"}` so the very-first-iteration deadline-already-past path returns a deterministic shape.

The other two line comments on `5f0be47` are codex restating its earlier P1 / P2 (those were already addressed in the previous push and the textual diff hasn't changed in those regions); no new action needed there.

@codex review

Soju06:
Pushed `5f609a6` for the new P1 codex flagged on `49c0b62` (`Enforce API-key limits before forwarding file endpoints`).

Real bypass: `/backend-api/files` and `/backend-api/files/{file_id}/uploaded` skipped `_enforce_request_limits`, so an API key could register and finalize unlimited upstream files without local rate-limit accounting. Other billable routes (`_transcribe_request`, `_responses`, `_compact_responses`) all reserve before forwarding and release in a `finally`.

Now both file endpoints reserve before forwarding, using synthetic models `files-create` / `files-finalize` to match the request-log entries written from `ProxyService._proxy_files_call`:

```python
reservation = await _enforce_request_limits(
    api_key,
    request_model=_FILES_CREATE_LIMIT_MODEL,  # "files-create"
    request_service_tier=None,
)
try:
    ...
finally:
    await _release_reservation(reservation)
```

The other two line comments on `49c0b62` are stale repeats of the P1/P2 already addressed in `5f0be47` and `49c0b62`; no new action there.

@codex review

Soju06:
Pushed `4527d66` for the new P1 codex flagged on `5f609a6` (`Preserve account affinity across file protocol calls`).

The upstream file protocol is account-scoped (`chatgpt-account-id` header), so a finalize on a different account from create would fail with not-found / 401, and a follow-up `/responses` referencing that `file_id` would behave the same. With multiple accounts and capacity-weighted routing this is observable.

**Fix shape (in-memory pin):**

- `ProxyService` gains `_file_account_pins: dict[str, tuple[str, float]]` (file_id -> (account_id, expires_at)) with a 5-minute TTL and opportunistic eviction.
- `create_file`: on success, record `file_id -> account_id` from the upstream response.
- `finalize_file`: look the pin up first and pass it as `preferred_account_id` to `_select_account_with_budget_compatible` -- the load balancer already honors that field as a sticky-style preference.
- `_proxy_files_call` now returns `(result, account_id)` instead of just `result` so the pin is recorded without smuggling fields through the JSON body.
- Cross-instance handoff is best-effort: if a finalize lands on a different replica with no pin, we fall back to a fresh selection. The upstream surfaces that as not-found, same shape as before.

**Regression test**: `test_backend_files_finalize_pins_to_create_account` imports two accounts, drives create -> finalize, and asserts both calls used the same upstream account_id.

**Verification**:
- ruff/format/ty clean on touched files
- `pytest tests/integration/test_proxy_files.py tests/integration/test_proxy_transcriptions.py tests/integration/test_proxy_responses.py tests/unit/test_files_client.py tests/unit/test_openai_requests.py` -- 139 passed

The other line comments on `5f609a6` are stale repeats of the earlier P1/P2 already addressed in `5f0be47` / `49c0b62` / `5f609a6`.

@codex review

Soju06:
Pushed `ce76d87` for the new P1 codex flagged on `4527d66` (`Route file_id responses to the creating account`).

The previous round wired the file_id pin into create/finalize, but `/responses` flows still ran fresh load-balancer selections. Since the upstream file API is account-scoped (`chatgpt-account-id`), a `/v1/responses` request referencing an `input_file.file_id` from one account that landed on a different upstream account would fail with not-found / 401.

**What changed:**

- New helper `extract_input_file_ids(input_value)` in `app/core/openai/requests.py` -- walks top-level items + nested role-message `content` parts and returns the set of referenced file_ids.
- New `ProxyService._resolve_file_account_for_responses(payload, headers)` -- looks up `_file_account_pins`, but only when no stronger affinity is set:
  - explicit `prompt_cache_key`
  - session header / turn_state header (codex_session)
  - `previous_response_id`

  Multi-file_id ties prefer the most-recently-pinned with a lexicographic tie-break for determinism.
- Wired into three `/responses` entry points:
  - `_stream_with_retry` (HTTP `/v1/responses` + `/backend-api/codex/responses`)
  - `_stream_via_http_bridge` (HTTP bridge variant)
  - `compact_responses` (`/responses/compact`)

  The websocket path is intentionally not wired -- Codex CLI's websocket flow already uses `codex_session_affinity` (turn_state header) which is a stronger signal that overrides the file_id pin anyway, and the websocket request payload is consumed as raw text rather than the parsed `ResponsesRequest` model the helper expects.

**Tests:**

- Unit (`tests/unit/test_openai_requests.py`): `extract_input_file_ids` covers string input, top-level + nested file_ids, dedup / blank / missing ids, and unrelated content types.
- Integration (`tests/integration/test_proxy_files.py`):
  - `test_v1_responses_with_input_file_id_pins_to_creating_account` -- imports two accounts, asserts the `/v1/responses` request with `input_file.file_id` lands on the creating account.
  - `test_v1_responses_prompt_cache_key_overrides_file_id_pin` -- verifies precedence: prompt_cache_key keeps its existing routing.

**OpenSpec:** new `Responses requests with input_file.file_id route to the upload's account` requirement + two scenarios under `add-backend-api-files-protocol/specs/responses-api-compat/spec.md`.

**Verification:**

- `uv run ruff check app tests` -- clean
- `uv run ruff format --check` -- clean
- `uv run ty check` -- clean on touched modules
- `pytest tests/integration/test_proxy_files.py tests/integration/test_proxy_responses.py tests/integration/test_openai_compat_features.py tests/integration/test_proxy_compact.py tests/unit/test_files_client.py tests/unit/test_openai_requests.py` -- **182 passed**

@codex review

Soju06:
Pushed `30ab726` to wire the file_id pin into the websocket path too -- you're right that upload-then-converse on the same turn is exactly the case where file_id-only affinity matters, and that's the dominant Codex CLI / Desktop flow.

**What changed**:

`_prepare_websocket_response_create_request` already parses the inbound `response.create` into a `ResponsesRequest` model and computes the affinity policy. After that block, look up the file_id pin and seed `request_state.preferred_account_id` when the helper resolves to an account:

```python
if request_state.preferred_account_id is None:
    request_state.preferred_account_id = await self._resolve_file_account_for_responses(\n        responses_payload, headers,\n    )
```

Same precedence rule as the HTTP / compact paths -- the helper short-circuits to `None` when stronger affinity (prompt_cache_key, session header, turn_state header, previous_response_id) is present, so existing routing wins.

**Test rewrite**: the previous `/v1/responses`-via-route regression was flaky because the capacity-weighted load balancer chooses non-deterministically and pin state leaks across tests in the same pytest process. Replaced with a helper-level test that:

- Pins `file_response_pin -> acc_resp_a` directly.
- Asserts the resolver returns `acc_resp_a` for a payload referencing it.
- Drives `_prepare_websocket_response_create_request` and asserts it threads the same pin into `request_state.preferred_account_id`.

That isolates the contract under verification and is deterministic.

**Verification**:
- ruff check / format -- clean
- ty check on touched modules -- clean
- `pytest tests/integration/test_proxy_files.py tests/integration/test_proxy_websocket_responses.py tests/integration/test_proxy_responses.py tests/integration/test_proxy_compact.py tests/integration/test_openai_compat_features.py tests/unit/test_files_client.py tests/unit/test_openai_requests.py` -- **218 passed**

@codex review

Soju06:
Pushed `b33bf87` for the new P1 codex flagged on `30ab726` (`Distinguish explicit and derived prompt cache keys`).

Real bug -- not just stale review noise:

`_sticky_key_for_responses_request` / `_sticky_key_for_compact_request` can *derive* a `prompt_cache_key` and assign it onto the payload when openai cache affinity is enabled. With the default config (cache affinity on, derivation on), `_prompt_cache_key_from_request_model(payload) is not None` was effectively always true even when the client did **not** send a cache key. So my precedence check was unconditionally short-circuiting and file-pin routing was silently skipped on the very first turn -- the upload-then-converse scenario you raised earlier.

**Fix**: use Pydantic's `model_fields_set` to honor only client-supplied cache keys.

```python
explicit_fields = getattr(payload, "model_fields_set", set())
if (
    "prompt_cache_key" in explicit_fields
    and _prompt_cache_key_from_request_model(payload) is not None
):
    return None  # client-supplied wins
# A derived cache key falls through to the file-pin lookup.
```

**Tests**:

- The two existing precedence tests still pass (explicit key blocks file pin).
- New regression `test_derived_prompt_cache_key_does_not_block_file_id_pin` simulates the affinity helper's derived assignment by setting the attribute and removing it from `model_fields_set` (which is what the helper effectively does), then asserts the resolver still returns the file_id pin.

**Drive-by**: replaced `payload.to_payload() | {"type": ...}` with explicit dict construction in the websocket prep test -- ty currently blocks on dict-union there and the CI `Type check (ty)` job was red on the previous push because of it. Now ty is clean again.

**Verification**:
- ruff check / format -- clean
- ty check -- clean
- `pytest tests/integration/test_proxy_files.py tests/integration/test_proxy_websocket_responses.py tests/integration/test_proxy_responses.py tests/integration/test_proxy_compact.py tests/integration/test_openai_compat_features.py tests/unit/test_files_client.py tests/unit/test_openai_requests.py` -- **219 passed**

The other line comments on `30ab726` are stale repeats of P1/P2 already addressed in earlier rounds.

@codex review

Soju06:
Pushed `4cdcdf0` for the new P1 codex flagged on `b33bf87` (`Allow file-id affinity despite synthetic turn-state headers`).

Real follow-up to my own incomplete first attempt at this. The websocket entry points (and the HTTP path) always inject an `x-codex-turn-state` header on first turns when the client did not supply one, via `ensure_downstream_turn_state` / `ensure_http_downstream_turn_state`. My resolver was treating *any* turn-state value as a client-supplied continuation marker, so file-pin routing was silently never applied to the upload-then-converse flow that actually motivated this whole subsystem.

**Fix**: synthesizer outputs match `(?:http_)?turn_<32 hex>`. Add `_is_synthesized_turn_state(value)` and short-circuit only on values that do NOT match that pattern.

```python
turn_state_value = _sticky_key_from_turn_state_header(headers)
if turn_state_value is not None and not _is_synthesized_turn_state(turn_state_value):
    return None  # client-supplied continuation marker wins
# synthesizer placeholder falls through to file-pin lookup
```

**Tests** (`test_synthesized_turn_state_does_not_block_file_id_pin`):
- Synthesized header (`turn_<hex>`) -- helper still returns the pin.
- Client-supplied header (`client-conversation-handle-42`) -- helper returns `None` so the conversation continuation keeps its routing.

**Verification**:
- ruff check / format -- clean
- ty check -- clean
- 220 tests passed across files / websocket / responses / compact / compat / unit suites

The `b33bf87` review's other line comments are stale repeats of the earlier rounds' findings (the bot still echoes them per-commit) -- already addressed.

@codex review

Soju06:
Pushed `b715170` for the new P1 codex flagged on `4cdcdf0` (`Extend file-account pin TTL for 512 MiB upload flows`).

Real point. The pin TTL was 5 minutes but the upload limit is 512 MiB. A slow client uplink can keep PUTting blob bytes to the SAS URL for longer than that, and the finalize / follow-up `/responses` then land on the wrong account because the pin already expired -- defeating the affinity guarantee on exactly the heavy-upload flows that need it most.

Bumped to 30 minutes. That covers a 512 MiB upload at ~280 KiB/s, well below typical broadband uplink, and keeps the table footprint negligible (each pin is a short string tuple; opportunistic eviction runs on every write).

The other line comments on `4cdcdf0` are stale repeats of P1/P2 already addressed in earlier rounds.

@codex review

Soju06:
Pushed `7356778` for the new P2 codex flagged on `b715170` (`Treat promptCacheKey alias as explicit affinity signal`).

Borderline-gap: the V1 normalization path (`/v1/responses` -> `V1ResponsesRequest.to_responses_request`) rewrites `promptCacheKey` -> `prompt_cache_key` before `ResponsesRequest.model_validate` so the canonical name lands in `model_fields_set` correctly. But the raw codex backend route (`/backend-api/codex/responses`) binds `ResponsesRequest` directly via FastAPI body, so a client posting the camelCase alias would have it preserved verbatim and `"prompt_cache_key" in model_fields_set` would be false -- silently making the file pin override an explicit continuation marker.

```python
explicit_cache_key = (
    "prompt_cache_key" in explicit_fields
    or "promptCacheKey" in explicit_fields
)
if explicit_cache_key and _prompt_cache_key_from_request_model(payload) is not None:
    return None
```

`_prompt_cache_key_from_request_model` already accepts either name when it reads the populated value, so the value-check still fires correctly.

No test churn -- the existing precedence regression covers the canonical key and the new branch is a spelling tolerance for the alias.

@codex review

Issue #471, opened 2026-04-23
Cached session_header bridge keeps reusing a failed upstream session, causing repeated 502 on /v1/responses   until a new conversation is started


nhdong1993:
## Summary
  When an upstream Codex session behind a cached `session_header` bridge fails, codex-lb keeps reusing the same
  cached bridge for subsequent requests from the same API key / conversation. As a result, every request returns
  `502 upstream_unavailable` with the same OpenAI `request_id`, and the only workaround is for the end user to
  start a brand-new chat/conversation in their client.

  ## Environment
  - codex-lb image: `ghcr.io/soju06/codex-lb:1.14.1`
  - Deployment: Docker (`docker run`, single container)
  - Host OS: Linux (Ubuntu, kernel 6.8)
  - Clients affected: OpenAI-compatible IDE plugins (e.g. IntelliJ AI Assistant, other tools that reuse a
  conversation id across requests) calling `/v1/responses` and `/v1/chat/completions`.

  1. Point any OpenAI-compatible client at codex-lb using an API key bound to a Codex account.
  2. Send several requests in the same chat/conversation until one request fails upstream (e.g. transient OpenAI
  `server_error`, or a WebSocket close reported as `received 1000 (OK); then sent 1000 (OK)`).
  3. From the same client conversation, send more requests immediately.

  ## Expected Behavior
  After an upstream failure that invalidates the session, codex-lb should evict the cached `session_header`
  bridge and create a fresh one on the next request. Transient upstream errors should not turn into a sticky
  failure for the caller.

  ## Actual Behavior
  - All follow-up requests on that bridge return `502` with the exact same OpenAI `request_id`, proving the same
  upstream session is being reused.
  - The user can only recover by starting a new chat/conversation in the client (which changes the cache key and
  forces a new bridge).
  - Observed on `POST /v1/responses` and, by extension, on `/v1/chat/completions` which wraps the same upstream.

  ## Logs
  Anonymized excerpt; account id and API key redacted. Note the three different codex-lb `request_id`s mapping to
   the same upstream OpenAI `request_id`, and the `send_failure` event on the cached bridge:

  proxy_error_response request_id= method=POST path=/v1/responses status=502 code=server_error
    message=An error occurred while processing your request ... Please include the request ID
  841c44e2-0aff-4485-8f7d-fab72f950743 ...
  proxy_error_response request_id= method=POST path=/v1/responses status=502 code=server_error
[...]

PR #301, opened 2026-04-02
Harden HTTP bridge previous_response_id continuity recovery

JoshuaRileyDev:
## Summary
- add a bridge-local previous-response recovery index keyed by (response.id, api_key)
- recover live HTTP bridge sessions for follow-up requests carrying previous_response_id before fail-closed continuity errors
- clean up previous-response mappings when bridge sessions close
- add integration coverage for recovery when request affinity drifts between turns
- add OpenSpec change artifacts for this behavior update

## Validation
- `uv run pytest -q tests/integration/test_http_responses_bridge.py -k "missing_turn_state_alias_with_previous_response_id_fails_closed or recovers_previous_response_without_turn_state_alias or send_failure_returns_previous_response_not_found"`
- result: 3 passed

Issue #285, opened 2026-04-01
OpenClaw + codex-lb: visible intermediate assistant messages can stop before tool calls


SHAREN:
﻿## TL;DR

I initially thought this was a pure `codex-lb` transport bug. After spending a lot more time debugging OpenClaw itself, my current understanding is more nuanced:

- I did hit real OpenClaw-side integration bugs first (`No API key for provider: codex-lb`, then `Failed to extract accountId from token`)
- I locally patched those in OpenClaw and moved from the older HTTP `/v1` path to the real Codex websocket path (`openai-codex-responses` -> `/backend-api/codex`)
- that improved the transport/auth layer, but it did **not** fully fix the original symptom
- the remaining symptom now looks more like an OpenClaw continuation/orchestration problem than a pure `codex-lb` transport problem

I am posting this because I may still be misunderstanding something, and I would really appreciate feedback from people who know this stack better.

---

## Full story

About a month ago I installed `codex-lb` for the first time and started using it with OpenClaw.

Very quickly I noticed a strange behavior: the agent could send a visible intermediate message like:

> "OK, I'll open the file and check it"

or

> "First I'll send an intermediate update, then I'll run ping"

but after that, nothing happened. The next step, where the model was supposed to make a tool call, simply never came. The reasoning chain just stopped on a normal assistant text response.

An important detail: this was **not** a case where tool calls never worked at all. Tool calls did work in general. The problem was narrower: when the model first emitted a visible intermediate message and was then supposed to continue into a tool call, that chain often broke.

At first I assumed this was a `codex-lb` problem. I spent a long time digging into `codex-lb` itself, wondering whether it was dropping tool calls, mishandling websocket traffic, or doing something wrong in the proxy layer. At that time I never got to the real root cause, so I dropped the investigation.

Recently I came back to it, but this time I focused much more on OpenClaw itself.

---

## What I found first: a separate OpenClaw-side auth bug

The first confirmed issue I hit was:

`No API key for provider: codex-lb`

At first this looked like `codex-lb` itself was rejecting the key. But after tracing the code, it turned out the problem was deeper and not actually in `codex-lb`.

What I found was roughly this:

- OpenClaw resolved the custom provider key correctly in its model registry / auth path
- somewhere in the embedded runtime session path, that key was effectively lost before the call reached `pi-ai`
- so `pi-ai` ended up seeing provider `codex-lb` without an `apiKey`
[...]

PR #516, opened 2026-04-29
fix(proxy): mask single previous response misses

Komzpa:
## Summary

- mask single-request anonymous HTTP bridge `previous_response_not_found` as an internal continuity-loss signal instead of leaking the raw upstream 400
- preserve the existing local previous-response rebind path by raising an internal `bridge_previous_response_not_found` marker only for the single anonymous HTTP bridge case
- retry direct WebSocket full-resend follow-ups without a stale `previous_response_id` when upstream loses a just-completed anchor before `response.created`
- sanitize direct WebSocket short continuations that hit upstream `previous_response_not_found` into `stream_incomplete` instead of replaying the same stale anchor and exposing the raw upstream 400
- avoid false local `no_plan_support_for_model` 503s after partial model-registry refreshes: account selection now only applies the plan filter when the registry has an authoritative entry for the requested model, and model refresh invalidates cached selection inputs
- mask any lower-layer public Responses or compact `previous_response_not_found` payload to retryable `stream_incomplete` so the missing response id cannot leak even if a lower recovery path misses it
- add OpenSpec change artifacts plus HTTP bridge, WebSocket, and model-registry/load-balancer regression coverage

## Root Cause

The HTTP bridge only rewrote upstream `previous_response_not_found` when the upstream event had a `response.id` or there were other pending requests. A single pending follow-up with an anonymous upstream error could therefore propagate the raw invalid-request error to Codex clients.

Live archive follow-up showed direct WebSocket Codex turns can receive `response.completed` for an anchor and then hit `previous_response_not_found` on the next turn seconds later on the same account/session. Full-resend turns can be recovered by reconnecting upstream and replaying the full payload without `previous_response_id`; short continuation turns cannot, because the only semantic payload is the stale anchor, so they now fail closed as a sanitized continuity-loss signal.

A DNS outage exposed a separate local-routing edge while the live container was running this PR stack: partial model registry data could make `responses/compact` apply a plan filter for a model the registry did not actually know yet, producing local 503 `No accounts with a plan supporting model 'gpt-5.5'` while adjacent `responses` calls on pro accounts were already succeeding.

## Fresh live incident check

Checked the live raw leak with `previous_response_id=resp_0bebc32c66e5a5990169f9d272ffbc8191801046ca030a311b` from `2026-05-05`:

- `2026-05-05 11:20:19 UTC`: upstream created `resp_0beb...d272...`.
- `2026-05-05 11:20:24 UTC`: upstream returned `response.completed` for that response.
- `2026-05-05 11:21:27 UTC`: the same response stream recorded `no close frame received or sent`.
- `2026-05-05 11:22:25 UTC`: the next turn sent `previous_response_id=resp_0beb...d272...` and got raw `previous_response_not_found` / `previous_response_id` / `400` downstream.

This was not a cross-account routing miss: the created/completed response, follow-up request, and raw error all used the same account family `c26b2ebf...`. A new account `fe2708b6...` appeared nearby in the archive, but for a different parallel request.

So this PR's desired behavior is fail-closed masking: when upstream loses continuity even on the owner account, downstream gets `502 stream_incomplete`, not raw `previous_response_not_found` or a `resp_*` id.

## Validation

- `.venv/bin/pytest -q tests/integration/test_proxy_websocket_responses.py` (37 passed, 2 pre-existing aiosqlite thread warnings)
- `.venv/bin/pytest -q tests/integration/test_proxy_websocket_responses.py::test_v1_responses_websocket_masks_short_previous_response_not_found_without_retry tests/integration/test_proxy_websocket_responses.py::test_backend_responses_websocket_masks_short_previous_response_not_found_without_retry tests/integration/test_proxy_websocket_responses.py::test_backend_responses_websocket_masks_previous_response_not_found_when_message_omits_response_id tests/integration/test_proxy_websocket_responses.py::test_v1_responses_websocket_retries_full_resend_previous_response_miss_without_anchor` (4 passed)
- `.venv/bin/pytest -q tests/unit/test_proxy_load_balancer_refresh.py tests/unit/test_graceful_degradation.py` (42 passed, 3 skipped)
- `.venv/bin/ruff check app/modules/proxy/service.py tests/integration/test_proxy_websocket_responses.py`
- `.venv/bin/ruff check app/modules/proxy/load_balancer.py app/core/openai/model_refresh_scheduler.py tests/unit/test_proxy_load_balancer_refresh.py`
- `.venv/bin/ruff format --check app/modules/proxy/service.py tests/integration/test_proxy_websocket_responses.py`
- `.venv/bin/ruff format --check app/modules/proxy/load_balancer.py app/core/openai/model_refresh_scheduler.py tests/unit/test_proxy_load_balancer_refresh.py`
- `npx --yes @fission-ai/openspec validate --specs` (19 passed)
- `git diff --check`
- `/home/kom/proj/codex-lb/.venv/bin/python -m pytest -q tests/unit/test_proxy_api_websocket_auth.py::test_public_previous_response_not_found_error_is_masked_to_stream_incomplete tests/unit/test_proxy_api_websocket_auth.py::test_public_previous_response_invalid_request_param_is_masked_to_stream_incomplete tests/integration/test_proxy_websocket_responses.py tests/integration/test_http_responses_bridge.py::test_v1_responses_http_bridge_rebinds_after_upstream_previous_response_not_found tests/integration/test_http_responses_bridge.py::test_v1_responses_http_bridge_masks_anonymous_previous_response_not_found_with_inflight_request` (41 passed, rerun on `2026-05-05`)
- `/home/kom/proj/codex-lb/.venv/bin/ruff check app/modules/proxy/service.py tests/unit/test_proxy_api_websocket_auth.py tests/integration/test_proxy_websocket_responses.py tests/integration/test_http_responses_bridge.py` (rerun on `2026-05-05`)
- `git diff --check` (rerun on `2026-05-05`)

GitHub CI for this PR head is green: 18/18 checks passed.

## Notes

- `make -f src/Makefile precommit` is not available in this repository; that Makefile belongs to the surrounding OpenClaw workspace.
- `.venv/bin/ty check app/modules/proxy/service.py tests/integration/test_proxy_websocket_responses.py` still reports the pre-existing `service.py:9018..9029` diagnostics that are handled by neighboring PR #517.

## Related issues
- Related to #210, #285, #471, #530, and #550 previous-response / Codex continuity reports.
- Supersedes the fail-closed/rebind subset of #301 for the current HTTP/WebSocket bridge paths.

### DIFF
--- a/app/core/openai/model_refresh_scheduler.py
+++ b/app/core/openai/model_refresh_scheduler.py
@@ -16,6 +16,7 @@ from app.db.models import Account, AccountStatus
 from app.db.session import get_background_session
 from app.modules.accounts.auth_manager import AuthManager
 from app.modules.accounts.repository import AccountsRepository
+from app.modules.proxy.account_cache import get_account_selection_cache
 
 logger = logging.getLogger(__name__)
 
@@ -96,6 +97,7 @@ class ModelRefreshScheduler:
                         len(per_plan_results),
                         total_models,
                     )
+                    get_account_selection_cache().invalidate()
                 else:
                     logger.warning("Model registry refresh failed for all plans")
         except Exception:
--- a/app/modules/proxy/api.py
+++ b/app/modules/proxy/api.py
@@ -1698,16 +1698,17 @@ async def _collect_responses(
     except ProxyResponseError as exc:
         await _release_reservation(reservation)
         error = _parse_error_envelope(exc.payload)
+        status_code, error = _mask_previous_response_not_found_error(error, default_status=exc.status_code)
         return _logged_error_json_response(
             request,
-            exc.status_code,
+            status_code,
             error.model_dump(mode="json", exclude_none=True),
             headers=rate_limit_headers,
         )
     if isinstance(response_payload, OpenAIResponsePayload):
         if response_payload.status == "failed":
             error_payload = _error_envelope_from_response(response_payload.error)
-            status_code = _status_for_error(error_payload.error)
+            status_code, error_payload = _mask_previous_response_not_found_error(error_payload)
             return _logged_error_json_response(
                 request,
                 status_code,
@@ -1718,7 +1719,7 @@ async def _collect_responses(
             content=response_payload.model_dump(mode="json", exclude_none=True),
             headers={**turn_state_headers, **rate_limit_headers},
         )
-    status_code = _status_for_error(response_payload.error)
+    status_code, response_payload = _mask_previous_response_not_found_error(response_payload)
     return _logged_error_json_response(
         request,
         status_code,
@@ -1806,9 +1807,10 @@ async def _compact_responses(
         )
     except ProxyResponseError as exc:
         error = _parse_error_envelope(exc.payload)
+        status_code, error = _mask_previous_response_not_found_error(error, default_status=exc.status_code)
         return _logged_error_json_response(
             request,
-            exc.status_code,
+            status_code,
             error.model_dump(mode="json", exclude_none=True),
             headers=rate_limit_headers,
         )
--- a/app/modules/proxy/load_balancer.py
+++ b/app/modules/proxy/load_balancer.py
@@ -416,7 +416,7 @@ class LoadBalancer:
                 allowed_account_ids = set(account_ids)
                 accounts = [account for account in accounts if account.id in allowed_account_ids]
             pre_model_filter_accounts = accounts
-            if model and (effective_limit_name is None or _mapped_model_has_registry_entry(model)):
+            if model and _mapped_model_has_registry_entry(model):
                 accounts = _filter_accounts_for_model(pre_model_filter_accounts, model)
             if model and not accounts:
                 if not all_accounts:
--- a/app/modules/proxy/service.py
+++ b/app/modules/proxy/service.py
@@ -1216,6 +1216,14 @@ class ProxyService:
                     and request_state.error_http_status_override is not None
                     and request_state.error_http_status_override >= 400
                 ):
+                    if request_state.previous_response_not_found_rewritten:
+                        raise ProxyResponseError(
+                            request_state.error_http_status_override,
+                            openai_error(
+                                "bridge_previous_response_not_found",
+                                "Upstream websocket closed before response.completed",
+                            ),
+                        )
                     raise ProxyResponseError(
                         request_state.error_http_status_override,
                         _openai_error_envelope_from_response_failed_payload(block_payload),
@@ -3016,6 +3024,19 @@ class ProxyService:
                     slim_summary["historical_images_slimmed"],
                 )
         request_state.request_text = text_data
+        if (
+            transport == _REQUEST_TRANSPORT_WEBSOCKET
+            and payload.previous_response_id is not None
+            and _http_bridge_payload_looks_like_full_resend(payload)
+        ):
+            fresh_upstream_payload = dict(upstream_payload)
+            fresh_upstream_payload.pop("previous_response_id", None)
+            request_state.fresh_upstream_request_text = json.dumps(
+                fresh_upstream_payload,
+                ensure_ascii=True,
+                separators=(",", ":"),
+            )
+            request_state.fresh_upstream_request_is_retry_safe = True
         _enforce_response_create_size_limit(request_state)
         return request_state, text_data
 
@@ -5650,9 +5671,11 @@ class ProxyService:
             status_request_state is not None
             and status_request_state.previous_response_id is not None
             and is_previous_response_not_found_event
-            and (response_id is not None or has_other_pending_requests)
         ):
             status_request_state.error_http_status_override = 502
+            status_request_state.previous_response_not_found_rewritten = (
+                response_id is None and not has_other_pending_requests
+            )
             event, payload, event_type, rewritten_text = _maybe_rewrite_websocket_previous_response_not_found_event(
                 request_state=status_request_state,
                 event=event,
@@ -6246,14 +6269,29 @@ class ProxyService:
             )
             retry_error_code = None
         if retry_error_code is not None:
-            upstream_control.reconnect_requested = True
             if retry_is_previous_response_not_found:
-                request_state.replay_count += 1
-                request_state.awaiting_response_created = True
-                request_state.response_id = None
-                upstream_control.suppress_downstream_event = True
-                upstream_control.replay_request_state = request_state
+                if not (
+                    request_state.fresh_upstream_request_is_retry_safe and request_state.fresh_upstream_request_text
+                ):
+                    # A short continuation depends entirely on the upstream
+                    # anchor. Replaying the same lost previous_response_id on a
+                    # new websocket just re-surfaces the raw upstream 400; only
+                    # full-resend payloads with a prepared fresh body can be
+                    # transparently retried.
+                    retry_error_code = None
+                else:
+                    upstream_control.reconnect_requested = True
+                    request_state.request_text = request_state.fresh_upstream_request_text
+                    request_state.previous_response_id = None
+                    request_state.proxy_injected_previous_response_id = False
+                    request_state.fresh_upstream_request_is_retry_safe = False
+                    request_state.replay_count += 1
+                    request_state.awaiting_response_created = True
+                    request_state.response_id = None
+                    upstream_control.suppress_downstream_event = True
+                    upstream_control.replay_request_state = request_state
             else:
+                upstream_control.reconnect_requested = True
                 request_state.replay_count += 1
                 request_state.awaiting_response_created = True
                 request_state.response_id = None
@@ -6264,7 +6302,8 @@ class ProxyService:
                     {"message": _websocket_event_error_message(event_type, payload) or "Upstream error"},
                     retry_error_code,
                 )
-            return downstream_text
+            if retry_error_code is not None:
+                return downstream_text
 
         await self._finalize_websocket_request_state(
             request_state,
@@ -8436,6 +8475,7 @@ class _WebSocketRequestState:
     error_param_override: str | None = None
     error_http_status_override: int | None = None
     response_event_count: int = 0
+    previous_response_not_found_rewritten: bool = False
     response_create_gate_acquired: bool = False
     response_create_gate: asyncio.Semaphore | None = None
     response_create_admission: AdmissionLease | None = None
@@ -10769,6 +10809,7 @@ def _http_bridge_should_attempt_local_previous_response_recovery(exc: ProxyRespo
     code = error.get("code")
     if code in {
         "bridge_owner_unreachable",
+        "bridge_previous_response_not_found",
         "previous_response_not_found",
         "bridge_instance_mismatch",
     }:
--- a/app/modules/proxy/service.py
+++ b/app/modules/proxy/service.py
@@ -3024,19 +3024,6 @@ class ProxyService:
                     slim_summary["historical_images_slimmed"],
                 )
         request_state.request_text = text_data
-        if (
-            transport == _REQUEST_TRANSPORT_WEBSOCKET
-            and payload.previous_response_id is not None
-            and _http_bridge_payload_looks_like_full_resend(payload)
-        ):
-            fresh_upstream_payload = dict(upstream_payload)
-            fresh_upstream_payload.pop("previous_response_id", None)
-            request_state.fresh_upstream_request_text = json.dumps(
-                fresh_upstream_payload,
-                ensure_ascii=True,
-                separators=(",", ":"),
-            )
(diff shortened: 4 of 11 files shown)


mgwals:
I validated this branch against a fresh local incident capture from `codex-cli 0.128.0` using a local `codex-lb:1.15.0-ws-retry` container.

Observed in `request_logs` over a 6h window:

- `1634` successes, `3` errors
- all 3 errors were `status=error`, `error_code=stream_incomplete`, `transport=http`
- message: `Upstream websocket closed before response.completed: no close frame received or sent`
- models/efforts: `gpt-5.5` with `high`/`xhigh`
- no raw account ids included here; account ids were only hashed in the local capture

This branch matches the mitigation I would want for the continuity-loss class: mask `previous_response_not_found`, retry only when the request is still safely replayable, and avoid blind retry once a semantic continuation depends only on a stale anchor.

Local validation on this PR branch:

- `uv run pytest -q tests/unit/test_proxy_api_websocket_auth.py::test_public_previous_response_not_found_error_is_masked_to_stream_incomplete tests/unit/test_proxy_api_websocket_auth.py::test_public_previous_response_invalid_request_param_is_masked_to_stream_incomplete tests/integration/test_proxy_websocket_responses.py::test_v1_responses_websocket_masks_short_previous_response_not_found_without_retry tests/integration/test_proxy_websocket_responses.py::test_backend_responses_websocket_masks_short_previous_response_not_found_without_retry tests/integration/test_proxy_websocket_responses.py::test_backend_responses_websocket_masks_previous_response_not_found_when_message_omits_response_id tests/integration/test_proxy_websocket_responses.py::test_v1_responses_websocket_retries_full_resend_previous_response_miss_without_anchor tests/integration/test_http_responses_bridge.py::test_v1_responses_http_bridge_rebinds_after_upstream_previous_response_not_found tests/integration/test_http_responses_bridge.py::test_v1_responses_http_bridge_masks_anonymous_previous_response_not_found_with_inflight_request` -> `8 passed`
- `uv run openspec validate --specs` -> `19 passed, 0 failed`
- `uv run ruff check ...` on changed proxy/test files -> passed
- `uv run ruff format --check ...` wanted to reformat `app/modules/proxy/service.py` locally with the current lockfile/tooling

Separate live smoke against the existing container, without deploying this PR: 60 small streamed `/v1/responses` requests (`20` each for low/high/xhigh) all reached `response.completed`, and no new error rows appeared after the smoke start.

Komzpa:
Cross-link note: #535 is a narrower overlapping PR for direct WebSocket `previous_response_not_found` masking.

This PR (#516) is the broader owned branch and should be treated as the canonical fix for our live incident class: HTTP bridge, direct WebSocket, compact/public masking, and the model-registry edge are all covered here.

Live deployment note: the current local container image is `codex-lb:weekly-recovery-safe-25770b8`, built from the live composite branch at `25770b8`. That stack includes this PR's proxy masking patch as commit `6e4cbc3`; its stable patch-id matches this PR head `368e287`, so the deployed proxy masking behavior corresponds to #516 even though the live commit hash differs due to the composite stack.

jumski:
Tested - resolves the `503 no_plan_support_for_model` false positives with opencode + codex-lb.

**To test locally:**

```bash
git clone https://github.com/Komzpa/codex-lb.git /tmp/codex-lb-pr516
git -C /tmp/codex-lb-pr516 checkout codex/mask-single-previous-response-miss
DOCKER_BUILDKIT=1 docker build -t codex-lb:pr516 /tmp/codex-lb-pr516
```

Point compose.yaml at `codex-lb:pr516` and restart.

Works well. Would love to see this merged.

Soju06:
@codex review

Soju06:
Thanks @Komzpa -- the live-archive trace, fail-closed framing for short continuations, and the orthogonal model-registry / plan-filter hardening are all in good shape. Codex review came back with one P1 that needs a fix before merging.

## P1 -- direct WebSocket safe full-resend retries are now skipped

The new guard in `_process_upstream_websocket_text`:

```python
# app/modules/proxy/service.py:6238
if retry_is_previous_response_not_found:
    if not (
        request_state.fresh_upstream_request_is_retry_safe
        and request_state.fresh_upstream_request_text
    ):
        # short continuation: anchor lost, no safe replay payload
        retry_error_code = None
    else:
        # full-resend retry path
        upstream_control.reconnect_requested = True
        request_state.request_text = request_state.fresh_upstream_request_text
        request_state.previous_response_id = None
        ...
```

`fresh_upstream_request_is_retry_safe` / `fresh_upstream_request_text` are only populated on the HTTP bridge preparation path. The direct WebSocket request path never sets them, so direct WebSocket follow-ups -- including safe full-resend payloads that this PR explicitly says it wants to keep recovering by reconnecting and replaying without `previous_response_id` -- now fall into the `retry_error_code = None` branch and surface `stream_incomplete` instead of being transparently retried.

This regresses one of the recovery cases this PR is trying to preserve, and contradicts the test name added in the same change (`test_v1_responses_websocket_retries_full_resend_previous_response_miss_without_anchor`), which I expect was authored against the HTTP-bridge-prepared full-resend variant rather than the direct WebSocket variant.

## Suggested fix

Populate the same retry-safety metadata on the direct WebSocket request path so the new guard treats full-resend variants identically across paths. Concretely:

1. Where the direct WebSocket path constructs the upstream request text (the `_prepare_responses_websocket_request` / equivalent site that already knows whether the payload is a full-resend, i.e. does not depend on the upstream anchor for semantics), set:

   ```python
   request_state.fresh_upstream_request_text = full_resend_text
   request_state.fresh_upstream_request_is_retry_safe = True
   ```

   for full-resend cases, and leave both unset (or explicitly `False`) for short continuations.

2. Keep the line 6238 guard as written.

3. Add a direct-WebSocket regression test mirroring the HTTP bridge one:

   ```python
   async def test_backend_responses_websocket_retries_full_resend_previous_response_miss_without_anchor():
       # full-resend payload via direct websocket
       # upstream emits previous_response_not_found
       # assert: reconnect + replay without previous_response_id
       # assert: downstream sees the second-attempt response.completed, not stream_incomplete
   ```

   And a sanity test confirming a direct-WebSocket short continuation (no full-resend payload) still ends with `stream_incomplete`:

   ```python
   async def test_backend_responses_websocket_short_continuation_previous_response_miss_returns_stream_incomplete():
       # direct websocket, short continuation (no fresh_upstream_request_text)
       # upstream emits previous_response_not_found
       # assert: downstream sees stream_incomplete, no retry
   ```

If the direct WebSocket path also has cases that legitimately can't be classified as retry-safe at request-preparation time (because retry-safety depends on what upstream just sent), an alternative is to track retry-safety on the request state at receive time rather than relying on a flag set at request prep time -- but the simpler path is to propagate the existing flag.

## Notes on scope

- HTTP bridge mask, the lower-layer public Responses / compact `previous_response_not_found` -> `stream_incomplete` guard, and the registry plan-filter `no_plan_support_for_model` 503 fix all read cleanly.
- The model registry plan-filter change is well-scoped: only apply the filter when the registry has an authoritative entry, and invalidate cached selection inputs on refresh. The fact that it's bundled with the previous-response masking is justified in your PR body (same live incident window), so I won't push to split it.

Once the P1 fix lands I'll re-run `@codex review` and queue for merge.

JeffKandt:
I let a local mitigation soak for several days and wanted to share the result because it may be useful for this PR.

Context: I initially staged PR #516 at the earlier head `368e287` and cut it over to live Codex Desktop traffic. That was necessary, but not sufficient in my local environment: I still saw raw downstream `previous_response_not_found` errors interrupt Desktop sessions. In at least one incident, `request_logs` had already normalized the request to `stream_incomplete`, while the raw upstream JSON still reached the client.

My local follow-up added a last-chance downstream WebSocket send / error-serializer sanitizer: if any outbound WebSocket payload still contains upstream `previous_response_not_found`, it is rewritten to fail closed as `502 stream_incomplete` before reaching the client.

After several days of normal use:

- `request_logs` has `0` raw `previous_response_not_found` / `Previous response with id ... not found` rows since the final local patch
- I have not seen the user-visible Desktop interruption recur
- the service still logs internal fail-closed catches for this class; I see 18 `continuity_fail_closed ... previous_response_not_found` events with hashed response ids, which suggests upstream still occasionally produces the miss but the downstream boundary prevents it leaking to the client
- remaining client-visible failures are masked shapes such as `stream_incomplete`, not raw missing-response-id errors

I have not yet validated the current PR head `6854f57` directly, and I see the maintainer P1 about direct WebSocket full-resend retries. I just wanted to pass along the local soak finding: in addition to the request-state / retry-path fixes, a final downstream WebSocket send/serializer mask appears valuable as a belt-and-suspenders guard, because the raw upstream payload can otherwise survive past request-log normalization.

Soju06:
Pushed `713ef42` to address codex review's P1 (direct WebSocket fresh-turn retry must not fall through to `stream_incomplete`).

## What the fix does

The retry path in `_process_upstream_websocket_text` only attempts a reconnect-and-replay for a request flagged `fresh_upstream_request_is_retry_safe` with a populated `fresh_upstream_request_text`. The HTTP-bridge prep path already sets those flags. The direct WebSocket prep path (`_prepare_websocket_response_create_request`) did not, so every direct-WebSocket recovery short-circuited into `retry_error_code = None` -- including the full-resend / fresh-turn variants this PR explicitly wants to keep recovering. That's the regression codex review flagged.

Mirror the classification in the direct WebSocket prep path: when a turn has no client-supplied `previous_response_id` and no proxy-injected anchor, populate both `fresh_upstream_request_text` and `fresh_upstream_request_is_retry_safe = True` so the masking path treats both transport variants identically.

## Why the classification is safe

The two conditions that gate the new flag — `payload.previous_response_id is None` and `not request_state.proxy_injected_previous_response_id` — together rule out the only path that could make a replay re-surface the same stale anchor:

- `payload.previous_response_id is None` => the client did not pin a specific upstream response, so a fresh upstream turn carries no `previous_response_id`.
- `not proxy_injected_previous_response_id` => codex-lb did not inject one either (the HTTP-bridge path is the only place that does, and it manages its own retry-safety classification).

Anything else (short continuation, anchored follow-up, proxy-injected anchor) leaves the flag at its default `False`, so the existing `retry_error_code = None` fail-closed path still applies.

## Tests

**New integration** (`tests/integration/test_proxy_websocket_responses.py`):
`test_v1_responses_websocket_marks_fresh_turn_as_retry_safe_at_prep_time` pins that connecting an upstream for a fresh-turn direct WebSocket request observes `fresh_upstream_request_is_retry_safe=True` and a populated `fresh_upstream_request_text` at prep time.

**Updated unit** (`tests/unit/test_proxy_utils.py`):
`test_process_upstream_websocket_text_retries_precreated_previous_response_not_found` now drives the retry path with the new retry-safety contract on the pending request state and asserts:
- The retry strips the stale `previous_response_id` (`pending_request.previous_response_id is None`).
- The retry uses the prepared retry-safe text (`pending_request.request_text == fresh_request_payload`).
- The retry-safety flag is consumed (`fresh_upstream_request_is_retry_safe is False`).
- `replay_count` is incremented, `upstream_control.replay_request_state is pending_request`, and `reconnect_requested is True`.

## Verification on this head

```
uv run ruff check app/modules/proxy/service.py tests/integration/test_proxy_websocket_responses.py tests/unit/test_proxy_utils.py     # clean
uv run ruff format --check app/modules/proxy/service.py tests/integration/test_proxy_websocket_responses.py tests/unit/test_proxy_utils.py  # clean
uv run ty check app/modules/proxy/service.py                                                                                                  # clean
uv run pytest tests/integration/test_proxy_websocket_responses.py -q                                                                          # 38 passed
uv run pytest tests/unit/test_proxy_utils.py tests/unit/test_proxy_http_bridge.py tests/unit/test_proxy_load_balancer_refresh.py -q          # 346 passed, 3 skipped
```

Re-triggering `@codex review`.

Soju06:
@codex review

Soju06:
Pushed `93043c6` to fix the ruff `Lint` failure on the new direct-WebSocket retry-safe test in `tests/unit/test_proxy_utils.py`:

- `E501`: wrapped the long docstring referencing the inverse test name.
- `F841`: dropped the unused `downstream_text` assignment (the test only asserts on `upstream_control` / `pending_request` state).

No behavior change; the test still passes locally (`pytest -q tests/unit/test_proxy_utils.py -k retries_precreated_previous_response_not_found`).

CI is now rerunning against the new head.

Soju06:
@all-contributors please add @Komzpa for code, test, doc

Soju06:
@all-contributors please add @Komzpa for code, bug

Soju06:
@all-contributors please add @mgwals for test, review

Soju06:
@all-contributors please add @jumski for test

Soju06:
@all-contributors please add @JeffKandt for test, review

PR #522, opened 2026-04-30
fix(api-keys): backfill new limit usage

### LINKED ISSUE #518: Limits are not applying after add limit rule to existing API key with current usage
Image

- steps:
-  add API key
-  use for payload (10k tokens as example)
-  add limit rule (100k tokens daily )
-----
- expected: 10k/100k 
- actually: 0/100k

mgwals:
## Summary
- Backfill newly added API key limits from successful request-log usage in the active lookback window.
- Preserve existing matching limit counters and keep `resetUsage` as the explicit zero-reset path.
- Add OpenSpec coverage and regression tests for total-token, model-filtered, reset, and non-success-log behavior.

Closes #518

## Validation
- `openspec validate --specs`
- `openspec validate backfill-api-key-limit-current-usage`
- `uv run pytest -q tests/integration/test_api_keys_trends_api.py::test_adding_limit_backfills_current_window_usage tests/integration/test_api_keys_trends_api.py::test_adding_model_scoped_limit_backfills_only_matching_model tests/integration/test_api_keys_trends_api.py::test_adding_limit_with_reset_usage_keeps_current_value_zero`
- `uv run pytest -q tests/unit/test_api_keys_service.py tests/integration/test_api_keys_trends_api.py`
- `uv run ruff check app/modules/api_keys tests/integration/test_api_keys_trends_api.py tests/unit/test_api_keys_service.py`
- `uv run ruff format --check app/modules/api_keys tests/integration/test_api_keys_trends_api.py tests/unit/test_api_keys_service.py`
- `uv run ty check`
- `codex-review-high --base origin/main`


Soju06:
@codex review

Soju06:
Thanks @mgwals -- the backfill design is correct (preserves the explicit `resetUsage` zero-reset path, only fires when a new limit is added on `update_key`, and the per-`limit_type` value extraction matches the live accrual surfaces). The OpenSpec change is in good shape too. One P2 from codex review to address before merging, plus a quick note on CI.

## P2 -- cost backfill diverges from live limit accrual

The cost path in `get_limit_usage_value` does:

```python
# app/modules/api_keys/repository.py:228
return int(round(float(value or 0.0) * 1_000_000))
```

which sums the raw `cost_usd` column first and then rounds the aggregate into microdollars. But the live limit increment path uses per-request truncation:

```python
# _calculate_cost_microdollars
int(cost_usd * 1_000_000)
```

These two methods diverge whenever individual `cost_usd` values aren't exact integer microdollars. The most common case is a long history of small-cost requests where each row's microdollar value rounds *down*:

- 100 requests at `cost_usd=0.0000005` -> live = 100 * int(0.5) = 0
- 100 requests at `cost_usd=0.0000005` -> backfill = int(round(100 * 0.0000005 * 1_000_000)) = 50

So a brand-new cost limit added today on a key with 100 such requests would start at `current_value=50`, while a key that already had the limit during that history would still be at `current_value=0`. Same usage, two different states, and the backfilled key could prematurely trigger limit refusal.

The reverse direction is also possible (rare but real): per-request `int(...)` truncation can lose a tiny fraction that, in aggregate, undercounts the limit by a few microdollars.

## Minimal fix

Mirror the live truncation pattern in the backfill query. Sum the **already-truncated** per-row microdollar values:

```python
elif limit_type == LimitType.COST_USD:
    value_expr = func.coalesce(
        func.cast(RequestLog.cost_usd * 1_000_000, Integer),
        0,
    )
```

(Adjust the cast for the actual SQLAlchemy column type if needed -- the point is to truncate per-row before summing.)

Then drop the post-sum `round(... * 1_000_000)` since the cast already produces microdollar integers:

```python
result = await self._session.execute(stmt)
value = result.scalar_one()
return int(value or 0)  # no special-case for COST_USD anymore
```

That way `get_limit_usage_value` for `COST_USD` returns the same total as if the limit had been counting `_calculate_cost_microdollars` all along.

## Tests to update

The existing `test_adding_limit_backfills_current_window_usage` covers the integer-token path. We'd want one focused cost-divergence test:

```python
async def test_adding_cost_limit_uses_per_request_truncation():
    # seed N request logs with cost_usd = 0.0000005 (rounds to 0 microdollars per row)
    # add a cost limit
    # assert current_value == 0 (live truncation total), not 50 (post-sum rounding)
```

And one symmetric test confirming non-trivial costs still aggregate correctly:

```python
async def test_adding_cost_limit_backfills_truncated_microdollars():
    # seed request logs with cost_usd = 0.0000015 each (1 microdollar per row after truncation)
    # add a cost limit
    # assert current_value == N (truncated per-row before sum)
```

## CI -- workflow approval

The full Actions workflow has not actually run on this PR yet (only GitGuardian completed -- this is the first-time-fork-contributor approval gate). I'll approve the run on the maintainer side after you push the cost-truncation fix; no action needed from you. The local `uv run pytest tests/unit/test_api_keys_service.py tests/integration/test_api_keys_trends_api.py` you already ran is enough for the next round.

## Not blocking otherwise

Everything else reads cleanly -- the OpenSpec change is good, the token-path math matches the live accrual, and the model-filter / non-success-log behavior is the right shape. Once the cost-truncation fix lands I'll re-run `@codex review` and approve the workflow.

mgwals:
Thanks, addressed the P2 cost-backfill issue in `c09c528`.

The cost backfill now converts each request-log row to truncated integer microdollars before summing, using `floor(cost_usd * 1_000_000)` so it matches the live per-request accrual path and avoids PostgreSQL cast rounding. I also added focused regression coverage for the `0.0000005` and `0.0000015` cases, and updated the OpenSpec change text to make that contract explicit.

Validation run locally:

- `uv run pytest -q tests/integration/test_api_keys_trends_api.py::test_adding_cost_limit_uses_per_request_truncation tests/integration/test_api_keys_trends_api.py::test_adding_cost_limit_backfills_truncated_microdollars` (`2 passed`)
- `uv run pytest -q tests/unit/test_api_keys_service.py tests/integration/test_api_keys_trends_api.py` (`40 passed`)
- `uv run ruff check app/modules/api_keys/repository.py tests/integration/test_api_keys_trends_api.py`
- `uv run ruff format --check app/modules/api_keys/repository.py tests/integration/test_api_keys_trends_api.py`
- `openspec validate --specs` (`19 passed`)
- `git diff --check`
- `codex-review-high --uncommitted` after the PostgreSQL truncation adjustment: no actionable bugs found.

I also saw the Codex connector environment comment; no branch change is needed for that.

Komzpa:
@codex review

Soju06:

Komzpa:
Revived on a maintainer-owned branch as #748 because this PR head has maintainerCanModify=false and cannot be updated here. Keeping credit to the original implementation in the replacement PR.

PR #524, opened 2026-04-30
fix(proxy): inject SSE keepalive comment frames to prevent client stream hangs [Codex getting stuck issue]

balakumardev:
## Summary

Add SSE comment-frame heartbeats on the client-facing streaming response so the TCP path between codex-lb and the client (e.g. the codex CLI) stays observable when the upstream is silent for an extended period.

## Problem

When an upstream account stalls mid-stream — sticky-routed conversation pinned to an account that hits a rate limit, network slowdown, or any pause between OpenAI events — the path between codex-lb and the codex CLI can go silent for a long time. Without traffic on the wire, half-open TCP sockets (NAT/firewall timeouts, SSH-tunneled or load-balanced clients) are not detected until the OS keepalive timer fires (typically minutes later), which manifests as the CLI hanging indefinitely with no output and no error.

The upstream-facing reader already has `stream_idle_timeout_seconds` to detect upstream stalls, but that does not help if the client-facing socket has gone half-open in the meantime: the timeout's terminal error event is written into a dead socket and never reaches the client.

## Fix

New `inject_sse_keepalives(source, interval_seconds)` async generator in `app/core/utils/sse.py`. It wraps an SSE event iterator and emits an SSE comment frame (`": keepalive\n\n"`) whenever the source has been idle for more than `interval_seconds`.

- New setting `sse_keepalive_interval_seconds: float = 10.0` (set to `0` to disable).
- The wrapper is applied at three client-facing `StreamingResponse` boundaries in `app/modules/proxy/api.py`:
  - chat completions streaming
  - responses stream — `StopAsyncIteration` branch
  - responses stream — normal flow

### Why SSE comment frames are safe

Lines beginning with `:` are mandated by the SSE spec ([WHATWG HTML Living Standard](https://html.spec.whatwg.org/multipage/server-sent-events.html#parsing-an-event-stream)) to be ignored by parsers. Every conformant SSE client — `EventSource`, Python `httpx-sse`, Rust `eventsource-stream`, Go `r3labs/sse`, the codex CLI's reqwest-based parser — drops these lines silently before any application code sees them. They will not be logged by the client, not shown in the UI, not counted as tokens, not interfere with `event:`/`data:` accumulation.

The frames are only injected between event blocks (never inside one), and only when the source has been idle for the configured interval — so a fast-streaming response sees zero overhead.

## Tests

Four new unit tests in `tests/unit/test_sse.py`:

- `test_inject_sse_keepalives_passes_through_when_disabled` — `interval_seconds=0` is a pass-through.
- `test_inject_sse_keepalives_no_pings_when_source_is_fast` — fast source emits no pings.
- `test_inject_sse_keepalives_emits_pings_on_idle_gap` — idle gap triggers comment frame.
- `test_inject_sse_keepalives_keepalive_frame_is_sse_comment` — frame format starts with `: ` and ends with `\n\n`.

```
$ pytest tests/unit/test_sse.py -v
tests/unit/test_sse.py::test_format_sse_event_serializes_payload PASSED
tests/unit/test_sse.py::test_inject_sse_keepalives_passes_through_when_disabled PASSED
tests/unit/test_sse.py::test_inject_sse_keepalives_no_pings_when_source_is_fast PASSED
tests/unit/test_sse.py::test_inject_sse_keepalives_emits_pings_on_idle_gap PASSED
tests/unit/test_sse.py::test_inject_sse_keepalives_keepalive_frame_is_sse_comment PASSED
```

## Production validation

Deployed to a personal codex-lb instance and run for 24 hours with a real codex CLI workload over an SSH tunnel:

- Previously-hanging sessions now either complete normally or fail fast with a real error event reaching the client.
- No regressions observed.
- Confirmed clean failover when an account hits its usage limit mid-stream — the client now sees the error and the next request re-pins to a fresh account, instead of the session hanging indefinitely.

## Compatibility & rollout

- Default `sse_keepalive_interval_seconds = 10.0` is conservative — chosen well below typical NAT idle timeouts (30–60s) and well above any reasonable per-event latency.
- Setting it to `0` disables the feature entirely.
- No protocol changes, no client changes required, no schema or contract changes.

## Out of scope (potential follow-up)

A separate change could evict a sticky bridge entry on `usage_limit_reached` mid-stream so a session restart immediately re-pins to a fresh account instead of replaying the same wedged account. That is orthogonal to this fix; this PR is strictly about preventing the silent client-side stall.

## Files changed

- `app/core/utils/sse.py` — new `inject_sse_keepalives` async generator and `SSE_KEEPALIVE_FRAME` constant.
- `app/core/config/settings.py` — new `sse_keepalive_interval_seconds` setting.
- `app/modules/proxy/api.py` — wrap three client-facing `StreamingResponse` iterators.
- `tests/unit/test_sse.py` — four new unit tests.

### DIFF
--- a/app/core/config/settings.py
+++ b/app/core/config/settings.py
@@ -134,6 +134,7 @@ class Settings(BaseSettings):
     proxy_request_budget_seconds: float = Field(default=600.0, gt=0)
     compact_request_budget_seconds: float = Field(default=75.0, gt=0)
     stream_idle_timeout_seconds: float = 300.0
+    sse_keepalive_interval_seconds: float = Field(default=10.0, ge=0)
     proxy_downstream_websocket_idle_timeout_seconds: float = Field(default=120.0, gt=0)
     # Applies to both upstream SSE event buffering and upstream websocket message
     # frames. Keep the default aligned with the common 16 MiB websocket ceiling so
--- a/app/core/utils/sse.py
+++ b/app/core/utils/sse.py
@@ -31,12 +31,15 @@ async def inject_sse_keepalives(
             yield chunk
         return
 
+    async def _next_chunk(it: AsyncIterator[str]) -> str:
+        return await it.__anext__()
+
     iterator = source.__aiter__()
     pending: asyncio.Task[str] | None = None
     try:
         while True:
             if pending is None:
-                pending = asyncio.create_task(iterator.__anext__())
+                pending = asyncio.create_task(_next_chunk(iterator))
             try:
                 chunk = await asyncio.wait_for(
                     asyncio.shield(pending),
--- a/app/core/utils/sse.py
+++ b/app/core/utils/sse.py
@@ -1,13 +1,62 @@
 from __future__ import annotations
 
+import asyncio
 import json
-from collections.abc import Mapping
+from collections.abc import AsyncIterator, Mapping
 
 from app.core.errors import ResponseFailedEvent
 from app.core.types import JsonValue
 from app.core.utils.json_guards import is_json_dict
 
 type JsonPayload = Mapping[str, JsonValue] | ResponseFailedEvent
+
+SSE_KEEPALIVE_FRAME = ": keepalive\n\n"
+
+
+async def inject_sse_keepalives(
+    source: AsyncIterator[str],
+    interval_seconds: float,
+) -> AsyncIterator[str]:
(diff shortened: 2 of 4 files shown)


Soju06:
@codex review

Soju06:
@all-contributors please add @balakumardev for code, test

PR #528, opened 2026-04-30
fix(model-registry): populate bootstrap models to prevent missing metadata on startup

### LINKED ISSUE #480: Model metadata for `gpt-5.5` not found. Defaulting to fallback metadata; this can degrade performance and cause   issues.
'Model metadata for `gpt-5.5` not found. Defaulting to fallback metadata; this can degrade performance and cause issues.'

I have updated to codex-lb 1.15.0,  OpenAI Codex (v0.124.0). 
I have checked and the accounts I have active in codex-lb all have access to gpt 5.5 if logging in independently in Codex App.

Both CLI and Codex don't show gpt 5.5.
/model in CLI showed: " Warning: OpenAI base URL is overridden to http://127.0.0.1:2455/backend-api/codex. Selecting models may not be supported or work properly."

Request Log in Dashboard is showing model "gpt-5.5 (default)"

`model = "gpt-5.5"
model_reasoning_effort = "xhigh"
model_provider = "codex-lb"

personality = "pragmatic"
approvals_reviewer = "user"
sandbox_mode = "workspace-write"
plan_mode_reasoning_effort = "xhigh"

[windows]
sandbox = "elevated"

[features]
multi_agent = true

[model_providers.codex-lb]
name = "OpenAI"
base_url = "http://127.0.0.1:2455/backend-api/codex"
wire_api = "responses"
supports_websockets = true
env_key = "CODEX_LB_API_KEY"

[sandbox_workspace_write]
network_access = true

[mcp_servers.knowns]
command = "cmd"
args = ["/c", "npx", "-y", "knowns", "mcp"]

[mcp_servers.knowns.tools.get_doc]
approval_mode = "approve"

[mcp_servers.knowns.tools.search]
approval_mode = "approve"

[mcp_servers.knowns.tools.set_project]
approval_mode = "approve"

[mcp_servers.knowns.tools.update_doc]
approval_mode = "approve"

[mcp_servers.knowns.tools.validate]
approval_mode = "approve"

[mcp_servers.playwright]
command = "npx"
args = ["@playwright/mcp@latest"]

[mcp_servers.playwright.tools.browser_close]
approval_mode = "approve"

[mcp_servers.playwright.tools.browser_navigate]
approval_mode = "approve"

[mcp_servers.playwright.tools.browser_tabs]
approval_mode = "approve"

[mcp_servers.chrome-devtools]
command = "npx"
args = ["-y", "chrome-devtools-mcp@latest"]

[mcp_servers.googleDeveloperKnowledge]
url = "https://developerknowledge.googleapis.com/mcp"
env_http_headers = { "X-Goog-Api-Key" = "GOOGLE_DEVELOPER_KNOWLEDGE_API_KEY" }

[mcp_servers.StitchMCP]
command = "node"
args = ['C:\Users\The Lord\.codex\stitch-proxy-runtime\start.mjs']

[mcp_servers.StitchMCP.env]
STITCH_API_KEY = ""

[projects.'\\?\C:\Users\The Lord\Desktop\luyentap']
trust_level = "trusted"

[projects.'C:\Users\The Lord']
trust_level = "trusted"

[projects.'C:\']
trust_level = "trusted"

[projects.'C:\Users\The Lord\OneDrive\Documents\Playground']
trust_level = "trusted"

[notice]
hide_full_access_warning = true
hide_rate_limit_model_nudge = true

[notice.model_migrations]
"gpt-5.3-codex" = "gpt-5.4"

[marketplaces.openai-bundled]
last_updated = "2026-04-24T03:21:28Z"
source_type = "local"
source = '\\?\C:\Users\The Lord\.codex\.tmp\bundled-marketplaces\openai-bundled'
`

sudhanthirapriya:
Problem

Fixes #480

On startup — before the first upstream model refresh completes — `ModelRegistry._bootstrap_models` was an empty dict `{}`. This caused `GET /models` and `GET /backend-api/codex/models` to return empty lists, making Codex CLI log:

> Model metadata for gpt-5.5 not found. Defaulting to fallback metadata; this can degrade performance and cause issues.

This affects:
- Fresh installs before any account is added
- Server restarts during the ~60s window before the first refresh cycle
- Accounts that temporarily fail token refresh

 Root Cause:

`_bootstrap_models` existed as a fallback path in `get_models_with_fallback()` but was never populated:

```python
# Before
self._bootstrap_models: dict[str, UpstreamModel] = {}

Fix:
Added _BOOTSTRAP_STATIC_MODELS with conservative static entries for the 5 most commonly used models. The registry now initializes from these instead of an empty dict.

# After
self._bootstrap_models: dict[str, UpstreamModel] = {m.slug: m for m in _BOOTSTRAP_STATIC_MODELS}
Models added: gpt-5.5, gpt-5.5-pro, gpt-5.4, gpt-5.4-mini, gpt-5.3-codex

### DIFF
--- a/app/core/openai/model_registry.py
+++ b/app/core/openai/model_registry.py
@@ -50,6 +50,117 @@ class ModelRegistrySnapshot:
 
 _BOOTSTRAP_WEBSOCKET_PREFERRED_MODEL_PATTERNS = ("gpt-5.5", "gpt-5.5-*", "gpt-5.4", "gpt-5.4-*")
 
+_REASONING_LEVELS_STANDARD = (
+    ReasoningLevel(effort="low", description="Low reasoning effort"),
+    ReasoningLevel(effort="medium", description="Medium reasoning effort"),
+    ReasoningLevel(effort="high", description="High reasoning effort"),
+)
+
+_REASONING_LEVELS_EXTENDED = (
+    ReasoningLevel(effort="low", description="Low reasoning effort"),
+    ReasoningLevel(effort="medium", description="Medium reasoning effort"),
+    ReasoningLevel(effort="high", description="High reasoning effort"),
+    ReasoningLevel(effort="xhigh", description="Extra high reasoning effort"),
+)
+
+# Static fallback models used before the first upstream registry refresh.
+# These ensure /models and /backend-api/codex/models return known slugs
+# immediately on startup (e.g. before any account is added or the first
+# refresh completes). Values are conservative best-effort approximations;
+# the live upstream data always takes precedence once available.
+_BOOTSTRAP_STATIC_MODELS: tuple[UpstreamModel, ...] = (
+    UpstreamModel(
+        slug="gpt-5.5",
+        display_name="GPT-5.5",
+        description="GPT-5.5",
+        context_window=1_050_000,
+        input_modalities=("text",),
+        supported_reasoning_levels=_REASONING_LEVELS_EXTENDED,
+        default_reasoning_level="high",
+        supports_reasoning_summaries=True,
+        support_verbosity=False,
+        default_verbosity=None,
+        prefer_websockets=True,
+        supports_parallel_tool_calls=False,
+        supported_in_api=True,
+        minimal_client_version=None,
+        priority=0,
+        available_in_plans=frozenset({"plus", "pro", "team", "business", "enterprise"}),
+    ),
+    UpstreamModel(
+        slug="gpt-5.5-pro",
+        display_name="GPT-5.5 Pro",
+        description="GPT-5.5 Pro",
+        context_window=1_050_000,
+        input_modalities=("text",),
+        supported_reasoning_levels=_REASONING_LEVELS_EXTENDED,
+        default_reasoning_level="high",
+        supports_reasoning_summaries=True,
+        support_verbosity=False,
+        default_verbosity=None,
+        prefer_websockets=True,
+        supports_parallel_tool_calls=False,
+        supported_in_api=True,
+        minimal_client_version=None,
+        priority=0,
+        available_in_plans=frozenset({"pro", "enterprise"}),
+    ),
+    UpstreamModel(
+        slug="gpt-5.4",
+        display_name="GPT-5.4",
+        description="GPT-5.4",
+        context_window=1_050_000,
+        input_modalities=("text",),
+        supported_reasoning_levels=_REASONING_LEVELS_EXTENDED,
+        default_reasoning_level="high",
+        supports_reasoning_summaries=True,
+        support_verbosity=False,
+        default_verbosity=None,
+        prefer_websockets=True,
+        supports_parallel_tool_calls=False,
+        supported_in_api=True,
+        minimal_client_version=None,
+        priority=0,
+        available_in_plans=frozenset({"plus", "pro", "team", "business", "enterprise"}),
+    ),
+    UpstreamModel(
+        slug="gpt-5.4-mini",
+        display_name="GPT-5.4 Mini",
+        description="GPT-5.4 Mini",
+        context_window=400_000,
+        input_modalities=("text",),
+        supported_reasoning_levels=_REASONING_LEVELS_STANDARD,
+        default_reasoning_level="medium",
+        supports_reasoning_summaries=True,
+        support_verbosity=False,
+        default_verbosity=None,
+        prefer_websockets=False,
+        supports_parallel_tool_calls=False,
+        supported_in_api=True,
+        minimal_client_version=None,
+        priority=0,
+        available_in_plans=frozenset({"free", "plus", "pro", "team", "business", "enterprise"}),
+    ),
+    UpstreamModel(
+        slug="gpt-5.3-codex",
+        display_name="GPT-5.3 Codex",
+        description="GPT-5.3 Codex",
+        context_window=272_000,
+        input_modalities=("text",),
+        supported_reasoning_levels=_REASONING_LEVELS_EXTENDED,
+        default_reasoning_level="high",
+        supports_reasoning_summaries=True,
+        support_verbosity=False,
+        default_verbosity=None,
+        prefer_websockets=True,
+        supports_parallel_tool_calls=False,
+        supported_in_api=True,
+        minimal_client_version=None,
+        priority=0,
+        available_in_plans=frozenset({"plus", "pro", "team", "business", "enterprise"}),
+    ),
+)
+
 
 class ModelRegistry:
     def __init__(self, *, ttl_seconds: float = 300.0) -> None:
@@ -57,7 +168,7 @@ class ModelRegistry:
             raise ValueError("ttl_seconds must be positive")
         self._ttl_seconds = ttl_seconds
         self._snapshot: ModelRegistrySnapshot | None = None
-        self._bootstrap_models: dict[str, UpstreamModel] = {}
+        self._bootstrap_models: dict[str, UpstreamModel] = {m.slug: m for m in _BOOTSTRAP_STATIC_MODELS}
         self._lock = anyio.Lock()
 
     def get_snapshot(self) -> ModelRegistrySnapshot | None:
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by Soju06** [comment]
Thanks @sudhanthirapriya -- the bootstrap-empty -\u003e empty registry response failure mode you describe is real. I have one ask before we merge.

## Issue: `gpt-5.5-pro` is not in the upstream registry response

I verified earlier (#482) against my dev codex-lb with a Pro account and a Plus account active that the upstream `/backend-api/codex/models` returns these slugs and **only** these slugs, identically across both plans:

```
gpt-5.5
gpt-5.4
gpt-5.4-mini
gpt-5.3-codex
gpt-5.3-codex-spark
gpt-5.2
codex-auto-review
```

There is no `gpt-5.5-pro` slug today. The `gpt-5.5` entry includes `pro` in `available_in_plans`, but no separate `gpt-5.5-pro` exists. Putting it in `_BOOTSTRAP_STATIC_MODELS` means a CLI hitting `/v1/models` during the bootstrap window will see a model that does not exist upstream, then fail when it tries to use it -- exactly the same class of "phantom model metadata" symptom #480 was about, just with a different shape.

I would either:

- Drop `gpt-5.5-pro` from the static set. `gpt-5.5` itself with `pro` in `available_in_plans` already covers Pro accounts.
- Or, if you want to mirror the actual upstream registry exactly, add `gpt-5.3-codex-spark`, `gpt-5.2`, and `codex-auto-review` and drop `gpt-5.5-pro`.

The first option is simpler and gives clients exactly the slugs upstream will hand back later -- happy with either.

## Smaller item -- not a blocker

Some of the static `available_in_plans` / `context_window` values are best-effort approximations. Once the first refresh lands the live values overwrite them, so this only affects the first ~5 second window. Could you cross-check `available_in_plans` for `gpt-5.4-mini` (your value adds `free`, which I cannot directly verify -- consider matching the conservative `{plus, pro, team, business, enterprise}` shape used by the other entries) and `context_window=1_050_000` for `gpt-5.5` / `gpt-5.4` if you have a reference for those?

If you would rather just drop `gpt-5.5-pro` and ship, I am happy with that. CI workflow approval is in -- once the static set is updated I will re-run `@codex review` and we can land it.

(Heads-up: I will not require all of the smaller cross-checks before merge; only the `gpt-5.5-pro` removal is a blocker. The `available_in_plans` and `context_window` values get clobbered by the first live refresh anyway.)

**Comment by Soju06** [comment]
Thanks @sudhanthirapriya, the framing is correct -- before the first registry refresh, `/models` and `/backend-api/codex/models` returning empty did cause the Codex CLI `Model metadata for gpt-5.5 not found` log, and a non-empty bootstrap is the right shape of fix. But the current bootstrap diverges from upstream in a few places that we have to clean up before merging, and the two CI failures are flagging exactly that gap.

I just pulled live upstream registry metadata from our dev instance against the actual ChatGPT backend (single Pro account, codex CLI client version `0.124.0`) and compared it line by line to the static dict in this PR. Posting the deltas so you don't have to reverse-engineer them.

## 1. `gpt-5.5-pro` is not in the upstream registry

Live response from `GET /backend-api/codex/models?client_version=0.124.0`:

```
gpt-5.5
gpt-5.4
gpt-5.4-mini
gpt-5.3-codex
gpt-5.3-codex-spark
gpt-5.2
codex-auto-review
```

No `gpt-5.5-pro` slug is returned by upstream for any plan, including Pro. This was independently observed in #482 as well. Bootstrapping it would advertise a slug clients can see in `/models` but that upstream then rejects when called, which is worse than the original `fallback metadata` warning -- it surfaces as a 5xx on a real request.

Please drop the `gpt-5.5-pro` entry.

## 2. Three real upstream slugs are missing from the bootstrap

The current bootstrap includes only `gpt-5.5`, `gpt-5.5-pro`, `gpt-5.4`, `gpt-5.4-mini`, `gpt-5.3-codex`. Upstream actually advertises:

```diff
  gpt-5.5
- gpt-5.5-pro       # phantom, drop
  gpt-5.4
  gpt-5.4-mini
  gpt-5.3-codex
+ gpt-5.3-codex-spark
+ gpt-5.2
+ codex-auto-review
```

Without the additional three, the bootstrap-only window still 503s clients that try those slugs. The whole point of the bootstrap is to cover the `~60s before first refresh` and any account-refresh-stalled state, so any model that upstream actually serves should be in there.

`codex-auto-review` is the upstream slug for the `@codex review` workflow we use in this repo; it has to be in bootstrap or PR reviews break for fresh installs.

## 3. The static field values need to be aligned with upstream

This is the larger one. The fields in the current static dict don't match what upstream returns for the same slugs. Concretely, for `gpt-5.5` upstream reports:

```json
{
  "default_reasoning_level": "medium",
  "support_verbosity": true,
  "default_verbosity": "low",
  "supports_parallel_tool_calls": true,
  "context_window": 272000,
  "input_modalities": ["text", "image"],
  "minimal_client_version": "0.124.0",
  "available_in_plans": ["plus", "pro", "prolite", "team", "business", "enterprise", "edu", "education", "k12", "go", "hc", "finserv", "free", "free_workspace", "quorum", "self_serve_business_usage_based", "enterprise_cbp_usage_based"],
  "prefer_websockets": true
}
```

The PR's static entry for `gpt-5.5` reports:

```json
{
  "default_reasoning_level": "high",
  "support_verbosity": false,
  "default_verbosity": null,
  "supports_parallel_tool_calls": false,
  "context_window": 1050000,
  "input_modalities": ["text"],
  "minimal_client_version": null,
  "available_in_plans": ["plus", "pro", "team", "business", "enterprise"],
  "prefer_websockets": true
}
```

Every field except `prefer_websockets` is wrong relative to upstream. The effects are real: `supports_parallel_tool_calls=false` will lose tool-call parallelism on the bootstrap path, `input_modalities=["text"]` will reject image inputs on freshly-restarted instances, `context_window=1050000` over-reports the limit, the narrower `available_in_plans` will keep free / prolite / edu accounts out of routing for the bootstrap window, and missing `minimal_client_version` will let older Codex clients hit the model and then break on payload features they don't actually support.

Please pull the per-model values straight from the upstream registry response. If you don't have a Pro account to hit it with, I can attach a sanitized JSON snapshot here (slug + every field except the long `base_instructions` blob) so you can paste the values in.

## 4. The two CI failures are expected and want test updates

Both failures are pointing at the same test pair:

```
FAILED tests/integration/test_v1_models.py::test_v1_models_empty_when_registry_not_populated
FAILED tests/integration/test_v1_models.py::test_backend_codex_models_empty_when_registry_not_populated
```

Those tests assert that `/models` returns an empty list when the registry has not been refreshed yet. With this PR they will (correctly) return the bootstrap models instead, so the tests are obsolete. Please:

- delete or rename them to reflect the new contract (e.g. `test_v1_models_returns_bootstrap_when_registry_not_populated`)
- add a positive assertion that the bootstrap list is what is returned, matching the upstream slug set in (2) so future drift fails CI

## 5. Optional: persist last-known-good snapshot

Outside the scope of this PR but worth a separate small follow-up: instead of relying on a hardcoded static dict, the registry could persist its last successful refresh as a snapshot (DB or local file) and rehydrate from that on startup. That removes the maintenance debt of keeping `_BOOTSTRAP_STATIC_MODELS` in sync every time upstream changes a field. Not blocking this PR; if you want to take that on after the static-dict fix, that would be a nice direction.

## Summary of asks for this PR

1. drop `gpt-5.5-pro` from the static dict.
2. add `gpt-5.3-codex-spark`, `gpt-5.2`, and `codex-auto-review` to the static dict using their upstream field values.
3. align all existing per-model fields (reasoning levels, verbosity, parallel tool calls, context window, input modalities, available_in_plans, minimal_client_version) with upstream.
4. rewrite or rename the two registry-empty tests so the new contract is enforced and CI is green again.

Once you push, I'll re-run `@codex review` and we can land it.

If pulling the upstream values is awkward to do on your side, ping me here and I'll attach a JSON snapshot you can copy from. Thanks again for picking the underlying issue up.

Komzpa:
@codex review

Soju06:

Komzpa:
Revived on a maintainer-owned branch as #750 because this PR head has maintainerCanModify=false and cannot be updated here. Keeping credit to the original implementation in the replacement PR.

Issue #479, opened 2026-04-24
Plus accounts showing quote exceeded


Chicken3veryDay:
<img width="1622" height="874" alt="Image" src="https://github.com/user-attachments/assets/c0a2be2d-b58e-4ef5-9efc-ec6421a8525b" />
All my team accounts are working properly, but my plus accounts have no requests being routed to them and are showing "Quota Exceeded", despite them being authed and having max of both 5h Window and 1 Week Window quotas, if you need more info please let me know.

dhpancor:
Happening to me too, in my Prolite account.

Edit: I stopped the container and started it again and it's working fine now.

Soju06:
FYI: Possibly resolved by #559 (merged in `ee747d3`). The patch recovers `quota_exceeded` accounts to `active` when a successful usage refresh reports available quota -- if your case was a Plus account stuck at `quota_exceeded` status while the dashboard showed remaining usage, this should clear it on the next usage refresh after upgrading to v1.16.x.

Could you re-test after the next release and confirm whether the Plus account still appears as quota_exceeded? If it now self-recovers, this can be closed; if not, please share a fresh `/api/accounts` snapshot so we can dig further.

Soju06:
Closing — this should be resolved by #559 (`fix(accounts): recover quota status from usage refresh`), which shipped in v1.17.0.

The fix makes the usage refresh path actively reconcile accounts stuck in `quota_exceeded` once the underlying quota window is available again, instead of relying solely on `reset_at` to trigger the transition. Plus accounts that were showing `quota_exceeded` while the upstream API was actually allowing requests should now recover on the next refresh cycle.

If you're on v1.17.0 (or later) and still see plus accounts incorrectly stuck in `quota_exceeded`, please open a new issue with:
- codex-lb version
- account plan (Plus / Pro / Team / etc.)
- the `/api/accounts` body for the affected account at the time of the incident

PR #529, opened 2026-04-30
fix(status): add auto heal for rate-limit/ quota exceed status after reset in the ui

huzky-v:
Address #479 

Probably fix the issue that the UI still show the stale status after the time resets
As currently the status update is more like `lazy update`, if there is no request to `codex-lb` or the account does not tide to any API, the status shown in UI will never be updated.
Not sure why the request does not get routed, but this would best effort reset the status, as a fallback, with the help of scheduler

This fix is to introduce the account status for those account status = rate-limit/ quota exceed in the scheduler

### DIFF
--- a/app/core/usage/refresh_scheduler.py
+++ b/app/core/usage/refresh_scheduler.py
@@ -8,9 +8,11 @@ from dataclasses import dataclass, field
 from typing import Protocol, cast
 
 from app.core.config.settings import get_settings
+from app.db.models import Account, AccountStatus, UsageHistory
 from app.db.session import get_background_session
 from app.modules.accounts.repository import AccountsRepository
 from app.modules.proxy.account_cache import get_account_selection_cache
+from app.modules.proxy.load_balancer import background_recovery_state_from_account
 from app.modules.proxy.rate_limit_cache import get_rate_limit_headers_cache
 from app.modules.usage import updater as usage_updater_module
 from app.modules.usage.repository import AdditionalUsageRepository, UsageRepository
@@ -18,9 +20,31 @@ from app.modules.usage.updater import UsageUpdater
 
 logger = logging.getLogger(__name__)
 
+_RECOVERABLE_ACCOUNT_STATUSES = frozenset({AccountStatus.RATE_LIMITED, AccountStatus.QUOTA_EXCEEDED})
+
 
 class _LeaderElectionLike(Protocol):
     async def try_acquire(self) -> bool: ...
+
+
+class _RecoverableAccountsRepository(Protocol):
+    async def update_status_if_current(
+        self,
+        account_id: str,
+        status: AccountStatus,
+        deactivation_reason: str | None = None,
+        reset_at: int | None = None,
+        blocked_at: int | None | object = None,
+        *,
+        expected_status: AccountStatus,
+        expected_deactivation_reason: str | None = None,
+        expected_reset_at: int | None = None,
+        expected_blocked_at: int | None | object = None,
+    ) -> bool: ...
+
+
+class _LatestUsageRepository(Protocol):
+    async def latest_by_account(self, window: str | None = None) -> dict[str, UsageHistory]: ...
 
 
 def _get_leader_election() -> _LeaderElectionLike:
@@ -74,6 +98,12 @@ class UsageRefreshScheduler:
                     accounts = await accounts_repo.list_accounts()
                     updater = UsageUpdater(usage_repo, accounts_repo, additional_usage_repo)
                     await updater.refresh_accounts(accounts, latest_usage)
+                    refreshed_accounts = await accounts_repo.list_accounts()
+                    await reconcile_recoverable_account_statuses(
+                        accounts_repo=accounts_repo,
+                        usage_repo=usage_repo,
+                        accounts=refreshed_accounts,
+                    )
                     await get_rate_limit_headers_cache().invalidate()
                     get_account_selection_cache().invalidate()
             except Exception:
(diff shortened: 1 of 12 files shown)


Soju06:
@codex review

Soju06:
Thanks @huzky-v -- the design itself is clean: reusing `_state_from_account` from a synthetic runtime, gating writes through `update_status_if_current`, and complementing the in-process recovery in #559 (which only covers `QUOTA_EXCEEDED` from the usage path, while this PR covers both `RATE_LIMITED` and `QUOTA_EXCEEDED` and survives cross-process / cold-start stuck rows). The OpenSpec design.md is genuinely good.

I checked merge-tree against current main (post-#559) and there is no text or semantic conflict -- the two paths are complementary (sync vs periodic / single-process vs cross-process). So functionally we're good to land this once one P1 is addressed.

## P1 -- AsyncSession concurrent use

Codex review caught this and it's correct:

```python
# app/core/usage/refresh_scheduler.py:130
latest_primary, latest_secondary = await asyncio.gather(
    usage_repo.latest_by_account(window="primary"),
    usage_repo.latest_by_account(window="secondary"),
)
```

`usage_repo` here is bound to a single `AsyncSession` (created in the surrounding `async with get_background_session()` block earlier in `_refresh_once`). SQLAlchemy explicitly documents that `AsyncSession` is not safe for concurrent use across tasks -- it carries per-session state (current transaction, identity map, dialect-level cursors) that breaks when two coroutines fence it at the same time.

When this hits, the failure mode is one of two SQLAlchemy errors:

- `InvalidRequestError: This session is already in use` (the friendly one)
- `StatementError` with a `MissingGreenlet` or a closed-cursor message (the less friendly one)

Either way the surrounding `try ... except Exception:` in `_refresh_once` swallows it and the whole reconcile pass plus the usage refresh in the same cycle become a no-op. Worst case the scheduler loops at the configured interval with both responsibilities silently failing.

## Suggested fix

Sequential calls are the simplest and match the SQLAlchemy contract:

```python
latest_primary = await usage_repo.latest_by_account(window="primary")
latest_secondary = await usage_repo.latest_by_account(window="secondary")
```

The two queries don't share a critical section that benefits from parallelism (no cross-DB joins, no upstream calls in between), and `latest_by_account` is a small indexed read. Sequential is the right shape here.

If we ever do want parallelism on a future variant -- for example to also pull additional-quota windows in the same cycle -- the right pattern is per-task sessions:

```python
async with get_background_session() as session_primary, get_background_session() as session_secondary:
    primary_repo = UsageRepository(session_primary)
    secondary_repo = UsageRepository(session_secondary)
    latest_primary, latest_secondary = await asyncio.gather(
        primary_repo.latest_by_account(window="primary"),
        secondary_repo.latest_by_account(window="secondary"),
    )
```

But that is overkill for two small reads -- the sequential pattern above is fine.

## Tests to update

The existing `test_usage_refresh_scheduler_recovery.py` uses a stub `StubUsageRepository` so it doesn't exercise the real session. Two options:

- add a focused unit test that constructs a real `UsageRepository` against an in-memory SQLite session and runs `reconcile_recoverable_account_statuses` against it. The point is to ensure the sequential pattern survives the SQLAlchemy contract, not to exercise the logic itself.
- or add a small note to `design.md` explaining the sequential read choice so future contributors don't re-introduce the `gather`.

Whichever is easier on your side.

## Not blocking otherwise

The rest of the PR is in good shape and the merge-tree against main is clean. Once you push the sequential read I'll re-run `@codex review` and we can land it. Thanks again for the fix -- the periodic-reconciliation safety net complements #559 well.

huzky-v:
@codex review

Soju06:
@codex review

Soju06:
Thanks @huzky-v -- the AsyncSession P1 (`asyncio.gather` -> sequential awaits) is fixed and the re-review picked up a different P1 that needs to be addressed before merging. This one is more impactful because it can re-introduce the 429 it's trying to recover from.

## P1 -- rate-limit cooldown is bypassed in recovery state

In `app/modules/proxy/load_balancer.py:1169`, `background_recovery_state_from_account` seeds `runtime.cooldown_until` with `blocked_at`:

```python
runtime.cooldown_until = blocked_at  # in the past as soon as we evaluate it
```

This causes the RATE_LIMITED branch of `_state_from_account` to treat the cooldown as expired the moment we evaluate the account, because `blocked_at` is always in the past. The flow that breaks:

1. Upstream rate-limits account; we set `status = RATE_LIMITED`, `blocked_at = T`, with a real retry delay (e.g., 30 minutes).
2. Some unrelated usage refresh writes a primary row at `T + 5min` while the account is still genuinely cooling down.
3. Recovery path runs `background_recovery_state_from_account`, seeds `cooldown_until = blocked_at = T` (in the past).
4. `_state_from_account` sees `cooldown_until` is expired, clears `runtime_reset`, flips the account back to `ACTIVE`.
5. We immediately route to the account and re-trigger 429 (and now the cooldown shows the new `blocked_at`, looping the same logic 30 minutes early every cycle).

The intent of the recovery scheduler is the opposite -- only flip an account back to `ACTIVE` after the real reset has elapsed. Seeding with `blocked_at` reads more like a typo for the actual reset time.

## Suggested fix

Seed `cooldown_until` with the real cooldown deadline, not the moment the block started:

```python
# app/modules/proxy/load_balancer.py:~1169
runtime.cooldown_until = (
    account.rate_limit_reset_at      # if a hard reset timestamp is known
    or (account.blocked_at + retry_delay)  # else derive from blocked_at + delay
)
```

(Exact field names depend on what `Account`/recovery state actually exposes -- the point is to use the reset deadline, not the block start.)

If the recovery state has no notion of "how long the cooldown is" because that info lives elsewhere, then the recovery path should **not** clear `runtime_reset` in this branch at all -- let the original RATE_LIMITED state's cooldown gate make the active/blocked decision.

## Tests to add

```python
def test_recovery_state_preserves_rate_limit_cooldown_when_block_recent():
    account = _make_account(
        status=AccountStatus.RATE_LIMITED,
        blocked_at=now - timedelta(minutes=5),
        rate_limit_reset_at=now + timedelta(minutes=25),
    )
    state = background_recovery_state_from_account(account)
    runtime = _state_from_account(account, state)
    assert runtime.status == AccountStatus.RATE_LIMITED  # not ACTIVE
    assert runtime.cooldown_until > now

def test_recovery_state_recovers_account_after_cooldown_elapsed():
    account = _make_account(
        status=AccountStatus.RATE_LIMITED,
        blocked_at=now - timedelta(hours=2),
        rate_limit_reset_at=now - timedelta(minutes=5),  # already past reset
    )
    state = background_recovery_state_from_account(account)
    runtime = _state_from_account(account, state)
    assert runtime.status == AccountStatus.ACTIVE
```

## Not blocking otherwise

- The `asyncio.gather` -> sequential await change for `latest_by_account` is correct (one shared `AsyncSession` per scheduler tick).
- The UI side of the auto-heal is fine; the issue is in the gating before we display "recovered".

Once this is fixed I'll re-run `@codex review` and queue for merge.

huzky-v:
@codex review

huzky-v:
@codex review

huzky-v:
@codex review

Soju06:
## Current-main smoke failure before merge

Live PR gates are green (`codex: ok`, 18/18 CI, `CLEAN/MERGEABLE`), but the PR no longer passes a focused no-commit merge smoke against the current `origin/main`.

Repro I ran:

```bash
git fetch origin main pull/529/head:refs/remotes/origin/pr-529 --prune
git worktree add --detach .worktree/review-pr-529 refs/remotes/origin/pr-529
cd .worktree/review-pr-529
git merge --no-commit --no-ff origin/main

uv run --locked openspec validate reconcile-background-account-status --strict
uv run --locked ruff check \
  app/core/usage/refresh_scheduler.py \
  app/modules/proxy/load_balancer.py \
  tests/integration/test_accounts_api_extended.py \
  tests/unit/test_load_balancer.py \
  tests/unit/test_request_locality.py \
  tests/unit/test_usage_refresh_scheduler_recovery.py
uv run --locked ty check \
  app/core/usage/refresh_scheduler.py \
  app/modules/proxy/load_balancer.py \
  tests/integration/test_accounts_api_extended.py \
  tests/unit/test_load_balancer.py \
  tests/unit/test_request_locality.py \
  tests/unit/test_usage_refresh_scheduler_recovery.py
uv run --locked pytest -q -ra \
  tests/integration/test_accounts_api_extended.py \
  tests/unit/test_load_balancer.py \
  tests/unit/test_request_locality.py \
  tests/unit/test_usage_refresh_scheduler_recovery.py \
  --durations=20
```

OpenSpec / ruff / ty pass, but pytest fails:

```text
FAILED tests/integration/test_accounts_api_extended.py::test_accounts_list_stale_rate_limited_status_recovers_after_background_reconcile

assert stale_account["status"] == "rate_limited"
E AssertionError: assert 'active' == 'rate_limited'
E   - rate_limited
E   + active
```

This looks like drift against current `main`: `/api/accounts` now returns the effective status as `active` for this expired/stale `rate_limited` account, while the PR test still expects the stale persisted `rate_limited` state before background reconciliation.

Could you rebase/merge current `main` and update the PR to align the behavior/test with the current `/api/accounts` effective-status semantics? After that, the focused pytest command above should pass on the merged tree.

Komzpa:
@codex review

huzky-v:
@codex review

Komzpa:
@codex review

Soju06:

Komzpa:
Revived on a maintainer-owned branch as #754 because this PR head has maintainerCanModify=false and cannot be updated here. Keeping credit to the original implementation in the replacement PR.

Issue #392, opened 2026-04-12
hermes agent compability tools settings and subagents json parsing error


cemalisaltik:
hey im just a beginner but i seem to have a problem connecting this beautiful repo I tried to build in last month it took me hours... So sorry for the beginner slop but the hermes does not seem to be able to call subagents and the output format etc i think makes this tool less useful in some cases when combined with these agent harnesses. I would be grateful to have a reply. Thank you for your time please keep up the good work!! 
I made gemini summarise my problem (not the whole thing this probably is the part where i made my self into a alice in wonderland hole. Sorry if it sounds dumb  : The core problem seems to be a Protocol Violation during the SSE (Server-Sent Events) streaming phase, especially on the /v1/chat/completions endpoint.

Observation: When the system encounters a rate limit or "No active accounts available" state (503), it injects a custom SSE event type: codex.rate_limits.

Root Cause: Strict OpenAI-compatible parsers (used by agentic frameworks like Hermes) expect the stream to start with a standard OpenAI handshake (like response.created or a data chunk). When they receive an unrecognized event type like codex.rate_limits before the first real data, it causes a RuntimeError: Expected to have received response.created before codex.rate_limits.

Furthermore, injecting these custom events into the stream seems to cause "Extra data" errors in tool-call parsing (e.g., Invalid JSON in tool call arguments: Extra data: line 1 column 182), as the metadata is being prepended/appended to the JSON body that the agent expects to be a clean tool-specification.

Suggested Fix: Implement a "Strict Compatibility Mode" for the /v1/ endpoints. The proxy should ideally suppress or filter out all codex.* event types from the SSE stream for clients that expect standard OpenAI wire format, ensuring telemetry doesn't pollute the response data.
🔀 preparing delegate_task…  
[subagent-1] ⚠️ Invalid JSON in tool call arguments for 'search_files': Extra data: line 1 column 182 (char 181)
[subagent-1] 🔄 Retrying API call (1/3)...
[subagent-0] ⚠️ Invalid JSON in tool call arguments for 'search_files': Extra data: line 1 column 179 (char 178)
[subagent-0] 🔄 Retrying API call (1/3)...
[subagent-1] ⚠️ Invalid JSON in tool call arguments for 'search_files': Extra data: line 1 column 182 (char 181)
[subagent-1] 🔄 Retrying API call (2/3)...
[subagent-0] ⚠️ Invalid JSON in tool call arguments for 'search_files': Extra data: line 1 column 172 (char 171)

[...]

PR #533, opened 2026-05-01
fix(proxy): emit response.created before buffered stream events

### LINKED ISSUE #473: Responses API SSE event order is incompatible with Hermes/OpenAI SDK clients
Hi, I’m trying to use Codex LB as the backend for Hermes Agent / Hermes WebUI.

My Hermes side is already configured specifically for this backend, not as generic OpenAI /v1:

provider: custom
api_mode: codex_responses
base_url: https://196.168.110.5:2285/backend-api/codex
model: gpt-5.4
So this is not a wrong-base-url issue.
What works:

GET /backend-api/codex/models -> 200
POST /backend-api/codex/responses -> 200

What does not exist / is not accepted:
GET /backend-api/codex/v1/models -> 404
POST /backend-api/codex/v1/responses -> 405
POST /backend-api/codex/chat/completions -> 405
POST /backend-api/codex/v1/chat/completions -> 405

So I correctly switched Hermes to the backend-api/codex + responses path.
The actual problem is protocol compatibility of the Responses SSE stream.

Observed Hermes/OpenAI SDK errors:

No response received: Expected to have received response.created before codex.rate_limits
API call failed after 3 retries: Expected to have received response.created before response.reasoning_summary_text.delta

This strongly suggests the SSE event ordering is incompatible with clients that rely on the OpenAI Responses streaming parser.
From direct probing, Codex LB emits events like:

codex.rate_limits
response.created
response.in_progress
...

So codex.rate_limits may appear before response.created.
That breaks clients which expect response.created to be the first response lifecycle event.

I also observed that even when sending:

"stream": false

the server still returns:
Content-Type: text/event-stream

instead of a normal JSON response body.
Expected behavior:

For Responses SSE, emit response.created before any custom or downstream events.
Do not emit codex.rate_limits before response.created.
Do not emit response.reasoning_summary_text.delta before response.created.
If stream=false, return a normal JSON response instead of SSE.
If codex.rate_limits is a custom extension, please either:
- move it after response.created, or
- make it optional/disable-able for compatibility.
Why this matters:
Many clients that support OpenAI Responses API do not just “parse whatever arrives”. They rely on the official/expected event ordering. Right now Codex LB appears reachable and authenticated correctly, but the stream format/order makes Hermes and likely other OpenAI-SDK-based clients fail.

Lotfree618:
## Summary

Fixes #473.

Related: #392, NousResearch/hermes-agent#8133.

Responses streams can currently forward non-terminal events before `response.created`. Strict OpenAI SDK Responses parsers expect `response.created` first so they can build the initial response snapshot; otherwise clients such as Hermes fail before consuming the stream.

This buffers pre-`response.created` non-terminal events per request, emits `response.created` first, then flushes the buffered events in their original order.

## Changes

- Add per-request buffering for pre-`response.created` events.
- Apply the ordering fix to both WebSocket and HTTP bridge stream paths.
- Add unit coverage for both paths.

## Testing

- `python -m pytest tests/unit/test_proxy_utils.py tests/unit/test_proxy_http_bridge.py`
- `ruff check app/modules/proxy/service.py tests/unit/test_proxy_utils.py tests/unit/test_proxy_http_bridge.py`

### DIFF
--- a/app/modules/proxy/service.py
+++ b/app/modules/proxy/service.py
@@ -5215,6 +5215,7 @@ class ProxyService:
             retry_text_data = request_state.fresh_upstream_request_text
         if request_state.replay_count >= 1:
             return False
+        request_state.pre_response_created_event_texts = []
         request_state.replay_count += 1
         _log_http_bridge_event(
             "retry_fresh_upstream",
@@ -5264,7 +5265,7 @@ class ProxyService:
                 return False
             request_text = request_state.request_text
             assert isinstance(request_text, str)
-            request_state.replay_count += 1
+            request_state.reset_for_replay()
         _log_http_bridge_event(
             "retry_precreated",
             session.key,
@@ -6155,15 +6156,11 @@ class ProxyService:
         if retry_error_code is not None:
             upstream_control.reconnect_requested = True
             if retry_is_previous_response_not_found:
-                request_state.replay_count += 1
-                request_state.awaiting_response_created = True
-                request_state.response_id = None
+                request_state.reset_for_replay()
                 upstream_control.suppress_downstream_event = True
                 upstream_control.replay_request_state = request_state
             else:
-                request_state.replay_count += 1
-                request_state.awaiting_response_created = True
-                request_state.response_id = None
+                request_state.reset_for_replay()
                 upstream_control.suppress_downstream_event = True
                 upstream_control.replay_request_state = request_state
                 await self._handle_stream_error(
@@ -8322,6 +8319,12 @@ class _WebSocketRequestState:
     input_item_count: int = 0
     input_full_fingerprint: str | None = None
 
+    def reset_for_replay(self) -> None:
+        self.pre_response_created_event_texts = []
+        self.replay_count += 1
+        self.awaiting_response_created = True
+        self.response_id = None
+
 
 @dataclass(frozen=True, slots=True)
 class _HTTPBridgeSessionKey:
@@ -8724,9 +8727,7 @@ async def _pop_replayable_precreated_websocket_request_state(
         if request_state.replay_count >= 1:
             return None
         pending_requests.popleft()
-    request_state.replay_count += 1
-    request_state.awaiting_response_created = True
-    request_state.response_id = None
+    request_state.reset_for_replay()
     return request_state
 
 
--- a/app/modules/proxy/service.py
+++ b/app/modules/proxy/service.py
@@ -205,6 +205,9 @@ async def _await_cancelled_task(
 
 
 _TEXT_DELTA_EVENT_TYPES = frozenset({"response.output_text.delta", "response.refusal.delta"})
+_RESPONSE_STREAM_TERMINAL_EVENT_TYPES = frozenset(
+    {"response.completed", "response.failed", "response.incomplete", "error"}
+)
 _TEXT_DONE_CONTENT_PART_TYPES = frozenset({"output_text", "refusal"})
 _REQUEST_TRANSPORT_HTTP = "http"
 _REQUEST_TRANSPORT_WEBSOCKET = "websocket"
@@ -5430,6 +5433,7 @@ class ProxyService:
         async with session.pending_lock:
             matched_request_state = None
             created_request_state = None
+            precreated_request_state = None
             has_other_pending_requests = False
             grouped_previous_response_request_states: list[_WebSocketRequestState] = []
             if event_type == "response.created":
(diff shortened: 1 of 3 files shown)


mgwals:
I validated this branch because I have a fresh client-side incident that includes a large burst of Codex CLI parser errors:

- `codex-cli 0.128.0`
- local `codex-lb:1.15.0-ws-retry`
- `codex-tui.log` contains `20308` occurrences of `OutputTextDelta without active item`
- one concrete incident had a `turn_aborted`, then a `continua` resume; the next assistant-visible output was serialized internal memory JSON starting with `{"rollout_summary":...}`

That last part is likely a Codex CLI/session-state bug rather than something this PR can fully solve, but this PR addresses the stream ordering class that can make strict Responses consumers lose their initial snapshot: non-terminal events before `response.created`.

Local validation on this PR branch:

- `uv run pytest -q tests/unit/test_proxy_utils.py tests/unit/test_proxy_http_bridge.py` -> `295 passed`
- `uv run openspec validate --specs` -> `19 passed, 0 failed`
- `uv run ruff check app/modules/proxy/service.py tests/unit/test_proxy_utils.py tests/unit/test_proxy_http_bridge.py` -> passed
- `uv run ruff format --check ...` wanted to reformat `app/modules/proxy/service.py` locally with the current lockfile/tooling

Separate live smoke against the existing container, without deploying this PR: 60 small streamed `/v1/responses` requests (`20` each for low/high/xhigh) all reached `response.completed`, and no new error rows appeared after the smoke start.

Soju06:
@codex review

Soju06:
Thanks @Lotfree618 -- the framing is right, the fix is small and matches what strict Responses-API parsers expect, and the helper extraction in `_should_buffer_pre_response_created_event` cleanly shares the buffering policy between the websocket and HTTP-bridge paths. Two things to address before merging.

## P1 -- buffer is not cleared on replay

Codex review caught a real one:

`pre_response_created_event_texts` is persisted on `_WebSocketRequestState` and flushed when `response.created` finally arrives. But the replay / reconnect paths (`_pop_replayable_precreated_websocket_request_state`, `_retry_http_bridge_precreated_request`, the fresh-upstream replay branches) reuse the same `_WebSocketRequestState` instance and resend the original payload. If the failed attempt observed any pre-`response.created` deltas, that buffer survives the retry and gets flushed into the *new* stream on the first `response.created`, duplicating/corrupting downstream output.

Concretely the failure mode is:

1. attempt A: upstream sends `response.in_progress` -> buffered
2. disconnect before `response.created`
3. replay on fresh upstream
4. attempt B: upstream sends `response.created` -> flushes the stale `response.in_progress` from attempt A, then attempt B's real events

Minimal fix: clear the buffer whenever the request is replayed.

```python
# in app/modules/proxy/service.py, anywhere a request is about to be retried on a fresh upstream
request_state.pre_response_created_event_texts = []
request_state.replay_count += 1
```

The two replay entry points to cover:

- `_retry_http_bridge_precreated_request` (`session.upstream.send_text` site)
- the websocket replay branch in `proxy_responses_websocket` that pops a replayable pre-created request and re-sends `request_state.request_text`

A small `_WebSocketRequestState.reset_for_replay()` helper would keep both sites symmetric and survive future fields that grow the same property.

Test we'd want:

```python
async def test_pre_response_created_buffer_is_dropped_on_replay():
    state = _make_websocket_request_state(awaiting_response_created=True)
    state.pre_response_created_event_texts = ['{"type": "response.in_progress"}']
    state.reset_for_replay()
    assert state.pre_response_created_event_texts == []
```

Plus an integration-style test that drives the websocket path through a disconnect-before-`response.created` cycle and asserts the second stream does not emit the stale event.

## CI -- workflow approval

The GitHub Actions workflow has not actually run on this PR yet (only GitGuardian completed). That's because the repo requires maintainer approval for first-time fork contributors' workflow runs. I'll approve the run on the maintainer side after you push the P1 fix -- no action needed from you.

If you already ran the suite locally with `uv run pytest` and `uv run ruff check` and they pass, that's enough on your end for the next round.

## Not blocking otherwise

The rest of the patch is solid. After the P1 fix lands and I approve the workflow, I'll re-run `@codex review` and we can merge.

Soju06:
Hi @Lotfree618 — thanks for the work on this!

Heads up: PR #639 (merged today) ended up covering the same underlying problem with a broader fix — it backfills `response.created` as part of a wider `/v1/responses` OpenAI-SDK-parseability normalization (drops `codex.*` events, synthesizes the create envelope, backfills terminal events). The buffered-stream-ordering case you targeted here is included in that change.

I think this PR is now superseded. If you agree, feel free to close it; otherwise let me know what behavior #639 doesn't cover and we'll take another look.

Issue #409, opened 2026-04-14
Dashboard, accounting: OpenAI post April 2nd Plus thinking time, account burn rate and credits cost


mws-weekend-projects:
According to OpenAI and other sources they changed how plus/pro/... accounts are calculating usage and limits and credits now got 5x more expensive - on par with API cost. Duration of deep thinking per 5h slot might be the new heavy limit in our accounts. A dashboard showing accurate consumption and burn rates (short and long term) are even mire important. It might be extremely useful having a detailed history of % of 5h and 7d stats together with each request in the database. Dashboard should show me the burn rates of my accounts past 5h and past 7d. Current Circle diagrams with "remaining 225" or so-ish might be misleading with the new regime.

Komzpa:
Closing this as completed by the current dashboard accounting work.

The dashboard now includes weekly credits pace and account burn projection surfaces, including 5h/7d burn-rate style visibility. That covers the core request here: making post-April account consumption and burn rates visible instead of relying only on the older remaining-credit circles.

There may still be follow-up dashboard refinements, but this issue's requested burn-rate/accounting surface is now present on `main`.

Issue #371, opened 2026-04-08
[UX] Display weekly/hourly credits


cani1989:
When I saw this label:
<img width="77" height="144" alt="Image" src="https://github.com/user-attachments/assets/63fbf114-96d2-4f49-a8c8-8e0a71dd0f82" />

My first question was: what do these numbers mean?

A clearer presentation would be:

- Use **Weekly Credits** instead of Weekly Remaining
- Show raw numbers like **7331** instead of abbreviated values like **7.33k**
- Show the current amount relative to the total, for example **7331/7560 credits**

PR #534, opened 2026-05-02
feat(dashboard): show weekly token pace

Komzpa:
## Summary

- add a dashboard Weekly pace card that compares actual weekly token consumption against schedule
- compute pace by summing token/credit budgets per account, using each account's own reset time and window
- model weekly reset replenishment as the full account budget, not only current unused credits
- hide Recovery options when the pool is on pace and no throttle/account action is needed
- preserve weekly capacity/remaining fields in the frontend schema and document the behavior in OpenSpec

## Why

A nearly empty account near its weekly reset should not look scary, and a healthy pool should not tell the operator to do recovery work. The card compares actual remaining tokens to expected remaining tokens per account, then derives display percentages from aggregate token totals instead of averaging account percentages.

## Live data sanity check

Checked the real codex-lb usage history around the "tokens were insufficient, then an account was added, then tokens were sufficient" case:

- `2026-05-03 18:13 UTC`: 6-account pool was danger, about 91.5% used vs 76.4% scheduled, short by about 18.8K credits.
- `2026-05-03 18:16 UTC`: after adding the unused Pro account, actual 7-account pool became on track, about 64.1% used vs 53.6% scheduled, projected low-water about 31.7K credits.
- Counterfactual without the new Pro account at `2026-05-03 18:16 UTC` stayed danger, short by about 18.7K credits.

That matches the intended continuity context: adding a real account turns the pool from insufficient to sufficient, and the card should stop showing Recovery options once the pool is safe.

## Validation

- `cd frontend && bun run test -- ./src/features/dashboard/utils.test.ts ./src/features/dashboard/components/weekly-credits-pace-card.test.tsx`
- `cd frontend && bun run typecheck`
- `cd frontend && bun run lint` (passes with two pre-existing warnings in `account-multi-select.tsx`)
- `cd frontend && bun run build` (passes with existing Vite chunk-size/dynamic-import warnings)
- `git diff --check`
- earlier branch validation also covered `.venv/bin/python -m pytest tests/integration/test_dashboard_overview.py::test_dashboard_overview_combines_data`

`openspec` CLI is not installed in this workspace, so I could not run local OpenSpec validation.

## Related issues
- Addresses part of #409 (weekly credit pace / burn-rate visibility).
- Addresses part of #371 (weekly credit display semantics).

### DIFF
--- a/app/core/usage/types.py
+++ b/app/core/usage/types.py
@@ -107,6 +107,7 @@ class UsageTrendBucket:
     samples: int
     reset_at: int | None = None
     window_minutes: int | None = None
+    recorded_at: datetime | None = None
 
 
 @dataclass(frozen=True)
--- a/app/core/usage/types.py
+++ b/app/core/usage/types.py
@@ -105,6 +105,8 @@ class UsageTrendBucket:
     window: str
     avg_used_percent: float
     samples: int
+    reset_at: int | None = None
+    window_minutes: int | None = None
 
 
 @dataclass(frozen=True)
--- a/app/modules/accounts/mappers.py
+++ b/app/modules/accounts/mappers.py
@@ -206,11 +206,8 @@ def build_account_usage_trends(
     # Group buckets by (account_id, window)
     grouped: dict[tuple[str, str], dict[int, float]] = {}
     secondary_schedule: dict[str, dict[int, tuple[int, int]]] = {}
-    secondary_bucket_keys = {(b.account_id, b.bucket_epoch) for b in buckets if b.window == "secondary"}
-    for b in buckets:
+    for b in _effective_usage_trend_buckets(buckets):
         is_weekly_primary = b.window == "primary" and usage_core.is_weekly_window_minutes(b.window_minutes)
-        if is_weekly_primary and (b.account_id, b.bucket_epoch) in secondary_bucket_keys:
-            continue
         window = "secondary" if is_weekly_primary else b.window
         key = (b.account_id, window)
         grouped.setdefault(key, {})[b.bucket_epoch] = b.avg_used_percent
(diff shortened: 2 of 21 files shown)


Soju06:
@codex review

Soju06:
Thanks @Komzpa -- the previous P1 about the first-reset `topUpCredits` model is fixed by `c4bd12d`, and the rest of the follow-ups (per-account windows, expire weekly credits at account reset, hide safe Recovery options, surface ahead pace) all read cleanly. Codex review on the new head surfaced one more P1 to address before merging.

## P1 -- de-duplication should prefer the freshest weekly bucket, not always secondary

The new weekly de-dup in `app/modules/accounts/mappers.py:213` unconditionally discards a `primary` bucket whenever a `secondary` bucket exists at the same timestamp, but elsewhere weekly primary vs secondary is resolved by freshness via `should_use_weekly_primary`. The two paths now disagree, so:

- During ingestion/backfill, `secondary` can legitimately lag `primary`. The account summary picks `primary` (fresher), but the new trend de-dup throws `primary` away and surfaces the stale `secondary` value.
- The dashboard trend/schedule line and the account summary window then disagree for the same account/timestamp -- exactly the kind of visual contradiction the weekly-pace card is trying to remove.

## Suggested fix

Apply the same `should_use_weekly_primary` selection criterion at the de-dup site instead of "always prefer secondary". Sketch:

```python
# app/modules/accounts/mappers.py:~213
def _dedupe_weekly_buckets(buckets: list[WeeklyBucket]) -> list[WeeklyBucket]:
    by_ts: dict[datetime, WeeklyBucket] = {}
    for bucket in buckets:
        existing = by_ts.get(bucket.window_start)
        if existing is None:
            by_ts[bucket.window_start] = bucket
            continue
        # Pick the same way the summary path does.
        if should_use_weekly_primary(primary=primary_of(bucket, existing), secondary=secondary_of(bucket, existing)):
            by_ts[bucket.window_start] = _primary_pick(bucket, existing)
        else:
            by_ts[bucket.window_start] = _secondary_pick(bucket, existing)
    return sorted(by_ts.values(), key=lambda b: b.window_start)
```

(Exact helper shape depends on how `should_use_weekly_primary` accepts its arguments in this codebase -- the point is to reuse it rather than open-code a preference.)

## Tests to add

One direct regression test pinning the rule:

```python
def test_weekly_dedup_prefers_fresher_primary_over_lagging_secondary():
    primary = _bucket(window_start=t0, kind="primary", remaining=1000, recorded_at=t_late)
    secondary = _bucket(window_start=t0, kind="secondary", remaining=2000, recorded_at=t_early)
    [picked] = _dedupe_weekly_buckets([primary, secondary])
    assert picked is primary  # fresher wins, regardless of kind
```

And the symmetric one:

```python
def test_weekly_dedup_prefers_fresher_secondary_when_primary_is_stale():
    primary = _bucket(window_start=t0, kind="primary", remaining=2000, recorded_at=t_early)
    secondary = _bucket(window_start=t0, kind="secondary", remaining=1000, recorded_at=t_late)
    [picked] = _dedupe_weekly_buckets([primary, secondary])
    assert picked is secondary
```

## Not blocking otherwise

- The frontend pace projection (post `c4bd12d`) lines up with the live data sanity check you ran (the "add Pro account -> safe" transition projection matches).
- Hiding Recovery options when on pace + the per-account window split are exactly the UX guardrails this card needs to be useful instead of noisy.

After the P1 fix lands I'll re-run `@codex review` and queue for merge.

Komzpa:
Closing this draft because the weekly token pace work was split into reviewable replacement PRs: #655 for backend/account data and #656 for the dashboard UI. The active stack should use those split PRs instead of this aggregate draft.

PR #535, opened 2026-05-02
fix(proxy): mask websocket previous response misses

mgwals:
## Summary

- Mask direct Responses WebSocket `previous_response_not_found` frames as retryable `stream_incomplete` instead of replaying the same stale `previous_response_id`.
- Keep the existing transparent replay path only for states that already have a verified safe fresh upstream request body.
- Add OpenSpec coverage and update WebSocket/unit tests around the sanitized failure contract.

## Existing tracker check

- Refs #221: existing open issue already covers this `previous_response_not_found` / `stream_incomplete` failure mode, so I did not open a duplicate issue.
- Also checked #516, which overlaps but is stale against current `main` and carries a broader old diff. This PR is a smaller current-main fix.

## Validation

- `.venv/bin/pytest -q tests/integration/test_proxy_websocket_responses.py tests/unit/test_proxy_utils.py`
- `.venv/bin/ruff check app/modules/proxy/service.py tests/integration/test_proxy_websocket_responses.py tests/unit/test_proxy_utils.py`
- `git diff --check`
- `npx --yes @fission-ai/openspec validate mask-websocket-previous-response-miss --strict`
- `npx --yes @fission-ai/openspec validate --specs`
- `codex-review-high --base origin/main`

### DIFF
--- a/app/modules/proxy/service.py
+++ b/app/modules/proxy/service.py
@@ -6126,14 +6126,24 @@ class ProxyService:
             )
             retry_error_code = None
         if retry_error_code is not None:
-            upstream_control.reconnect_requested = True
             if retry_is_previous_response_not_found:
-                request_state.replay_count += 1
-                request_state.awaiting_response_created = True
-                request_state.response_id = None
-                upstream_control.suppress_downstream_event = True
-                upstream_control.replay_request_state = request_state
+                if not (
+                    request_state.fresh_upstream_request_is_retry_safe and request_state.fresh_upstream_request_text
+                ):
+                    retry_error_code = None
+                else:
+                    upstream_control.reconnect_requested = True
+                    request_state.request_text = request_state.fresh_upstream_request_text
+                    request_state.previous_response_id = None
+                    request_state.proxy_injected_previous_response_id = False
+                    request_state.fresh_upstream_request_is_retry_safe = False
+                    request_state.replay_count += 1
+                    request_state.awaiting_response_created = True
+                    request_state.response_id = None
+                    upstream_control.suppress_downstream_event = True
+                    upstream_control.replay_request_state = request_state
             else:
+                upstream_control.reconnect_requested = True
                 request_state.replay_count += 1
                 request_state.awaiting_response_created = True
                 request_state.response_id = None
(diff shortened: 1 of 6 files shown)


Komzpa:
Follow-up validation for the live `previous_response_not_found` leak we saw today.

I checked a real raw leak with `previous_response_id=resp_0bebc32c66e5a5990169f9d272ffbc8191801046ca030a311b` from `2026-05-05`.

What happened:

- `2026-05-05 11:20:19 UTC`: upstream created `resp_0beb...d272...`.
- `2026-05-05 11:20:24 UTC`: upstream returned `response.completed` for that response.
- `2026-05-05 11:21:27 UTC`: the same response stream recorded `no close frame received or sent`.
- `2026-05-05 11:22:25 UTC`: the next turn sent `previous_response_id=resp_0beb...d272...` and got raw `previous_response_not_found` / `previous_response_id` / `400` downstream.

I also checked whether this was a cross-account routing miss after adding a new account. It was not: the created/completed response, the follow-up request, and the raw error all used the same account family `c26b2ebf...`. A new account `fe2708b6...` appears nearby in the archive, but for a different parallel request.

So the desired behavior for this PR is fail-closed masking: when upstream loses continuity even on the owner account, downstream should see `502 stream_incomplete`, not the raw upstream `previous_response_not_found` or `resp_*` id.

Local focused validation on this PR head:

- `/home/kom/proj/codex-lb/.venv/bin/python -m pytest -q tests/integration/test_proxy_websocket_responses.py tests/unit/test_proxy_utils.py` (`247 passed`)
- `/home/kom/proj/codex-lb/.venv/bin/ruff check app/modules/proxy/service.py tests/integration/test_proxy_websocket_responses.py tests/unit/test_proxy_utils.py`
- `git diff --check`

I tried to update the PR body with this, but this PR head is not editable by my account (`UpdatePullRequest` permission denied), so recording it here instead.

Komzpa:
Correction / cross-link: the broader owned PR for this topic is #516 (`Komzpa/codex/mask-single-previous-response-miss`). This PR (#535) is a narrower current-main mask for the direct WebSocket miss.

The live deployment we are testing locally is not this PR head directly; it is a composite branch image (`codex-lb:weekly-recovery-safe-25770b8`) that includes the same patch as #516 (`fix(proxy): mask single previous response misses`) plus the dashboard weekly pace fixes from #534.

I verified the live composite commit and #516 have the same stable patch-id for the proxy masking change, so #516 is the PR to treat as the canonical broader fix. The incident evidence I posted above also now appears in #516's PR body.

mgwals:
Thank you for validating.

Yes, I agree. I was focusing on fixing the WebSocket miss specifically. I've closed this as superseded by #516.

PR #541, opened 2026-05-03
fix(proxy): handle model fetch timeouts

Komzpa:
## Summary

- convert upstream model-list timeouts into `ModelFetchError(504)` responses
- preserve the existing upstream status/body handling for classified model-list failures
- add a regression test for `asyncio.TimeoutError` from the model fetch client

## Why

Transient upstream `/models` stalls should be reported as a gateway timeout instead of bubbling up as an unexpected traceback. This keeps proxy reliability failures classified consistently for callers and logs.

This was split out of #534 so the weekly credits dashboard PR no longer carries an unrelated proxy reliability fix.

## Validation

- `/home/kom/proj/codex-lb/.venv/bin/python -m pytest tests/unit/test_model_fetcher.py`

## Related issues
- Related to #480 model metadata/fetch reliability, but only covers upstream model-list timeout classification.

### DIFF
--- a/app/core/clients/model_fetcher.py
+++ b/app/core/clients/model_fetcher.py
@@ -111,12 +111,17 @@ async def fetch_models_for_plan(
     timeout = aiohttp.ClientTimeout(total=_FETCH_TIMEOUT_SECONDS)
     session = get_http_client().session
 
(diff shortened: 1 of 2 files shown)


Soju06:
@codex review

Soju06:

Issue #284, opened 2026-03-31
`/v1/chat/completions` duplicates tool call `function.arguments` when proxying Responses API tool calls


elmakus:
## Summary

  `codex-lb` appears to duplicate tool call arguments in OpenAI-compatible `/v1/chat/completions`.

  I can reproduce this with a minimal direct request to `codex-lb`, without Honcho/Hermes in the loop.

  ## Minimal repro

  Request:

  ```json
  {
    "model": "gpt-5.4-mini",
    "messages": [
      {"role": "system", "content": "You must use the provided tool exactly once."},
      {"role": "user", "content": "Call the tool with city=Zurich and unit=C."}
    ],
    "tools": [
      {
        "type": "function",
        "function": {
          "name": "get_weather",
          "description": "Get weather for a city.",
          "parameters": {
            "type": "object",
            "properties": {
              "city": {"type": "string"},
              "unit": {"type": "string", "enum": ["C", "F"]}
            },
            "required": ["city", "unit"],
            "additionalProperties": false
          }
        }
      }
    ],
    "tool_choice": "required",
    "stream": false
  }
```
  ## Actual result

  The final /v1/chat/completions response contains:

  "arguments":"{\"city\":\"Zurich\",\"unit\":\"C\"}{\"city\":\"Zurich\",\"unit\":\"C\"}{\"city\":\"Zurich\",\"unit\":
  \"C\"}"

  This breaks OpenAI-compatible clients that expect function.arguments to be one valid JSON string.

  ## Stream observations

  With stream=true, the chat-completions stream emits:

  - incremental argument deltas
  - then a full argument snapshot again on response.function_call_arguments.done
  - then a full argument snapshot again on response.output_item.done

  This makes downstream clients prone to duplicated tool names/arguments.

  ## Expected result

  - stream=false: final tool_calls[0].function.arguments should be exactly one valid JSON string
  - stream=true: tool call chunks should not re-emit already completed full argument snapshots in a way that causes
  duplication

  ## Notes

  This looks like an adapter issue in the Responses -> Chat Completions mapping layer, not a model issue.
  I reproduced it with both gpt-5.4 and gpt-5.4-mini.

  Possibly related:

  - #113
  - #87
  - #86
  - #85

Issue #113, opened 2026-02-27
Intermittent missing tool_calls in streamed /v1/chat/completions responses (OpenClaw + codex-lb)


SHAREN:
## Summary
Intermittent missing `tool_calls` in streamed `/v1/chat/completions` responses. The model sometimes returns only text (e.g. “Preparing diagnostics execution / I’ll run commands now”) and then the workflow appears to stall because no tool call is emitted.

## Environment
- OpenClaw: Telegram channel workflow
- Proxy: codex-lb (OpenAI proxy)
- Endpoint: `POST /v1/chat/completions`
- Request: `stream=true`, `tools` list present
- Model: `gpt-5.3-codex`
- Diagnostics: mitmproxy reverse proxy `127.0.0.1:2456 -> 127.0.0.1:2455`

## Problem description
Under normal production load, we sometimes observe:
- Assistant sends a normal text reply (often implying a tool step will follow).
- No tool execution happens afterward (no `exec` / tool step).
- No follow-up message is produced; the session appears to go silent.

## Steps to reproduce
1. Run OpenClaw using codex-lb as the OpenAI proxy.
2. Use Telegram to send a prompt that should trigger a tool loop (commands via `exec` tool).
3. Observe that intermittently the model replies with text, but no tool calls are emitted.

## Expected behavior
When a prompt requires tools, the streamed response should include `tool_calls` and end with `finish_reason: "tool_calls"` (followed by tool execution in the client workflow).

## Actual behavior
In intermittent cases captured via mitmproxy:
- Response completes normally with `data: [DONE]`
- Final chunk shows `finish_reason: "stop"`
- No `tool_calls` are present in the streamed response (searching response shows none)
- As a result, OpenClaw performs no tool step and the conversation appears to stall

### 1) codex-lb transport/streaming appears stable
- Baseline SSE test via `curl -N` against codex-lb shows:
  - `finish_reason: "tool_calls"`
  - `data: [DONE]`
- Stress test: 200 SSE runs via `curl -N` (with `tools` and `tool_choice:"required"`)
  - 0 “bad” SSE samples (none missing `[DONE]` or missing `finish_reason:"tool_calls"`)
  - `.../sse/bad` directory is empty

### 2) Request contains `tools`, but no `tool_choice`
In the captured OpenClaw request body (mitmproxy):
- `"tools": [...]` is present (includes `exec`, etc.)
- `"tool_choice"` is not found in the request body

### 3) “Silent” cases are completed responses without tool calls
[...]

PR #543, opened 2026-05-03
fix(chat): resolve parallel tool_call argument duplication via output_index routing

### LINKED ISSUE #542: Parallel tool_calls still receive duplicated arguments (PR #287 fix incomplete — item_id not resolved in _extract_tool_call_fields)
## Summary

PR #287 (commit `41ceb4f`) fixed single-tool-call argument triplication (issue #284) but **did not fix parallel tool_calls**. When the upstream Responses API returns multiple parallel function calls, all `tool_calls[].function.arguments` in the Chat Completions response are identical — the second (or Nth) call's arguments are lost and replaced with a copy of another call's arguments.

This is a **data corruption bug** — downstream agents silently execute the wrong actions with wrong parameters. It affects both streaming and non-streaming paths.

**Version:** v1.15.0 (latest release, 2026-04-24)  
**Severity:** Critical — causes silent data loss in agentic workflows  
**Related:** #284 (incomplete fix), #392 (same root cause, Hermes manifestation), #113, #473

---

### Request (multi-tool-call scenario)

```bash
curl -s -X POST http://<codex-lb-host>:2455/v1/chat/completions \
  -H "Content-Type: application/json" \
  -H "Authorization: Bearer <key>" \
  -d '{
    "model": "gpt-5.4-mini",
    "messages": [{"role":"user","content":"User said: I biked 20km today and had pasta for dinner. Record these observations."}],
    "tools": [{
      "type": "function",
      "function": {
        "name": "record_observation",
        "description": "Record an observation about the user",
        "parameters": {
          "type": "object",
          "properties": {
            "category": {"type": "string", "enum": ["food","mood","activity","health","preference"]},
            "content": {"type": "string"},
            "confidence": {"type": "number", "minimum": 0, "maximum": 1}
          },
          "required": ["category","content","confidence"]
        }
      }
    }],
    "tool_choice": "required",
    "max_tokens": 500
  }'
```

### Expected response

Two tool_calls with **distinct** arguments:

```json
{
  "tool_calls": [
    {
      "id": "call_abc...",
      "function": {
        "name": "record_observation",
        "arguments": "{\"category\":\"activity\",\"content\":\"Biked 20km today\",\"confidence\":0.95}"
      }
    },
    {
      "id": "call_def...",
      "function": {
        "name": "record_observation",
        "arguments": "{\"category\":\"food\",\"content\":\"Had pasta for dinner\",\"confidence\":0.9}"
      }
    }
  ]
}
```

### Actual response

Two tool_calls with **identical** arguments (activity observation is lost):

```json
{
  "tool_calls": [
    {
      "id": "call_WTLDhKAufqYoKx7Po6NFtLhS",
      "function": {
        "name": "record_observation",
        "arguments": "{\"category\":\"food\",\"confidence\":0.97,\"content\":\"Had pasta for dinner today.\"}"
      }
    },
    {
      "id": "call_2TDf5kHITXfqsSIjT1mjPP6t",
      "function": {
        "name": "record_observation",
        "arguments": "{\"category\":\"food\",\"confidence\":0.97,\"content\":\"Had pasta for dinner today.\"}"
      }
    }
  ]
}
```

### Consistency

Reproduced **5 times** across different prompts — 100% reproducible. The bug also manifests with `"stream": false` (non-streaming path), confirming it's in the core translation logic, not just the streaming reassembly.

Single tool_call responses work correctly — the bug is isolated to parallel (2+) tool_calls.

---

### Location

**File:** `app/core/openai/chat_responses.py`  
**Functions:** `_extract_tool_call_fields()` (line ~592-608), `ToolCallIndex.index_for()` (line ~120-132)

### Mechanism

The Responses API emits `response.function_call_arguments.delta` and `response.function_call_arguments.done` events that identify which tool call they belong to using the **`item_id`** field (e.g., `"fc_01ee746e..."`). However, `_extract_tool_call_fields()` only checks for `call_id`, `tool_call_id`, and `id` — it does **NOT** check for `item_id`.

When processing argument delta/done events:
1. `call_id` = `None` (not present at the top level of these events)
2. `name` = `None` (not present in delta events)
3. `_tool_call_key(None, None)` returns `None`
4. `ToolCallIndex.index_for(key=None)` returns **0** as the default

**Result:** ALL argument delta/done events for ALL parallel tool calls are routed to index 0 (the first tool call's slot).

### Event sequence trace

```
Event                           | item_id       | Routed to | Correct?
─────────────────────────────────────────────────────────────────────
output_item.added   [idx=0]    | fc_...FIRST   | index 0   | ✓ (has call_id)
func_call_args.delta [idx=0]   | fc_...FIRST   | index 0   | ✓ (accidental: key=None → default 0)
func_call_args.done  [idx=0]   | fc_...FIRST   | index 0   | ✓ (accidental)
output_item.done    [idx=0]    | fc_...FIRST   | index 0   | ✓ (has call_id)
                                                            
output_item.added   [idx=1]    | fc_...SECOND  | index 1   | ✓ (has call_id)
func_call_args.delta [idx=1]   | fc_...SECOND  | index 0   | ✗ BUG (key=None → default 0)
func_call_args.done  [idx=1]   | fc_...SECOND  | index 0   | ✗ BUG (overwrites first call's args)
output_item.done    [idx=1]    | fc_...SECOND  | index 1   | ✓ (has call_id, but copies from idx 0's state)
```

The first tool call's arguments are accidentally correct (default index happens to be 0). The second tool call's arguments overwrite index 0 (via `mode=replace` on the `.done` event), and then `output_item.done` for the second call copies from the now-corrupted index 0 state into index 1. Both slots end up with the same arguments.

### Why PR #287 didn't catch this

PR #287's tests only covered **single tool call** deduplication (preventing delta+done+output_item.done triplication). The test `test_stream_chat_chunks_preserves_tool_call_state` uses two calls but only with delta events (no `.done` snapshots), which doesn't exercise the routing bug.

---

### Change 1: Add `item_id` to the field extraction fallback chain

In `_extract_tool_call_fields()`, add `candidate.get("item_id")` to the `call_id` resolution:

```python
# Before:
call_id = (
    candidate.get("call_id")
    or candidate.get("tool_call_id")
    or candidate.get("id")
)

# After:
call_id = (
    candidate.get("call_id")
    or candidate.get("tool_call_id")
    or candidate.get("id")
    or candidate.get("item_id")  # Responses API delta/done events use item_id
)
```

### Change 2: Register `item.id` as an alias in ToolCallIndex

When processing `output_item.added` or `output_item.done` events (which carry both `item.id` and `item.call_id`), register `item.id` as an alias that maps to the same index as `item.call_id`:

```python
# In ToolCallIndex:
def register_alias(self, alias_key: str, canonical_key: str) -> None:
    """Register an additional key that maps to the same index as canonical_key."""
    if canonical_key in self._map:
        self._map[alias_key] = self._map[canonical_key]
```

Then after processing `output_item.added`:
```python
# Register item_id → same index as call_id
if item_id and call_id:
    idx = self._tool_call_index.index_for(call_id, name)
    item_key = _tool_call_key(item_id, None)
    call_key = _tool_call_key(call_id, None)
    if item_key and call_key:
        self._tool_call_index.register_alias(item_key, call_key)
```

### Change 3: Don't default to index 0 when key is None

Instead of silently returning 0 for unknown keys, either raise an error or use a monotonically increasing fallback:

```python
def index_for(self, call_id: str | None, name: str | None) -> int:
    key = _tool_call_key(call_id, name)
    if key is None:
        # Fallback: assign next available index instead of always 0
        return self._next_index  # or log a warning
    ...
```

### Alternative (simpler, following OpenAI's reference implementation)

The OpenAI Python SDK's own Responses stream handler uses `output_index` directly:

```python
# From openai-python/src/openai/lib/streaming/responses/_responses.py:
elif event.type == "response.function_call_arguments.delta":
    output = snapshot.output[event.output_index]
    if output.type == "function_call":
        output.arguments += event.delta
```

If the raw SSE events already carry `output_index`, the most robust fix is to **key the accumulator on `output_index`** rather than trying to resolve `item_id` → `call_id` → index. This is the approach used by LiteLLM (after their similar bug fix in PR #17652) and PromptKit.

---

## Missing Test Cases

The following scenarios should be added to the test suite:

1. **Two parallel tool_calls, same function name** — the exact reproduction case
2. **Three+ parallel tool_calls** — ensure indexing scales beyond 2
3. **Parallel tool_calls with different function names** — verify no regression
4. **Non-streaming path with parallel tool_calls** — currently untested
5. **Interleaved delta events** (e.g., `delta[0]`, `delta[1]`, `delta[0]`, `delta[1]`) — some models emit arguments in round-robin

---

## Impact

- **Agentic workflows are broken** — any agent that issues parallel tool calls (which is the default behavior for GPT-4o/5.x when multiple actions are appropriate) gets corrupted results
- **Silent data loss** — there's no error, the response looks valid, but the arguments are wrong
- **Issue #392 is the same bug** — Hermes users report "Invalid JSON in tool call arguments: Extra data: line 1 column 182" which is the doubled-arguments symptom
- **Affects all downstream consumers** — Claude Code, Cursor, Hermes, custom agents, anything using the `/v1/chat/completions` endpoint with tools

---

## Similar Bugs in Other Projects (for reference)

| Project | Issue | Root Cause | Fix |
|---------|-------|-----------|-----|
| LiteLLM | [#17246](https://github.com/BerriAI/litellm/issues/17246) | `output_item.done` for text fired `is_finished=True`, dropping subsequent tool_calls | Only signal finish on `response.completed` |
| Ollama | [#15457](https://github.com/ollama/ollama/issues/15457) | All tool_call chunks emitted with `index: 0` | Map `output_index` → sequential `tool_calls[N].index` |
| LangChain | [#32562](https://github.com/langchain-ai/langchain/issues/32562) | Annotation events bumped shared index counter | Separate state machines for annotations vs. tool args |
| PromptKit | [#940](https://github.com/AltairaLabs/PromptKit/issues/940) | `item_id` lookup failed, no fallback to `output_index` | Key accumulator on `output_index` directly |

---

## Environment

- codex-lb v1.15.0 (installed from source, Python 3.13)
- Upstream: OpenAI Responses API (gpt-5.4-mini)
- OS: Debian 13 (Proxmox LXC)
- FastAPI/uvicorn

klaascommerce:
## Summary

Fixes #542 — parallel `tool_calls` receive duplicated/identical `function.arguments` when proxying Responses API → Chat Completions format.

**Root cause:** `response.function_call_arguments.delta/done` events identify their target tool call via `item_id`, but `_extract_tool_call_fields()` only checked `call_id`, `tool_call_id`, and `id`. With `item_id` unrecognized, `_tool_call_key()` returned `None`, and `index_for()` defaulted all events to **index 0** — overwriting the first call's arguments and duplicating them across all slots.

### 1. Alias registration
When processing `output_item.added/done` events (which carry both `item.id` and `item.call_id`), register `item.id` as an alias mapping to the same tool call index as `call_id`. Subsequent delta/done events (carrying only `item_id`) then resolve to the correct index.

### 2. `output_index` routing
Every Responses API event carries `output_index`. Added `index_for_output_index()` which uses this as the primary routing key, providing robust fallback even if IDs are missing or arrive in unexpected order. This follows the pattern used by [OpenAI's own reference implementation](https://github.com/openai/openai-python/blob/722d3fff/src/openai/lib/streaming/responses/_responses.py) and matches the fix approach in [LiteLLM #17652](https://github.com/BerriAI/litellm/pull/17652). The `output_index` mapping is now registered in **both** the key-based and keyless code paths, preventing out-of-order events from splitting one logical tool call into phantom indices.

### 3. Routing vs. emitted ID separation
`item_id` is used **only for index routing** via a separate `routing_id` variable — it is never returned as `call_id` on the `ToolCallDelta` and never leaks into the public Chat Completions response. This prevents internal `fc_xxx` Responses API IDs from appearing in streaming chunks; clients only see proper `call_xxx` IDs.

### Unit tests (4 new, all 24 pass)
- `test_collect_completion_parallel_tool_calls_distinct_arguments` — 2 calls, same function, distinct args
- `test_collect_completion_three_parallel_tool_calls` — 3 calls, verifies indexing scales
- `test_collect_completion_parallel_different_functions` — 2 calls, different function names
- `test_collect_completion_parallel_calls_item_id_equals_call_id` — edge case where `item.id == call_id`

### End-to-end (live gpt-5.4-mini)

**Non-streaming:**
```
Before fix: {"category":"food","content":"Had pasta"} × 2  (duplicated)
After fix:  {"category":"activity","content":"Biked 20 km today"} +
            {"category":"food","content":"Ate pasta for dinner"}   ✓
```

**Streaming (fc_ ID leak fixed):**
```
Before fix: id alternates between call_xxx and fc_xxx per tool call
After fix:  id=call_xxx on first chunk only, id=None on subsequent chunks  ✓
```

Tested streaming, non-streaming, single-tool (regression), multi-tool with 2+ calls, and 3-call scenarios.

## Related issues
- #284 — Original report, closed with incomplete fix (PR #287 only addressed single-call triplication)
- #392 — Same root cause manifesting as "Invalid JSON in tool call arguments: Extra data" in Hermes
- #113 — Potentially related streaming tool_call issues

## Similar bugs fixed in other projects
| Project | Issue | Fix |
|---------|-------|-----|
| LiteLLM | [#17246](https://github.com/BerriAI/litellm/issues/17246) | Only signal finish on `response.completed`, not `output_item.done` |
| Ollama | [#15457](https://github.com/ollama/ollama/issues/15457) | Map `output_index` → sequential `tool_calls[N].index` |
| LangChain | [#32562](https://github.com/langchain-ai/langchain/issues/32562) | Separate state machines for annotations vs. tool args |
| PromptKit | [#940](https://github.com/AltairaLabs/PromptKit/issues/940) | Key accumulator on `output_index` directly |

## Test plan
- [x] All 20 existing unit tests pass (no regressions)
- [x] 4 new unit tests covering parallel tool_calls
- [x] E2E: multi-tool with same function name → distinct arguments
- [x] E2E: single tool call still works (regression)
- [x] E2E: non-streaming multi-tool → distinct arguments
- [x] E2E: streaming no longer leaks internal `fc_` IDs
- [x] E2E: 3 parallel tool calls with system message → all distinct

### DIFF
--- a/app/core/openai/chat_responses.py
+++ b/app/core/openai/chat_responses.py
@@ -149,7 +149,10 @@ class ToolCallIndex:
         if key not in self.indexes:
             self.indexes[key] = self.next_index
             self.next_index += 1
-        return self.indexes[key]
+        idx = self.indexes[key]
+        if output_index is not None and output_index not in self.output_index_map:
+            self.output_index_map[output_index] = idx
+        return idx
 
     def register_alias(self, alias_id: str, index: int) -> None:
         """Register an additional ID that maps to the same index."""
@@ -659,7 +662,14 @@ def _tool_call_delta_from_payload(payload: Mapping[str, JsonValue], indexer: Too
         return None
     call_id, name, arguments, tool_type = fields
     output_index = _resolve_output_index(payload)
-    index = indexer.index_for_output_index(output_index, call_id, name)
+    routing_id = call_id
+    if routing_id is None:
+        item = payload.get("item")
+        routing_id = _first_str(
+            payload.get("item_id"),
+            item.get("id") if isinstance(item, dict) else None,
+        )
+    index = indexer.index_for_output_index(output_index, routing_id, name)
     _register_item_id_alias(payload, index, indexer)
     return ToolCallDelta(
         index=index,
@@ -701,14 +711,12 @@ def _extract_tool_call_fields(
         candidate.get("call_id"),
         candidate.get("tool_call_id"),
         candidate.get("id"),
-        candidate.get("item_id"),
     )
     if call_id is None and delta_map is not None:
         call_id = _first_str(
             delta_map.get("id"),
             delta_map.get("call_id"),
             delta_map.get("tool_call_id"),
-            delta_map.get("item_id"),
         )
 
     name = _first_str(candidate.get("name"), candidate.get("tool_name"))
--- a/app/core/openai/chat_responses.py
+++ b/app/core/openai/chat_responses.py
@@ -122,6 +122,7 @@ ChatCompletionResult = ChatCompletion | OpenAIErrorEnvelope
 class ToolCallIndex:
     indexes: dict[str, int] = field(default_factory=dict)
     next_index: int = 0
+    output_index_map: dict[int, int] = field(default_factory=dict)
 
     def index_for(self, call_id: str | None, name: str | None) -> int:
         key = _tool_call_key(call_id, name)
@@ -131,6 +132,35 @@ class ToolCallIndex:
             self.indexes[key] = self.next_index
             self.next_index += 1
         return self.indexes[key]
+
+    def index_for_output_index(self, output_index: int | None, call_id: str | None, name: str | None) -> int:
+        """Resolve tool call index, preferring output_index mapping when available."""
+        if output_index is not None and output_index in self.output_index_map:
+            return self.output_index_map[output_index]
+        key = _tool_call_key(call_id, name)
+        if key is None:
+            if output_index is not None:
+                idx = self.output_index_map.get(output_index, self.next_index)
+                if output_index not in self.output_index_map:
+                    self.output_index_map[output_index] = idx
+                    self.next_index = max(self.next_index, idx + 1)
+                return idx
+            return 0
+        if key not in self.indexes:
+            self.indexes[key] = self.next_index
+            self.next_index += 1
+        return self.indexes[key]
+
+    def register_alias(self, alias_id: str, index: int) -> None:
+        """Register an additional ID that maps to the same index."""
+        key = f"id:{alias_id}"
+        if key not in self.indexes:
+            self.indexes[key] = index
+
+    def register_output_index(self, output_index: int, index: int) -> None:
+        """Map a Responses API output_index to a tool call index."""
+        if output_index not in self.output_index_map:
+            self.output_index_map[output_index] = index
 
 
 @dataclass
@@ -589,6 +619,38 @@ def _coerce_number(value: JsonValue) -> int | float | None:
     return None
 
 
+def _resolve_output_index(payload: Mapping[str, JsonValue]) -> int | None:
+    """Extract output_index from a Responses API event payload."""
+    val = payload.get("output_index")
+    if isinstance(val, int):
+        return val
+    return None
+
+
+def _register_item_id_alias(payload: Mapping[str, JsonValue], index: int, indexer: ToolCallIndex) -> None:
+    """Register item.id from output_item events as an index alias.
+
+    The Responses API emits output_item.added/done events with item.id
+    (e.g. "fc_01ee...") and item.call_id (e.g. "call_kI8y...").
+    Subsequent function_call_arguments.delta/done events reference the
+    tool call via item_id (= item.id), not call_id.  By registering
+    item.id as an alias when we first see the output_item event, later
+    delta/done events can resolve to the correct index.
+    """
+    item = payload.get("item")
+    if not isinstance(item, dict):
+        return
+    item_id = item.get("id")
+    call_id = item.get("call_id")
+    if isinstance(item_id, str) and item_id:
+        if item_id != call_id:
+            indexer.register_alias(item_id, index)
+
+    output_index = _resolve_output_index(payload)
+    if output_index is not None:
+        indexer.register_output_index(output_index, index)
+
+
 def _tool_call_delta_from_payload(payload: Mapping[str, JsonValue], indexer: ToolCallIndex) -> ToolCallDelta | None:
     if not _is_tool_call_event(payload):
         return None
@@ -596,7 +658,9 @@ def _tool_call_delta_from_payload(payload: Mapping[str, JsonValue], indexer: Too
     if fields is None:
         return None
     call_id, name, arguments, tool_type = fields
-    index = indexer.index_for(call_id, name)
+    output_index = _resolve_output_index(payload)
+    index = indexer.index_for_output_index(output_index, call_id, name)
+    _register_item_id_alias(payload, index, indexer)
     return ToolCallDelta(
         index=index,
         call_id=call_id,
@@ -637,12 +701,14 @@ def _extract_tool_call_fields(
         candidate.get("call_id"),
         candidate.get("tool_call_id"),
         candidate.get("id"),
+        candidate.get("item_id"),
     )
     if call_id is None and delta_map is not None:
         call_id = _first_str(
             delta_map.get("id"),
             delta_map.get("call_id"),
             delta_map.get("tool_call_id"),
+            delta_map.get("item_id"),
         )
 
     name = _first_str(candidate.get("name"), candidate.get("tool_name"))
--- a/app/core/openai/chat_responses.py
+++ b/app/core/openai/chat_responses.py
@@ -134,10 +134,23 @@ class ToolCallIndex:
         return self.indexes[key]
 
     def index_for_output_index(self, output_index: int | None, call_id: str | None, name: str | None) -> int:
-        """Resolve tool call index, preferring output_index mapping when available."""
-        if output_index is not None and output_index in self.output_index_map:
-            return self.output_index_map[output_index]
+        """Resolve tool call index, preferring output_index mapping when available.
+
+        Always records the resolved index against every routing key that is
+        present (``output_index``, ``call_id``/``name``) so a later event for
+        the same logical tool call that only carries one of those identifiers
+        does not get assigned a new slot. Without this, a stream that first
+        routes by ``item_id`` + ``output_index`` and later emits a
+        ``call_id``-only event without ``output_index`` (e.g. legacy delta
+        formats) would split one tool call into two ``tool_calls[]`` slots and
+        reintroduce the duplication this change fixes.
+        """
         key = _tool_call_key(call_id, name)
+        if output_index is not None and output_index in self.output_index_map:
+            idx = self.output_index_map[output_index]
+            if key is not None and key not in self.indexes:
+                self.indexes[key] = idx
+            return idx
         if key is None:
             if output_index is not None:
                 idx = self.output_index_map.get(output_index, self.next_index)
--- a/openspec/changes/fix-parallel-tool-call-arguments-routing/proposal.md
+++ b/openspec/changes/fix-parallel-tool-call-arguments-routing/proposal.md
@@ -1,0 +1,24 @@
+## Why
+Issue #542 reports that `/v1/chat/completions` corrupts parallel tool-call arguments. The Responses API emits `response.function_call_arguments.delta` / `.done` events that identify their owning call only via `item_id` (e.g. `"fc_..."`). The previous `ToolCallIndex.index_for` keyed only on `call_id`/`name`, so argument events for the second and subsequent parallel calls fell back to index `0` and overwrote the first call's payload. Downstream agents then silently executed the wrong actions with corrupted parameters.
+
+The fix uses a three-layer routing strategy to be robust against ordering and identifier variations the upstream may emit:
+
+1. **`output_index` routing (primary).** Every Responses API event carries a stable `output_index`, matching the routing key used by OpenAI's own reference client (`openai-python`) and the LiteLLM proxy (BerriAI/litellm#17652). Once `output_index` is observed, all events for that slot resolve through the same `output_index_map` regardless of which identifier they carry.
+2. **`call_id` / `name` keying (existing).** Preserved for legacy Chat-Completions-style event streams and any path that already routed by `call_id`.
+3. **`item_id` alias (fallback).** When `output_item.added` / `.done` events expose both `item.id` (e.g. `fc_...`) and `item.call_id` (e.g. `call_...`), the `item.id` is registered as an alias to the same slot, so later argument-only events that only carry `item_id` resolve correctly even if their `output_index` was not seen.
+
+A small secondary fix in the same change prevents the upstream `fc_...` item id from leaking into the public `tool_calls[].id` / `call_id` fields exposed to clients: clients continue to see the upstream `call_...` value and never the internal `fc_...` routing handle.
+
+## What Changes
+- Add `index_for_output_index(output_index, call_id, name)` on `ToolCallIndex` that prefers a registered `output_index` mapping and falls back to the existing `call_id`/`name` keying. The original `index_for` is preserved unchanged for callers that have no `output_index` (chat completions legacy).
+- Add `register_alias(alias_id, index)` and `register_output_index(output_index, index)` helpers, plus a new `output_index_map: dict[int, int]` field, so the indexer can be primed from `output_item.added` / `.done` events.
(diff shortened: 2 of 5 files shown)


klaascommerce:
@Soju06 — this PR is ready for review. Quick summary of the two commits:

**Commit 1:** Core fix — adds `item_id` alias registration, `output_index` routing, and `item_id` to the extraction fallback chain. Fixes the parallel tool_call argument duplication reported in #542.

**Commit 2:** Addresses self-review feedback — separates the routing ID from the emitted public `call_id` so that internal `fc_xxx` Responses API IDs never leak into Chat Completions streaming chunks, and ensures `output_index` is registered in the key-based code path (preventing out-of-order event issues).

All 24 unit tests pass (20 existing + 4 new). End-to-end validated against live gpt-5.4-mini — streaming and non-streaming both produce correct, distinct arguments with proper `call_xxx` IDs.

This also resolves the root cause behind #392 (Hermes "Extra data" JSON parse errors from doubled arguments).

Soju06:
@codex review

Soju06:
Thanks @klaascommerce -- the three-layer defense pattern with `output_index` as the primary routing key is the right shape and matches what LiteLLM / Ollama / LangChain landed for the same upstream behavior. The cross-reference table in the PR body is genuinely useful prior art.

Codex review on `9223a02` left a single P2 we want to fold in before merging, and the GitHub Actions workflow hasn't actually run on the PR yet (only GitGuardian completed). I'll handle the CI approval separately; here's the P2 first.

## P2 -- `index_for_output_index` skips alias registration on known indexes

Codex spotted that `index_for_output_index` returns immediately when `output_index` is already mapped, so it never records a newly observed `call_id` / `name` key for the same tool call. This breaks layer 1 (alias registration) when events arrive in mixed shapes within one stream:

1. `response.function_call_arguments.delta` arrives carrying `item_id` and `output_index` -> routes correctly via `output_index`; layer 2 wins
2. later, `response.output_tool_call.delta` arrives carrying only `call_id` (no `output_index`) -> the existing keyless path on `_extract_tool_call_fields` sees `call_id` as a fresh key, allocates a new index, and the same logical tool call splits into two slots

The existing tests do cover `response.output_tool_call.delta` for some shapes, but the new tests in this PR don't combine that event with an earlier `output_index`-routed delta in the same stream, so this didn't get caught.

Minimal fix: when `index_for_output_index` finds an existing mapping, also register the `call_id` / `name` keys at that index before returning early. Something like:

```python
def index_for_output_index(
    output_index: int | None,
    key: tuple[str, ...] | None,
    name: str | None,
) -> int:
    if output_index is not None and output_index in output_index_to_index:
        index = output_index_to_index[output_index]
        if key is not None:
            key_to_index.setdefault(key, index)
        if name is not None:
            name_to_index.setdefault(name, index)
        return index
    ...
```

(Pseudocode -- adjust to the actual locals in the closure.) The point is: any pre-existing index lookup should still teach the alias maps about new keys it sees, so a later keyless event lands on the same slot.

Test we'd want:

```python
def test_parallel_tool_calls_mixed_output_index_and_call_id_events_share_slot():
    events = [
        # routes via output_index + item_id
        {"type": "response.function_call_arguments.delta", "output_index": 0, "item_id": "fc_a", "delta": "{\"x\":"},
        # later, only call_id (no output_index) for the same logical call
        {"type": "response.output_tool_call.delta", "call_id": "call_a", "delta": "1}"},
    ]
    # expect the two deltas to concatenate on tool_calls[0], not split into [0] and [1]
```

## CI workflow approval

This is a first-time-contributor PR from a fork, so the GitHub Actions workflow needs maintainer approval to actually run. I'll do that on the maintainer side once you push the P2 fix -- no action needed from you. If the workflow already ran for you locally with `uv run pytest` and you've got it green, that's enough.

## Not blocking otherwise

The rest of the PR reads clean -- the routing-vs-emitted-ID separation, the 3-layer defense, and the four new unit tests are all in the right shape. Once the P2 fix and the matching test are in I'll re-run `@codex review` and approve the CI workflow.

Soju06:
Pushed a docs-only follow-up (`df51a9f`) on top of your branch to document the routing rules from this PR in an OpenSpec change folder, mirroring the structure used by the duplicate proposal #587 and the prior responses-api-compat changes in this repo.

## What landed on the branch

`openspec/changes/fix-parallel-tool-call-arguments-routing/`:

- `proposal.md` — explains the three-layer routing strategy you implemented (`output_index` primary, `call_id` / `name` keying, `item_id` alias fallback), references the OpenAI reference client and LiteLLM #17652, and notes the secondary `fc_` -> public `call_id` leak guard.
- `specs/responses-api-compat/spec.md` — adds three new scenarios under the existing "Tool call events and output items are preserved" requirement:
  - parallel tool calls route arguments by `output_index`
  - parallel tool calls also resolve through `item_id` aliases
  - internal `item_id` never leaks into the public call identifier
- `tasks.md` — task list checked off against the changes already on the branch.

No code changes -- the wording is descriptive of what the branch already does, so the existing implementation and tests stay authoritative.

Verified locally:

```
npx @fission-ai/openspec validate fix-parallel-tool-call-arguments-routing
# Change 'fix-parallel-tool-call-arguments-routing' is valid
npx @fission-ai/openspec validate --specs
# Totals: 19 passed, 0 failed (19 items)
```

## Why this PR (not #587)

After comparing both PRs against #542, this branch is the more robust fix:

- Three-layer defense (`output_index` primary + `call_id` keying + `item_id` alias) vs single-layer item_id aliasing in #587.
- `output_index` routing matches OpenAI's own reference Responses streaming client and LiteLLM's equivalent fix, so the routing key tracks an upstream-stable identifier rather than a sometimes-missing one.
- Bonus fix in `9223a02` (no `fc_` leak into the public `call_id` surface) that #587 doesn't cover.
- Backward compatible: `index_for` keeps its original signature, so the legacy Chat-Completions-only path doesn't change.

I'll trigger `@codex review` against this new head and queue for merge from here. Closing #587 with a thanks note that points at this PR.

Soju06:
@codex review

Soju06:
@codex review

Soju06:
@codex review

Soju06:
Pushed P2 fix (`fb17a2e`) addressing codex review's repeated finding on `index_for_output_index`:

> `index_for_output_index` returns immediately when `output_index` is known, so it never records a newly observed `call_id`/`name` key for that same tool call.

The new behavior records the resolved slot against every routing key that's present (`output_index`, `call_id` / `name`) — so a stream that first routes via `item_id` + `output_index` and later emits a `call_id`-only event without `output_index` (legacy delta formats) still resolves to the same `tool_calls[]` slot.

Regression test `test_collect_completion_registers_call_id_when_output_index_already_mapped` pins the exact ordering that previously fragmented one call into two slots.

Local verification on this branch:

```
uv run ruff check app/core/openai/chat_responses.py tests/unit/test_chat_response_mapping.py     # clean
uv run ruff format --check app/core/openai/chat_responses.py tests/unit/test_chat_response_mapping.py  # clean
uv run pytest tests/unit/test_chat_response_mapping.py -q                                          # 25 passed
```

Re-triggered `@codex review` against the new head.

Soju06:
Pushed `faf4511` to clear the `Type check (ty)` CI failure.

The earlier ty diagnostics on this branch were `possibly-missing-attribute` warnings against `tool_calls[i].function.arguments` / `.name` accesses in the new parallel tool-call tests, because the pydantic `ChatToolCallFunction` model declares both `function` and its fields as `Optional` for upstream compatibility. The adapters always populate them by the time a tool call is surfaced to the client, but ty (correctly) does not infer that.

Added two small type-narrowing helpers in the test module — `_tool_call_args` and `_tool_call_name` — and routed every parallel-tool-call assertion through them. No production model changes; behavior is identical.

Local re-verification on this head:

```
uv run ruff check app/core/openai/chat_responses.py tests/unit/test_chat_response_mapping.py     # clean
uv run ruff format --check app/core/openai/chat_responses.py tests/unit/test_chat_response_mapping.py  # clean
uv run ty check tests/unit/test_chat_response_mapping.py                                          # All checks passed!
uv run pytest tests/unit/test_chat_response_mapping.py -q                                          # 25 passed
```

Soju06:
@codex review

Soju06:
@codex review

Soju06:
@all-contributors please add @klaascommerce for code, test

Soju06:
@all-contributors please add @ozpool for ideas, doc

PR #552, opened 2026-05-06
fix(openspec): satisfy strict purpose validation

Komzpa:
## Summary
- replace placeholder Purpose text in 13 OpenSpec capability specs with concrete descriptions
- keep requirements unchanged while making strict spec validation pass

## Validation
- `openspec validate --specs --strict --no-interactive` -> 19 passed, 0 failed

PR #545, opened 2026-05-04
feat(accounts): add account priorities

cani1989:
fixes #540 

## Summary

- add account priority levels (`gold`, `silver`, `bronze`) end-to-end across API, DB, proxy selection, and dashboard/account views
- keep priorities behind a settings toggle so the feature can be enabled or disabled without changing account data
- add a per-account priority editor in the accounts page and priority-aware sorting in the dashboard/accounts list

## Why

Priority-aware selection makes it possible to prefer certain accounts before falling back to usage/reset-based ordering. The settings toggle keeps the new behavior opt-in, while still letting admins assign and change priorities independently of routing strategy.

## Validation

- `npm test -- --run src/features/accounts/components/accounts-page.test.tsx`
- `npm test -- --run src/features/dashboard/components/account-cards.test.tsx src/features/dashboard/components/account-card.test.tsx`
- backend integration/unit coverage added for account priority updates, dashboard ordering, settings persistence, and load-balancer selection

PR #553, opened 2026-05-06
feat(accounts): add manual routing policy

### LINKED ISSUE #428: Feature request: allow manual account priority / ordered fallback routing
## Summary

It would be very useful to support manual account priority, so requests can follow a user-defined order instead of only using the current balancing strategies (`capacity_weighted`, `usage_weighted`, `round_robin`).

## Problem

Right now, the available routing strategies are useful for balancing, but they do not allow a strict operator-defined usage order.

A common real-world case is:

- 4 `Plus` accounts
- 1 `Pro` account

In this setup, I may want to fully consume or prefer the `Plus` accounts first, and only start using the `Pro` account when the `Plus` pool is exhausted, rate-limited, or otherwise unavailable.

At the moment, this is not possible in a strict way:

- `capacity_weighted` tends to favor higher-capacity plans such as `Pro`
- `round_robin` still sends traffic to `Pro` before `Plus` is exhausted
- `usage_weighted` also does not provide a true ordered fallback model
- sticky routing can further hide the intended behavior

## Requested behavior

Please add a way to define a manual priority order for accounts.

Example desired behavior:

1. Try account/group A first
2. If unavailable, fall back to account/group B
3. Then C, and so on

"Unavailable" could include cases such as:

- rate limited
- quota exceeded
- paused
- deactivated
- model not supported
- filtered out by API key/account scope

## Example use cases

- Prefer `Plus` accounts before `Pro`
- Keep a more expensive account as a last-resort fallback
- Prefer personal accounts before team/workspace accounts
- Reserve one specific account for overflow only

## Possible implementation ideas

A few possible approaches:

### Option 1: new routing strategy

Add a routing mode such as `manual_priority` or `ordered_fallback`.

### Option 2: per-account priority field

Allow each account to have a numeric priority, with lower number = higher priority.

### Option 3: explicit ordered list in settings

Allow operators to drag-and-drop accounts into an ordered list used by the selector.

## Nice-to-have

It would be especially useful if this could work together with:

- API key assigned accounts
- sticky sessions
- model availability filtering

For example, when an API key is scoped to a subset of accounts, the manual order could apply within that subset.

## Why this matters

This would give operators much more predictable control over cost, quota burn strategy, and fallback behavior. Right now the system is optimized for balancing, but not for intentional "use these first, keep those as backup" workflows.

### LINKED ISSUE #540: Account priority
Account option to prioritize accounts over others. I can imagine different ways:
- <del>Groups
  - Prod
  - Test
  - Dev
  - BizDev
- <del>Buckets
  - Project A/B/C
  - Work A/B
  - Personal C</del>
- Categories
  - High 
  - Medium
  - Low  

I think High, Medium, Low is easy to understand and visually easy to represent (gold, silver, bronze)

I have my own fork with:

Image

Also, I added a Feature Toggle in Settings: On/Off.

Komzpa:
## Summary

- Add a focused per-account manual routing policy: `burn_first`, `normal`, and `preserve`.
- Add per-additional-quota routing policy overrides: `inherit`, `burn_first`, `normal`, and `preserve`.
- Default the `codex_spark` additional quota limit to `burn_first`, so Spark quota can be burned independently while standard Codex calls keep using ordinary account routing.
- Apply account and additional-quota policy after model/additional-quota eligibility filtering and before ordinary usage/reset/capacity balancing.
- Expose account policy through the accounts API/dashboard controls, and additional-quota policy through routing settings plus account usage badges.
- Keep OAuth re-authorization of a deactivated exact account identity in-place instead of creating `__copyN` when import-without-overwrite is enabled.
- Document the selection contract with an OpenSpec change.

## Relationship to #545 and #567

This is the narrow version of the account-priority/routing behavior discussed in #545. It intentionally avoids the reset-display, quota-appearance, Docker/dev-deploy, and broader dashboard sorting changes from that PR so the routing contract can be reviewed independently.

#567 builds on this PR for opportunistic worker admission and traffic classes. The Spark/additional-quota routing policy lives here so the policy UI and routing contract are reviewed together with manual account routing.

## Re-auth behavior

`importWithoutOverwrite` still preserves active duplicate imports as separate `__copyN` accounts. A re-authorized account with the same generated account id that is currently `deactivated` is treated as healing that account: fresh tokens/status/plan are written to the existing row, while manual routing policy and account history stay attached to the original account.

## Validation

- `openspec validate add-manual-account-routing-policy --strict --no-interactive`
- `/home/kom/proj/codex-lb/.venv/bin/python -m ruff check app/core/balancer/logic.py app/db/models.py app/db/alembic/versions/20260509_010000_add_additional_quota_routing_policies.py app/modules/proxy/load_balancer.py app/modules/settings app/modules/accounts app/modules/usage/additional_quota_keys.py tests/unit/test_load_balancer.py tests/integration/test_load_balancer_integration.py tests/integration/test_settings_api.py tests/integration/test_additional_usage_flow.py`
- `/home/kom/proj/codex-lb/.venv/bin/python -m pytest tests/unit/test_load_balancer.py tests/integration/test_load_balancer_integration.py tests/integration/test_settings_api.py tests/integration/test_additional_usage_flow.py tests/integration/test_migrations.py -q`
- `cd frontend && bun run test -- src/features/settings/schemas.test.ts src/features/settings/components/routing-settings.test.tsx src/features/accounts/components/account-usage-panel.test.tsx`
- `cd frontend && bun run build`
- `uv run ty check`
- `/home/kom/proj/codex-lb/.venv/bin/python -m ruff format --check ... && git diff --check`
- Previously for the original routing-policy/re-auth scope: `bun run typecheck`, `bun run lint`, and targeted OAuth/repository tests listed in earlier revisions of this PR.

Note: full `openspec validate --specs --strict --no-interactive` still fails on current `main` for pre-existing placeholder Purpose text; that is fixed separately in #552 and is not included here to keep this PR based directly on `main`.

## Related issues

- Fixes #428
- Fixes #540
- Supersedes the routing-policy part of #545 with a narrower review scope.


Soju06:
@codex review

Soju06:
Thanks @Komzpa -- the policy/health-tier ordering, the OAuth re-auth heal path, and the `codex_spark` default-burn_first decision all look right. Codex review landed two P2 input-validation findings that I think are worth folding in before merge given the routing contract is touching the accounts API and settings update paths:

## P2 -- normalize legacy account.routing_policy in the mapper

`account.routing_policy` flows through `accounts/mappers.py` verbatim. The frontend schema now validates `routingPolicy` as a strict enum (`normal | burn_first | preserve`), so any pre-existing row, manual SQL edit, or rolled-back deploy that left a value outside that set will make `GET /api/accounts` fail schema validation and break the dashboard accounts page entirely. Treat anything outside the enum as `normal` at the mapper boundary:

```python
_VALID_ROUTING_POLICIES = {"normal", "burn_first", "preserve"}

def _coerce_routing_policy(value: str | None) -> str:
    if value not in _VALID_ROUTING_POLICIES:
        return "normal"
    return value
```

Then call `_coerce_routing_policy(account.routing_policy)` in the mapper. The migration default is `"normal"` so today's data is fine, but downgrade/upgrade cycles or operator overrides should not be able to brick the page.

## P2 -- reject invalid additional_quota_routing_policies entries

The override normalizer in `app/modules/usage/additional_quota_keys.py:243` silently drops unknown quota keys and coerces unknown policy values to `"inherit"`. So a typo like `{"codex_spark": "burnfrist"}` returns 200 from the settings update endpoint and silently disables the `burn_first` default for Spark -- the operator thinks they re-applied the same policy with a typo and traffic shape silently changes.

The accounts mapper case is about tolerating *historical* bad data, but settings updates are *new* user input -- those should fail loud:

- Unknown quota slug -\u003e raise a 422-style ClientPayloadError listing valid slugs (or at least the registered ones from `config/additional_quota_registry.json`).
- Unknown policy value -\u003e raise a 422-style ClientPayloadError listing `inherit | burn_first | normal | preserve`.

The existing settings update flow already handles `ValidationError` paths, so wiring a clean error message is straightforward. Tests:

```python
def test_additional_quota_policy_rejects_unknown_slug()
def test_additional_quota_policy_rejects_unknown_policy_value()
```

These should each assert a 422 (or whatever shape the existing settings update uses for client validation) and that the stored `additional_quota_routing_policies_json` is unchanged.

## Not blocking

The OpenSpec change, the migrations, and the routing-policy pool wiring all look right. Only the two input-validation P2s above. Once those are in I'll re-run `@codex review`. Holding the merge until the new commit lands.

(Heads-up: #555 is still red in your stack -- one unit test on `test_stream_responses_suppresses_duplicate_http_tool_call_bursts` -- already commented there. Doesn't block this PR; #553 is independent.)

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Soju06:

Komzpa:
Revived on a maintainer-owned branch as #761 because this PR head has maintainerCanModify=false and cannot be updated here. Keeping credit to the original implementation in the replacement PR.

PR #555, opened 2026-05-06
fix(proxy): trim Codex websocket full-replay continuations

Komzpa:
## Summary
- trim Codex websocket full-replay continuations by anchoring rebuilt requests to the completed previous response
- keep websocket replay continuity state local to this review slice
- leave startup stream-error handling and side-effect tool-call replay suppression to follow-up split PRs #585 and #586

## Validation
- `uv run ruff check app/modules/proxy/service.py app/core/clients/proxy_websocket.py tests/unit/test_proxy_utils.py tests/integration/test_proxy_websocket_responses.py tests/unit/test_proxy_websocket_client.py`
- `uv run ty check app/modules/proxy/service.py app/core/clients/proxy_websocket.py tests/unit/test_proxy_utils.py tests/integration/test_proxy_websocket_responses.py tests/unit/test_proxy_websocket_client.py`
- `uv run pytest tests/unit/test_proxy_utils.py tests/integration/test_proxy_websocket_responses.py tests/unit/test_proxy_websocket_client.py -q -k 'websocket or replay or trim or continuity or duplicate_tool_call'`\n\nStack note: this PR is now the replay-core slice only. Merge #585 and #586 after it for the pieces split out of the old oversized branch.

### DIFF
--- a/app/modules/proxy/service.py
+++ b/app/modules/proxy/service.py
@@ -581,6 +581,8 @@ class ProxyService:
         effective_payload = payload
         proxy_injected_previous_response_id = False
         fresh_upstream_request_text: str | None = None
+        previous_response_trimmed_input_count: int | None = None
+        previous_response_trimmed_input_fingerprint: str | None = None
         if durable_lookup is not None:
             bridge_session_key = _HTTPBridgeSessionKey(
                 durable_lookup.canonical_kind,
@@ -622,6 +624,13 @@ class ProxyService:
                     cache_key_family=bridge_session_key.affinity_kind,
                     model_class=_extract_model_class(payload.model) if payload.model else None,
                 )
+        if effective_payload.previous_response_id is not None and isinstance(effective_payload.input, list):
+            previous_response_input_items = cast(list[JsonValue], effective_payload.input)
+            trimmed_input_items = _trim_websocket_previous_response_input_items(previous_response_input_items)
+            if len(trimmed_input_items) != len(previous_response_input_items):
+                previous_response_trimmed_input_count = len(previous_response_input_items)
+                previous_response_trimmed_input_fingerprint = _fingerprint_input_items(previous_response_input_items)
+                effective_payload = effective_payload.model_copy(update={"input": trimmed_input_items})
         request_state, text_data = self._prepare_http_bridge_request(
             effective_payload,
             headers,
@@ -631,6 +640,19 @@ class ProxyService:
         )
         if downstream_turn_state is not None:
             request_state.session_id = _normalize_session_id(downstream_turn_state)
+        if previous_response_trimmed_input_count is not None:
+            request_state.input_item_count = previous_response_trimmed_input_count
+            request_state.input_full_fingerprint = previous_response_trimmed_input_fingerprint
+            logger.info(
+                "http_bridge_previous_response_input_trimmed request_id=%s original_items=%s trimmed_to=%s "
+                "previous_response_id=%s",
+                request_state.request_id,
+                previous_response_trimmed_input_count,
+                len(cast(list[JsonValue], effective_payload.input))
+                if isinstance(effective_payload.input, list)
+                else None,
+                effective_payload.previous_response_id,
+            )
         request_state.transport = _REQUEST_TRANSPORT_HTTP
         request_state.request_stage = _http_bridge_request_stage(
             headers=headers,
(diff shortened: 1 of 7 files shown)


Soju06:
Heads up @Komzpa -- CI is red on `9434018`:

```
FAILED tests/unit/test_proxy_utils.py::test_stream_responses_suppresses_duplicate_http_tool_call_bursts
  Left contains one more item:
  {'item': {'arguments': '{"session_id":1,"chars":"","yield_time_ms":1000}',
            'call_id': 'call_replayed',
            'name': 'write_stdin',
            'type': 'function_call'},
   'response_id': 'resp_tool_replayed',
   'type': 'response.output_item.done'}
```

Looks like the duplicate-burst guard is suppressing one more event than the test (or vice versa) expected -- probably an off-by-one or window-boundary case where a legitimate replayed-tool-call event under `resp_tool_replayed` is now folded into the dedupe instead of being passed through. This PR's body explicitly says replays under fresh/anonymous response ids should still be suppressed within the burst window, so the question is whether the test fixture is wrong (expected the *new* behavior to keep that item) or the implementation is wrong (suppressing too aggressively when the response_id changes).

Could you take another pass and either:

- Update the test fixture if the new suppression boundary is the intended behavior (i.e. `resp_tool_replayed` is a duplicate burst and should be dropped), and add a sibling test case that asserts the same call text appearing later as a *fresh model decision* still gets through. The PR body promises this is the contract.
- Or tighten the dedupe so it correctly distinguishes replay-burst from a re-issued tool call when the response_id changes inside the burst window.

Also kicking off `@codex review` on the current head so any other findings show up alongside.

PostgreSQL test failure is downstream of the same unit assertion -- once the unit test stabilizes that should clear too.

Will hold the merge until CI is green and a fresh review is in. The other PRs in your stack are landing as they get reviewed.

@codex review

Soju06:
@codex review

Soju06:
Thanks @Komzpa -- the websocket replay-continuity work is the right shape and the test coverage looks careful. One more codex review pass on `f40c33b` surfaced two P2s in the new continuity / leading-context paths that I'd like to fold in before merging. The PR is mergeable as-is, but the continuity surface is sensitive enough that I want the guards explicit rather than relying on the broader fail-closed path.

## P2 #1 -- anchor double-accept on ambiguous pre-`response.created` failure

`app/modules/proxy/service.py:3020` injects `previous_response_id` so a fresh full-resend looks like a continuation. That is the right intent for the trim, but the direct websocket replay path (`_pop_replayable_precreated_websocket_request_state` + replay in `proxy_responses_websocket`) still forwards `request_state.request_text` verbatim on ambiguous pre-`response.created` failure. So in the failure window:

- proxy sends anchored continuation A1 -> upstream may or may not have accepted it before the disconnect
- proxy reconnects, the websocket replay path resends A1 verbatim (same anchored payload)
- if upstream accepted both, continuity forks and model/tool work duplicates

The HTTP-bridge counterpart already guards this class: it only retries injected anchors by switching back to the stored unanchored full replay, and it fails closed for client-supplied or oversized anchors. The websocket injection path needs equivalent protection. Two options:

a. **Reuse the same protection on websocket replay**: when popping a replayable pre-created request, if the request was proxy-anchored, replay the *stored fresh unanchored* payload instead of the anchored `request_text`. Same shape as `_pop_replayable_precreated_http_bridge_request_state` already does.

b. **Refuse to replay anchored requests on the WS path** and surface `stream_incomplete` instead. This is what the PR body says is the intended behavior for client-supplied / oversized anchors, but it currently does not extend to proxy-injected anchors. Re-using the unanchored fresh resend (option a) is friendlier to the client.

## P2 #2 -- preserve client-provided leading assistant items

`app/modules/proxy/service.py:3175` drops every leading `assistant` / `function_call` / `reasoning` item whenever `previous_response_id` is set on the request. That is correct for proxy-injected continuations where we know the prefix is a stale replay, but Responses API clients can legitimately send leading assistant-role context as new input on the next turn (e.g. prefilled assistant scaffolding, partial tool plans, externally curated continuation context). Silent rewriting changes model behavior.

Suggested narrow:

- only run the leading-item trim when the proxy itself injected the `previous_response_id` for this request (the `proxy_injected_previous_response_id` flag already exists on `request_state`)
- for client-supplied `previous_response_id`, forward the leading context verbatim

## Tests to add

- WS replay double-accept guard: simulate a proxy-injected anchored request, force a pre-`response.created` disconnect, assert the replay uses the saved unanchored full resend rather than the anchored `request_text`.
- Client-supplied `previous_response_id` with leading assistant input: assert the leading items are forwarded verbatim and not stripped.

## Not blocking

The rest of the PR is solid -- the response-id-scoped tool-call dedupe, the serialization-time tool-call trim, the continuity-metadata wait/fail-closed, and the unanchored-only retry semantics all read cleanly. Just the two newly-introduced injection paths above need the same guards the HTTP bridge already has.

After you push, I'll re-run `@codex review`. Holding the merge until then.

Komzpa:
Closing this draft because the replay/continuity work was split into reviewable replacement PRs: #650 for the continuity core, #651 for websocket previous-response trim handling, and #652 for HTTP bridge replay trimming. The active stack should track the split PRs.

PR #558, opened 2026-05-06
fix(proxy): support Pro Lite account routing

### LINKED ISSUE #557: Prolite accounts show zero aggregate remaining credits
## Summary

Accounts whose upstream `plan_type` is reported as `prolite` can show valid per-account quota percentages while the dashboard aggregate remaining-credit donuts show `0` capacity / `0` remaining.

This appears to happen because `prolite` is not recognized as an account plan with known credit capacity.

## Observed behavior

A backup `store.db` from a v1.15.0 deployment contains one active account with:

- `accounts.status = active`
- `accounts.plan_type = prolite`
- latest `usage_history` rows for both `primary` and `secondary`
- `usage_history.used_percent = 0.0` for both windows
- API keys have no `allowed_models`, `enforced_model`, or account assignment restrictions

In the dashboard:

- the account card shows the account as active
- the account card shows 5h and weekly quota as fully remaining
- the aggregate "5h Remaining" and "Weekly Remaining" donuts show `Remaining 0` and `Total 0`

Requests around the same time failed before selecting an account with:

```text
No available accounts. Service is operating in degraded mode: all upstream accounts are unavailable
```

## Expected behavior

If `prolite` is a valid upstream ChatGPT plan type, codex-lb should map it to a known capacity bucket, likely equivalent to `pro`, so that:

- aggregate dashboard remaining credits are non-zero when usage is available
- per-account remaining credits are computed
- capacity-weighted routing does not treat the account as having unknown/zero capacity

## Code pointers

`prolite` is currently not present in account plan normalization:

```py
ACCOUNT_PLAN_TYPES = {
    "free",
    "plus",
    "pro",
    "team",
    "business",
    "enterprise",
    "edu",
}
```

And it is not present in the dashboard/routing capacity tables:

```py
PLAN_CAPACITY_CREDITS_PRIMARY = {
    "free": 33.75,
    "plus": 225.0,
    "business": 225.0,
    "team": 225.0,
    "edu": 225.0,
    "pro": 1500.0,
    "enterprise": 1500.0,
}
```

`capacity_for_plan("prolite", ...)` therefore returns `None`, and `summarize_usage_window()` skips those rows when computing aggregate capacity/used credits.

## Possible fix

Treat `prolite` as a recognized account plan and map its capacity to the Pro capacity values, or add a plan alias layer so `prolite -> pro` for capacity and routing calculations while preserving the raw displayed plan if desired.

## Related context

Issue #479 has a comment from another user mentioning the same `Prolite account` class of problem, though that issue is framed as quota exceeded / routing behavior rather than aggregate capacity display.

aruis:
## Summary

- Recognize upstream `prolite` account plan values without rewriting stored account rows.
- Map Pro Lite accounts to Pro-equivalent primary and secondary usage capacities for dashboard totals and capacity-weighted routing.
- Treat Pro Lite as Pro-equivalent when model registry plan gates allow `pro`, so Pro-gated models can select Pro Lite accounts.
- Add OpenSpec change artifacts and regression coverage for auth normalization, usage summaries, dashboard overview, and proxy account selection.

Fixes #557

## Validation

- `uv run pytest -q tests/unit/test_auth.py tests/unit/test_usage.py tests/unit/test_proxy_load_balancer_refresh.py tests/integration/test_dashboard_overview.py`
- `uv run ruff check app/core/plan_types.py app/core/usage/__init__.py app/modules/proxy/load_balancer.py tests/unit/test_auth.py tests/unit/test_usage.py tests/unit/test_proxy_load_balancer_refresh.py tests/integration/test_dashboard_overview.py`
- `uv run ty check app/core/plan_types.py app/core/usage/__init__.py app/modules/proxy/load_balancer.py tests/unit/test_auth.py tests/unit/test_usage.py tests/unit/test_proxy_load_balancer_refresh.py tests/integration/test_dashboard_overview.py`

## Notes

- `uv run openspec validate --specs` could not run locally because the `openspec` executable is not installed in this environment.
- A local amd64 Docker test image was exercised against the reported Pro Lite account, confirming dashboard quota totals and `gpt-5.4` request routing after proxy configuration was corrected.

### DIFF
--- a/app/core/plan_types.py
+++ b/app/core/plan_types.py
@@ -6,6 +6,7 @@ ACCOUNT_PLAN_TYPES: Final[set[str]] = {
     "free",
     "plus",
     "pro",
+    "prolite",
     "team",
     "business",
     "enterprise",
@@ -20,6 +21,10 @@ RATE_LIMIT_PLAN_TYPES: Final[set[str]] = {
     "education",
     "quorum",
     "k12",
+}
+
+ACCOUNT_PLAN_EQUIVALENTS: Final[dict[str, frozenset[str]]] = {
+    "prolite": frozenset({"pro"}),
 }
 
 
@@ -62,3 +67,13 @@ def normalize_rate_limit_plan_type(value: str | None) -> str | None:
         return None
     normalized = cleaned.lower()
     return normalized if normalized in RATE_LIMIT_PLAN_TYPES else None
+
+
+def account_plan_matches_allowed(value: str | None, allowed_plans: set[str] | frozenset[str]) -> bool:
+    normalized = normalize_account_plan_type(value)
+    if normalized is None:
+        return False
+    normalized_allowed = {plan.lower() for plan in allowed_plans}
+    if normalized in normalized_allowed:
+        return True
+    return bool(ACCOUNT_PLAN_EQUIVALENTS.get(normalized, frozenset()) & normalized_allowed)
--- a/app/core/plan_types.py
+++ b/app/core/plan_types.py
@@ -70,10 +70,14 @@ def normalize_rate_limit_plan_type(value: str | None) -> str | None:
 
 
 def account_plan_matches_allowed(value: str | None, allowed_plans: set[str] | frozenset[str]) -> bool:
-    normalized = normalize_account_plan_type(value)
-    if normalized is None:
+    cleaned = _clean_plan_type(value)
+    if cleaned is None:
         return False
     normalized_allowed = {plan.lower() for plan in allowed_plans}
-    if normalized in normalized_allowed:
+    normalized = normalize_account_plan_type(cleaned)
+    candidate = normalized or cleaned.lower()
+    if candidate in normalized_allowed:
         return True
+    if normalized is None:
+        return False
     return bool(ACCOUNT_PLAN_EQUIVALENTS.get(normalized, frozenset()) & normalized_allowed)
--- a/app/core/plan_types.py
+++ b/app/core/plan_types.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import Final
+from typing import AbstractSet, Final
 
 ACCOUNT_PLAN_TYPES: Final[set[str]] = {
     "free",
@@ -69,7 +69,7 @@ def normalize_rate_limit_plan_type(value: str | None) -> str | None:
     return normalized if normalized in RATE_LIMIT_PLAN_TYPES else None
 
 
-def account_plan_matches_allowed(value: str | None, allowed_plans: set[str] | frozenset[str]) -> bool:
+def account_plan_matches_allowed(value: str | None, allowed_plans: AbstractSet[str]) -> bool:
     cleaned = _clean_plan_type(value)
     if cleaned is None:
         return False
--- a/app/core/usage/__init__.py
+++ b/app/core/usage/__init__.py
@@ -22,7 +22,7 @@ PLAN_CAPACITY_CREDITS_PRIMARY = {
     "team": 225.0,
     "edu": 225.0,
     "pro": 1500.0,
-    "prolite": 1500.0,
+    "prolite": 1125.0,
     "enterprise": 1500.0,
 }
 
@@ -33,7 +33,7 @@ PLAN_CAPACITY_CREDITS_SECONDARY = {
     "team": 7560.0,
     "edu": 7560.0,
     "pro": 50400.0,
-    "prolite": 50400.0,
+    "prolite": 37800.0,
     "enterprise": 50400.0,
 }
 
--- a/app/core/usage/__init__.py
+++ b/app/core/usage/__init__.py
@@ -22,6 +22,7 @@ PLAN_CAPACITY_CREDITS_PRIMARY = {
     "team": 225.0,
     "edu": 225.0,
     "pro": 1500.0,
+    "prolite": 1500.0,
     "enterprise": 1500.0,
 }
 
@@ -32,6 +33,7 @@ PLAN_CAPACITY_CREDITS_SECONDARY = {
     "team": 7560.0,
     "edu": 7560.0,
     "pro": 50400.0,
+    "prolite": 50400.0,
     "enterprise": 50400.0,
 }
 
--- a/app/modules/proxy/load_balancer.py
+++ b/app/modules/proxy/load_balancer.py
@@ -26,6 +26,7 @@ from app.core.balancer import (
 from app.core.balancer.types import UpstreamError
 from app.core.config.settings import get_settings
 from app.core.openai.model_registry import get_model_registry
+from app.core.plan_types import account_plan_matches_allowed
 from app.core.resilience.circuit_breaker import are_all_account_circuit_breakers_open
 from app.core.resilience.degradation import get_status as get_degradation_status
 from app.core.resilience.degradation import set_degraded, set_normal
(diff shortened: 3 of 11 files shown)


Soju06:
@codex review

Soju06:
Thanks @aruis -- the prolite plumbing looks right and `Fixes #557` is on the money. Codex review flagged one real regression that I think we should address before merge:

## P1 -- `account_plan_matches_allowed` rejects unknown plans

The old `_filter_accounts_for_model` did a direct string comparison:

```python
allowed_plans = ...  # from upstream model.available_in_plans
return [a for a in accounts if a.plan_type in allowed_plans]
```

The new helper short-circuits to `False` when `normalize_account_plan_type(value)` returns `None`:

```python
def account_plan_matches_allowed(value, allowed_plans):
    normalized = normalize_account_plan_type(value)
    if normalized is None:
        return False  # <-- regression
    ...
```

`normalize_account_plan_type` only knows the values currently in `ACCOUNT_PLAN_TYPES`. So if OpenAI later ships a new plan slug (the next `prolite`-style addition), accounts on that plan will silently fail model gating for *every* model, including ones whose registry entry explicitly lists the new plan in `available_in_plans`. This is the regression direction.

Suggested shape -- preserve a direct string fallback so unknown-but-explicitly-allowed plans still work:

```python
def account_plan_matches_allowed(value, allowed_plans):
    cleaned = _clean_plan_type(value)
    if not cleaned:
        return False
    normalized = canonicalize_account_plan_type(cleaned)
    normalized_allowed = {plan.lower() for plan in allowed_plans}
    candidate = (normalized or cleaned.lower())
    if candidate in normalized_allowed:
        return True
    if normalized is not None and ACCOUNT_PLAN_EQUIVALENTS.get(normalized, frozenset()) & normalized_allowed:
        return True
    return False
```

(I.e. fall through to the lowercased raw string when normalization can't resolve it, instead of dropping the account.)

A quick regression test in `tests/unit/test_plan_types.py` would help -- something like:

```python
def test_unknown_plan_passes_when_explicitly_allowed():
    assert account_plan_matches_allowed("future_plan", {"future_plan", "plus"})

def test_unknown_plan_blocked_when_not_in_allowed():
    assert not account_plan_matches_allowed("future_plan", {"plus"})
```

Once that's in I'll re-run `@codex review` and if it lands clean we can merge. The rest of the PR (capacity entries, equivalents map, openspec deltas) is fine.

Heads-up that I dropped a couple of follow-up notes on #557 too:

- The current `prolite -> pro` capacity values mean prolite accounts get the same routing weight as Pro, even though their actual quotas may be smaller. If routing telemetry shows prolite over-utilization later, calibrating capacity values down (or treating prolite as roughly `plus * 5`) would be a clean follow-up.
- sharenla's point #4 (stale `quota_exceeded` / `rate_limited` not auto-recovering) looks like a separate issue from prolite -- the recovery logic in `_compute_account_state` already runs on every selection. If anyone has a concrete repro it deserves its own ticket.

Will re-trigger review after the regression fix.

aruis:
Addressed the P1 regression in `eaa044d`:

- `account_plan_matches_allowed` now falls back to the lowercased raw plan slug when normalization does not recognize a future upstream plan.
- Unknown plans still remain blocked unless the model registry explicitly lists that exact plan.
- Added `tests/unit/test_plan_types.py` coverage for explicitly allowed and blocked unknown plans, plus Pro Lite equivalence.

Local validation:

- `uv run pytest -q tests/unit/test_auth.py tests/unit/test_plan_types.py tests/unit/test_usage.py tests/unit/test_proxy_load_balancer_refresh.py tests/integration/test_dashboard_overview.py`
- `uv run ruff check app/core/plan_types.py tests/unit/test_plan_types.py tests/unit/test_usage.py tests/unit/test_proxy_load_balancer_refresh.py tests/integration/test_dashboard_overview.py`
- `uv run ty check app/core/plan_types.py tests/unit/test_plan_types.py tests/unit/test_usage.py tests/unit/test_proxy_load_balancer_refresh.py tests/integration/test_dashboard_overview.py`

Soju06:
@codex review

Soju06:
Thanks @aruis -- the P1 fix and the raw-fallback for unknown plans both look correct, codex review is clean across the latest commit, and the local docker repro you mentioned is a strong signal. Before merging I'd like to fold one capacity nuance @sharenla raised in #557 so we don't ship Pro Lite with an over-provisioned weight.

## What I'd like changed

```python
# app/core/usage/__init__.py
PRIMARY_CAPACITY = {
    ...,
    "pro": 1500.0,
-   "prolite": 1500.0,
+   "prolite": 1125.0,   # plus x5 long-term level
    "enterprise": 1500.0,
}

SECONDARY_CAPACITY = {
    ...,
    "pro": 50400.0,
-   "prolite": 50400.0,
+   "prolite": 37800.0,  # plus x5 long-term level
    "enterprise": 50400.0,
}
```

## Why

Per @sharenla's [#557 follow-up](https://github.com/Soju06/codex-lb/issues/557), Pro Lite Codex usage is product-documented as roughly **Plus x5** long-term:

- `prolite_primary  = plus_primary  x 5  = 225  x 5  = 1125`
- `prolite_secondary = plus_secondary x 5 = 7560 x 5 = 37800`

There is also a Pro Lite / Pro mid-tier promo running through May 31 with Codex usage closer to Plus x10, but as @sharenla notes, baking the temporary promo into the long-term default is product-maintenance debt -- the conservative default is the permanent Plus x5 baseline. Operators who want to track the promo can override the constants out-of-band; defaulting to Pro (1500 / 50400) is x6.7 over the long-term Pro Lite ceiling and x10+ over what Plus customers would expect parity with.

This only affects derived credit totals / remaining-credit donuts and capacity-weighted routing weights -- upstream `used_percent` is still the source of truth for how much of the account window is consumed (sharenla's wording), so percent-based routing decisions stay correct either way. But for the dashboard math and capacity-weighted routing this is worth getting right at merge.

## Tests

The two existing tests that exercise capacity will need their expected numbers updated to the new values:

- `tests/unit/test_usage.py` -- `prolite` capacity assertions
- `tests/integration/test_dashboard_overview.py` -- aggregate dashboard numbers when a Pro Lite account is present

The plan-matching / model-gating tests stay the same -- those don't depend on the capacity numbers, only on `ACCOUNT_PLAN_EQUIVALENTS`.

## After this push

I'll re-run `@codex review` once and merge. Sorry for one more round; the rest of the PR is in great shape.

aruis:
Updated the PR to fold in the Pro Lite capacity nuance from #557:

- `prolite` capacity is now Plus x5: `1125.0` primary / `37800.0` secondary.
- Dashboard and usage regression expectations now assert those values.
- Model eligibility remains Pro-equivalent for Pro-gated models.
- OpenSpec change text now distinguishes Plus x5 capacity from Pro-equivalent model eligibility.

Local validation:

```bash
uv run pytest -q tests/unit/test_auth.py tests/unit/test_usage.py tests/unit/test_plan_types.py tests/unit/test_proxy_load_balancer_refresh.py tests/integration/test_dashboard_overview.py
uv run ruff check app/core/usage/__init__.py tests/unit/test_auth.py tests/unit/test_usage.py tests/unit/test_plan_types.py tests/unit/test_proxy_load_balancer_refresh.py tests/integration/test_dashboard_overview.py
uv run ty check app/core/usage/__init__.py tests/unit/test_auth.py tests/unit/test_usage.py tests/unit/test_plan_types.py tests/unit/test_proxy_load_balancer_refresh.py tests/integration/test_dashboard_overview.py
```

All passed locally. `openspec validate --specs` still cannot run in my local environment because the `openspec` executable is not installed.

Soju06:
@codex review

Soju06:
@codex review

Soju06:
@all-contributors please add @aruis for code, test, bug

PR #559, opened 2026-05-06
fix(accounts): recover quota status from usage refresh

Komzpa:
## Summary

- Recover `quota_exceeded` accounts to `active` when a successful usage refresh reports available quota.
- Clear stale `reset_at` / `blocked_at` together with the status recovery.
- Cover both ordinary primary-window recovery and free weekly-only accounts whose upstream 7d window is normalized into the secondary window.

## Why

An account can keep a stale `quota_exceeded` status even after the usage refresh has fresh upstream evidence that quota is available again. The dashboard then shows a contradictory state: 100% remaining usage but a quota-exceeded badge, and routing still treats the account as unavailable.

#529 covers a related timer-based recovery path after reset time passes. This patch handles the evidence-based path: if upstream usage refresh succeeds and reports `used_percent < 100`, the account is usable again.

## Validation

- `uv run pytest tests/unit/test_usage_updater.py -k 'recovers_quota_exceeded' -q`
- `uv run ruff check app/modules/usage/updater.py tests/unit/test_usage_updater.py`
- `uv run ty check app/modules/usage/updater.py tests/unit/test_usage_updater.py`

### DIFF
--- a/app/modules/usage/updater.py
+++ b/app/modules/usage/updater.py
@@ -468,9 +468,7 @@ class UsageUpdater:
             secondary is None
             or not _window_has_available_quota(secondary)
             or any(_window_is_exhausted(window) for window in windows)
-            or not any(
-                _window_has_available_quota(window) for window in windows
-            )
+            or not any(_window_has_available_quota(window) for window in windows)
         ):
             return
 
--- a/app/modules/usage/updater.py
+++ b/app/modules/usage/updater.py
@@ -464,8 +464,13 @@ class UsageUpdater:
         if account.status != AccountStatus.QUOTA_EXCEEDED or not self._auth_manager:
             return
         windows = [window for window in (primary, secondary) if window is not None]
-        if any(_window_is_exhausted(window) for window in windows) or not any(
-            _window_has_available_quota(window) for window in windows
+        if (
+            secondary is None
+            or not _window_has_available_quota(secondary)
+            or any(_window_is_exhausted(window) for window in windows)
+            or not any(
+                _window_has_available_quota(window) for window in windows
+            )
         ):
             return
 
--- a/app/modules/usage/updater.py
+++ b/app/modules/usage/updater.py
@@ -17,7 +17,7 @@ from app.core.clients.usage import UsageFetchError, fetch_usage
 from app.core.config.settings import get_settings
 from app.core.crypto import TokenEncryptor
 from app.core.plan_types import coerce_account_plan_type
-from app.core.usage.models import AdditionalRateLimitPayload, UsagePayload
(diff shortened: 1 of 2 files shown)


Soju06:
@codex review

Soju06:
Thanks @Komzpa -- the timer-vs-evidence split with #529 is the right shape and the free-weekly test case is a nice catch. Codex review just landed one P1 finding on `2a9e236` that I think is real:

## P1 -- recovery condition is too permissive

Current logic:

```python
windows = [w for w in (primary, secondary) if w is not None]
if not any(_window_has_available_quota(window) for window in windows):
    return
```

This recovers the account if **either** window reports `used_percent < 100`. But for an account whose upstream-driving quota is the secondary (weekly) window -- a Plus account that has burned through its 7d budget while the 5h short-term window naturally cycles back to low usage -- this means we recover to ACTIVE while `secondary >= 100`, route a request to it, and immediately hit `usage_limit_reached` again upstream.

The recovery condition needs evidence that the *quota-driving* window has cleared, not just any window. Two reasonable shapes:

**Option A** (simple): require *both* windows to be available when both are present.

```python
windows = [w for w in (primary, secondary) if w is not None]
if not windows:
    return
if not all(_window_has_available_quota(window) for window in windows):
    return
```

This is the most conservative -- a quota_exceeded account stays exceeded until upstream evidence shows neither window is at the cap. Slightly over-conservative when primary is the only meaningful signal, but never wrong.

**Option B** (precise): preserve the existing free-weekly-only behavior by checking presence per window and rejecting recovery when any *present* window is still saturated. Same as A in practice, but the comment makes the intent clear:

```python
# Recover only when no present window is still at or above 100% --
# we cannot trust which window is the quota-driving signal here, so
# any window at the cap means the account is still blocked upstream.
```

I'd go with A; the implementation is identical, B is just a commentary tweak.

## Test additions to lock the contract

```python
def test_usage_refresh_does_not_recover_when_secondary_still_exceeded()
    # primary.used_percent = 0, secondary.used_percent = 100 -> stays QUOTA_EXCEEDED

def test_usage_refresh_does_not_recover_when_primary_still_exceeded()
    # primary.used_percent = 100, secondary.used_percent = 0 -> stays QUOTA_EXCEEDED
```

The two existing tests (`recovers_quota_exceeded` and `recovers_quota_exceeded_free_weekly`) keep passing under either shape since both have the relevant window at 0%.

Once that's in I'll re-run `@codex review` and we can land it. The dashboard-mismatch fix from this PR is a real user pain so I'd like to get it through quickly.

Soju06:
@codex review

Soju06:

PR #560, opened 2026-05-06
fix(proxy): slim oversized response.create history

### LINKED ISSUE #556: Large response.create payloads still 413 after slimming; dump path unwritable on macOS uv install
## Summary

`codex-lb` v1.15.0 can still return `413 Payload Too Large` for oversized Codex `response.create` requests even though there is already historical payload slimming logic. On macOS local/uv installs, the debug dump path also appears to be hardcoded to a Docker/Linux path that is not writable.

## Environment

- `codex-lb` version: `1.15.0`
- Install/run mode: `uv tool`, running as a macOS LaunchAgent
- Endpoint: `http://127.0.0.1:2455/backend-api/codex/responses`
- Client: Codex CLI using Responses/WebSocket support through `codex-lb`

## Observed error

A Codex request failed with:

```text
unexpected status 413 Payload Too Large: response.create is too large for upstream websocket (18483980 bytes > 15728640 bytes). Reduce historical images/screenshots or compact the thread.
```

Relevant logs:

```text
Large response.create prepared request_id=ws_a193c8a257464863b8522ff5c3b7695e request_log_id=b973d908-4c60-4bb3-860a-34a890a96da0 transport=http bytes=18483980 previous_response_id=None
proxy_error_response request_id=b973d908-4c60-4bb3-860a-34a890a96da0 method=POST path=/backend-api/codex/responses status=413 code=payload_too_large message=response.create is too large for upstream websocket (18483980 bytes > 15728640 bytes). Reduce historical images/screenshots or compact the thread.
```

The debug dump also failed:

```text
PermissionError: [Errno 13] Permission denied: '/var/lib/codex-lb'
```

## What seems to be happening

`_slim_response_create_payload_for_upstream` removes historical inline images and large historical tool outputs, but if the remaining historical input items are still over the upstream websocket budget, `_enforce_response_create_size_limit` still raises 413.

There is already a helper named `_response_create_history_omission_notice_item`, but it does not appear to be used in the slimming path.

Separately, `_OVERSIZED_RESPONSE_CREATE_DUMP_DIR` is hardcoded to:

```py
Path("/var/lib/codex-lb/debug/response-create-dumps")
```

That makes sense for Docker, but not for a local macOS `uv tool`/LaunchAgent install where the process cannot create `/var/lib/codex-lb`.

## Expected behavior

- If image/tool-output slimming is not enough, `codex-lb` should omit the oldest historical input items until the serialized `response.create` fits the websocket budget, while preserving the recent/current user input.
- The oversized response-create dump directory should default to the app data dir outside containers, e.g. `~/.codex-lb/debug/response-create-dumps`, or be configurable via settings/env.

## Local workaround tested

I patched my local install to:

1. Continue slimming by dropping oldest historical input items after the existing image/tool-output slimming if the payload still exceeds `_UPSTREAM_RESPONSE_CREATE_MAX_BYTES`.
2. Move oversized dumps to:

```py
Path.home() / ".codex-lb" / "debug" / "response-create-dumps"
```

A synthetic oversized historical payload that was >15 MiB then slimmed successfully under the guard.

Komzpa:
## Summary

- Continue `response.create` slimming by omitting the oldest historical input items when image/tool-output slimming is still over the websocket budget.
- Preserve the recent user-turn suffix and add a single assistant notice with the omitted item count.
- Move oversized `response.create` debug dumps under the default app home directory so local user installs use `~/.codex-lb` while containers keep `/var/lib/codex-lb`.

Fixes #556.

## Validation

- `openspec validate --specs`
- `uv run pytest tests/unit/test_proxy_utils.py -k 'slim_response_create or oversized_response_create_dump_dir' -q`
- `uv run ruff check app/modules/proxy/service.py tests/unit/test_proxy_utils.py`
- `uv run ty check app/modules/proxy/service.py tests/unit/test_proxy_utils.py`

### DIFF
--- a/app/core/clients/proxy.py
+++ b/app/core/clients/proxy.py
@@ -1430,27 +1430,36 @@ def _omit_historical_response_input_items_to_fit(
     if not slimmed_historical:
         return payload, 0
 
-    omitted_count = 0
-    remaining_historical = list(slimmed_historical)
-    candidate_payload = payload
-    while remaining_historical:
-        omitted_count += 1
-        remaining_historical = remaining_historical[1:]
+    def candidate_with_omitted_count(omitted_count: int) -> JsonObject:
         candidate_payload = dict(payload)
         candidate_payload["input"] = [
             _response_create_history_omission_notice_item(omitted_count),
-            *remaining_historical,
+            *slimmed_historical[omitted_count:],
             *recent,
         ]
+        return candidate_payload
+
+    low = 1
+    high = len(slimmed_historical)
+    best: tuple[JsonObject, int] | None = None
+    while low <= high:
+        omitted_count = (low + high) // 2
+        candidate_payload = candidate_with_omitted_count(omitted_count)
         if _serialized_json_size(candidate_payload) <= max_bytes:
-            return candidate_payload, omitted_count
+            best = (candidate_payload, omitted_count)
+            high = omitted_count - 1
+        else:
+            low = omitted_count + 1
+
+    if best is not None:
+        return best
 
     recent_only_payload = dict(payload)
     recent_only_payload["input"] = list(recent)
     if _serialized_json_size(recent_only_payload) <= max_bytes:
-        return recent_only_payload, omitted_count
+        return recent_only_payload, len(slimmed_historical)
 
-    return candidate_payload, omitted_count
+    return candidate_with_omitted_count(len(slimmed_historical)), len(slimmed_historical)
 
 
 def _serialized_json_size(payload: JsonObject) -> int:
--- a/app/core/clients/proxy.py
+++ b/app/core/clients/proxy.py
@@ -1430,26 +1430,31 @@ def _omit_historical_response_input_items_to_fit(
     if not slimmed_historical:
         return payload, 0
 
-    def candidate_with_omitted_count(omitted_count: int) -> JsonObject:
+    def candidate_with_omitted_count(prefix_omitted_count: int) -> tuple[JsonObject, int]:
+        remaining_historical, orphaned_outputs = _drop_orphaned_function_call_outputs(
+            slimmed_historical[prefix_omitted_count:],
+            recent,
+        )
+        omitted_count = prefix_omitted_count + orphaned_outputs
         candidate_payload = dict(payload)
         candidate_payload["input"] = [
             _response_create_history_omission_notice_item(omitted_count),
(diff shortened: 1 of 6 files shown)


Soju06:
@codex review

Soju06:
Thanks @Komzpa -- the historical-item omission and `DEFAULT_HOME_DIR` move both look right and #556 will close cleanly. Codex review landed one P2 finding I think is worth a small fix:

## P2 -- recent-only fallback when notice + recent still over budget

In `_omit_historical_response_input_items_to_fit`, after every historical item has been removed the loop still prepends `_response_create_history_omission_notice_item`. If `notice + recent` is itself over `max_bytes` (small history items, large notice text, or already-large `recent`), the function returns the oversized payload anyway and the caller still 413s -- even though dropping the notice would have fit.

Suggested addition right before the function returns the final `candidate_payload`:

```python
if _serialized_json_size(candidate_payload) > max_bytes:
    # Final fallback: notice + recent itself is oversized; try recent
    # alone. Drops the omission disclosure but lets the request go
    # through instead of hard-failing the user-visible turn.
    recent_only = dict(payload)
    recent_only["input"] = list(recent)
    if _serialized_json_size(recent_only) <= max_bytes:
        return recent_only, omitted_count
```

The trade-off is that the user no longer sees the "omitted N items" notice, but they also do not get a 413 in a case where the request could have succeeded. If you'd rather keep the notice always and just log a warning + still 413, that's a defensible call too -- the current behavior is silently returning oversized.

Quick regression to lock the contract:

```python
def test_slim_response_create_drops_notice_when_notice_plus_recent_oversized():
    # tiny historical items, large enough recent + notice text combined > max_bytes,
    # but recent alone <= max_bytes
    # assert summary['historical_items_omitted'] == len(history)
    # assert _serialized_json_size(slimmed_payload) <= max_bytes
    # assert no notice item in slimmed_payload['input']
```

Not a hard blocker -- if you'd rather keep this PR scoped to the original ask and ship as-is, fine, the current behavior is no worse than today's pre-PR 413. Up to you. The dump-dir move and the omission-pass core both look right.

If you push the fallback I'll re-run `@codex review` and we can land it. Otherwise let me know and I'll go ahead and merge with the P2 as a known soft edge.

Soju06:
@codex review

Soju06:
@codex review

Soju06:

Soju06:
@codex review

PR #561, opened 2026-05-06
fix(proxy): narrow budget-safe gate to primary usage

### LINKED ISSUE #446: [follow-up] Narrow `_state_above_budget_threshold` primary-only gating
Follow-up to #421 (merged as `e632d94`).

## Context

In #421 we introduced `_state_above_budget_threshold(state, threshold)` which returns `True` if either `used_percent` (primary) **or** `secondary_used_percent` exceeds the configured budget threshold. That is used by `_select_account_preferring_budget_safe()` to filter accounts before falling back to the normal `select_account()` path.

[@chatgpt-codex-connector flagged a narrow P1 edge case](https://github.com/Soju06/codex-lb/pull/421#discussion_r3114781748) that survived the merge:

When **all** candidate accounts cross the threshold, the preferred-safe filter collapses to the full pool. In that degraded state, `usage_weighted` ranking uses `_usage_sort_key = (secondary_used, primary_used, ...)`, so:

```
A: primary 99%, secondary  5%  -> sort key (5, 99, ...)
B: primary 10%, secondary 96%  -> sort key (96, 10, ...)
min(A, B) -> A   ← near-exhausted primary wins, usage_limit_reached risk
```

## Proposed fix

Narrow the gate to the **primary** window only (which is the one that actually throws `usage_limit_reached`):

```python
def _state_above_budget_threshold(state: AccountState, budget_threshold_pct: float) -> bool:
    return state.used_percent is not None and state.used_percent > budget_threshold_pct
```

Alternative: keep the `any(...)` shape but add a secondary tiebreaker that penalizes high primary in the sort key when all candidates are pressured. More invasive, probably unnecessary.

## Impact

- Reproduces only on `usage_weighted` strategy
- Only when every candidate is `> threshold` (already the degraded fallback path)
- Orthogonal to the tool-payload and sticky-reallocation fixes in #421, which are unaffected

## Tests to add

- `test_proxy_load_balancer_refresh.py::test_budget_safe_filter_ignores_secondary_only_pressure_when_primary_safe`
- `test_proxy_load_balancer_refresh.py::test_budget_safe_fallback_does_not_pick_near_exhausted_primary_under_usage_weighted`

Komzpa:
## Summary

- narrows the budget-safe Responses routing hard gate to primary-window usage only
- keeps secondary/weekly usage as a routing strategy signal instead of a hard exclusion signal
- makes the degraded all-primary-pressured `usage_weighted` fallback prefer lower primary pressure before lower weekly usage
- adds OpenSpec coverage and regression tests for the #446 edge cases

Closes #446.

## Verification

- `openspec validate narrow-budget-safe-primary-gate --strict`
- `UV_PROJECT_ENVIRONMENT=/home/kom/.venvs/codex-lb-fix-446 uv run pytest tests/unit/test_proxy_load_balancer_refresh.py -k 'budget_safe or reads_cached_usage_once' -q`\n- `UV_PROJECT_ENVIRONMENT=/home/kom/.venvs/codex-lb-fix-446 uv run ruff check app/modules/proxy/load_balancer.py tests/unit/test_proxy_load_balancer_refresh.py`\n- `UV_PROJECT_ENVIRONMENT=/home/kom/.venvs/codex-lb-fix-446 uv run ruff format --check app/modules/proxy/load_balancer.py tests/unit/test_proxy_load_balancer_refresh.py`\n- `UV_PROJECT_ENVIRONMENT=/home/kom/.venvs/codex-lb-fix-446 uv run ty check app/modules/proxy/load_balancer.py tests/unit/test_proxy_load_balancer_refresh.py`

### DIFF
--- a/app/core/balancer/logic.py
+++ b/app/core/balancer/logic.py
@@ -260,10 +260,7 @@ def select_account(
             selected = _select_capacity_weighted(candidate_pool)
     else:
         if primary_first_usage_weighted:
-            selected = min(
-                effective_pool,
-                key=_primary_reset_first_sort_key if prefer_earlier_reset else _primary_usage_sort_key,
-            )
+            selected = min(effective_pool, key=_primary_usage_sort_key)
         else:
             selected = min(effective_pool, key=_reset_first_sort_key if prefer_earlier_reset else _usage_sort_key)
     return SelectionResult(selected, None)
--- a/app/core/balancer/logic.py
+++ b/app/core/balancer/logic.py
@@ -77,6 +77,13 @@ def _usage_sort_key(state: AccountState) -> tuple[float, float, float, str]:
     return secondary_used, primary_used, last_selected, state.account_id
 
 
+def _primary_usage_sort_key(state: AccountState) -> tuple[float, float, float, str]:
+    primary_used = state.used_percent if state.used_percent is not None else 0.0
+    secondary_used = state.secondary_used_percent if state.secondary_used_percent is not None else primary_used
+    last_selected = state.last_selected_at or 0.0
+    return primary_used, secondary_used, last_selected, state.account_id
+
+
 def _reset_bucket_days(state: AccountState, current: float) -> int:
     if state.secondary_reset_at is None:
         return UNKNOWN_RESET_BUCKET_DAYS
@@ -105,6 +112,7 @@ def select_account(
     routing_strategy: RoutingStrategy = "capacity_weighted",
     allow_backoff_fallback: bool = True,
     deterministic_probe: bool = False,
+    primary_first_usage_weighted: bool = False,
 ) -> SelectionResult:
     """Select an eligible account by applying availability checks and routing strategy.
 
@@ -127,6 +135,8 @@ def select_account(
             account exists.
         deterministic_probe: Whether capacity-weighted routing should use a
             deterministic probe order instead of random weighted choice.
+        primary_first_usage_weighted: Whether usage-weighted routing should
+            rank by primary-window pressure before secondary-window pressure.
 
     Returns:
         A ``SelectionResult`` containing the selected ``AccountState`` and no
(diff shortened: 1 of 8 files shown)


Soju06:
@codex review

Soju06:
Thanks @Komzpa -- the #446 narrowing is exactly right. Codex review just landed two P1 findings on `b10cabe` that I think are real and worth addressing before merge:

## P1 -- bypassing availability filtering

The new all-primary-pressured branch iterates `sorted(state_list, key=_primary_budget_pressure_sort_key)` and calls `select_account([state], ...)` per single-state list. `select_account`'s availability filtering (RATE_LIMITED / QUOTA_EXCEEDED reset, cooldown suppression, error-backoff) does run on each single-state call, but the *ordering* of which states get tried first comes from the outer sort, not from the selector. So under `usage_weighted` with all candidates primary-pressured, an account that is currently unavailable can win the iteration just because its primary usage is the lowest.

## P1 -- bypassing health-tier preference

Same shape: `select_account([state])` cannot reason about `healthy` vs `probing` vs `draining` when there's only one state in the list, so the pool-level `healthy > probing > draining` preference that `select_account` would normally apply is silently lost. A draining account with slightly lower primary usage can now win over a healthy account in the all-pressured fallback.

## Suggested shape

Keep the gate-narrowing change as-is (that's the right fix for #446), but rework the all-pressured fallback so it stays inside one `select_account` call:

- Either: include health tier and availability flags in `_primary_budget_pressure_sort_key` (e.g. `(health_tier_rank, availability_blocked, primary_used, secondary_used, last_selected, account_id)`) and let `min(states, key=...)` produce the final ordering before calling `select_account([sorted_state_list], ...)` once.
- Or, push the primary-first rule into the existing `usage_weighted` ranking inside `select_account` so it applies naturally instead of being a separate fallback path. That keeps health/availability/cooldown logic in one place.

The second option is cleaner architecturally, the first is less invasive. Either is fine.

Could you also add a regression test for each P1:

```python
def test_all_primary_pressured_fallback_skips_unavailable_account()
def test_all_primary_pressured_fallback_prefers_healthy_over_draining()
```

Tests can stub `AccountState.cooldown_until` / `error_backoff_until` / status `RATE_LIMITED` for the unavailability case, and `health_tier=draining` vs `healthy` for the second.

The OpenSpec proposal copy is correct -- both scenarios (`secondary-only pressure stays eligible` and `primary-first ordering in the fallback`) hold under either fix shape, so no spec change required.

Ping me with the new commit and I'll re-run `@codex review`.

Soju06:
@codex review

Soju06:

PR #564, opened 2026-05-07
fix(proxy): expose drain status for graceful deploys

Komzpa:
## Summary
- keep drain control endpoints reachable while the server is already draining
- expose `/internal/drain/status` with `draining`, `bridge_drain_active`, and `in_flight` counters for deploy tooling
- exclude health/drain observer endpoints from the in-flight counter so drain polling does not keep itself alive
- update the Helm `preStop` hook to start drain and poll `/internal/drain/status` instead of only sleeping a fixed interval

This is a graceful-drain primitive, not a full single-container zero-downtime deploy story: websocket sessions still need an old backend kept alive until they close, and local Docker deployments need a separate stable frontend/proxy for true blue/green cutover.

## Tests
- `UV_PROJECT_ENVIRONMENT=/home/kom/.venvs/codex-lb-routing-policy uv run pytest tests/unit/test_graceful_shutdown.py tests/unit/test_health_probes.py tests/unit/test_helm_external_secrets.py -q` (Helm rendering tests skipped locally because `helm` is not installed in this environment)
- `UV_PROJECT_ENVIRONMENT=/home/kom/.venvs/codex-lb-routing-policy uv run ruff check app/core/middleware/inflight.py app/modules/health/api.py tests/unit/test_graceful_shutdown.py tests/unit/test_health_probes.py tests/unit/test_helm_external_secrets.py`
- `UV_PROJECT_ENVIRONMENT=/home/kom/.venvs/codex-lb-routing-policy uv run ty check app/core/middleware/inflight.py app/modules/health/api.py tests/unit/test_graceful_shutdown.py tests/unit/test_health_probes.py tests/unit/test_helm_external_secrets.py`
- `git diff --check`

### DIFF
--- a/app/core/middleware/inflight.py
+++ b/app/core/middleware/inflight.py
@@ -8,7 +8,19 @@ from starlette.types import ASGIApp, Receive, Scope, Send
 _DRAIN_ALLOWED_HTTP_PATHS = frozenset(
     {
         "/health/live",
+        "/internal/drain/start",
+        "/internal/drain/status",
         "/internal/bridge/responses",
+    }
+)
+
+_IN_FLIGHT_EXCLUDED_HTTP_PATHS = frozenset(
+    {
+        "/health/live",
+        "/health/ready",
+        "/health/startup",
+        "/internal/drain/start",
+        "/internal/drain/status",
     }
 )
 
@@ -39,6 +51,10 @@ class InFlightMiddleware:
                 },
             )
             await response(scope, receive, send)
+            return
+
+        if path in _IN_FLIGHT_EXCLUDED_HTTP_PATHS:
+            await self.app(scope, receive, send)
             return
 
         shutdown_state.increment_in_flight()
--- a/app/modules/health/api.py
+++ b/app/modules/health/api.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 from datetime import timedelta
 from hashlib import sha256
+from ipaddress import ip_address
 
 from fastapi import APIRouter, HTTPException, Request
 from sqlalchemy import select as sa_select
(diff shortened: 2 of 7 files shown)


Soju06:
@codex review

Soju06:
@codex review

Soju06:
Thanks @Komzpa -- the polling-based preStop and the in-flight exclusion for health/drain endpoints are both right. Codex review caught one real availability/security regression I'd like to address before merge:

## P1 -- private-network trust is too wide

The widening from loopback-only to `address.is_private` lets any workload sharing the cluster CIDR (or any RFC1918 / CGN-NAT host on a flat lab network) call `POST /internal/drain/start` and immediately flip readiness to 503 for normal traffic. In a multi-tenant cluster or any environment where the pod is not the only thing on `10.0.0.0/8` / `172.16.0.0/12` / `192.168.0.0/16`, that's a denial-of-service surface.

The Helm `preStop` hook calls `http://127.0.0.1:{httpPort}`, so it's already on loopback inside the same pod. So the original `{"127.0.0.1", "::1", "localhost"}` allowlist already covers the in-pod use case. What was the failure mode that motivated widening it? If there's a legit reason (sidecar pattern, host-network mode, etc.), let's solve it more narrowly.

Suggested shapes, ranked by least-invasive first:

**1. Revert to loopback-only**

If the in-pod loopback path works in your environment, revert `_is_internal_client_host` to the original allowlist. Cleanest.

**2. Env-configurable allowlist**

If there *is* a sidecar or DaemonSet caller that needs access, add an env like `CODEX_LB_DRAIN_TRUSTED_CIDRS` (CSV of CIDR ranges, default empty) and only widen access when an operator explicitly opts in:

```python
def _is_internal_client_host(client_host):
    if client_host in {"127.0.0.1", "::1", "localhost"}:
        return True
    if client_host is None:
        return False
    trusted = _get_trusted_drain_cidrs()  # env-driven
    if not trusted:
        return False
    try:
        addr = ip_address(client_host)
    except ValueError:
        return False
    return any(addr in net for net in trusted)
```

**3. Shared secret / bearer**

If we want to keep this purely network-level, an env-driven `CODEX_LB_DRAIN_TOKEN` checked via header (when set) gives operators a cluster-side knob without trusting the network at all.

Option 1 is what I'd push for. The Helm preStop already loops over loopback so I don't see what we lose. Option 2 is the right escape hatch if you have a legit non-loopback caller you want to keep working.

The rest of the PR (status endpoint, in-flight middleware excludes, drain endpoints reachable while draining, helm preStop with status polling, README update) all look correct.

Let me know which shape you want and I'll re-run `@codex review` after the new commit. Don't merge yet.

Soju06:
@codex review

Soju06:
@codex review

Soju06:

PR #570, opened 2026-05-09
fix(proxy): route, normalize and fail-fast input_image file references

Soju06:
## Why

Empirical reproduction on a local dev instance shows that `/v1/responses` requests carrying an attached file via `input_image.file_id` (the natural client shape after `POST /backend-api/files`) hang for the full `CODEX_LB_PROXY_REQUEST_BUDGET_SECONDS` (default 600s) before returning `502 upstream_request_timeout`. Three independent faults on the same surface combine to produce the loop:

1. **`extract_input_file_ids()` is too narrow** -- it only recognizes `{"type":"input_file","file_id":...}`. It misses both `{"type":"input_image","file_id":...}` and `{"type":"input_image","image_url":"sediment://file_*"}`. So first-turn file-pin routing does not engage for image attachments and the request can land on an account that did not register the upload.
2. **The proxy forwards `input_image.file_id` as-is** even though the upstream contract expects image references via `image_url` and accepts `sediment://<file_id>` URIs rather than a `file_id` field on `input_image`. (See `codex-rs/codex-api/src/files.rs::OPENAI_FILE_URI_PREFIX = "sediment://"` and `openai_file_uri(file_id)`.) Upstream consequently terminates the websocket cleanly (`close_code=1000`) before emitting any `response.*` event.
3. **The HTTP responses bridge classifies that clean close as a replayable pre-created race** and schedules `retry_precreated`/`retry_fresh_upstream` repeatedly. The same payload retries forever against a fresh upstream session that will reject it the same way, until the proxy budget is exhausted.

Together those three produce: request routed without the file pin → upstream rejects before `response.created` → bridge replays on a fresh websocket → repeat until budget burns.

## Repro from a local dev instance

```
$ docker compose -f docker-compose.dev.yml up -d --build
$ KEY=$(cat dev-api-key.txt)
$ INIT=$(curl -sS -X POST http://127.0.0.1:2456/backend-api/files \
    -H "Authorization: Bearer $KEY" -H "Content-Type: application/json" \
    -d '{"file_name":"red32.png","file_size":97,"use_case":"codex"}')
$ FID=$(echo "$INIT" | jq -r .file_id)
$ curl -sS -X PUT "$(echo $INIT | jq -r .upload_url)" \
    -H "x-ms-blob-type: BlockBlob" -H "Content-Type: image/png" \
    --data-binary @red32.png
$ curl -sS -X POST "http://127.0.0.1:2456/backend-api/files/$FID/uploaded" \
    -H "Authorization: Bearer $KEY" -H "Content-Type: application/json" -d '{}'
$ curl -sS --max-time 25 -X POST http://127.0.0.1:2456/v1/responses \
    -H "Authorization: Bearer $KEY" -H "Content-Type: application/json" \
    -d "{\"model\":\"gpt-5.4\",\"input\":[{\"role\":\"user\",\"content\":[
        {\"type\":\"input_text\",\"text\":\"What color? one word.\"},
        {\"type\":\"input_image\",\"file_id\":\"$FID\"}]}]}"
```

Before this PR: 5/5 attempts hang for ~10 minutes, end with `502 upstream_request_timeout`, request_log shows `Proxy request budget exhausted`.

After this PR: 5/5 attempts return `502 upstream_rejected_input` (`Upstream rejected the request before response.created (close_code=1000)`) within ~2 seconds. Inline base64 image attachments and text-only requests are unchanged.

### A. Recognize `input_image` file references in the file-pin extractor

`app/core/openai/requests.py`:

- `_is_input_file_with_id` now also matches `input_image` items that carry either a string `file_id` or an `image_url` with the `sediment://` URI prefix.
- `extract_input_file_ids` walks both top-level items and nested role-message `content` parts and now produces the union of `input_file.file_id`, `input_image.file_id`, and `input_image.image_url == "sediment://<file_id>"`.

### B. Normalize `input_image.file_id` to the canonical upstream form before forwarding

`app/modules/proxy/service.py`:

- New helper `_resolve_and_normalize_responses_file_references(payload, headers)` runs before sticky/prompt-cache derivation in the HTTP, websocket, and compact preparation paths. It (1) consults the existing file pin, then (2) rewrites `{type:"input_image",file_id:"file_*"}` parts into `{type:"input_image",image_url:"sediment://file_*"}`. Non-`sediment://` `image_url` values (HTTPS download URLs, `data:image/...` URLs) are left untouched.
- The normalization only touches the file reference itself. No other content is trimmed, slimmed, or reshaped. (See [#569](https://github.com/Soju06/codex-lb/pull/569) for why we are deliberately not doing history rewriting on the proxy.)
- The same `file_preferred_account_id` is now plumbed through the websocket prepare path, the HTTP responses path, and the compact path so the file pin resolves once per request and is reused.

### C. Fail-fast classifier for clean upstream closes with no events

`app/modules/proxy/service.py`:

- New `_classify_upstream_close(close_code, *, response_events_seen) -> Literal["transient","rejected"]`. Returns `"rejected"` for `close_code=1000` with zero `response.*` events received, otherwise `"transient"` so existing replay logic stays untouched.
- `_WebSocketRequestState.response_event_count` is incremented from the upstream event handlers on every `response.*` event we surface (created, partial, completed, failed, etc.).
- `_retry_http_bridge_precreated_request` now consults the classifier. When the close is `rejected`, it sets a 502 `upstream_rejected_input` override on the request state and returns without scheduling another retry. The transient path is unchanged.

### Tests

- `tests/unit/test_openai_requests.py`
  - `extract_input_file_ids` recognizes `input_image.file_id`
  - `extract_input_file_ids` recognizes `input_image.image_url: sediment://<file_id>`
  - non-`sediment://` image URLs are ignored
  - `input_file` items without a `file_id` are ignored
- `tests/unit/test_proxy_utils.py`
  - file-pin resolver returns pinned account for `input_image.file_id`
  - file-pin resolver returns pinned account for `input_image.image_url: sediment://...`
  - HTTP bridge path normalizes `input_image.file_id` before sticky derivation
  - clean close with zero `response.*` events is classified as rejected
  - HTTP bridge pre-created replay is suppressed for that rejected-close case

### OpenSpec docs

Updated the active `add-backend-api-files-protocol/` change in place (the change ships with the same release, so amending its proposal/spec keeps the docs honest):

- proposal.md — accepts `input_image.file_id` + `sediment://` URIs and documents the normalization and the clean-close classifier
- specs/responses-api-compat/spec.md — adds three accepted shapes, normalization scenario, file-pin scenario for image references, and a "rejected upstream close" requirement
- tasks.md — three new completed entries

## Verification

Local on the fix branch:

```
$ uv run --frozen ruff format --check app tests    # 426 files already formatted
$ uv run --frozen ty check app                     # All checks passed!
$ uv run --frozen pytest tests/unit -q             # 1486 passed, 3 skipped
```

Plus the empirical repro above against a rebuilt dev container.

## Limits worth flagging

- Upstream still reject `sediment://<file_id>` references when posted directly to `/v1/responses` from outside a Codex CLI MCP context (the close_code=1000 we observe). The proxy now surfaces that as a clean 502 instead of burning the budget, but actually getting the upstream to accept these references is a question for the codex CLI ↔ upstream contract, not for codex-lb. Tracking that separately is sensible; this PR is about not silently lying to the client about why a request hung.
- The integration test suite was not run end-to-end here. Targeted integration tests for the HTTP responses bridge and `tests/integration/test_proxy_files.py` were exercised and pass.

## Out of scope

- History compaction on the proxy (see #569).
- Adding a separate `/responses/compact` re-entry path for image attachments (not currently broken; this PR only updates the file-pin and normalization helpers it shares).
- Frontend dashboard changes.


Soju06:
Closing in favour of a deeper fix. After studying the upstream codex source (`codex-rs/utils/image/src/lib.rs`) it's clear that codex never puts `input_image.file_id` or `sediment://file_xxx` into a Responses payload -- the upload protocol is for MCP tool argument rewrite only, and conversation images are always inline `data:image/...;base64,...` URLs (resized to 2048 max with Triangle filter). The normalization in this PR was therefore wishful: rewriting `file_id` to `sediment://` keeps producing the upstream close_code=1000 we observed because upstream doesn't accept that shape on the Responses surface either.

A new PR will replace this one and resolve the actual file-pin -> inline data URL pipeline so attached uploads turn into the same `data:image/...;base64,...` shape codex uses (resize, format whitelist, byte passthrough where safe). Re-using the close_code=1000 fail-fast classifier from this PR there.

Issue #568, opened 2026-05-09
follow-up: side effects of response.create history slimming (#560)


Soju06:
## Background

#560 (`fix(proxy): slim oversized response.create history`, merged in `e42af5e`) added a second slimming pass that omits the oldest historical input items when image/tool-output slimming is still over the upstream WebSocket budget. It also injects a single assistant-role notice (`[codex-lb omitted N historical input items]`) and the slimming runs on both the HTTP and WebSocket `response.create` paths.

The PR fixes a real `413` failure mode (#556 is closed by it), and on net the new behavior is strictly better than the previous hard 413. But there are several second-order effects that deserve their own tracking issue. None of these block #556's fix; they are follow-ups to consider before tightening this further.

### 1. prompt-cache invalidation on slimmed turns

The recent `/responses` work (#515 ↔ in-flight pin / cache key affinity in `_resolve_file_account_for_responses`) leans on the upstream prompt cache to keep input-token costs down. When `_omit_historical_response_input_items_to_fit` drops historical items, the resulting `input` prefix is no longer identical to the previous turn, so any explicit or derived `prompt_cache_key` produces a cache miss and the entire conversation re-bills as fresh input tokens.

This is invisible to the user — they only see the slimming notice — but on long heavy-image threads it can multiply input cost on the very turn where slimming fires, and again on every later turn that further drops items.

Possible mitigations to evaluate:

- Recompute / mark `prompt_cache_key` after slimming so we explicitly attach a new cache identity once the prefix changes (so the *next* turn caches against the slimmed history rather than re-missing every turn).
- Surface `historical_items_omitted` (already in the slim summary) on the dashboard request log entry so operators can correlate slimming against unexpected token spikes.

### 2. assistant-role omission notice can mislead the model

The slimmed payload prepends `_response_create_history_omission_notice_item` as an assistant-role item. Subsequent turns will see "I [the assistant] omitted N historical input items" as part of the assistant's own prior speech, which can:

- nudge the model to apologize or hedge ("Sorry I lost earlier context") in user-visible answers,
- look like a checkpoint that history is *known* to the assistant when it isn't, allowing confident hallucination of dropped content,
- conflict with model grounding behavior tuned around what assistant-role messages mean.

A few alternative shapes worth measuring:

- system-role notice instead of assistant-role,
- user-role developer-style preamble notice (matches OpenAI's recommended developer-instruction patterns),
- silent omission (no notice) — less explainable but no role-confusion risk,
[...]

PR #571, opened 2026-05-09
fix(proxy): inline-rewrite input_image file references using codex-faithful image pipeline

Soju06:
## TL;DR

`/v1/responses` payloads carrying `{"type":"input_image","file_id":"file_*"}` (or `{"type":"input_image","image_url":"sediment://file_*"}`) used to hang for the full proxy request budget (~10 min) and then surface `502 upstream_request_timeout`. After this PR they round-trip correctly:

```
$ curl -X POST /v1/responses -d '{
    "model":"gpt-5.4","input":[{"role":"user","content":[
      {"type":"input_text","text":"What color? one word."},
      {"type":"input_image","file_id":"file_xxx"}
    ]}]}'
# → status: completed, output: "red"  (≤4s)
```

## Why the previous attempt (#570) was wrong

#570 tried to "normalize" `input_image.file_id` to `sediment://<file_id>` and forward it. That was wishful: studying upstream codex source (`codex-rs/utils/image/src/lib.rs`, `codex-rs/protocol/src/models.rs`, `codex-rs/core/src/mcp_openai_file.rs`, `codex-rs/core/src/tools/handlers/view_image.rs`) makes the contract clear:

- `ContentItem::InputImage { image_url: String, detail }` is the only shape upstream Responses accepts. `image_url` is always a string -- there is no `file_id` field on `input_image`. codex never puts `file_id` or `sediment://...` into a Responses `input_image` part.
- The `/backend-api/files` upload protocol exists for **MCP tool argument rewrite** (Apps SDK `_meta["openai/fileParams"]` flows). codex uploads a local file once, then plumbs `{file_id, sediment://<file_id>, download_url, mime_type, file_name, file_size_bytes}` into the JSON arguments the tool itself receives. The upstream tool consumes the file. The `sediment://` URI never reaches a Responses `input_image`.
- For prompt images (paste / drop / `--image`), codex reads the bytes locally, runs `load_for_prompt_bytes()`, encodes a `data:{mime};base64,{b64}` URL, and embeds *that* into `ContentItem::InputImage.image_url`. Upstream sees inline base64.

So the only way for codex-lb to support attached uploads on `/v1/responses` is to do what the codex client does locally: take the upload, resolve the bytes, image-process them faithfully, and inline-rewrite the part to `image_url: "data:..."` before forwarding. That's what this PR does.

## Three faults, one chain

1. `extract_input_file_ids` and the `_resolve_file_account_for_responses` lookup only matched `input_file.file_id`. They missed both `input_image.file_id` and `input_image.image_url=sediment://...`, so first-turn pin routing did not engage for image attachments and the request could land on an account that did not own the upload.
2. `_file_account_pins` only stored `(account_id, expires_at)`. Even if the pin lookup found the right account, codex-lb had no `download_url` to pull the bytes back and rewrite to inline.
3. The HTTP responses bridge classified the inevitable upstream rejection (`close_code=1000` with zero `response.*` events) as transient, then looped through `retry_precreated`/`retry_fresh_upstream` until the proxy request budget exhausted (~10 min).

This PR fixes all three in the right layers.

### `app/core/clients/image_processor.py` (new)

Codex-faithful prompt image processor. Mirrors `codex-rs/utils/image/src/lib.rs::load_for_prompt_bytes` exactly:

- whitelists PNG / JPEG / GIF / WebP source formats
- preserves PNG / JPEG / WebP bytes verbatim when the image already fits within `2048×2048`
- re-encodes GIF as PNG
- resizes oversized images to `2048×2048` with Pillow `Image.Resampling.LANCZOS` (closest fit to upstream's Triangle filter under Pillow 10+)
- JPEG quality 85, WebP lossless on resized output (matches codex)
- 32-entry in-process LRU cache keyed by `sha1(bytes) + mode`

Pillow was already a transitive dep through `python-multipart`; this PR adds it as a direct dependency.

### `app/core/clients/files.py` (extended)

- New `fetch_file_bytes(download_url, expected_mime, max_bytes)`. Downloads finalize SAS blobs with a hard byte cap so a single attachment cannot blow the websocket frame budget after base64 expansion.

### `app/core/openai/requests.py` (extended)

- `_input_image_file_reference()` covers both `input_image.file_id` and `input_image.image_url` starting with `sediment://`.
- `extract_input_file_ids()` now sees both `input_file` and uploaded `input_image` references so the file pin still drives routing precedence.
- `extract_input_image_file_references()` returns `(item_index, content_index, file_id)` triples so the rewriter knows exactly which parts to mutate without touching anything else.

### `app/modules/proxy/service.py`

- `_FilePinEntry(account_id, download_url, mime_type, file_name, expires_at)` replaces the old tuple. `create_file()` still pins immediately (so finalize stays on the same account); `finalize_file()` upgrades the entry with `download_url` / `mime_type` / `file_name` once the upstream finalize returns `status="success"`.
- Pin TTL is now `min(_FILE_ACCOUNT_PIN_TTL_SECONDS, SAS se= expiry)`. Expired pins fail closed instead of attempting a stale fetch.
- `_lookup_file_pin()` and `_resolve_file_account()` share the same evict-on-read path.
- `_rewrite_input_image_file_references(payload, headers, request_id)`:
  - finds only `input_image.file_id` / `sediment://file_*` parts
  - looks up the pin (fails the whole request `400 file_not_found` if any reference is missing or expired)
  - fetches bytes via `fetch_file_bytes` (16 MiB cap; over-cap is `400 file_too_large`)
  - runs the codex-faithful image processor (decode failures → `400 invalid_image`)
  - replaces the targeted part with `{type:"input_image", image_url:"data:...", detail}` (preserves an existing `detail`, defaults to `"auto"`)
  - leaves all non-targeted conversation content byte-for-byte untouched (same forwarding-proxy discipline as #569)
  - emits a synthetic `image-inline-rewrite` request-log row so operators can correlate inline rewrites with downstream requests
- Wired into:
  - `/v1/responses` HTTP path
  - `/backend-api/codex/responses` HTTP bridge prepare path
  - websocket `response.create` prepare path
  - `/responses/compact`
- `_classify_upstream_close(close_code, *, response_events_seen)` returns `"rejected"` only for `close_code=1000` with zero `response.*` events seen, otherwise `"transient"`.
- `_WebSocketRequestState.response_event_count` is incremented from the upstream event handlers on every `response.*` event we surface to the downstream stream.
- `_HTTPBridgeSession.last_upstream_close_code` tracks the most recent close code; `_retry_http_bridge_precreated_request` consults the classifier and on `"rejected"` sets `error_code_override="upstream_rejected_input"`, `error_http_status_override=502`, plus a descriptive message including the close code, then returns False (no retry, no replay).

### Tests

- `tests/unit/test_image_processor.py`: passthrough for PNG/JPEG/WebP under 2048, resize for wide/tall PNG, JPEG/WebP resize, ORIGINAL mode, GIF→PNG re-encode, BMP/TIFF/garbage rejection, cache identity on duplicate input.
- `tests/unit/test_files_client.py`: `fetch_file_bytes` happy path and both `Content-Length`-based and streaming `file_too_large` paths.
- `tests/unit/test_openai_requests.py`: `_input_image_file_reference` for `file_id`, `sediment://...`, `data:` URL, https URL; `extract_input_image_file_references` walks multi-message paths.
- `tests/unit/test_proxy_utils.py`: `_lookup_file_pin` live + expired, single `input_image.file_id` rewrite, multiple-reference rewrite (file_id and `sediment://...` mixed), missing pin → 400 `file_not_found`, oversized download → 400 `file_too_large`, non-image content preserved, returned account_id matches the upload owner, `_classify_upstream_close` truth table, `_retry_http_bridge_precreated_request` retry suppression on a rejected clean close.

### OpenSpec

Amended `add-backend-api-files-protocol/` in place (the change ships in the same release as #515): `proposal.md`, `tasks.md`, and `specs/responses-api-compat/spec.md` now document the accepted shapes (`input_file.file_id`, `input_image.file_id`, `input_image.image_url=sediment://...`), the inline-rewrite contract (whitelist, 2048×2048 cap, byte passthrough rules, JPEG q85, WebP lossless), the 16 MiB cap, the missing-pin failure mode, the `do not touch unrelated conversation content` rule, and the clean-close fail-fast classifier with its 502 `upstream_rejected_input` surface.

## Verification

Local on the fix branch:

```
$ uv run --frozen ruff format --check app tests    # 428 files already formatted
$ uv run --frozen ty check app                     # All checks passed!
$ uv run --frozen pytest tests/unit -q             # 1506 passed, 3 skipped
```

Plus an empirical reproduction against a rebuilt dev container with a freshly uploaded PNG:

```
input_image.file_id           attempt 1: completed | red
input_image.file_id           attempt 2: completed | red
input_image.file_id           attempt 3: completed | red
input_image image_url=sediment://...  completed | red
input_image image_url=data:image/png;base64,...  completed | Red   (regression OK)
text-only request                              completed (regression OK)
input_image.file_id=file_does_not_exist        400 file_not_found (immediate, descriptive message)
```

The `openspec validate add-backend-api-files-protocol --strict --no-interactive` step is documented in `SUMMARY.md` as "could not verify" because the `openspec` CLI is not installed in the dev workspace.

## Out of scope

- History compaction on the proxy (see #569 / #568 for why we deliberately do not do that).
- Conversation persistence / archive (separate proposals).
- Frontend dashboard changes.
- Anything other than the explicit `input_image` upload-reference rewrite path.


Soju06:
Pushed two follow-up commits (`1f8cc03`, `e76623e`) on top of the original after empirical multi-image testing showed the original commit still hit upstream WebSocket limits.

## What the new commits add

### `1f8cc03` — auto-route large rewritten payloads over HTTP transport

The first commit fixed the rewrite, so a single attached image works end-to-end. But the upstream WebSocket frame budget (~16 MiB after base64 expansion) was still being hit by multi-image payloads. Codex itself does not have this problem because the official client uses HTTP `POST` for `/backend-api/codex/responses`. codex-lb already has both transports available; the auto-mode selector just didn't consider payload size.

Added `_ws_transport_payload_budget_bytes(settings) = max_sse_event_bytes - 2 MiB headroom` and a new `payload_size_estimate_bytes` parameter on `_resolve_stream_transport`. When `upstream_stream_transport == "auto"` and the rewritten payload exceeds the budget, transport switches to HTTP. Explicit `websocket` / `http` overrides still win.

### `e76623e` — bypass HTTP responses bridge for oversized payloads

Local repro showed the previous commit still wasn't enough on the default code path. `/v1/responses` actually goes through the HTTP responses bridge (`_stream_via_http_bridge`), which uses an internal upstream WebSocket session and ignores `_resolve_stream_transport`. The bridge wraps requests for prompt-cache continuity / sticky-thread routing -- those are great for normal traffic but not negotiable against an upstream frame limit.

Added a payload-size guard at the top of `_stream_responses` (the bridge selector). When the post-rewrite payload exceeds the WebSocket frame budget AND the bridge is enabled, the request is dispatched with a per-request `dataclasses.replace(runtime_config, enabled=False)` -- bridge state stays intact globally, the single oversized request falls through to `_stream_with_retry` -> `_resolve_stream_transport` -> raw HTTP, and subsequent smaller requests continue to use the bridge normally.

Also consolidated `_rewrite_input_image_file_references` to run exactly once per request at the top of `_stream_responses`, then pass the rewritten payload + `file_account_id` through to whichever branch handles the request.

## Empirical verification on the rebuilt dev container

| scenario | before this PR | with all 3 commits |
|---|---|---|
| text-only | ok (~2-4s) | ok (~4s) |
| 1 small inline base64 image | ok (~2s) | ok (~2s) |
| 1 medium PNG (~3 MB) by file_id | hang | ok (~4s) |
| 5 PNG (~3 MB each, 15 MB raw, 20 MB base64) by file_id | 502 stream_incomplete (1009) | ok (~10s) |
| 8 PNG (~3 MB each, 24 MB raw, 32 MB base64) by file_id | 502 stream_incomplete (1009) | ok (~10s) |
| missing file_id | hang -> 502 budget | 400 `file_not_found` (immediate) |

`request_logs.transport=http` confirms the bridge bypass actually fires and the request goes out over the raw HTTP transport. Five `image-inline-rewrite` rows with byte counts match the uploaded file sizes.

## Local checks (still clean after both follow-ups)

```
$ uv run --frozen ruff format --check app tests    # 428 files already formatted
$ uv run --frozen ty check app                     # All checks passed!
$ uv run --frozen pytest tests/unit -q             # 1490+ passed (added test for the new bypass branch)
```

Plus integration suites `test_proxy_responses.py`, `test_proxy_websocket_responses.py`, `test_http_responses_bridge.py` (locally, on the modified branch).

## Out of scope (still)

- Bridge implementation itself stays unchanged. The bypass is per-request only.
- No changes to history slimming or any unrelated conversation content (#569 / #568 still hold).

Soju06:
@codex review

Soju06:
Addressed all three findings in commit `b41870c`:

**P1 (line 222) - inline image cap derived from upstream limit**

`_INLINE_IMAGE_MAX_BYTES` was a fixed 16 MiB raw cap. After base64 expansion (~33%) plus the JSON envelope, a 12 MiB raw image already exceeded the 15 MiB default `upstream_response_create_max_bytes` guard, even though our own code accepted it. Now derived as:

```python
def _inline_image_raw_byte_cap(upstream_max_bytes: int) -> int:
    encoded_budget = max(0, upstream_max_bytes - _INLINE_IMAGE_NON_IMAGE_HEADROOM_BYTES)  # 1 MiB headroom
    raw_cap = (encoded_budget * 3) // 4  # inverse base64 expansion
    return max(256 * 1024, raw_cap)  # 256 KiB floor for thumbnails
```

At default upstream max (15 MiB) the raw cap is now ~10.5 MiB; on the dev container with the 32 MiB env override the cap is 23.2 MiB, both fitting the upstream guard with the 1 MiB headroom.

**P1 (line 2131) - account resolved from rewritten payload**

`_resolve_file_account_for_responses()` was running on the original payload, so an `input_image.file_id` pin (which becomes an inline data URL after rewrite) could win against a surviving `input_file.file_id` pin in mixed-attachment requests, sending upstream the wrong `chatgpt-account-id`. Resolution now runs on the rewritten payload first; if no surviving file references remain (single-image case), it falls back to the rewritten image's pin so single-image requests keep their upload-owner affinity.

Added regression test `test_rewrite_input_image_file_references_resolves_account_from_surviving_input_file` that pins `file_doc -> acc_doc` (input_file) and `file_pic -> acc_pic` (input_image) and asserts the resolver returns `acc_doc` after rewrite.

**P2 - DecompressionBombError caught**

`process_for_prompt_bytes()` now catches `PIL.Image.DecompressionBombError` alongside `UnidentifiedImageError` / `OSError`, surfacing as `ImageProcessingError` -> 400 `invalid_image` instead of leaking as a 500.

Added regression test that pins `Image.MAX_IMAGE_PIXELS` very low and feeds a small valid PNG to trigger the bomb path deterministically.

**Verification (local on b41870c)**

- `uv run --frozen ruff check app tests`
- `uv run --frozen ruff format --check app tests`
- `uv run --frozen ty check app tests`
- `uv run --frozen pytest tests/unit -q` -> 1515 passed, 3 skipped

Empirical re-run on the dev container (`upstream_response_create_max_bytes = 32 MiB` env, computed inline cap 23.2 MiB):
- 5 images x ~3 MB attached by file_id -> `status: completed`, text `5`, ~15 s
- text-only -> `status: completed`, text `OK`

Re-running `@codex review` once CI lands.

@codex review

Soju06:
@codex review

PR #574, opened 2026-05-10
refactor(proxy): reject input_image.file_id with 400, keep transport / classifier hardening

Soju06:
## Why

Follow-up to #571. Reading the upstream contract empirically (direct calls to `https://chatgpt.com/backend-api/codex/responses` with the dev account's `chatgpt-account-id` + access token) shows:

- `{"type":"input_image","file_id":"file_*"}` → upstream returns `Files [file_*] were not found`
- `{"type":"input_image","image_url":"sediment://file_*"}` → upstream returns `Invalid 'input[0].content[1].image_url'. Expected a valid URL, but got a value with an invalid format`
- `{"type":"input_image","image_url":"data:image/png;base64,..."}` → ✅ works (this is what codex CLI actually sends)

So the upstream Responses API only accepts inline `data:` URLs for `input_image`. codex CLI matches that contract: prompt images are read locally, run through `codex-rs/utils/image/src/lib.rs::load_for_prompt_bytes`, and embedded as inline `data:` URLs (`codex-rs/protocol/src/models.rs::local_image_content_items_with_label_number`). The `/backend-api/files` upload protocol (#515) is for **MCP tool argument rewrite** (`codex-rs/core/src/mcp_openai_file.rs::rewrite_mcp_tool_arguments_for_openai_files`), not for `/v1/responses` `input_image`.

The inline rewrite added in #571 was a charitable conversion from a shape upstream rejects into a shape it accepts, but:

- **No known upstream consumer produces the rejected shape.** codex CLI has never sent `input_image.file_id`. We could not find any client in `Soju06/codex-lb` issues that does either.
- **A silent rewrite hides the contract** from any future client that would otherwise learn about the inline-`data:` requirement and adapt. The proxy ends up papering over a client bug while obscuring it.
- **All the maintenance cost** (per-attachment 16 MiB cap, base64 expansion math, file-pin re-resolution after rewrite, `_INLINE_IMAGE_*` constants) buys behaviour that a real client could just do client-side, the same way codex CLI already does.

So this PR removes the rewrite and replaces it with a clean `400 unsupported_input_image_format` carrying a hint pointing the caller at inline `data:` URLs. Everything else from #571 that has independent value stays.

### Removed

- `_rewrite_input_image_file_references` and the `_INLINE_IMAGE_*` byte-cap helpers in `app/modules/proxy/service.py`.
- The rewrite call sites in `_stream_http_bridge_or_retry`, `_stream_via_http_bridge`, `_stream_with_retry`, `compact_responses`.
- The Pillow / image_processor import in `service.py`.
- Tests that exercised the removed rewrite.

### Added

- `_raise_for_unsupported_input_image_references()` early-return guard inside the existing `/v1/responses` flow. Raises `ProxyResponseError(400, openai_error("unsupported_input_image_format", "..."))` with a message that documents the supported shape and the file-upload protocol's intended scope (MCP tool arguments).
- Regression tests in `tests/unit/test_proxy_utils.py`:
  - `_stream_http_bridge_or_retry` rejects `input_image.file_id` with `400 unsupported_input_image_format`
  - `_stream_http_bridge_or_retry` rejects `input_image.image_url=sediment://...` with `400 unsupported_input_image_format`
  - `_stream_http_bridge_or_retry` continues to route `input_file.file_id` via the existing file-pin path (no rejection)
  - The bridge-bypass test was rewritten to exercise the size guard directly without going through the removed rewriter

### Kept (independently valuable, unchanged in spirit from #571)

- `_classify_upstream_close()` + `_WebSocketRequestState.response_event_count` + `_retry_http_bridge_precreated_request()` fail-fast for `close_code=1000` with zero `response.*` events. Defence in depth against any other shape upstream rejects cleanly.
- `_resolve_stream_transport()` HTTP auto-fallback when `auto` + payload exceeds the WebSocket frame budget (still relevant for large multi-image inline `data:` URL requests on raw HTTP).
- `_stream_responses` per-request HTTP responses bridge bypass via `dataclasses.replace(runtime_config, enabled=False)` for oversized payloads (same reasoning).
- `app/core/clients/image_processor.py` — codex-faithful prompt image module. Reusable utility; keeping it does not pull in any service-side complexity.
- `app/core/clients/files.py::fetch_file_bytes()` — general SAS download helper with size cap.
- `_FilePinEntry`'s extended `download_url` / `mime_type` / `file_name` fields. Cheap to keep; lets a future caller reuse the pin without re-implementing the metadata persistence.
- `extract_input_image_file_references` / `_input_image_file_reference` helpers in `app/core/openai/requests.py`. The new reject path uses them to detect the unsupported shape.

## OpenSpec

`openspec/changes/add-backend-api-files-protocol/` was amended in place:

- proposal.md — replaced "rewrite uploaded `input_image` references" with the rejection requirement.
- specs/responses-api-compat/spec.md — accepted shapes shrunk to `input_file.file_id` (verbatim) and inline `data:` URLs; new `Scenario: input_image file_id is rejected before forwarding` and `Scenario: sediment upload URL is rejected before forwarding`. The HTTP-fallback / bridge-bypass / clean-close-classifier scenarios stay (now phrased without "rewritten" since there is no rewrite).
- tasks.md — task 11 reworded to reflect the rejection contract; tasks 13/14 reworded similarly.

## Verification

Local on the refactor branch:

```
$ uv run --frozen ruff format --check app tests    # 428 files already formatted
$ uv run --frozen ty check app tests               # All checks passed!
$ uv run --frozen pytest tests/unit -q             # 1510 passed, 3 skipped
```

Empirical verification on the rebuilt dev container:

| Shape | Result |
|---|---|
| `input_image.file_id` | `400 unsupported_input_image_format` (immediate, helpful message) |
| `input_image.image_url=sediment://...` | `400 unsupported_input_image_format` (immediate) |
| `input_image.image_url=data:image/png;base64,...` (codex CLI shape) | `200 completed`, model answers correctly |
| text-only | `200 completed` |
| `POST /backend-api/files` 3-step upload | success (protocol still wired) |

Diff stats: **+188 / -453, 6 files**. Net code reduction.

## Why this is the right shape, not a step backward from #571

#571 added real value in three layers; this PR keeps all three:

1. Transport: `_resolve_stream_transport` still picks HTTP for oversized payloads on `auto`.
2. Bridge selector: `_stream_responses` still bypasses the bridge per-request when the payload is over the WebSocket frame budget.
3. Bridge close handling: `_classify_upstream_close` still fails fast on the cleanly-rejected case.

What it removes is one layer that was charitable to a shape no client actually sends. Replacing it with an explicit `400` makes the contract visible.

## Out of scope

- The `image_processor.py` module is left in place for future use (it is a faithful port of `codex-rs/utils/image/src/lib.rs::load_for_prompt_bytes` and could power a future client-side / batch helper).
- No changes to the `/backend-api/files` upload protocol surface itself.
- No changes to history slimming or any unrelated conversation content (#569 / #568 still hold).


Soju06:
Pushed an audit follow-up commit (`ea7296f`).

Asked myself: are the bits I left behind in the previous commit actually used in production code, or did I justify them as 'reusable utility' and walk away? Audit:

- `app/core/clients/image_processor.py` — production callers: 0
- `app/core/clients/files.py::fetch_file_bytes()` — production callers: 0
- `_FilePinEntry.download_url` / `.mime_type` / `.file_name` — written by `_pin_file_account` and `finalize_file`, read by nothing on this branch
- `_pin_expiry_from_download_url` — only called by `_pin_file_account`, only relevant when those extended fields were used

So yes, those were dead code. 'Reusable utility' was a lazy excuse. Removed:

- entire `app/core/clients/image_processor.py` and `tests/unit/test_image_processor.py`
- `fetch_file_bytes` and its three tests
- `_FilePinEntry` reverted to `(account_id, expires_at)` shape
- `_pin_expiry_from_download_url` SAS-aware helper
- `_pin_file_account`'s `download_url=` / `mime_type=` / `file_name=` keyword args (back to the original 2-arg signature)
- `finalize_file`'s extraction of the finalize metadata fields
- `pyproject.toml` direct `pillow` dependency (transitive coverage stays through python-multipart)

Kept only what has verified production callers:

- `_classify_upstream_close` + retry suppression
- `_resolve_stream_transport` HTTP auto-fallback
- `_stream_responses` bridge bypass for oversized payloads
- `_raise_for_unsupported_input_image_references` and the `extract_input_image_file_references` / `_input_image_file_reference` helpers
- `create_file` / `finalize_file` / `_proxy_files_call` (the actual #515 protocol)

Diff stats from this commit alone: **+9 / -539, 11 files**. Combined with the previous refactor commit on this branch the PR is now **+197 / -992**.

Local checks: `uv run --frozen ruff check app tests` ✅, `ruff format --check` ✅, `ty check app tests` ✅, `pytest tests/unit -q` ✅ (1493 passed, 3 skipped).

Empirical verification on the rebuilt dev container, all green:
- `input_image.file_id` → 400 `unsupported_input_image_format`
- `input_image.image_url=sediment://...` → 400 `unsupported_input_image_format`
- inline base64 image → 200 'Red'
- text-only → 200 OK
- `POST /backend-api/files` 3-step upload → success
- 5 multi-image inline base64 (~20 MiB serialized payload) → 200 '5' in 17 s (bridge-bypass + HTTP transport fallback intact)

OpenSpec amended to drop the "image_processor / fetch_file_bytes remain reusable utilities" sentence.

Soju06:
@codex review

PR #579, opened 2026-05-10
fix(quota): 5h quota of free acc should be 0

huzky-v:
螢幕截圖 2026-05-10 21 59 46

now it is `33.75`, which will cause the pie chart showing a `dead` proportion of 5h quota


Soju06:
@codex review

Soju06:
@all-contributors please add @huzky-v for code

PR #584, opened 2026-05-11
feat: add omit priority function to API key

### LINKED ISSUE #546: Enforcing default service tier does not work
It would be nice to have the ability to disable fast mode either globally or per api token. Preferably converting the requests to normal mode, but if that is not possible, blocking fast mode requests is fine also.

huzky-v:
Fixes #546 

`Enforce service tier` is just forcing the upstream `service_tier`, so a flag `omit priority request` is added to detect the user priority request and downgrade it accordingly.
螢幕截圖 2026-05-11 23 15 19

This flag is to `downgrade` the request from `priority` to no `service_tier`, forcing the request to run on a non `priority` tier when it is enabled.

螢幕截圖 2026-05-11 23 15 00

The request log list will mark the request `Requested Priority but omitted` for this case.


Soju06:
@codex review

Soju06:
Thanks @huzky-v -- the design is clean (orthogonal to `enforced_service_tier`, request-log surfacing of the omitted tier, OpenSpec coverage extending the same change folder, decent unit/integration test fan-out). One P2 from codex review to address before merging.

## P2 -- `None` vs missing in requested-service-tier metadata

`get_policy_requested_service_tier` currently can't distinguish "client explicitly didn't send `service_tier`" (`_REQUESTED_SERVICE_TIER_ATTR == None`) from "metadata never set" (`_REQUESTED_SERVICE_TIER_ATTR` absent):

```python
# app/modules/proxy/request_policy.py:115
def get_policy_requested_service_tier(payload) -> str | None:
    value = getattr(payload, _REQUESTED_SERVICE_TIER_ATTR, None)
    return value if isinstance(value, str) else getattr(payload, "service_tier", None)
```

When `value` is the literal `None` set by `apply_api_key_enforcement`, the `isinstance(value, str)` branch fails and it falls through to `payload.service_tier` -- but by that point `payload.service_tier` has already been rewritten by `enforced_service_tier` (and could also be `None` after `omit_priority_request` downgrade). The end result is that the request log reports the **enforced** tier as if the client had requested it.

### Reproducer

- API key with `enforced_service_tier = "priority"`
- Client request omits `service_tier` entirely
- Expected: `requested_service_tier = None` in logs/metadata
- Actual: `requested_service_tier = "priority"`

### Fix

Use `_MISSING` (which you already define) as the sentinel so explicit `None` is preserved:

```python
def get_policy_requested_service_tier(payload) -> str | None:
    value = getattr(payload, _REQUESTED_SERVICE_TIER_ATTR, _MISSING)
    if value is _MISSING:
        return getattr(payload, "service_tier", None)
    if value is None or isinstance(value, str):
        return value
    # defensive fallback if some other type sneaks in
    return getattr(payload, "service_tier", None)
```

And add a focused unit test:

```python
def test_requested_service_tier_preserved_as_null_when_enforced():
    payload = _make_responses_request(service_tier=None)
    api_key = _make_api_key(enforced_service_tier="priority")
    apply_api_key_enforcement(payload, api_key)
    assert payload.service_tier == "priority"
    assert get_policy_requested_service_tier(payload) is None
    assert get_policy_service_tier_omitted(payload) is False
```

Plus the symmetric one for `omit_priority_request` downgrading:

```python
def test_requested_service_tier_recorded_when_priority_omitted():
    payload = _make_responses_request(service_tier="priority")
    api_key = _make_api_key(omit_priority_request=True)
    apply_api_key_enforcement(payload, api_key)
    assert payload.service_tier is None
    assert get_policy_requested_service_tier(payload) == "priority"
    assert get_policy_service_tier_omitted(payload) is True
```

## Not blocking otherwise

The rest reads cleanly:

- The omit happens after `enforced_service_tier` is applied, so combining the two (enforce a tier and then strip priority) collapses correctly into a single resolved tier.
- The Alembic migration follows the existing idempotent pattern (inspect columns before `add_column`).
- Request-log mapping + frontend recent-requests-table changes surface the omitted state to the user.
- OpenSpec change folder is extended in place rather than fragmenting into a new change.

After the P2 fix lands I'll re-run `@codex review` and queue for merge.

huzky-v:
@codex review

huzky-v:
@codex review

huzky-v:
I think it may be redundant to #611, which should fix it in easier way (force `downgrade`)
I will work that instead.

PR #586, opened 2026-05-12
fix(proxy): suppress duplicate side-effect tool calls

Komzpa:
Draft split from #555.

## Purpose
- Suppress replayed side-effect tool-call events in the HTTP stream path, including replays that change `call_id`.
- Keep independent downstream responses scoped by `response_id` so separate turns are not suppressed as duplicates.
- Rewrite duplicate nested side-effect operations inside `multi_tool_use.parallel` before forwarding the batch to the client.
- Limit input-history side-effect dedupe to previous-response replay/continuity payloads, not first-attempt list inputs.

## Validation
- Rebased on current `main`.
- `ruff format --check`, `ruff check`, and `ty check` on touched backend/test files passed locally.
- Focused pytest for downstream tool-call dedupe, HTTP stream replay, same-response replay, and previous-response input replay passed locally.
- Exact-head GitHub CI is green for `172f032bb306a712119a29ee157a0ce5d6fd6060`.
- Fresh `@codex review` completed for `172f032bb306a712119a29ee157a0ce5d6fd6060`; no inline review comments were reported.

## Relationship
- Replaces the side-effect dedupe portion of #555.
- #555 remains Draft and should not be merged as one large PR.

### DIFF
--- a/app/core/clients/proxy.py
+++ b/app/core/clients/proxy.py
@@ -935,18 +935,21 @@ def _normalize_stream_event_payload(payload: dict[str, JsonValue]) -> dict[str, 
         for key in ("plan_type", "resets_at", "resets_in_seconds"):
             if key in detail:
                 event["response"]["error"][key] = detail[key]
-        return event
+        return cast(dict[str, JsonValue], event)
     if event_type == "error":
         message = _extract_upstream_message(payload) or "Upstream websocket error"
         code = payload.get("code")
         error_type = payload.get("error_type") or payload.get("type")
-        return response_failed_event(
-            _normalize_error_code(
-                code if isinstance(code, str) else None, error_type if isinstance(error_type, str) else None
+        return cast(
+            dict[str, JsonValue],
+            response_failed_event(
+                _normalize_error_code(
+                    code if isinstance(code, str) else None, error_type if isinstance(error_type, str) else None
+                ),
+                message,
+                error_type=error_type if isinstance(error_type, str) and error_type != "error" else "server_error",
+                response_id=get_request_id(),
             ),
-            message,
-            error_type=error_type if isinstance(error_type, str) and error_type != "error" else "server_error",
-            response_id=get_request_id(),
         )
     return payload
 
--- a/app/core/clients/proxy.py
+++ b/app/core/clients/proxy.py
@@ -939,15 +939,17 @@ def _normalize_stream_event_payload(payload: dict[str, JsonValue]) -> dict[str, 
     if event_type == "error":
         message = _extract_upstream_message(payload) or "Upstream websocket error"
         code = payload.get("code")
-        error_type = payload.get("error_type") or payload.get("type")
+        raw_error_type = payload.get("error_type")
+        error_type = raw_error_type if isinstance(raw_error_type, str) else None
         return cast(
             dict[str, JsonValue],
             response_failed_event(
                 _normalize_error_code(
-                    code if isinstance(code, str) else None, error_type if isinstance(error_type, str) else None
+                    code if isinstance(code, str) else None,
+                    error_type,
                 ),
                 message,
-                error_type=error_type if isinstance(error_type, str) and error_type != "error" else "server_error",
+                error_type=error_type or "server_error",
                 response_id=get_request_id(),
             ),
         )
--- a/app/core/clients/proxy.py
+++ b/app/core/clients/proxy.py
@@ -84,6 +84,13 @@ _SSE_EVENT_TYPE_ALIASES = {
     "response.audio.delta": "response.output_audio.delta",
     "response.audio_transcript.delta": "response.output_audio_transcript.delta",
 }
+_RESPONSE_STREAM_TERMINAL_EVENT_TYPES = frozenset(
+    {
+        "response.completed",
+        "response.failed",
+        "response.incomplete",
+    }
+)
 
 _SSE_READ_CHUNK_SIZE = 1 * 1024
 _IMAGE_INLINE_MAX_BYTES = 8 * 1024 * 1024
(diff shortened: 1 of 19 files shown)


Soju06:
Thanks @Komzpa -- the side-effect tool-call dedupe is exactly the right shape (per-response id cache, side-effect allowlist, `multi_tool_use.parallel` nested rewriting, distinct read-only / later-response-id calls intact), and the OpenSpec change folder is in place. The 566 lines of unit coverage are also a great safety net. Codex review picked up two issues that should land before merging.

## P1 -- `wait_agent` dedupe must not raise on malformed targets

`app/modules/proxy/service.py:~8653` calls `sorted(targets)` on tool-call payload that is model-generated JSON. If the model emits `functions.wait_agent` with mixed-type targets (or any otherwise non-comparable list), `sorted` raises `TypeError` and aborts the whole stream-processing path. That converts what's supposed to be a best-effort dedupe into a hard stream failure on input the proxy has no control over.

Suggested fix: feature-detect comparability, or just sort safely with a key that's always comparable, then fall back to the original list on failure:

```python
def _normalize_wait_agent_targets(targets):
    if not isinstance(targets, list):
        return targets
    try:
        return sorted(targets, key=lambda t: (t.__class__.__name__, str(t)))
    except (TypeError, ValueError):
        return list(targets)  # fall through; dedupe is best-effort
```

And a regression to pin behavior:

```python
def test_wait_agent_dedupe_tolerates_mixed_type_targets():
    parallel = {
        "call_id": "call_x",
        "name": "multi_tool_use.parallel",
        "arguments": json.dumps({
            "tool_uses": [
                {"recipient_name": "functions.wait_agent", "parameters": {"targets": [1, "a", {}]}}
            ]
        }),
    }
    # should not raise, should forward the event unchanged (or at worst with a no-op dedupe)
    result = _maybe_dedupe_parallel_tool_use(parallel, dedupe_cache={})
    assert result is not None
```

## P2 -- normalize top-level `exec_command` / `write_stdin` keys before dedupe

The top-level side-effect path keys on the raw `arguments` string:

```python
# app/modules/proxy/service.py:~8619
dedupe_key = (call_name, arguments_str)
```

while the nested `multi_tool_use.parallel` path already strips volatile fields (`yield_time_ms`, `max_output_tokens`, etc.) before keying. That asymmetry defeats the dedupe in the exact replay scenario this PR is targeting:

- Same response_id, same `exec_command` body, but a replayed event with `yield_time_ms: 5000` vs `yield_time_ms: 6000` -> two distinct keys -> both events forwarded -> command executed twice.

Suggested fix: share the volatile-field-stripping helper between the top-level and nested paths, and key on the normalized arguments dict (or its canonical-JSON dump):

```python
_SIDE_EFFECT_VOLATILE_ARG_KEYS = frozenset({"yield_time_ms", "max_output_tokens", "timeout_ms"})

def _side_effect_dedupe_key(call_name: str, arguments: dict) -> str:
    operative = {k: v for k, v in arguments.items() if k not in _SIDE_EFFECT_VOLATILE_ARG_KEYS}
    return f"{call_name}::{json.dumps(operative, sort_keys=True, separators=(',', ':'))}"
```

Then both top-level (`exec_command`/`write_stdin`) and the parallel inner-dedupe use the same function.

Test to pin:

```python
def test_top_level_exec_command_dedupes_across_yield_time_ms_differences():
    # same response_id, two events with identical exec_command `command` but different `yield_time_ms`
    state = _make_dedupe_state(response_id="resp_a")
    first = _exec_event(call_id="call_a", command="echo hi", yield_time_ms=5000)
    second = _exec_event(call_id="call_b", command="echo hi", yield_time_ms=6000)
    assert _consume_side_effect_event(first, state) is not None      # forwarded
    assert _consume_side_effect_event(second, state) is None         # suppressed
```

## Not blocking otherwise

- Scope (split from #555) is clean.
- Cache limit (`_WEBSOCKET_TOOL_CALL_DEDUPE_CACHE_LIMIT = 1024`) is reasonable for per-response keying and won't grow unbounded.
- Allowlist for side-effect tool names is conservative -- read-only calls and later-response ids are still passed through.

After the P1 fix and (ideally) the P2 normalization land, I'll re-run `@codex review` and queue for merge.

Soju06:
@codex review

Soju06:
@codex review

Komzpa:
@codex review

Soju06:
@Komzpa second-round `@codex review` on `d971742` confirms the reconnect-survival fix is correct (`session.seen_tool_call_keys` is now session-scoped, the regression `test_http_bridge_tool_call_dedupe_survives_upstream_reconnect` reproduces and pins the reconnect path). Good.

One new P1 from this round to address before merge — found at `app/modules/proxy/tool_call_dedupe.py:278` inside `dedupe_replayed_side_effect_input_items`:

> `call_keys.get(index)` returning `None` immediately `continue`s, which leaves the previous side-effect key active; a later `write_stdin`/`exec_command` with the same canonical args is then treated as a replay and removed even if a read-only tool call happened in between.

Concretely, today the loop only resets `first_call_id_by_key` / `last_side_effect_key` on `replayed_input_segment_boundary(item)`. A non-boundary, non-side-effect item (a `read_file` / `search` / a tool output between two `write_stdin` calls) returns `key=None` and we `continue` without touching the state, so the next side-effect call with the same canonical key matches `last_side_effect_key` and gets dropped.

Failure mode in production: in a mixed turn `side-effect → read-only → same-args side-effect`, the second legitimate operation gets removed and the tool-call/output pair desyncs for the continuation request. This is exactly the class of bug the dedupe is supposed to *prevent*, except inverted (dropping real work instead of duplicate work).

Suggested fix — reset the side-effect chain state whenever the current item is **not** a side-effect tool call:

```python
for index, item in enumerate(input_items):
    if isinstance(item, dict) and replayed_input_segment_boundary(item):
        first_call_id_by_key.clear()
        last_side_effect_key = None
    key = call_keys.get(index)
    if key is None:
        last_side_effect_key = None   # <-- non-side-effect breaks the replay chain
        continue
    # ... existing side-effect logic
```

(`first_call_id_by_key` is keyed on canonical side-effect identity and survives correctly across a single read-only item; only `last_side_effect_key` needs to clear, because it encodes "the immediately previous side-effect operation".)

Regression to pin:

```python
def test_dedupe_replayed_side_effect_keeps_repeat_after_read_only_intermission():
    """side-effect A -> read-only R -> side-effect A (same args) must keep both A's."""
    input_items = [
        _function_call("call_a1", name="write_stdin", arguments=json.dumps({"chars": "hi"})),
        _function_call_output("call_a1", output="ok"),
        _function_call("call_r",  name="read_file", arguments=json.dumps({"path": "/tmp/x"})),
        _function_call_output("call_r", output="..."),
        _function_call("call_a2", name="write_stdin", arguments=json.dumps({"chars": "hi"})),
        _function_call_output("call_a2", output="ok"),
    ]
    result, removed = dedupe_replayed_side_effect_input_items(input_items)
    assert removed == 0
    assert [item.get("call_id") for item in result if isinstance(item, dict)] == [
        "call_a1", "call_a1", "call_r", "call_r", "call_a2", "call_a2",
    ]
```

(Adjust the `_function_call*` helpers to match the existing test scaffolding; the assertion shape is what matters.)

Pair it with the existing `test_dedupe_replayed_side_effect_collapses_repeat_with_same_args` to make sure the "drop the byte-identical adjacent replay" path still fires when there's no read-only item in between.

After this lands I'll re-run `@codex review` and queue for merge.

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

PR #587, opened 2026-05-12
fix(chat): route parallel tool-call arguments by item_id

### LINKED ISSUE #542: Parallel tool_calls still receive duplicated arguments (PR #287 fix incomplete — item_id not resolved in _extract_tool_call_fields)

ozpool:
## Summary
- Fixes #542. When upstream emits two or more parallel `function_call` items, the Responses API streams `response.function_call_arguments.delta` / `.done` events that identify their owning call only via `item_id` (e.g. `fc_...`). The previous `ToolCallIndex.index_for` keyed on `call_id` / `name` only, so all argument events for the 2nd+ parallel calls fell back to index `0` and overwrote the first call's arguments. Downstream agents silently executed the wrong actions with corrupted parameters.
- Extend `ToolCallIndex.index_for` to accept an `item_id` and alias it to the same slot as the `call_id` / `name` key the first time both are seen on an `output_item.added` / `.done` event. Argument-only events then resolve to the correct slot via the `item_id` alias.
- Single-tool-call paths, Chat-Completions-style events (which carry `call_id` directly), and raw `/v1/responses` forwarding are unchanged.
- OpenSpec change `fix-parallel-tool-call-arguments-routing` documents the new requirement and mirrors the structure of the prior accepted PR #287 change.

## Validation
- `uv run pytest tests/unit/test_chat_response_mapping.py tests/integration/test_proxy_chat_completions.py tests/integration/test_openai_compat_features.py tests/integration/test_proxy_responses.py` — 102 passed
- `uv run ruff check app/core/openai/chat_responses.py tests/unit/test_chat_response_mapping.py` — clean
- `uv run ty check app/core/openai/chat_responses.py` — clean
- Added regression tests `test_collect_completion_routes_parallel_tool_calls_by_item_id` and `test_stream_chat_chunks_routes_parallel_tool_calls_by_item_id` exercising realistic `output_item.added` + `function_call_arguments.delta/.done` + `output_item.done` sequences for two parallel calls

Fixes #542

### DIFF
--- a/app/core/openai/chat_responses.py
+++ b/app/core/openai/chat_responses.py
@@ -123,14 +123,37 @@ class ToolCallIndex:
     indexes: dict[str, int] = field(default_factory=dict)
     next_index: int = 0
 
-    def index_for(self, call_id: str | None, name: str | None) -> int:
-        key = _tool_call_key(call_id, name)
-        if key is None:
-            return 0
-        if key not in self.indexes:
-            self.indexes[key] = self.next_index
+    def index_for(
+        self,
+        call_id: str | None,
+        name: str | None,
+        item_id: str | None = None,
+    ) -> int:
+        call_key = _tool_call_key(call_id, name)
+        item_key = _tool_call_item_key(item_id)
+
+        # If either key already maps to an index, reuse it. The Responses API
+        # streams output_item.added (which carries call_id) and
+        # function_call_arguments.delta/done (which only carry item_id) for the
+        # same logical tool call; binding both keys to one slot prevents
+        # parallel tool calls from collapsing to index 0 (see issue #542).
+        existing: int | None = None
+        if call_key is not None:
+            existing = self.indexes.get(call_key)
+        if existing is None and item_key is not None:
+            existing = self.indexes.get(item_key)
+
+        if existing is None:
+            if call_key is None and item_key is None:
+                return 0
+            existing = self.next_index
             self.next_index += 1
-        return self.indexes[key]
+
+        if call_key is not None:
+            self.indexes.setdefault(call_key, existing)
+        if item_key is not None:
+            self.indexes.setdefault(item_key, existing)
+        return existing
 
 
 @dataclass
@@ -596,7 +619,8 @@ def _tool_call_delta_from_payload(payload: Mapping[str, JsonValue], indexer: Too
     if fields is None:
         return None
     call_id, name, arguments, tool_type = fields
-    index = indexer.index_for(call_id, name)
+    item_id = _extract_item_id(payload)
+    index = indexer.index_for(call_id, name, item_id=item_id)
     return ToolCallDelta(
         index=index,
         call_id=call_id,
@@ -701,6 +725,28 @@ def _tool_call_key(call_id: str | None, name: str | None) -> str | None:
         return f"id:{call_id}"
     if name:
         return f"name:{name}"
+    return None
+
+
+def _tool_call_item_key(item_id: str | None) -> str | None:
+    if item_id:
+        return f"item:{item_id}"
+    return None
+
+
+def _extract_item_id(payload: Mapping[str, JsonValue]) -> str | None:
+    # response.function_call_arguments.delta/.done carry the owning item
+    # identifier at the top level as `item_id`.
+    raw = payload.get("item_id")
+    if isinstance(raw, str) and raw:
+        return raw
+    # response.output_item.added/.done nest the item; its `id` matches the
+    # `item_id` emitted by the argument events for the same call.
+    item = _as_mapping(payload.get("item"))
+    if item is not None:
+        raw = item.get("id")
+        if isinstance(raw, str) and raw:
+            return raw
     return None
 
 
--- a/openspec/changes/fix-parallel-tool-call-arguments-routing/proposal.md
+++ b/openspec/changes/fix-parallel-tool-call-arguments-routing/proposal.md
@@ -1,0 +1,12 @@
+## Why
+Issue #542 reports that `/v1/chat/completions` corrupts parallel tool-call arguments. The Responses API emits `response.function_call_arguments.delta` / `.done` events that identify their owning call only via `item_id` (e.g. `"fc_..."`). The current `ToolCallIndex.index_for` keys on `call_id`/`name` only, so argument events for the second and subsequent parallel calls fall back to index `0` and overwrite the first call's payload. Downstream agents then silently execute the wrong actions.
+
+## What Changes
+- Route Responses tool-call events through a `ToolCallIndex` that also recognises `item_id` as an indexing key.
+- Establish an alias between the `output_item` `call_id` and the corresponding `item_id` the first time both are seen, so subsequent argument-only events resolve to the same slot.
+- Preserve the existing behaviour for Chat-Completions-style events that carry `call_id` directly with no `item_id`.
+- Add regression coverage for parallel tool-call routing in both the streaming and non-streaming `/v1/chat/completions` adapters.
+
+## Impact
+- Restores correct `tool_calls[].function.arguments` per call when upstream emits multiple parallel function calls.
+- No effect on single-tool-call paths, raw `/v1/responses` forwarding, or non-tool text handling.
--- a/openspec/changes/fix-parallel-tool-call-arguments-routing/specs/responses-api-compat/spec.md
+++ b/openspec/changes/fix-parallel-tool-call-arguments-routing/specs/responses-api-compat/spec.md
@@ -1,0 +1,17 @@
+## MODIFIED Requirements
+### Requirement: Tool call events and output items are preserved
+If the upstream model emits tool call deltas or output items, the service MUST forward those events in streaming mode and MUST include tool call items in the final response output for non-streaming mode.
+
+#### Scenario: Tool call emitted
+- **WHEN** the upstream emits a tool call delta event
+- **THEN** the service forwards the delta event and includes the finalized tool call in the completed response output
+
+#### Scenario: Chat Completions tool arguments avoid snapshot duplication
+- **WHEN** `/v1/chat/completions` maps Responses tool-call events that include incremental deltas and later finalized snapshots for the same tool call
+- **THEN** the final `tool_calls[].function.arguments` value is exactly one valid JSON string for that tool call
+- **AND** the adapter MUST NOT append full snapshot payloads on top of already-collected incremental argument deltas
+
+#### Scenario: Parallel tool calls route arguments by item_id
+- **WHEN** `/v1/chat/completions` maps Responses events for two or more parallel function calls whose `response.function_call_arguments.delta` / `.done` events identify the owning call only via `item_id`
+- **THEN** each call's arguments are routed to its own `tool_calls[]` slot using `item_id` as the indexing key
+- **AND** the adapter MUST alias the `item_id` to the same slot as the `call_id` once both identifiers have been observed, so subsequent argument-only events for that call do not collapse to index `0`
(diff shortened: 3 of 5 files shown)


Soju06:
Thanks @ozpool for picking up #542 -- closing in favor of #543, which lands a more layered fix on the same routing problem.

## Why #543 is going in

Both PRs identify the same root cause (Responses API `response.function_call_arguments.delta` / `.done` events carry only `item_id`, so the previous indexer collapsed parallel calls to index 0). The mechanisms differ:

- **#543 uses `output_index` as the primary routing key**, with `call_id`/`name` keying and an `item_id` alias fallback. `output_index` is on every Responses event and matches the routing key in OpenAI's own reference client and LiteLLM (BerriAI/litellm#17652). This means routing is correct even when an event arrives without `item_id`, when the upstream reorders events, or when only one of the two identifiers is present.
- **#587 uses `item_id` aliasing alone.** Correct for the issue as reported, but less robust if a future stream variant arrives without a preceding `output_item.added` to establish the alias.

Two extra reasons #543 fits this repo:

- It includes a secondary fix preventing the internal `fc_...` item id from leaking into the public `tool_calls[].id` / `call_id` surface.
- It keeps the existing `ToolCallIndex.index_for(call_id, name)` signature intact for the Chat-Completions-only path, so legacy callers don't change.

## What I borrowed from your PR

I really liked that you bundled an OpenSpec change folder with the fix -- #543 originally didn't have one. I just pushed `openspec/changes/fix-parallel-tool-call-arguments-routing/` to #543's branch, mirroring your structure but rewording the proposal and spec scenarios to describe the three-layer routing #543 implements. The change validates cleanly against `@fission-ai/openspec validate --specs`.

So the OpenSpec contribution from your PR effectively lands as well, just on #543's branch.

## Credit

Once #543 merges I'll make sure you get an `all-contributors` entry for code (and your OpenSpec layout). Thanks again for the clean PR and the scoped tests -- both made it easier to compare and decide.

If you want to follow up with anything else on the chat_responses routing path or elsewhere, happy to review.

ozpool:
Okay, thanks @Soju06

PR #594, opened 2026-05-12
fix(api): add back the account selection on api creation

huzky-v:
Currently only edit API could configure account pool
This PR is going to add back the ability to configure the account pool while creating the API key

螢幕截圖 2026-05-12 23 34 40


huzky-v:
@codex review

PR #598, opened 2026-05-13
fix(dashboard-auth): reject passwords longer than bcrypt's 72-byte limit

### LINKED ISSUE #597: Error 500 when setting password
Logs show "password cannot be longer than 72 bytes", but frontend just errors "Unexpected error"

Soju06:
Fixes #597.

## Root cause

`bcrypt.hashpw` enforces a hard 72-byte input limit on the encoded password (bcrypt 5.x raises `ValueError("password cannot be longer than 72 bytes ...")`). `_hash_password` in `app/modules/dashboard_auth/service.py` calls `bcrypt.hashpw` directly, and the two API entry points that lead there only validate `len(password) < 8` before hashing.

So a password whose UTF-8 encoded length exceeds 72 bytes propagated as an unhandled `ValueError` and surfaced as a 500 with a generic "Unexpected error" on the dashboard frontend (the symptom reported in #597).

## Fix

- Add a shared `_validate_password_length` helper in `app/modules/dashboard_auth/api.py` that enforces both the existing 8-character minimum and a new 72-UTF-8-byte maximum.
- Route both `POST /api/dashboard-auth/password/setup` and `POST /api/dashboard-auth/password/change` through it, so the two endpoints validate identically before hashing.
- On the new failure case, return `DashboardValidationError` (HTTP 422) with code `password_too_long` and a message that references the 72-byte limit, so the dashboard can render a clear, actionable error.

The limit is measured against the **UTF-8 encoded byte length**, not codepoint count, so multi-byte characters (emoji, non-ASCII letters) count for as many bytes as bcrypt itself counts. This matches bcrypt's own ceiling exactly.

## Why 72 bytes specifically

`bcrypt` (the underlying password hashing primitive used by all major language bindings) silently truncates input after 72 bytes when hashing on most platforms, but the Python `bcrypt` 5.x package explicitly rejects oversize inputs with `ValueError` instead of silently truncating. Anything that exceeds bcrypt's 72-byte ceiling cannot be hashed at all, so the right place to surface that is at the API layer, not as an unhandled `ValueError`.

## Tests

Added four integration tests in `tests/integration/test_dashboard_password_auth.py`:

```
tests/integration/test_dashboard_password_auth.py::test_password_setup_rejects_overlong_password PASSED
tests/integration/test_dashboard_password_auth.py::test_password_setup_accepts_72_byte_password PASSED
tests/integration/test_dashboard_password_auth.py::test_password_setup_counts_utf8_bytes_not_codepoints PASSED
tests/integration/test_dashboard_password_auth.py::test_password_change_rejects_overlong_new_password PASSED
```

Verification on this branch:

```
uv run ruff check app/modules/dashboard_auth/api.py tests/integration/test_dashboard_password_auth.py     # clean
uv run ruff format --check app/modules/dashboard_auth/api.py tests/integration/test_dashboard_password_auth.py  # clean
uv run ty check app/modules/dashboard_auth/api.py tests/integration/test_dashboard_password_auth.py        # clean
uv run pytest tests/integration/test_dashboard_password_auth.py tests/unit/test_dashboard_auth_api.py tests/unit/test_dashboard_auth_password_service.py -q   # 16 passed
npx @fission-ai/openspec validate validate-password-length                                                 # change is valid
```

## OpenSpec

Added an OpenSpec change folder `openspec/changes/validate-password-length/` capturing the new requirement under the `admin-auth` capability (length is bounded by bcrypt's input limit, measured in UTF-8 bytes, applied to both `setup` and `change`).

## Migration / Compatibility

- No schema or runtime migration.
- No change to bcrypt behavior or to how passwords are stored.
- Existing valid passwords are unaffected; the new check matches bcrypt's existing ceiling, so any password that could already be set is still accepted.

## Hotfix candidate for v1.17.1

This is a small, scoped fix on the dashboard-auth API surface with no runtime impact. It is a good candidate for the next patch release.

### DIFF
--- a/app/modules/dashboard_auth/api.py
+++ b/app/modules/dashboard_auth/api.py
@@ -153,6 +153,23 @@ def _ensure_password_management_enabled(request: Request) -> None:
         )
 
 
+# bcrypt enforces a hard 72-byte limit on the input password; anything longer
+# raises ``ValueError`` from ``bcrypt.hashpw`` and surfaces as a 500 to the
+# client. Validate the encoded length here so the API returns a clear 400.
+_MAX_PASSWORD_BYTES = 72
+
+
+def _validate_password_length(password: str) -> None:
+    if len(password) < 8:
+        raise DashboardValidationError("Password must be at least 8 characters")
+    if len(password.encode("utf-8")) > _MAX_PASSWORD_BYTES:
+        raise DashboardValidationError(
+            f"Password must be at most {_MAX_PASSWORD_BYTES} bytes when encoded as UTF-8. "
+            "Note that multi-byte characters (e.g. emoji, non-ASCII letters) count for more than one byte.",
+            code="password_too_long",
+        )
+
+
 @router.get("/session", response_model=DashboardAuthSessionResponse)
 async def get_dashboard_auth_session(
     request: Request,
@@ -212,8 +229,7 @@ async def setup_password(
         if validation_status != "valid":
             raise DashboardAuthError("Invalid dashboard bootstrap token.", code="invalid_bootstrap_token")
     password = payload.password.strip()
-    if len(password) < 8:
-        raise DashboardValidationError("Password must be at least 8 characters")
+    _validate_password_length(password)
     try:
         await context.service.setup_password(password)
     except PasswordAlreadyConfiguredError as exc:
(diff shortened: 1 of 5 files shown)


Soju06:
@codex review

PR #600, opened 2026-05-13
fix(auth): deactivate account on refresh-time token_expired

### LINKED ISSUE #383: Access token expiration (token_expired) does not deactivate account, so expired accounts remain routable
## Summary
When an account starts returning `token_expired` / HTTP 401 repeatedly, `codex-lb` keeps the account in `active` state instead of removing it from routing.
As a result, the load balancer continues sending API traffic to an account that is clearly unusable, causing repeated request failures until the account is manually paused.
## Expected behavior
If an account consistently fails with authentication errors that indicate an expired/unusable session, it should stop being routable automatically.
At minimum, after repeated `token_expired` failures, the account should be moved out of the active pool (`deactivated` or `paused`) so it is no longer selected for API requests.
## Actual behavior
The account remains `active` even after many repeated failures.
Observed in production:
- account email: `email@telegmail.com`
- account id: `email_3e2ba4c18bda`
The account only stopped receiving traffic after being manually paused.
### Runtime logs
The container logs show repeated auth failures for the same account:
- `Token refresh failed ... status=401`
- `Usage fetch failed ... Provided authentication token is expired. Please try signing in again.`
- `Model fetch 401 retry failed account=email_3e2ba4c18bda plan=team`
### Request logs
Recent `request_logs` rows for the account show repeated:
- `status = error`
- `error_code = token_expired`
- `error_message = Provided authentication token is expired. Please try signing in again.`
The same request IDs appear multiple times, which suggests retries are happening, but the account still remains routable.
## Why this happens
`token_expired` is not treated as a permanent failure.
Current permanent failure codes are defined as:
- `refresh_token_expired`
- `refresh_token_reused`
- `refresh_token_invalidated`
- `account_suspended`
- `account_deleted`
Relevant code:
- `app/core/balancer/logic.py`
- `app/modules/proxy/service.py`
Deactivation only happens when the error code is in `PERMANENT_FAILURE_CODES`.
So if upstream returns `token_expired`, the code records an error / retries, but does **not** deactivate the account.
## Impact
This allows clearly broken accounts to remain in the active routing pool and continue failing user requests.
In practice, an expired account can generate many failed requests before an operator notices and pauses it manually.
## Suggested fix
One of the following should happen:
1. Treat repeated `token_expired` as a permanent authentication failure and deactivate the account.
2. Convert `token_expired` into a stronger terminal auth state after a small threshold.
3. Validate/refresh tokens during manual reactivation, and immediately re-deactivate if refresh/auth still fails permanently.
4. Ensure background usage/model fetch auth failures can also transition the account out of `active` when they repeatedly prove the session is unusable.
## Reproduction
1. Have an account whose access token/session is no longer valid.
2. Keep the account in `active`.
3. Send repeated API requests through `codex-lb`.
4. Observe repeated `token_expired` request log entries and 401-related runtime logs.
5. Observe that the account still remains `active` instead of being removed from routing.
## Notes
This is especially confusing because the dashboard/token status can show the token as expired, while routing still treats the account as usable.

## P.S. 
I am not sure why this has not been considered important so far, but my real-world case shows that this token can expire and may not recover on its own until the account is re-authenticated.

Soju06:
Fixes #383.

## Root cause

When the OAuth refresh endpoint returns error code `token_expired` (HTTP 401 body `{"error": {"code": "token_expired"}}` with message `"Provided authentication token is expired. Please try signing in again."`), the existing code path treats it as a transient failure:

- `PERMANENT_FAILURE_CODES` in `app/core/balancer/logic.py` only lists `refresh_token_expired`, `refresh_token_reused`, `refresh_token_invalidated`, `account_deactivated`, `account_suspended`, `account_deleted`. It does **not** include `token_expired`.
- `classify_refresh_error("token_expired")` therefore returns `False`, so `RefreshError` is built with `is_permanent=False`.
- `AuthManager.refresh_account` only deactivates on a permanent error.
- `usage._should_deactivate_for_usage_error` only deactivates on HTTP 402/404, on the existing permanent code set, or on the "account has been deactivated" message hint.

Net effect: refresh keeps failing, the account stays `ACTIVE`, and the load balancer keeps selecting it. The reporter ran into exactly this with `email_3e2ba4c18bda` until they manually paused it.

## Why `token_expired` at the refresh boundary is permanent

This is not a transient error in practice:

- Access-token-only expiry would have caused the refresh call itself to return a fresh token pair (HTTP 200). The refresh wouldn't fail.
- A `token_expired` arriving at the **refresh boundary** means the refresh token (or the session it represents) is no longer usable.
- That is the same shape as `refresh_token_expired` / `refresh_token_invalidated` -- just a different error code on the upstream side.

So treating it as permanent matches what is actually happening on the upstream side, and matches the reporter's observation that the account only recovers after re-authentication.

## Fix

Add `token_expired` to `PERMANENT_FAILURE_CODES` with the message `"Authentication token expired - re-login required"`. This single change flows through both deactivation paths without further plumbing:

- `AuthManager.refresh_account` -- via `classify_refresh_error("token_expired") -> True`
- usage-refresh `_should_deactivate_for_usage_error` -- via its `code in PERMANENT_FAILURE_CODES` check

So a refresh-time `token_expired` now correctly deactivates the account on both the request-driven refresh path and the background usage refresh path. The reporter's affected account would now be deactivated on the next failing refresh and stop draining client requests.

Re-authentication clears the deactivation as it does today for the other permanent codes.

## Tests

```
tests/unit/test_auth_refresh.py::test_classify_refresh_error_token_expired_is_permanent PASSED
tests/unit/test_auth_manager.py::test_refresh_account_deactivates_when_upstream_returns_token_expired PASSED
```

The auth-manager regression test drives `refresh_account` against a stubbed `refresh_access_token` raising `RefreshError("token_expired", ..., classify_refresh_error("token_expired"))`, then asserts:

- the resulting `RefreshError` has `is_permanent=True` and `code="token_expired"`
- the account row is updated to `AccountStatus.DEACTIVATED`
- the deactivation reason surfaces the re-login / expired wording

Verification on this branch:

```
uv run ruff check app/core/balancer/logic.py tests/unit/test_auth_refresh.py tests/unit/test_auth_manager.py     # clean
uv run ruff format --check app/core/balancer/logic.py tests/unit/test_auth_refresh.py tests/unit/test_auth_manager.py  # clean
uv run ty check app/core/balancer/logic.py tests/unit/test_auth_refresh.py tests/unit/test_auth_manager.py        # clean
uv run pytest tests/unit/test_auth_refresh.py tests/unit/test_auth_manager.py tests/unit/test_load_balancer.py -q   # 74 passed
npx @fission-ai/openspec validate deactivate-on-token-expired                                                       # change is valid
```

## OpenSpec

Added `openspec/changes/deactivate-on-token-expired/` capturing the new behavior under the existing `usage-refresh-policy` capability (refresh-time and usage-refresh-time `token_expired` both deactivate; classification scenario; deactivation scenario; reason wording requirement).

## Hotfix candidate for v1.17.1 / v1.17.2

Small, scoped change to the permanent-failure classification, with two unit regressions and an OpenSpec entry. No runtime/migration impact. Pairs naturally with #598 (the bcrypt 72-byte fix) for the next patch release.

### DIFF
--- a/app/core/balancer/logic.py
+++ b/app/core/balancer/logic.py
@@ -14,6 +14,12 @@ PERMANENT_FAILURE_CODES = {
     "refresh_token_expired": "Refresh token expired - re-login required",
     "refresh_token_reused": "Refresh token was reused - re-login required",
     "refresh_token_invalidated": "Refresh token was revoked - re-login required",
+    # ``token_expired`` from the OAuth refresh endpoint means the refresh
+    # request itself failed because the refresh token (or the session it
+    # belonged to) is no longer usable -- access-token-only expiry would have
+    # returned a fresh token pair instead. Treat it as a permanent failure so
+    # the account stops being routed to until it is re-authenticated.
+    "token_expired": "Authentication token expired - re-login required",
     "account_deactivated": "Account has been deactivated",
     "account_suspended": "Account has been suspended",
     "account_deleted": "Account has been deleted",
--- a/openspec/changes/deactivate-on-token-expired/proposal.md
+++ b/openspec/changes/deactivate-on-token-expired/proposal.md
@@ -1,0 +1,22 @@
+## Why
+Issue #383 reports that accounts whose upstream session has expired keep returning `token_expired` (HTTP 401 with `{"error": {"code": "token_expired"}}`) on every request, yet codex-lb keeps the account in `ACTIVE` state. The load balancer continues to route real client traffic to it, every request fails the same way, and the only mitigation is for an operator to manually pause the account.
+
+The reporter's diagnosis is correct against the current code:
+
+- `PERMANENT_FAILURE_CODES` in `app/core/balancer/logic.py` only lists `refresh_token_expired`, `refresh_token_reused`, `refresh_token_invalidated`, `account_deactivated`, `account_suspended`, and `account_deleted`. It does **not** include `token_expired`.
+- `classify_refresh_error("token_expired")` therefore returns `False`, so `RefreshError` for this case is built with `is_permanent=False`.
+- `AuthManager.refresh_account` only deactivates the account when the error is permanent, and `usage._should_deactivate_for_usage_error` only deactivates on HTTP 402/404 or the existing permanent set, so a refresh-time `token_expired` slips through both paths.
+- Result: refresh keeps failing, the account stays `ACTIVE`, and the load balancer keeps selecting it. The reporter ran into exactly this with `email_3e2ba4c18bda` until they manually paused it.
+
+The cause is specifically that `token_expired` is reaching codex-lb on the **token refresh** endpoint. Access-token-only expiry would have caused the refresh call itself to return a fresh token pair, not to fail with `token_expired`. So a `token_expired` at the refresh boundary unambiguously means the refresh token (or the session it represents) is no longer usable, which is the same shape as `refresh_token_expired` and the other entries already classified as permanent.
+
+## What Changes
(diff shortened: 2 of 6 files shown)


Soju06:
@codex review

PR #601, opened 2026-05-13
fix(proxy): classify upstream overloaded_error as retryable transient

### LINKED ISSUE #565: Agent stops after transient upstream overload error instead of retrying
## Summary

When the upstream API returns a transient overload error:

> Our servers are currently overloaded. Please try again later

the Codex agent stops progressing and does not appear to retry the request. This makes long-running agent sessions fragile during temporary OpenAI/server overloads.

## Actual behavior

After receiving the overload error, the agent gets stuck or stops the current operation. It does not automatically retry the same request.

## Expected behavior

The agent should treat this as a transient retryable upstream/server error and retry automatically with a small bounded retry budget, for example:

- initial attempt
- retry 1 after short backoff
- retry 2 after short backoff

If all attempts fail, then surface the error to the user.

## Error observed

```text
Our servers are currently overloaded. Please try again later
```

## Why this matters
This error is usually temporary and not caused by the user prompt, tools, repository state, or agent logic. Without retries, the agent can stop in the middle of an otherwise recoverable task.

## Suggested fix
Classify upstream overload / retry-later messages as transient retryable failures.

## Suggested matching examples:

HTTP 5xx with OpenAI-style error payload
- server_error
- upstream_error with message containing:
-- servers are currently overloaded
-- please try again later

## Suggested behavior:
- retry only before the failed request has produced meaningful downstream output
- use bounded retries, e.g. 2 extra attempts
- use exponential backoff/jitter
- avoid retrying non-transient 4xx errors
- preserve the original error after retry exhaustion

## Workaround
A downstream proxy can implement this retry behavior, but it would be better for Codex itself to handle this transient upstream failure natively.

ozpool:
## Summary
- Fixes #565. Long-running agent sessions stop mid-task when the upstream returns `Our servers are currently overloaded. Please try again later`. OpenAI delivers that as `error.code = \"overloaded_error\"`, which is missing from `_TRANSIENT_CODES` in `app/modules/proxy/helpers.py`.
- The envelope often arrives without a 5xx HTTP status — streamed Responses traffic returns `200` before the error frame hits the wire, and some upstream variants downgrade the JSON 503 to 200 to keep the SSE channel open. With both code-set membership and the `http_status >= 500` fallback failing, the classifier returns `non_retryable`. `failover_decision` only retries / fails over for `rate_limit`, `quota`, or `retryable_transient` (`app/core/balancer/logic.py:361`), so the agent is failed back to the client without any account failover and stops.
- Add `\"overloaded_error\"` to `_TRANSIENT_CODES`, exercise the regression in `tests/unit/test_failover_foundation.py` using the `http_status=None` shape, and document the requirement under the `responses-api-compat` capability.
- No public surface change. Clients still see the upstream error envelope verbatim; only codex-lb's internal failover decision is widened by one code that is, by definition, transient.

## Validation
- `uv run pytest tests/unit/test_failover_foundation.py` — 44 passed
- `uvx ruff check app/modules/proxy/helpers.py tests/unit/test_failover_foundation.py` — clean
- `uv run ty check app/modules/proxy/helpers.py` — clean
- `openspec validate treat-upstream-overloaded-as-retryable` — valid

## Notes
- Same classifier feeds both `/v1/responses` and `/v1/chat/completions` failover, so the fix benefits both surfaces even though the documented requirement is parked under `responses-api-compat` (where the issue surfaced).
- Intentionally minimal: this PR widens the transient set by one code. It does not change retry budget, backoff, or any client-facing payload. If you'd prefer broader message-pattern matching as a defence in depth for hosts that strip the structured `code`, happy to follow up in a separate change.

Fixes #565

### DIFF
--- a/app/modules/proxy/helpers.py
+++ b/app/modules/proxy/helpers.py
@@ -36,7 +36,7 @@ PLAN_TYPE_PRIORITY = (
 
 _RATE_LIMIT_CODES = frozenset({"rate_limit_exceeded", "usage_limit_reached"})
 _QUOTA_CODES = frozenset({"insufficient_quota", "usage_not_included", "quota_exceeded"})
-_TRANSIENT_CODES = frozenset({"server_error", "upstream_error", "stream_incomplete"})
+_TRANSIENT_CODES = frozenset({"server_error", "upstream_error", "stream_incomplete", "overloaded_error"})
 
 
 def classify_upstream_failure(
--- a/openspec/changes/treat-upstream-overloaded-as-retryable/proposal.md
+++ b/openspec/changes/treat-upstream-overloaded-as-retryable/proposal.md
@@ -1,0 +1,19 @@
+## Why
+
+Issue #565 reports that long-running Codex agent sessions stop mid-task when the upstream returns a transient `"Our servers are currently overloaded. Please try again later"` error. The codex-lb classifier in `app/modules/proxy/helpers.py:classify_upstream_failure` keys its retryable-transient decision on `_TRANSIENT_CODES = {"server_error", "upstream_error", "stream_incomplete"}` plus any 5xx HTTP status.
+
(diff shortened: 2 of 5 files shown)


Soju06:
@codex review

Soju06:
@all-contributors please add @ozpool for code, test

Soju06:
@all-contributors please add @ozpool for code

PR #610, opened 2026-05-14
fix(proxy): accept /backend-api/codex/v1/<rest> as alias for /backend-api/codex/<rest>

### LINKED ISSUE #267: Strip /v1/ sent by badly behaved applications
Some applications like appending `v1` to the base url for some reason. 
When fed with `https://example.com/backend-api/codex/` they contact `https://example.com/backend-api/codex/v1/models` which severely aggravates my homicidal thoughts while I consider WHY on earth they would ask the user the base url and then NOT treat it like the base url it should be.

Please, for the love of god, just strip `/v1/` from api calls such that `https://example.com/backend-api/codex/v1/models` starts working.

ozpool:
## Summary

Resolves #267.

Some OpenAI-compatible clients unconditionally append `/v1/` to whatever base URL the operator configured. When the configured base already terminates at `/backend-api/codex`, those clients end up hitting `/backend-api/codex/v1/models`, `/backend-api/codex/v1/responses`, etc. codex-lb does not register that shape (OpenAI-style routes live at top-level `/v1/<rest>`, Codex-style at `/backend-api/codex/<rest>`), so the request 404s with no operator-facing signal.

## What changes

- New small path-rewrite middleware (`app/core/middleware/path_rewrite.py`) collapses `/backend-api/codex/v1/<rest>` to `/backend-api/codex/<rest>` by mutating `scope["path"]` and `scope["raw_path"]` together before routing.
- The aliasing only triggers when `<rest>` is non-empty — the exact `/backend-api/codex/v1` and `/backend-api/codex` paths are left alone, and the canonical top-level `/v1/<rest>` namespace is unaffected.
- Registered as the outermost HTTP middleware in `app/main.py` so the firewall, metrics, request-id, etc. all see the canonical path the rest of the stack already understands.
- OpenSpec change folder under `openspec/changes/strip-codex-v1-prefix/` adds the contract under `responses-api-compat`.

## Test plan

- [x] `pytest tests/unit/test_path_rewrite_middleware.py` — canonical no-op, alias rewrite, idempotent re-application, edge sentinels (`/backend-api/codex`, `/backend-api/codex/v1`, top-level `/v1/...`)
- [x] `pytest tests/integration/test_path_rewrite_alias.py` — `GET /backend-api/codex/v1/models` returns the same payload as `GET /backend-api/codex/models`
- [x] `pytest tests/integration/test_api_firewall_middleware.py tests/integration/test_health_and_errors.py` — existing middleware/firewall tests unaffected
- [x] `ruff check`, `ty check` clean on the new files

### DIFF
--- a/app/core/middleware/__init__.py
+++ b/app/core/middleware/__init__.py
@@ -1,10 +1,12 @@
 from app.core.middleware.api_firewall import add_api_firewall_middleware
 from app.core.middleware.dashboard_auth_proxy import add_dashboard_auth_proxy_middleware
+from app.core.middleware.path_rewrite import add_backend_api_codex_v1_alias_middleware
 from app.core.middleware.request_decompression import add_request_decompression_middleware
 from app.core.middleware.request_id import add_request_id_middleware
 
 __all__ = [
     "add_api_firewall_middleware",
+    "add_backend_api_codex_v1_alias_middleware",
     "add_dashboard_auth_proxy_middleware",
     "add_request_decompression_middleware",
     "add_request_id_middleware",
--- a/app/core/middleware/path_rewrite.py
+++ b/app/core/middleware/path_rewrite.py
@@ -1,0 +1,78 @@
+"""Path-rewrite middleware for backwards-compatible `/v1/` URL handling.
+
+Some OpenAI-compatible clients unconditionally append ``/v1/`` to
+whatever base URL the operator configured. When the configured base URL
+already terminates at ``/backend-api/codex`` (codex-lb's Codex-style
+entry point), those clients end up hitting
+``/backend-api/codex/v1/<rest>`` -- a shape codex-lb does not register
+because the OpenAI-style endpoints are mounted at the top-level
+``/v1/<rest>`` and the Codex-style endpoints at
+``/backend-api/codex/<rest>``.
+
+This middleware collapses the duplicated ``/v1`` segment in-place by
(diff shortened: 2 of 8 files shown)


Komzpa:
@codex review

ozpool:
Thanks for the catch — that's a real production gap, not just theoretical. Pushed `a55b223` switching the alias to a pure ASGI middleware so both `http` and `websocket` scopes get the rewrite.

The old shape was `@app.middleware("http")`, which Starlette only invokes for HTTP scopes. Codex-style websocket endpoints in `app/modules/proxy/api.py` are exposed under `/backend-api/codex` (and `/v1`), so a client that appends `/v1` to a `/backend-api/codex` base URL was getting a 404 on the WS handshake even though the matching HTTP request would have succeeded after this PR — which defeats the point of the alias.

The new `BackendApiCodexV1AliasMiddleware` is a plain ASGI class:
- rewrites `scope["path"]` and `scope["raw_path"]` on both `http` and `websocket` scopes
- passes `lifespan` and any other scope types straight through
- copies the scope before mutation so ASGI servers that reuse scope instances across calls don't get poisoned

Four direct-invocation tests in `tests/unit/test_path_rewrite_middleware.py` cover the new shape (http rewrite, websocket rewrite, lifespan passthrough, copy-on-rewrite). All 20/20 path-rewrite tests pass locally.

Komzpa:
@codex review

PR #611, opened 2026-05-14
fix(api-keys): map enforced service_tier auto/default to outbound omission

### LINKED ISSUE #546: Enforcing default service tier does not work

ozpool:
## Summary

Resolves #546.

Setting an API key's `enforced_service_tier` to `default` or `auto` previously broke every request through that key: the request failed upstream with `invalid_request_error Unsupported service_tier: default` (or `... auto`). codex-lb accepts both values at the API-key surface as conventions for "let upstream pick", but the wire-level equivalent of "let upstream pick" is the **absence** of the `service_tier` field, not the literal string.

## What changes

- In `apply_api_key_enforcement` (`app/modules/proxy/request_policy.py`), when `enforced_service_tier` resolves to `auto` or `default`, set `payload.service_tier = None` instead of the literal value. Real upstream tiers (`priority`, `flex`) keep propagating as the literal value unchanged.
- Extend the existing `api_key_service_tier_enforced` log line with `outbound_service_tier=` so operators can confirm enforcement reached upstream as expected.
- OpenSpec change folder under `openspec/changes/fix-enforced-service-tier-default/` adds the contract under `api-keys`.

## Test plan

- [x] `pytest tests/unit/test_proxy_utils.py` (228 tests pass; 3 new assertions: `default`/`auto` omission paths, plus a `flex` propagation regression guard)
- [x] Existing `test_apply_api_key_enforcement_overrides_service_tier_aliases_to_priority` continues to pass — the alias-to-`priority` rewrite path is unchanged.
- [x] `ruff check`, `ty check` clean on the touched files.

### DIFF
--- a/app/modules/proxy/request_policy.py
+++ b/app/modules/proxy/request_policy.py
@@ -27,6 +27,15 @@ logger = logging.getLogger(__name__)
 _UNSUPPORTED_UPSTREAM_REASONING_EFFORTS: frozenset[str] = frozenset({"minimal"})
 _DEFAULT_REASONING_EFFORT_FALLBACK = "low"
 
+# Service tier values codex-lb accepts at the API-key surface but that the
+# ChatGPT/Codex backend rejects with ``Unsupported service_tier: <value>``.
+# Semantically both ``auto`` and ``default`` mean "let upstream pick" -- the
+# same thing as omitting the field entirely -- so when an enforced API-key
+# policy resolves to one of these, we forward the request without a
+# ``service_tier`` instead of sending a literal that fails upstream. See
+# https://github.com/Soju06/codex-lb/issues/546
+_UPSTREAM_OMIT_SERVICE_TIERS: frozenset[str] = frozenset({"auto", "default"})
+
 
 def validate_model_access(api_key: ApiKeyData | None, model: str | None) -> None:
     if api_key is None:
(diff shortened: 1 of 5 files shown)


Soju06:
@codex review

Soju06:
Thanks @ozpool -- the framing is exactly right (the wire-level equivalent of "let upstream pick" is the absence of the field, not the literal string), and the OpenSpec change folder is in place. Two items to address before merging.

## P2 from codex review -- log noise after the `auto/default -> None` remap

The new mapping changes the outbound value but the log guard still compares against the original enforced literal:

```python
# app/modules/proxy/request_policy.py:~99
if requested_service_tier != api_key.enforced_service_tier:
    logger.info(
        "api_key_service_tier_enforced request_id=%s key_id=%s "
        "requested_service_tier=%s enforced_service_tier=%s "
        "outbound_service_tier=%s",
        ...
    )
```

For an API key enforced to `auto` or `default`, requests that already omit `service_tier` (`requested_service_tier=None`) now log `api_key_service_tier_enforced` on every call -- because `None != "auto"`. But the *outbound* tier is `None` either way, so the enforcement didn't actually change anything. Operators will see this as a high-volume false-positive enforcement signal.

Suggested guard: compare against the effective outbound tier instead of the literal:

```python
if requested_service_tier != effective_service_tier:
    logger.info(
        "api_key_service_tier_enforced request_id=%s key_id=%s "
        "requested_service_tier=%s enforced_service_tier=%s "
        "outbound_service_tier=%s",
        get_request_id(),
        api_key.id,
        requested_service_tier,
        api_key.enforced_service_tier,
        effective_service_tier,
    )
```

Now the log only fires when the outbound tier actually differs from what the client sent, which is the signal operators care about. The `enforced_service_tier` literal is still in the log line for diagnostic context.

Regression to add against the existing `tests/unit/test_proxy_utils.py` test block:

```python
def test_enforced_auto_does_not_log_when_request_already_omits_service_tier(caplog):
    # request omits service_tier (the common case), key enforces "auto"
    # outbound is None either way -> no api_key_service_tier_enforced log
    with caplog.at_level(logging.INFO):
        apply_api_key_enforcement(make_payload(service_tier=None), make_api_key(enforced_service_tier="auto"))
    assert not any("api_key_service_tier_enforced" in r.getMessage() for r in caplog.records)
```

And the symmetric one confirming the log still fires when the outbound actually differs (client sent `priority`, key enforces `default` -> outbound `None`).

## CI: Lint (ruff) fail on `tests/unit/test_proxy_utils.py`

The Lint job is red on this head:

```
Would reformat: tests/unit/test_proxy_utils.py
1 file would be reformatted, 434 files already formatted
```

Running `uv run ruff format tests/unit/test_proxy_utils.py` locally should land the formatter delta. Once that's pushed the rest of the merge gates pass cleanly.

## Not blocking otherwise

- The mapping decision (`auto`/`default` -> outbound `None`, `priority`/`flex` literal-through) is correct.
- `outbound_service_tier=` in the log line is exactly the operator-visibility piece that was missing -- keep it.

After the P2 fix and `ruff format` land I'll re-run `@codex review` and queue for merge.

Komzpa:
@codex review

Issue #620, opened 2026-05-14
Complete deferred Images API fan-out and observability tasks


whoisjayd:
## Suggested labels
`enhancement`, `proxy`, `images-api`, `needs-design`

## Confidence / impact / size
- Confidence: Medium
- Impact: Medium to high
- Size: Medium to large
- OpenSpec needed: Yes. This changes Images API behavior and should update the active images compatibility OpenSpec change.

## Summary
The OpenAI-compatible `/v1/images` implementation currently hard-rejects `n > 1` because client-side fan-out is not implemented. The OpenSpec task list also has deferred observability/validation items.

## Evidence
- `app/modules/proxy/images_service.py:107-113` explicitly documents and asserts that `n > 1` fan-out is not implemented.
- `tests/unit/test_images_schemas.py:200-212` documents the hard cap and says it should be relaxed when fan-out is introduced.
- `openspec/changes/add-images-api-compat/tasks.md:38` leaves structured logs / Prometheus metrics deferred.
- `openspec/changes/add-images-api-compat/tasks.md:60` leaves strict OpenSpec validation deferred because the CLI was not available in that workspace.

## Problem
Clients that request multiple generated images with `n > 1` cannot be served through the compatibility endpoint. Observability is also intentionally minimal compared to the rest of the proxy surface.

## Proposed implementation
This should likely be split into smaller PRs:

1. Design/spec PR:
   - Define expected `n > 1` behavior in OpenSpec.
   - Clarify whether fan-out should be sequential or concurrent.
   - Define partial-failure behavior.
   - Define request-log/metrics expectations.
2. Implementation PR:
   - Fan out `n` requests to the image generation tool.
   - Preserve response shape and ordering.
   - Bound concurrency and cost exposure.
   - Add structured logging/metrics if accepted by maintainers.

## Acceptance criteria
- [ ] OpenSpec describes multi-image fan-out behavior and failure modes.
- [ ] `n > 1` requests return multiple images in a client-compatible response.
- [ ] Existing `n = 1` behavior remains unchanged.
- [ ] Tests cover `n = 1`, `n > 1`, unsupported model/size/quality/style, and partial upstream failure behavior.
- [ ] Observability expectations are either implemented or explicitly documented as out of scope.

## Suggested verification
```bash
uv run pytest tests/unit/test_images_schemas.py tests/unit/test_images_translation.py tests/integration/test_proxy_images.py -q
openspec validate --specs
uvx ruff check .
uv run ty check
```

[...]

PR #628, opened 2026-05-14
spec(proxy): define images API fan-out behavior

whoisjayd:
## Summary
- Define non-streaming `/v1/images/*` fan-out semantics for `n > 1` in the existing Images API OpenSpec change.
- Keep streaming `n > 1` explicitly rejected until streaming fan-out is specified separately.
- Add implementation tasks for validation, fan-out orchestration, usage aggregation, and tests.

## Stack
- Part 1 of 2 for #620.
- Follow-up implementation PR will branch from this PR after spec/design feedback.

## Verification
- `git diff --check`
- `uv run openspec validate --specs` could not run locally because the `openspec` executable is not available in this environment.

Closes no issue yet; this is the spec/design slice for #620.


whoisjayd:
@Soju06 can you please review so i can proceed with implementation part

PR #634, opened 2026-05-14
fix(proxy): replay HTTP bridge quota failures

Komzpa:
## Summary
- retry pre-created HTTP bridge requests when upstream returns transparent replay quota/rate errors before `response.created`
- fail closed with `upstream_unavailable` for owner-pinned `previous_response_id` quota failures instead of leaking raw account quota details downstream
- retire only owner-quota-drained HTTP bridge sessions before closing their stale upstream, while preserving existing `previous_response_not_found` keepalive behavior

## Validation
- `uvx ruff check .`
- `uvx ruff format --check .`
- `uv run ty check`
- `uv run pytest -q tests/unit/test_proxy_http_bridge.py`
- `uv run pytest -q -ra -vv -o faulthandler_timeout=300 -o faulthandler_exit_on_timeout=true --timeout=180 --timeout-method=thread --durations=20 tests/integration/test_http_responses_bridge.py tests/integration/test_proxy_websocket_responses.py`
- `git diff --check`
- `codex review --base origin/main` clean after review-found edge fixes

## Live note
A live aggregate containing this proxy fix is already deployed locally as `codex-lb:live-stack-httpbridge-quota-ce89d54`; this PR keeps the upstream change scoped to the HTTP bridge quota replay behavior.

### DIFF
--- a/app/modules/proxy/service.py
+++ b/app/modules/proxy/service.py
@@ -5221,6 +5221,25 @@ class ProxyService:
         _release_websocket_response_create_gate(request_state, session.response_create_gate)
         await self._release_websocket_reservation(request_state.api_key_reservation)
         request_state.api_key_reservation = None
+        await self._retire_http_bridge_after_drain_if_ready(session)
+        return True
+
+    async def _retire_http_bridge_after_drain_if_ready(self, session: "_HTTPBridgeSession") -> bool:
+        if not (session.upstream_control.reconnect_requested and session.upstream_control.retire_after_drain):
+            return False
+        async with session.pending_lock:
+            should_reconnect = not session.pending_requests
+        if not should_reconnect:
+            return False
+
+        session.closed = True
+        try:
+            await session.upstream.close()
+        except Exception:
+            logger.debug(
+                "Failed to close HTTP bridge upstream for reconnect",
+                exc_info=True,
+            )
         return True
 
     async def _relay_http_bridge_upstream_messages(
@@ -5269,19 +5288,8 @@ class ProxyService:
                 if message.kind == "text" and message.text is not None:
                     session.last_upstream_close_code = None
                     await self._process_http_bridge_upstream_text(session, message.text)
-                    if session.upstream_control.reconnect_requested and session.upstream_control.retire_after_drain:
-                        async with session.pending_lock:
-                            should_reconnect = not session.pending_requests
-                        if should_reconnect:
-                            session.closed = True
-                            try:
-                                await session.upstream.close()
-                            except Exception:
-                                logger.debug(
-                                    "Failed to close HTTP bridge upstream for reconnect",
-                                    exc_info=True,
-                                )
-                            break
+                    if await self._retire_http_bridge_after_drain_if_ready(session):
+                        break
                     continue
 
                 session.last_upstream_close_code = message.close_code
--- a/app/modules/proxy/service.py
+++ b/app/modules/proxy/service.py
@@ -4993,6 +4993,12 @@ class ProxyService:
                     502,
                     openai_error("upstream_unavailable", "HTTP responses session bridge is closed"),
                 )
+        if session.upstream_control.retire_after_drain:
+            await self._retire_http_bridge_after_drain_if_ready(session)
+            raise ProxyResponseError(
+                502,
+                openai_error("upstream_unavailable", "HTTP responses session bridge is retiring"),
+            )
         await self._maybe_prewarm_http_bridge_session(
             session,
             request_state=request_state,
--- a/app/modules/proxy/service.py
+++ b/app/modules/proxy/service.py
@@ -5269,6 +5269,19 @@ class ProxyService:
                 if message.kind == "text" and message.text is not None:
                     session.last_upstream_close_code = None
                     await self._process_http_bridge_upstream_text(session, message.text)
+                    if session.upstream_control.reconnect_requested and session.upstream_control.retire_after_drain:
+                        async with session.pending_lock:
+                            should_reconnect = not session.pending_requests
+                        if should_reconnect:
+                            session.closed = True
+                            try:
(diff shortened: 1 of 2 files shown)


Soju06:
@codex review

Soju06:
@Komzpa codex review's P1 catch on `app/modules/proxy/service.py:~5719` is real: the new `upstream_unavailable` masking + `retire_after_drain` block is gated on `retry_error_code is not None`, but `_websocket_precreated_retry_error_code()` returns `None` whenever `has_other_pending_requests=True`. With the default HTTP bridge `queue_limit > 1`, a `previous_response_id` request hitting `usage_limit_reached` while another request is queued in the same session bypasses both:

- the rewrite to `upstream_unavailable` -> the raw upstream quota metadata still leaks downstream
- the `retire_after_drain` toggle -> the owner-pinned session keeps getting reused on a stale upstream

That breaks the second and third bullets of this PR's summary on exactly the configuration this PR is supposed to fix.

The two branches inside the guarded block already split on `status_request_state.previous_response_id is None` vs `is not None`, and only the `previous_response_id is not None` branch needs to run when there are other pending requests (the `previous_response_id is None` branch is the transparent-replay path that legitimately needs an isolated request).

Suggestion: separate the two intents at the guard. Keep `retry_error_code` as the gate for the replay branch, and gate the owner-pinned rewrite + retire-after-drain branch on the request shape directly:

```python
# pseudocode -- exact symbol names per current file
if status_request_state is not None and status_request_state.previous_response_id is not None \
        and status_request_state.preferred_account_id is not None \
        and _is_owner_pinned_quota_error(event_type, payload):
    status_request_state.error_http_status_override = 502
    session.upstream_control.reconnect_requested = True
    session.upstream_control.retire_after_drain = True
    event, payload, event_type, rewritten_text = (
        _rewrite_websocket_previous_response_owner_unavailable_event(
            request_state=status_request_state,
        )
    )
    event_block = f"data: {rewritten_text}\n\n"
elif retry_error_code is not None and not is_previous_response_not_found_event:
    # existing transparent-replay path (no `previous_response_id`)
    ...
```

Where `_is_owner_pinned_quota_error` covers the same shapes `_websocket_precreated_retry_error_code` already classifies as quota / rate-limit transient. That keeps the replay path single-request-only (where it has to be) while letting the owner-pinned masking path fire regardless of queue depth.

Regression to add against `tests/unit/test_proxy_http_bridge.py`:

```python
async def test_http_bridge_masks_owner_pinned_quota_error_with_queued_requests():
    # owner-pinned request_state (previous_response_id + preferred_account_id)
    # plus another queued request in the same session
    # upstream emits usage_limit_reached before response.created
    # assert: downstream_text envelope = upstream_unavailable, no raw quota details
    # assert: session.upstream_control.retire_after_drain is True
```

The rest of the change reads clean -- the replay path itself, the symmetric `_WebSocketUpstreamControl.retire_after_drain` field, and the upstream-reader retirement on drain when `pending_requests` is empty.

After the P1 fix lands I'll re-run `@codex review` and we can merge.

Soju06:
@codex review

Soju06:
@Komzpa second-round `@codex review` on `032e024` confirms the P1 fix is correct: the owner-pinned masking guard now classifies on request shape alone (not `has_other_pending_requests`), the if/elif ordering is right, and `test_http_bridge_masks_owner_pinned_quota_error_with_queued_requests` exercises the exact queue-depth-2 scenario. Good.

One P2 left and I think it's worth addressing in this PR before merge:

> The new retirement path only runs immediately after processing a text frame (`_relay_http_bridge_upstream_messages`), so `reconnect_requested + retire_after_drain` is never acted on if the remaining pending requests are drained by `_detach_http_bridge_request` (for example, client cancellation) and upstream sends no more messages.

Concretely: owner-pinned `usage_limit_reached` masks the failed request, sets `retire_after_drain=True`, and `_detach_http_bridge_request` drains the remaining queued requests on the cancellation path. The reader is parked in `upstream.receive()` with an empty queue, no further text frame arrives, the drain-then-retire check inside `_relay_http_bridge_upstream_messages` never fires, and the next `_get_or_create_http_bridge_session` lookup happily returns the quota-drained owner session.

Narrow trigger (cancellation + silent upstream), but it directly undermines the third bullet of this PR's summary — owner-quota-drained sessions are supposed to be retired before close, and they aren't on this path.

Suggestion — either:

1. Have `_detach_http_bridge_request` (or whichever helper observes `queued_request_count == 0` post-detach) check `session.upstream_control.retire_after_drain` and, if set, mark the session closed + close the upstream so the reader unblocks and the session-pool entry is dropped. OR
2. In `_get_or_create_http_bridge_session`, refuse to reuse a session whose `upstream_control.retire_after_drain` is set — treat it the same as `closed=True` for lookup purposes.

(1) is the matching half of the existing `_relay_http_bridge_upstream_messages` drain-then-retire path; (2) is cheaper but leaves the reader hanging on a doomed upstream until idle TTL.

Regression to add (mirrors the queued-quota test but exercises the cancellation path):

```python
async def test_http_bridge_retire_after_drain_closes_session_on_cancellation():
    # owner-pinned request_state + one queued request_state on a shared session
    # upstream emits usage_limit_reached
    #   -> owner-pinned masking + retire_after_drain=True, owner request finalized
    # client cancels the queued request -> _detach_http_bridge_request drains it
    # no further upstream frames
    # assert: session.closed is True (or, for option 2:
    #         subsequent _get_or_create_http_bridge_session lookup with the same
    #         affinity key returns a NEW session, not the quota-drained one)
```

After this lands I'll re-run `@codex review` and we can merge.

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

PR #635, opened 2026-05-14
fix(oauth): start device polling on oauth start

### LINKED ISSUE #110: Device Code OAuth can stay forever in pending if frontend misses /api/oauth/complete
## Environment
- codex-lb image: `ghcr.io/soju06/codex-lb:latest`
- Run mode: Docker
- Client access: via Tailscale (`100.x.x.x`)
- Container ports: `2455`, `1455`
- Container user: `uid=1000(app)`
- Data dir: `/var/lib/codex-lb` (docker volume)

## Problem
Adding an account via **Device Code** often hangs at “Waiting for authorization” forever, even after entering the code successfully on `auth.openai.com/codex/device`.

On the same server, `odrobnik/codex-account-switcher-skill` works fine, so this is not a global auth/network issue.

## What I observed
From `codex-lb` logs:
- `POST /api/oauth/start` is called
- then many `GET /api/oauth/status`
- but frontend/user flow may miss `POST /api/oauth/complete` (or it may fail client-side), and then status remains `pending` forever

When I manually call `/api/oauth/complete` with valid `deviceAuthId + userCode`, flow can proceed and eventually becomes `success`.

So the backend currently depends on `complete` being called to start polling.

## Root cause (as debugged)
In `app/modules/oauth/service.py`, device polling is started from `complete_oauth()` only, not from `start_oauth()` device branch.

If frontend/proxy/session misses `POST /api/oauth/complete`, backend never polls token endpoint and status never leaves `pending`.

## Expected behavior
Device flow should be robust:
- once `/api/oauth/start` (device mode) returns `deviceAuthId + userCode`, backend should start polling automatically.
- `/api/oauth/complete` may remain optional/compat API, but should not be required for normal success.

## Suggested fix
Start polling task directly in `_start_device_flow()` after saving state:
- create `DevicePollContext(...)`
- set `state.poll_task = asyncio.create_task(self._poll_device_tokens(...))`
- keep existing guards in `complete_oauth()` to avoid duplicate poll tasks.

This makes Device Code work out-of-the-box even if frontend misses `complete`.

## Why this matters
Right now users need local patches/workarounds to add accounts reliably.
A release-level fix would prevent recurring manual patching after updates.

## Extra note
This is unrelated to `~/.codex/auth.json` (used by Codex CLI/account-switcher).
`codex-lb` stores tokens in its own DB (`/var/lib/codex-lb/store.db`), and file permissions were OK.

Komzpa:
## Summary
- start Device Code OAuth token polling as soon as `/api/oauth/start` initializes the device flow
- keep `/api/oauth/complete` as a duplicate-safe compatibility endpoint
- add an OpenSpec delta and regression coverage for start-only device OAuth

Fixes https://github.com/Soju06/codex-lb/issues/110

## Validation
- `uv run python -m pytest -q -ra tests/integration/test_oauth_flow.py -o faulthandler_timeout=300 -o faulthandler_exit_on_timeout=true --timeout=180 --timeout-method=thread`
- `uv run python -m ruff check app/modules/oauth/service.py tests/integration/test_oauth_flow.py`
- `uv run python -m ty check app/modules/oauth/service.py tests/integration/test_oauth_flow.py`
- `git diff --check`

### DIFF
--- a/app/modules/oauth/service.py
+++ b/app/modules/oauth/service.py
@@ -165,22 +165,10 @@ class OauthService:
                 return OauthCompleteResponse(status="success")
             if state.method != "device":
                 return OauthCompleteResponse(status="pending")
-            if state.poll_task and not state.poll_task.done():
-                return OauthCompleteResponse(status="pending")
-            if not state.device_auth_id or not state.user_code or not state.expires_at:
+            if not self._ensure_device_poll_task_locked(state):
                 state.status = "error"
                 state.error_message = "Device code flow is not initialized."
                 return OauthCompleteResponse(status="error")
-
-            interval = state.interval_seconds if state.interval_seconds is not None else 0
-            interval = max(interval, 0)
-            poll_context = DevicePollContext(
-                device_auth_id=state.device_auth_id,
-                user_code=state.user_code,
-                interval_seconds=interval,
-                expires_at=state.expires_at,
-            )
-            state.poll_task = asyncio.create_task(self._poll_device_tokens(poll_context))
             return OauthCompleteResponse(status="pending")
 
     async def _start_browser_flow(self) -> OauthStartResponse:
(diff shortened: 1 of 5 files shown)


Soju06:
@codex review

PR #686, opened 2026-05-17
fix(dashboard): derive quota status from usage windows

Komzpa:
## Summary
- derive account dashboard status from the current usage windows before rendering
- show `quota_exceeded` when weekly/secondary usage is exhausted even if the persisted account row still says `active`
- add regression coverage for the “5h remaining, weekly exhausted” dashboard case
- sync the editable package version in `uv.lock` with `pyproject.toml`

## Validation
```bash
uvx ruff check app/modules/accounts/mappers.py tests/integration/test_dashboard_overview.py
uv run --locked ty check app/modules/accounts/mappers.py tests/integration/test_dashboard_overview.py
uv run --locked pytest -q -ra tests/integration/test_dashboard_overview.py::test_dashboard_overview_derives_quota_status_from_current_weekly_usage tests/integration/test_dashboard_overview.py::test_dashboard_overview_maps_weekly_only_primary_to_secondary
```

### DIFF
--- a/app/modules/accounts/mappers.py
+++ b/app/modules/accounts/mappers.py
@@ -6,9 +6,10 @@ from app.core import usage as usage_core
 from app.core.auth import DEFAULT_PLAN, extract_id_token_claims
 from app.core.crypto import TokenEncryptor
 from app.core.plan_types import coerce_account_plan_type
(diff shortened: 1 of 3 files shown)


Komzpa:
@codex review

PR #688, opened 2026-05-18
fix(proxy): persist request logs outside cancellation

Komzpa:
## Summary
- persist proxy request logs in a separate shielded task so request-scope cancellation does not cancel the database write mid-flight
- preserve caller `CancelledError` semantics after scheduling the background persistence task
- keep tracking background request-log tasks until they finish, with explicit warning logs if the task itself is cancelled or fails
- add a regression test that cancels the caller while `add_log` is blocked and verifies the request log still gets written

## Validation
- `uv run pytest tests/unit/test_proxy_utils.py::test_write_request_log_continues_after_caller_cancellation`
- `uv run pytest tests/unit/test_proxy_utils.py -k 'write_request_log_continues_after_caller_cancellation or stream_responses_logs_actual_service_tier_and_requested_tier_trace'`
- `uvx ruff check app/modules/proxy/service.py tests/unit/test_proxy_utils.py`
- `uvx ruff format --check app/modules/proxy/service.py tests/unit/test_proxy_utils.py`
- `git diff --check`

### DIFF
--- a/app/modules/proxy/service.py
+++ b/app/modules/proxy/service.py
@@ -398,6 +398,7 @@ class ProxyService:
         self._file_account_pin_lock = asyncio.Lock()
         self._http_bridge_lock = anyio.Lock()
         self._work_admission: WorkAdmissionController | None = None
+        self._request_log_tasks: set[asyncio.Task[None]] = set()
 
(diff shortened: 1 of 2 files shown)


Komzpa:
@codex review

Issue #678, opened 2026-05-16
docs: FAQ entry on /wham/usage vs Codex Desktop Settings UI divergence during OpenAI-side reset events


Soju06:
## Context

Follow-up from #676. We've now seen the same confusion pattern across several reports during OpenAI-side team reset events: operators see Codex Desktop **Settings → Account** UI show a reset, but codex-lb (and Codex CLI's own quota pill, for the same reason) still shows the account as rate-limited or quota-exceeded for some time afterwards.

This is not a bug in codex-lb. The Settings UI and the `/wham/usage` rate-limiter endpoint are different OpenAI-side data sources, and they can legitimately disagree during a reset window. codex-lb only reads `/wham/usage`.

## Proposal

Add a FAQ entry (in `docs/` and/or README troubleshooting section) covering:

### Why does codex-lb show my account as `rate_limited` when Codex Desktop says it's reset?

Cover:

1. **Which endpoint codex-lb uses**: `GET https://chatgpt.com/backend-api/wham/usage`, with link to [`app/core/clients/usage.py`](https://github.com/Soju06/codex-lb/blob/main/app/core/clients/usage.py).
2. **How status is derived**: link to [`app/core/usage/quota.py`](https://github.com/Soju06/codex-lb/blob/main/app/core/usage/quota.py) and a one-paragraph summary of the `>=100` → `rate_limited` / `quota_exceeded` rule and the auto-recovery on `< 100`.
3. **Why Settings UI and `/wham/usage` can disagree**: they are backed by different OpenAI data sources; the rate-limiter (`/wham/usage`) updates lazily, typically on the next chargeable request or when its internal window crosses `reset_at`. The Settings UI is fed by a different account/quota view that updates earlier on team-side reset events.
4. **What you can do**: wait for the next request through that account to wake the upstream limiter, or (once #677 lands) hit the dashboard's `Probe` button.
5. **How to verify it's upstream, not codex-lb**: include the `curl` repro:

   ````bash
   ACCESS_TOKEN=...
   ACCOUNT_ID=...   # chatgpt-account-id (UUID), not codex-lb's id

   curl -s https://chatgpt.com/backend-api/wham/usage \
     -H "Authorization: Bearer ${ACCESS_TOKEN}" \
     -H "chatgpt-account-id: ${ACCOUNT_ID}" \
     -H "Accept: application/json" | jq '.rate_limit'
   ````

[...]

Issue #677, opened 2026-05-16
feat(dashboard): per-account force-probe action to wake upstream limiter after reset events


Soju06:
## Context

Follow-up from #676. After OpenAI-side team reset events, there is a window where:

- Codex Desktop **Settings → Account** UI has already flipped to the reset state
- but `https://chatgpt.com/backend-api/wham/usage` still returns the pre-reset `used_percent` (the rate-limiter's internal counter has not yet been re-evaluated)

During that window, codex-lb faithfully mirrors `/wham/usage` (verified in #676), so the affected account stays in `rate_limited` / `quota_exceeded` until upstream updates. The operator has nothing to do but wait.

A real `responses.create` request through codex-lb against the affected account usually causes the upstream limiter to re-evaluate the window and start returning `used_percent < 100`, after which codex-lb auto-recovers on the next refresh tick (60s default).

## Proposal

Add a one-shot **force-probe** action exposed on the account row in the dashboard (and as an admin API endpoint), which:

1. Bypasses load-balancer scoring (account is excluded from selection while `rate_limited`/`quota_exceeded`, so it can't probe itself organically).
2. Sends a single minimal request — a `responses.create` with `max_output_tokens=1` (or equivalent cheapest path) — through codex-lb's existing proxy pipeline, pinned to the specific account.
3. Triggers an immediate usage refresh against `/wham/usage` after the probe completes.
4. Surfaces the result inline (HTTP status, returned `used_percent`, new account status if changed).

## UX

- Per-account row in `/accounts` dashboard: small button labeled `Probe` (icon-only is fine), only enabled for `rate_limited` / `quota_exceeded` / status mismatch states.
- Click -> confirm dialog (this consumes 1 minimal call against the account's quota) -> progress spinner -> outcome toast.

## Admin API

```
POST /api/accounts/{account_id}/probe
```

Response:
```json
{
  "request_id": "...",
  "probe_status_code": 200,
  "primary_used_percent_before": 100.0,
  "primary_used_percent_after": 0.0,
  "account_status_before": "rate_limited",
  "account_status_after": "active"
}
```

## Why this is worth doing even though it doesn't fix upstream

- Gives operators a concrete action during reset events — no more "is anything happening?" anxiety
- Doubles as a debug tool for `/wham/usage` vs Settings UI disagreements in general
- The reset-event pattern looks like it's becoming a recurring thing on OpenAI's side, so the support volume is non-trivial

## Out of scope

[...]

Issue #676, opened 2026-05-16
bug(accounts): Codex team reset event leaves Plus account quota stale/inconsistent between Codex UI and codex-lb


NN1985:
### Pre-flight checklist

- [x] I searched existing issues for similar reports.
- [x] I'm on the latest patch release of my minor version.

### codex-lb version

1.17.0

### Affected account plan(s)

Plus

### Symptom category

Quota / usage values are wrong

### What does the dashboard / logs show?

```markdown
Running codex-lb `1.17.0`
Docker image: `ghcr.io/soju06/codex-lb:1.17.0`
Image revision: `fd0fd085b2a813b4ed89f362f5aa0a43f84b21dd`

This was first observed on `1.15.0`, then I upgraded cleanly to `1.17.0`; the behavior persists.

After the OpenAI/Codex team reset event for paid Codex usage limits, the direct Codex Desktop account menu shows an affected Plus account as reset:

- 5h: `99% remaining`
- Weekly: `100% remaining`

But codex-lb still reports stale/dry quota state for the same account.

Sanitized `/api/accounts` snapshot from `2026-05-16 ~20:35 UTC`:

| account | status | plan | 5h remaining % | weekly remaining % | 5h reset | weekly reset | windows |
|---|---|---|---:|---:|---|---|---|
| account A | active | plus | 15 | 36 | 2026-05-16T23:32:42Z | 2026-05-21T23:50:28Z | 300 / 10080 |
| account B | active | plus | 71 | 47 | 2026-05-17T00:17:53Z | 2026-05-22T02:06:38Z | 300 / 10080 |
| account C | rate_limited | plus | 0 | 25 | 2026-05-16T23:03:43Z | 2026-05-21T05:39:51Z | 300 / 10080 |
| account D | active | plus | 99 | 10 | 2026-05-17T01:35:56Z | 2026-05-18T04:02:08Z | 300 / 10080 |
| account E | active | plus | 99 | 4 | 2026-05-17T01:35:56Z | 2026-05-18T02:48:05Z | 300 / 10080 |

For account C, Codex Desktop direct UI shows reset, but codex-lb still shows:
- status: `rate_limited`
- 5h remaining: `0%`
- weekly remaining: `25%`

For account E, re-importing with a fresh token briefly showed weekly as blank/`--`, then codex-lb populated the same stale weekly value again.

Recent DB `usage_history` rows show codex-lb is actively refreshing these values, not just serving an old frontend cache:


[...]

PR #690, opened 2026-05-18
fix(proxy): cap selector "Try again in" hint at 300s

AnobleSCM:
## Summary

When \`select_account\` cannot return a candidate, the surfaced \`Rate limit exceeded. Try again in {N}s\` value is currently driven directly by \`min(reset_at)\` of the quota-exceeded pool (or \`min(cooldown_until)\` of all states) — see \`app/core/balancer/logic.py:227-235\`. After OpenAI-side reset events where \`/wham/usage\` propagates lazily (the same pattern documented in #676 / #677 / #678), \`N\` can reach ~89,872s (~25h). Clients (Codex CLI, OpenCode, OpenClaw) then back off for hours when codex-lb's own 60s \`/wham/usage\` refresh + 120s per-status cooldown threshold would have recovered the account inside a 5-minute window.

This change caps the user-visible retry hint at \`SELECTOR_RETRY_HINT_MAX_SECONDS = 300\`. \`AccountState.reset_at\` and \`AccountState.cooldown_until\` are unchanged and continue to drive selection, telemetry, and dashboard reads — only the string returned alongside \`SelectionResult(account=None, ...)\` is clamped.

This is the operator-facing UX gap that fed several of the reports in #676 / #678: operators see a \"wait 25 hours\" message and assume they are hard-blocked, when in reality codex-lb will auto-recover on the next refresh tick. With the clamp, clients reattempt inside the auto-recovery window and the symptom resolves naturally.

## Test plan

- [x] \`uv run pytest tests/unit/test_load_balancer.py -k 'select_account or retry_hint'\` — 27 passed, 37 deselected
- [x] \`uv run ruff check app/core/balancer/logic.py tests/unit/test_load_balancer.py\` — clean
- [x] 3 new regression tests in \`tests/unit/test_load_balancer.py\`:
  - \`test_select_account_caps_quota_exceeded_retry_hint\` — \`reset_at = now + 89_872\` clamps to 300s
  - \`test_select_account_preserves_short_quota_exceeded_retry_hint\` — \`reset_at = now + 60\` surfaces as 60s (no clamp)
  - \`test_select_account_caps_cooldown_retry_hint\` — \`cooldown_until = now + 86_400\` clamps to 300s
- [x] Existing \`test_select_account_reports_cooldown_wait_time\` still passes

OpenSpec change folder: \`openspec/changes/clamp-selector-retry-hint/\` (modifies the \`query-caching\` capability spec to add the clamp scenarios).

## Notes

- The clamp is a UX-only change. Hard upstream caps (legitimate multi-day weekly resets) still produce repeated rejections; clients retry, see the same 503, and back off via their own exponential backoff. Net behavior is no worse than today's repeated requests against a hard-capped account, and substantially better when the upstream state has actually recovered.
- No public API or schema change. The error envelope shape is preserved.
- Refs #676 / #678.

### DIFF
--- a/app/core/balancer/logic.py
+++ b/app/core/balancer/logic.py
@@ -228,11 +228,11 @@ def select_account(
                 reset_candidates = [s.reset_at for s in quota_exceeded if s.reset_at]
                 if reset_candidates:
                     wait_seconds = max(0, min(reset_candidates) - int(current))
-                    return SelectionResult(None, f"Rate limit exceeded. Try again in {wait_seconds:.0f}s")
+                    return SelectionResult(None, _format_retry_hint(wait_seconds))
             cooldowns = [s.cooldown_until for s in all_states if s.cooldown_until and s.cooldown_until > current]
             if cooldowns:
                 wait_seconds = max(0.0, min(cooldowns) - current)
-                return SelectionResult(None, f"Rate limit exceeded. Try again in {wait_seconds:.0f}s")
+                return SelectionResult(None, _format_retry_hint(wait_seconds))
             return SelectionResult(None, "No available accounts")
 
     def _reset_first_sort_key(state: AccountState) -> tuple[int, float, float, float, str]:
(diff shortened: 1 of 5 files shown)


Komzpa:
@codex review

Soju06:
@all-contributors please add @AnobleSCM for code, test

PR #691, opened 2026-05-18
feat(accounts): add per-account force-probe admin endpoint

AnobleSCM:
## Summary

Implements the per-account force-probe action proposed in #677.

After OpenAI-side reset events, accounts can get stuck in \`rate_limited\` / \`quota_exceeded\` because \`/wham/usage\` propagates lazily. The selector excludes blocked accounts, so the stuck account cannot wake the upstream limiter via organic traffic — operators today have no first-class action besides waiting for the natural window roll or editing the DB.

This adds \`POST /api/accounts/{account_id}/probe\` (dashboard auth). The endpoint:

1. Looks up the account (404 if missing, 409 with code \`account_not_probable\` if status is \`paused\` / \`deactivated\`).
2. Decrypts the account access token and sends one minimal \`responses.create\` directly to \`{upstream_base_url}/codex/responses\` (\`stream=true\`, \`store=false\`, \`max_output_tokens=1\`). The call **bypasses load-balancer scoring** (does not touch \`LoadBalancer.select_account\`); this is the whole point — we are intentionally pinning a single request to one specific account.
3. Triggers an immediate \`UsageUpdater.force_refresh\` so the post-probe \`/wham/usage\` snapshot is persisted.
4. Reloads account + latest primary/secondary usage and returns the before/after snapshot:

\`\`\`json
{
  \"status\": \"probed\",
  \"accountId\": \"...\",
  \"probeStatusCode\": 200,
  \"primaryUsedPercentBefore\": 100.0,
  \"primaryUsedPercentAfter\": 0.0,
  \"secondaryUsedPercentBefore\": 58.0,
  \"secondaryUsedPercentAfter\": 58.0,
  \"accountStatusBefore\": \"rate_limited\",
  \"accountStatusAfter\": \"active\"
}
\`\`\`

Notes:

- The access token is **never logged** (covered by a regression test).
- \`AuditService.log_async(\"account_probed\", ...)\` records every successful probe alongside the existing \`account_created\` / \`account_deleted\` audit entries.
- Network failures return \`probe_status_code: 0\` as a sentinel so the value is distinguishable from any real HTTP status the upstream might return.
- Uses \`UsageRepository.latest_entry_for_account(account_id, window=...)\` for single-account snapshots (not the dict-returning \`latest_by_account\`) — efficient single-row lookup.
- **No frontend / dashboard button in this PR** — the contextual UI is intentionally deferred to a follow-up so the API surface lands cleanly with a focused review. The button work is staged at AnobleSCM:#6 and can be opened upstream once this one merges.

## Test plan

- [x] \`uv run pytest tests/unit/test_accounts_service_probe.py tests/integration/test_accounts_api_probe.py -q\` — 11 passed
- [x] \`uv run ruff check app/modules/accounts tests/unit/test_accounts_service_probe.py tests/integration/test_accounts_api_probe.py\` — clean
- [x] Unit tests cover: missing account → \`None\` path, paused / deactivated rejection, before/after snapshot capture, default-model fallback, access-token-never-logged invariant, network-failure status surfaced as \`0\` sentinel
- [x] Integration tests cover: 404 for missing, 409 for paused, 200 with snapshot for active, default model when body omitted

OpenSpec change folder: \`openspec/changes/add-account-probe-endpoint/\` (extends the \`usage-refresh-policy\` capability spec).

Refs #676 / #677 / #678.

Part of #677 (admin API endpoint only; dashboard button remains follow-up). Refs #676 and #678.


Komzpa:
@codex review

Soju06:
Confirming the codex review finding at \`app/modules/accounts/service.py:335\` is real and worth addressing before merge.

The shared client built in [\`app/core/clients/http.py\`](https://github.com/Soju06/codex-lb/blob/main/app/core/clients/http.py) sets \`trust_env=True\` explicitly on the main session for exactly this reason — operators behind \`HTTP_PROXY\` / \`HTTPS_PROXY\` (corporate networks, Tailscale exit nodes, egress proxies, etc.) need outbound requests to honor those env vars. \`aiohttp.ClientSession()\` defaults to \`trust_env=False\`, so the bare session created in \`_send_probe_request\` will silently bypass the env-proxy path while every other upstream call in codex-lb (\`fetch_usage\`, the proxy service, etc.) still works.

In those deployments the probe endpoint will appear permanently broken — always returning \`probeStatusCode: 0\` with the network-failure sentinel — even though the operator's other upstream traffic is fine. That makes the failure mode hard to triage.

Suggested fix: route the probe through the shared client instead of a one-off session. Two reasonable shapes:

1. Reuse \`get_http_client().session\` directly:

   \`\`\`python
   from app.core.clients.http import get_http_client

   client = get_http_client().session
   try:
       async with client.post(url, headers=headers, json=body, timeout=timeout) as resp:
           return resp.status
   except (aiohttp.ClientError, asyncio.TimeoutError) as exc:
       ...
   \`\`\`

2. Or, if you want a dedicated short-lived session for the probe (some operators may want it isolated from the retry-pooled client), at minimum pass \`trust_env=True\` so env proxies still work:

   \`\`\`python
   async with aiohttp.ClientSession(trust_env=True) as session:
       ...
   \`\`\`

Option 1 is preferred — it matches how \`fetch_usage\` and the rest of the upstream surface already work, gets connection pooling for free, and means we never have to remember to set \`trust_env\` again on new upstream call sites. If you go this way please also drop the import of \`aiohttp.ClientSession\` if it becomes unused.

A targeted test that asserts the probe uses the shared client (e.g. by patching \`get_http_client\` and verifying the patched session was called) would also prevent regression here.

Otherwise the PR looks good — bypass-selection design is exactly what #677 was asking for, the audit log entry and "access token never logged" invariant are nice touches.

Soju06:
Confirming the second codex review finding at \`app/modules/accounts/service.py:262\` is also real. Walked the freshness gate end-to-end.

\`refresh_accounts\` ([updater.py:211](https://github.com/Soju06/codex-lb/blob/main/app/modules/usage/updater.py#L211)) does:

\`\`\`python
latest = latest_usage.get(account.id)
bypass_freshness = _quota_recovery_should_bypass_freshness(account, latest=latest)
if not bypass_freshness and _latest_usage_is_fresh(latest, now=now, interval_seconds=interval):
    continue
\`\`\`

And \`_quota_recovery_should_bypass_freshness\` ([updater.py:754](https://github.com/Soju06/codex-lb/blob/main/app/modules/usage/updater.py#L754)) only returns True when \`account.status == AccountStatus.QUOTA_EXCEEDED\` with cooldown expired. Every other status — including the common cases \`rate_limited\` (the one the probe is mainly designed for) and \`active\` — does **not** bypass.

So the failure mode is: any account whose primary \`usage_history\` row was written within \`usage_refresh_interval_seconds\` (default 60s, and the periodic refresh hits this aggressively) → probe call goes out fine → \`refresh_accounts\` skips the fetch entirely → \`latest_entry_for_account\` returns the same pre-probe row → the response's \`primary_used_percent_after\` equals \`primary_used_percent_before\` and \`accountStatusAfter\` equals \`accountStatusBefore\` even when the probe successfully woke the limiter upstream.

For the \`rate_limited\` reset-event case the bot called out, that's the entire happy path — the probe will look "ineffective" to the operator most of the time.

Also worth noting: even passing \`latest_usage={}\` doesn't fix this. The \`latest is None\` branch in \`refresh_accounts\` then checks \`_last_successful_refresh.get(account.id)\` ([updater.py:241](https://github.com/Soju06/codex-lb/blob/main/app/modules/usage/updater.py#L241)) and respects the same interval gate. So you can't sidestep the gate by lying about freshness either.

### Suggested fix

The clean way is a dedicated forced-refresh path that bypasses both gates. Two reasonable shapes:

1. **Expose a \`force=True\` parameter on \`refresh_accounts\`** that skips both \`_latest_usage_is_fresh\` and the \`_last_successful_refresh\` interval check. Single-account semantics, easy to test.

   \`\`\`python
   await self._usage_updater.refresh_accounts([account], {}, force=True)
   \`\`\`

2. **Add a single-account \`force_refresh(account)\` entry point** on \`UsageUpdater\` that calls \`_refresh_account\` directly. The internal \`_refresh_account\` already does the right thing (token decrypt, 401 retry, deactivate-on-client-error, plan sync, usage write). Wrapping it for external callers keeps the bulk path untouched.

   \`\`\`python
   await self._usage_updater.force_refresh(account)
   \`\`\`

Either works; option 2 is slightly cleaner because the probe never goes through bulk-loop bookkeeping that wasn't meant for single-account refresh, and it keeps \`refresh_accounts\` semantics intact (bulk + interval-gated, used by the periodic loop and the accounts service \`upsert\` flow).

### Test to pin this

Worth adding a regression test where the periodic loop has already populated \`_last_successful_refresh\` / a fresh primary row for the account being probed, then asserting that the probe-triggered refresh **still** hits \`fetch_usage\`. Without this test, future refactors of the freshness gate could re-introduce the same silent-skip behavior.

This is more impactful than the \`trust_env\` finding from the other review thread, because in the proxy case the probe at least loudly returns \`probeStatusCode: 0\`. In this case the probe appears to succeed (\`probeStatusCode: 200\`) while the snapshot quietly lies, which is the worse failure mode for an operator-facing diagnostic action.

Komzpa:
@codex review

Soju06:
@codex review

Soju06:
@codex review

Soju06:
@codex review

Soju06:
@codex review

Soju06:
@codex review

Soju06:
@codex review

Soju06:
@codex review

Soju06:
Merge-gate hold: implementation, tests, OpenSpec validation, CI, and Codex review look good, but the PR body only has `Refs #676 / #677 / #678` and `closingIssuesReferences` is empty.

Because this adds a behavior-changing admin endpoint and appears to address those tracked issues, please add the appropriate `Fixes #...` / `Closes #...` reference for the issue(s) this PR fully resolves, or clarify that the refs are intentionally partial. Code-wise I did not find a blocker.

Soju06:
@codex review

Komzpa:
@codex review

Soju06:
@codex review

Soju06:
Hermes blocker on current head (`198cdf4`): the probe still does not immediately recover the main `rate_limited` happy path that #677 is meant to unlock.

`AccountsService.probe_account()` sends the direct probe and then calls `UsageUpdater.force_refresh(account)` (`app/modules/accounts/service.py:323-334`). `force_refresh()` writes fresh `/wham/usage` rows through `_refresh_account()` (`app/modules/usage/updater.py:317-329`), but `_refresh_account()` only calls `_recover_quota_status_from_usage()` and that recovery helper returns immediately unless `account.status == QUOTA_EXCEEDED` (`app/modules/usage/updater.py:496`, `app/modules/usage/updater.py:541`). It never applies the existing `RATE_LIMITED -> ACTIVE` recovery logic from `reconcile_recoverable_account_statuses()`.

I verified this with the current code: a `rate_limited` account with a forced post-probe usage payload of `primary=0.0`, `secondary=10.0` writes both fresh usage rows but remains `rate_limited` with no status update. That contradicts the issue/API example (`account_status_before: "rate_limited"`, `account_status_after: "active"`) and leaves the selector excluding the account until a later background scheduler reconciliation happens.

Suggested fix: after the forced usage refresh, run the same recoverable-status reconciliation for this account (or make `_refresh_account()` handle `RATE_LIMITED` with the same policy as `background_recovery_state_from_account()`), and add a regression where `POST /api/accounts/{id}/probe` on a `rate_limited` account whose refreshed primary usage is below 100 returns/persists `accountStatusAfter: "active"`.

Also note there is already an unresolved current-head Codex thread about refreshing stale account tokens before sending the probe; I am not duplicating that one here, but it remains a separate blocker.

Soju06:
@codex review

Soju06:
@codex review

Komzpa:

Current #865 status:

Issue #624, opened 2026-05-14
Add tests for runtime connect-address resolution


whoisjayd:
## Suggested labels
`test`, `backend`, `settings`, `good first issue`

## Confidence / impact / size
- Confidence: Medium
- Impact: Medium
- Size: Small
- OpenSpec needed: No unless expected behavior changes.

## Summary
The runtime connect-address endpoint determines what address the dashboard should show clients. It has nuanced request-header behavior but no obvious focused tests.

## Evidence
- `app/modules/settings/api.py:48` defines `_resolve_runtime_connect_address(request: Request) -> str`.
- `app/modules/settings/api.py:95-97` exposes `GET /runtime/connect-address`.
- The scan did not find focused tests for this helper/endpoint in `tests/integration/test_settings_api.py`.

## Problem
Connect-address calculation can be sensitive to host headers, forwarded headers, local vs remote access, and deployment topology. Regressions here can produce incorrect setup instructions in the dashboard.

## Proposed tests
- Direct request with a normal `Host` header returns the expected address.
- Forwarded host/proto behavior matches the intended trusted-proxy configuration, if supported by the helper.
- Missing or unusual host values fall back safely according to current behavior.

## Acceptance criteria
- [ ] Tests cover the public endpoint or helper for representative request header cases.
- [ ] No production behavior changes unless a bug is uncovered.
- [ ] Tests document current intended behavior clearly.

## Suggested verification
```bash
uv run pytest tests/integration/test_settings_api.py -q
uvx ruff check .
uv run ty check
```

## Contributor note
Read `_resolve_runtime_connect_address` carefully before writing tests and assert current behavior, not guessed behavior.

Komzpa:
Closed with #712

PR #712, opened 2026-05-19
test(settings): add coverage for runtime connect-address resolution

ozpool:
## Summary

\`_resolve_runtime_connect_address\` in \`app/modules/settings/api.py\` backs the \`/api/settings/runtime/connect-address\` endpoint that the dashboard surfaces to operators as "the address to point clients at". Resolution is small but nuanced (env override / non-loopback IPv4 / hostname DNS / loopback fallback) and a regression here produces wrong setup instructions in the UI. There were no focused tests for the helper or the endpoint before.

Added \`tests/unit/test_settings_connect_address.py\` with 14 parametrized cases exercising every documented branch:

- \`CODEX_LB_CONNECT_ADDRESS\` env override wins over request host
- Whitespace-only env value is treated as unset
- Non-loopback IPv4 host is returned verbatim (no DNS call)
- Hostname that resolves to a non-loopback IPv4 returns the resolved IP
- DNS that returns \`127.0.0.1\` alongside a real IP picks the real IP
- DNS that returns only loopback IPs falls back to the hostname
- Unresolvable hostname falls back to the hostname
- Loopback hosts (\`localhost\`, \`127.0.0.1\`, \`::1\`, \`[::1]\`) + missing / empty host all return the \`<codex-lb-ip-or-dns>\` placeholder

\`socket.getaddrinfo\` is monkeypatched on the settings module so the suite is fully offline. Pure test addition — no production behaviour change.

Refs #624.

## Test plan
- [x] \`uv run --frozen pytest tests/unit/test_settings_connect_address.py\` — 14 passed
- [x] Full unit suite still green (1825 passed, 39 skipped locally)

### DIFF
--- a/tests/unit/test_settings_connect_address.py
+++ b/tests/unit/test_settings_connect_address.py
@@ -1,0 +1,146 @@
+"""Tests for `_resolve_runtime_connect_address` (issue #624).
+
+The helper feeds the `/api/settings/runtime/connect-address` endpoint
+which the dashboard surfaces to operators as "the address to point
+clients at". Bugs here produce wrong setup instructions in the UI, so
+the resolution rules deserve focused coverage even though the helper
+itself is small.
+
+Behaviour exercised (mirrors `app/modules/settings/api.py:48-63`):
+
+  1. The `CODEX_LB_CONNECT_ADDRESS` env override wins over everything.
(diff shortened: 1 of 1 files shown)


Komzpa:
@codex review

PR #729, opened 2026-05-20
fix(proxy): replay pre-visible websocket drops

Komzpa:
## Summary
- keep websocket connect/refresh transport failures on the deterministic account-failover path instead of surfacing the first account timeout
- avoid misclassifying inner upstream websocket connect timeouts as proxy request budget exhaustion unless the outer proxy budget elapsed
- replay a request once when upstream closes after `response.created` but before any visible output, while still refusing replay after visible output or ambiguous response events

## Testing
- `uv run pytest tests/unit/test_proxy_utils.py -q -k 'websocket_connect or connect_proxy_websocket or open_upstream_websocket or pop_replayable_created or pop_replayable_precreated or replays_precreated_request_after_upstream_close_race'`
- `uv run pytest tests/unit/test_proxy_utils.py::test_proxy_responses_websocket_replays_precreated_request_after_upstream_close_race -q`
- `uv run ruff check app/modules/proxy/service.py tests/unit/test_proxy_utils.py`
- `git diff --check origin/main..HEAD`

### DIFF
--- a/app/cli.py
+++ b/app/cli.py
@@ -17,10 +17,10 @@ def _parse_args() -> argparse.Namespace:
     parser.add_argument(
         "--timeout-keep-alive",
         type=int,
-        default=int(os.getenv("UVICORN_TIMEOUT_KEEP_ALIVE", "300")),
+        default=int(os.getenv("UVICORN_TIMEOUT_KEEP_ALIVE", "7200")),
         help=(
             "Seconds to keep idle HTTP connections open. Codex CLI reuses local "
-            "connections for compact POSTs; uvicorn's 5s default can leave the "
+            "connections for large compact POSTs; short keepalive windows can leave the "
             "client writing to a stale socket before the request reaches the app."
         ),
     )
--- a/app/core/clients/proxy.py
+++ b/app/core/clients/proxy.py
@@ -2241,6 +2241,18 @@ async def _stream_responses_with_session(
                 ),
             )
             return
+        total_elapsed_seconds = max(0.0, now - pre_request_started_at)
+        if request_total_timeout is None or total_elapsed_seconds < request_total_timeout:
+            error_code = "upstream_unavailable"
+            error_message = str(exc) or "Request to upstream timed out"
+            yield format_sse_event(
+                response_failed_event(
+                    "upstream_unavailable",
+                    error_message,
+                    response_id=get_request_id(),
+                ),
+            )
+            return
         error_code = "upstream_request_timeout"
         error_message = "Proxy request budget exhausted"
         yield format_sse_event(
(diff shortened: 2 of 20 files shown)


Komzpa:
@codex review

Komzpa:
@codex review

Soju06:
@codex review

Soju06:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Soju06:
@codex review

Soju06:
@codex review

Komzpa:
@codex review

Soju06:
Thanks for the iteration here. I found two blockers that should be resolved before merge even though CI and the Codex label are currently green.

## 1. Created-only websocket replay can break downstream response-id consistency

`_websocket_request_can_replay_before_visible_output()` allows replay after `response.created` has already been observed, as long as no visible output was sent yet:

- `app/modules/proxy/service.py:10883-10906`

But `_prepare_websocket_request_state_for_visible_output_replay()` then clears the known response id and leaves created-event suppression disabled:

- `app/modules/proxy/service.py:10925-10930`

The downstream suppression hook exists only when `suppress_next_created_downstream` is true:

- `app/modules/proxy/service.py:6964-6966`

I could not find any assignment setting `suppress_next_created_downstream = True` for this replay path.

Impact: a client can receive `response.created(resp_old)`, then after transparent replay receive another `response.created(resp_new)` for the same logical downstream stream. If replay/reconnect fails after the state mutation, the terminal failure can also lose the already-emitted `resp_old` id because `response_id` was cleared.

Minimal state-transition check:

```text
can_replay= True
prepared= True
response_id= None
suppress_next_created_downstream= False
awaiting_response_created= True
```

Please either preserve the original downstream response id through the replay, or explicitly suppress/rewrite the replayed `response.created` and keep terminal events id-consistent. This also needs a focused regression test for the “created emitted, no visible output, replayed upstream emits created/then terminal or replay fails” path.

## 2. Conversation archive now drops records, conflicting with the current OpenSpec SSOT

The current spec says the opt-in archive must preserve full upstream conversation traffic and apply synchronous write backpressure when the writer queue is saturated:

- `openspec/specs/proxy-runtime-observability/spec.md:29-30`

This PR changes `_enqueue_record()` to drop records when byte-budget reservation fails:

- `app/core/conversation_archive.py:181-192`

That is a behavior/operability change from non-lossy archive backpressure to lossy archive sampling/drop under pressure, but the active OpenSpec change only covers timeout classification / multiplexed pending preservation / SSE liveness.

Please either restore the non-lossy backpressure semantics, or update the relevant OpenSpec requirement to explicitly allow lossy archive drops and document the audit-fidelity tradeoff.

## Notes

I also verified the current PR gates are green (`CLEAN/MERGEABLE`, 21 checks complete, no failed/pending checks, `🤖 codex: ok`). These two issues are semantic/spec blockers rather than CI failures.

Soju06:
@codex review

Soju06:
@codex review

Soju06:
@codex review

Soju06:
@codex review

Komzpa:
@codex review

Soju06:
@codex review

PR #744, opened 2026-05-21
fix(codex_version): fall back to npm registry when GitHub is rate-limited

### LINKED ISSUE #664: Stale 'client_version' fallback causes 503 'no_plan_support_for_model' for newer models (e.g. gpt-5.5) when GitHub releases API rate-limits
## Summary

When `app/core/clients/codex_version.py` cannot fetch the latest codex release from `api.github.com/repos/openai/codex/releases/latest` (e.g. anonymous-rate-limited with HTTP 403), it logs:

```
GitHub releases API returned HTTP 403
GitHub fetch failed and no cached version; falling back to settings default 0.101.0
```

and falls back to `Settings.model_registry_client_version` (default `0.101.0`). The model registry refresh (`app/core/openai/model_refresh_scheduler.py` → `clients/model_fetcher.fetch_models_for_plan`) then asks upstream `/codex/models?client_version=0.101.0`. Upstream gates newer model slugs behind newer client versions, so the response excludes e.g. `gpt-5.5`.

As a result, `ModelRegistry.model_plans` never gets a `gpt-5.5` entry, `_filter_accounts_for_model` in `modules/proxy/load_balancer.py:1168` returns an empty set, and every Codex request for `gpt-5.5` is rejected with:

```
proxy_error_response status=503 code=no_plan_support_for_model
  message=No accounts with a plan supporting model 'gpt-5.5'
```

…even though the actual Codex CLI is on `0.130.0` and the account's ChatGPT subscription does support the model (calling `/backend-api/codex/models?client_version=0.130.0` through the LB returns 200 with the model present).

## Reproduction

1. Run `ghcr.io/soju06/codex-lb:1.17.0` on a host where anonymous GitHub API requests are 403'ed.
2. Connect a current Codex CLI (e.g. `@openai/codex@0.130.0`) configured with `model = \"gpt-5.5\"`.
3. Observe startup log line:
   - `GitHub fetch failed and no cached version; falling back to settings default 0.101.0`
4. Every `POST /backend-api/codex/responses` returns 503 with code `no_plan_support_for_model`.

Setting `CODEX_LB_MODEL_REGISTRY_CLIENT_VERSION=0.130.0` and recreating the container fully resolves the symptom.

## Suggested fixes

A few low-effort options, any one would help:

1. **Bump the baked-in default `model_registry_client_version` on each release** so it tracks the recent `@openai/codex` version (or read it from a generated file at build time).
2. **Add a secondary version source.** `npm view @openai/codex version` against `https://registry.npmjs.org` is not rate-limited the way the GitHub anonymous API is, and a static fetch of `https://registry.npmjs.org/@openai/codex/latest` returns JSON containing `\"version\"`.
3. **Cache the fetched version on disk** (e.g. inside the data volume) so a single successful lookup survives container restarts and survives transient 403s for the cache TTL.
4. **Honor the client's own `client_version` query string** as a hint for which model list to refresh (only when greater than the configured default), so a newer client effectively bootstraps the registry.

## Workaround

For anyone hitting this, pass `-e CODEX_LB_MODEL_REGISTRY_CLIENT_VERSION=<your-codex-cli-version>` to `docker run`:

```sh
docker run -d \\
  -e CODEX_LB_DASHBOARD_BOOTSTRAP_TOKEN=... \\
  -e CODEX_LB_MODEL_REGISTRY_CLIENT_VERSION=0.130.0 \\
  -p 2455:2455 -p 1455:1455 \\
  -v codex-lb-data:/var/lib/codex-lb \\
  ghcr.io/soju06/codex-lb:1.17.0
```

## Environment

- codex-lb image: `ghcr.io/soju06/codex-lb:1.17.0`
- Codex CLI: `@openai/codex@0.130.0`
- Host: macOS, Docker Desktop / Colima
- Account plan: `pro` (3 accounts, all `plan_type=pro` in `store.db`)

Related: #480 ("Model metadata for `gpt-5.5` not found. Defaulting to fallback metadata") looks like the same underlying cause.

ozpool:
## Summary

Adds the npm registry (`https://registry.npmjs.org/@openai/codex/latest`) as a secondary upstream source for `CodexVersionCache`, so a 403 from the anonymous GitHub API no longer poisons the model registry into rejecting any model gated behind a newer Codex client version.

## Why

When `app/core/clients/codex_version.py` cannot reach
`api.github.com/repos/openai/codex/releases/latest` (e.g. anonymous
rate-limit on a shared VPS / Docker host), it falls back to the baked-in
`Settings.model_registry_client_version` (currently `0.101.0`). The
model refresh then calls
`/codex/models?client_version=0.101.0`, which excludes any slug
upstream gates behind newer versions (e.g. `gpt-5.5`), and every proxy
request for those models is rejected with:

```
proxy_error_response status=503 code=no_plan_support_for_model
  message=No accounts with a plan supporting model 'gpt-5.5'
```

…even though the user is on a current Codex CLI and the account does
support the model. The reporter confirmed that bumping
`CODEX_LB_MODEL_REGISTRY_CLIENT_VERSION` manually fully resolves the
symptom.

## Approach

`https://registry.npmjs.org/@openai/codex/latest` is the cleanest alternate
source (option 2 in the issue) — it is not anonymously rate-limited the
way the GitHub API is and tracks the same release cadence (the npm
publish ships with the GitHub release). Order is now:

1. GitHub releases (existing primary source).
2. npm registry (new secondary source).
3. Stale cached value, if any.
4. Settings default (`model_registry_client_version`).

`_fetch_from_github` and `_fetch_from_npm` share the same aiohttp
session and 10s total timeout, and both apply the strict
`^\d+\.\d+\.\d+$` version regex so prerelease tags from npm (e.g.
`0.130.0-rc.1`) cannot poison the cache.

## Tests

`uv run pytest tests/unit/test_codex_version.py -x` — 17 passed.
Full unit suite: `uv run pytest tests/unit -x -q` — 1895 passed, 39 skipped.

New cases in `tests/unit/test_codex_version.py`:

- GitHub 403 → npm wins.
- GitHub returns invalid name → npm wins.
- GitHub 403 + npm prerelease version → settings default.
- GitHub 403 + npm payload missing `version` field → settings default.
- GitHub OK → npm is not consulted.
- GitHub failure + npm failure + previously populated cache → stale cache.

Updated the failure-path warning message to reflect the multi-source
fallback. Existing single-source tests still pass because they share
one mock response across all calls, which matches both the original
single-source behaviour when GitHub succeeds and the multi-source no-op
when it fails.

## Test plan

- [x] `uv run pytest tests/unit/test_codex_version.py -x`
- [x] `uv run pytest tests/unit -x -q`
- [x] `uv run ruff check app/core/clients/codex_version.py tests/unit/test_codex_version.py`
- [x] `uv run ruff format --check app/core/clients/codex_version.py tests/unit/test_codex_version.py`

Closes #664

### DIFF
--- a/app/core/clients/codex_version.py
+++ b/app/core/clients/codex_version.py
@@ -12,6 +12,7 @@ from app.core.config.settings import get_settings
 logger = logging.getLogger(__name__)
 
 _GITHUB_RELEASES_URL = "https://api.github.com/repos/openai/codex/releases/latest"
+_NPM_REGISTRY_URL = "https://registry.npmjs.org/@openai/codex/latest"
 _FETCH_TIMEOUT_SECONDS = 10.0
 _VERSION_RE = re.compile(r"^\d+\.\d+\.\d+$")
 
@@ -43,12 +44,18 @@ class CodexVersionCache:
 
             # Fallback: stale cache value
             if self._cached_version is not None:
-                logger.warning("GitHub fetch failed; using stale cached version %s", self._cached_version)
+                logger.warning(
+                    "Upstream version sources failed; using stale cached version %s",
+                    self._cached_version,
+                )
                 return self._cached_version
 
             # Fallback: settings default
             fallback = get_settings().model_registry_client_version
-            logger.warning("GitHub fetch failed and no cached version; falling back to settings default %s", fallback)
+            logger.warning(
+                "Upstream version sources failed and no cached version; falling back to settings default %s",
+                fallback,
+            )
             return fallback
 
     async def invalidate(self) -> None:
@@ -57,15 +64,30 @@ class CodexVersionCache:
             self._cached_at = 0.0
 
     async def _fetch_latest_version(self) -> str | None:
+        timeout = aiohttp.ClientTimeout(total=_FETCH_TIMEOUT_SECONDS)
         try:
-            timeout = aiohttp.ClientTimeout(total=_FETCH_TIMEOUT_SECONDS)
             async with aiohttp.ClientSession(timeout=timeout, trust_env=True) as session:
-                headers = {"Accept": "application/vnd.github+json"}
-                async with session.get(_GITHUB_RELEASES_URL, headers=headers) as resp:
-                    if resp.status != 200:
-                        logger.warning("GitHub releases API returned HTTP %d", resp.status)
-                        return None
-                    data = await resp.json(content_type=None)
+                version = await self._fetch_from_github(session)
+                if version is not None:
+                    return version
+
+                version = await self._fetch_from_npm(session)
+                if version is not None:
+                    return version
+        except Exception:
+            logger.warning("Failed to fetch latest Codex release from upstream sources", exc_info=True)
+            return None
+
+        return None
+
+    async def _fetch_from_github(self, session: aiohttp.ClientSession) -> str | None:
+        try:
+            headers = {"Accept": "application/vnd.github+json"}
+            async with session.get(_GITHUB_RELEASES_URL, headers=headers) as resp:
+                if resp.status != 200:
+                    logger.warning("GitHub releases API returned HTTP %d", resp.status)
+                    return None
+                data = await resp.json(content_type=None)
         except Exception:
             logger.warning("Failed to fetch latest Codex release from GitHub", exc_info=True)
             return None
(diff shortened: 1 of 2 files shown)


Soju06:
@codex review

PR #751, opened 2026-05-21
fix(proxy): make Codex Spark quota gating plan-aware

Komzpa:
## Summary
Revives #513 on a maintainer-owned branch because the original PR head has `maintainerCanModify=false`.

Credit: original implementation by @nguyendkn in #513; this branch rebases and updates that work on current `main`.

- Keeps `codex_spark` in the quota registry, but makes model quota eligibility plan-aware.
- Allows Plus-like plans that do not provide additional quota data while keeping Pro/Team-style accounts fail-closed when their additional quota data is missing or stale.

## Validation
- Focused backend/unit/integration suite passed: 54 passed, 3 skipped
- `uv run ruff check ...` passed
- `uv run ty check ...` passed
- `git diff --check origin/main...HEAD` passed

Revives #513


Soju06:
@codex review

PR #572, opened 2026-05-09
feat(proxy): route cyber-flagged work to authorized accounts

Komzpa:
Draft stacked split from #567/#572 cleanup.

## Purpose
- Add a persistent account flag for Trusted Access for Cyber / security-work authorization, with dashboard toggle support.
- Filter account selection to security-authorized accounts when retrying a cyber-flagged upstream response.
- Detect the upstream cyber warning before surfacing it to clients and retry once on the authorized pool.
- Keep `no_security_work_authorized_accounts` advisory: when the authorized pool is empty, fall back to normal account selection instead of surfacing it as a terminal proxy error.
- Preserve sticky failover behavior after a retryable upstream rate-limit response by excluding the failed account from the next stream attempt.
- Add Alembic migration, backend/frontend schemas, and focused regression tests.

## Validation
- Head: `142142a0328a2b64a09fbd27babcd2bb2f5ae32c`.
- Exact-head GitHub CI is green.
- Current-head `@codex review` has been requested after the rebuild and is pending.
- `uv run codex-lb-db --db-url sqlite:////tmp/codex-lb-pr572-check.db upgrade head && uv run codex-lb-db --db-url sqlite:////tmp/codex-lb-pr572-check.db check`
- `uv run pytest tests/unit/test_proxy_utils.py tests/unit/test_proxy_load_balancer_refresh.py tests/unit/test_load_balancer.py tests/integration/test_proxy_websocket_responses.py` (397 passed, 3 skipped)
- `uv run ruff check app/modules/proxy app/modules/accounts app/modules/config app/db tests/unit/test_proxy_utils.py tests/unit/test_proxy_load_balancer_refresh.py tests/unit/test_load_balancer.py tests/integration/test_proxy_websocket_responses.py`
- `uv run ty check app/modules/proxy app/modules/accounts app/modules/config app/db tests/unit/test_proxy_utils.py tests/unit/test_proxy_load_balancer_refresh.py tests/unit/test_load_balancer.py tests/integration/test_proxy_websocket_responses.py`

## Relationship
- Rebuilt directly on current `main` after the prerequisite stack slices were merged.
- #567 remains Draft and should not be merged as one large PR.
- This branch carries only the security-work authorization slice and the retry failover repair needed by that slice.

PR #761, opened 2026-05-21
feat(accounts): add manual routing policy

Komzpa:
## Summary
Revives #553 on a maintainer-owned branch because the original PR head has `maintainerCanModify=false`.

Credit: original implementation by @Komzpa in #553; this branch rebases and updates that work on current `main`.

- Adds manual account routing policy controls across backend/frontend/settings paths.
- Aligns migration history with current `main`, keeps settings override lookup isolated from dashboard DB state, and updates frontend trend schema defaults.

## Validation
- `uv run ruff check ...` passed
- `uv run pytest -q ...` passed: 164 passed, 3 skipped
- `cd frontend && bun run test -- src/features/accounts src/features/settings` passed: 76 passed
- `git diff --check` passed
- `make migration-check` passed: `migration_policy=ok`, `schema_drift=none`
- `openspec validate --specs` passed: 21 passed

## OpenSpec
- Manual routing policy requirements remain included and valid.

Revives #553

PR #716, opened 2026-05-19
feat(proxy): restore opportunistic traffic class on main

Komzpa:
## Summary

This is the consolidated routing/quota/account-dashboard branch rebased onto current `main`.

The branch keeps the original opportunistic traffic-class restoration and folds the overlapping routing/quota work that was otherwise split across several open PRs. The goal is one reviewable main-based PR for account identity, additional-quota routing, sticky/fill-first selection, manual account routing policy, reset-window preference, single-account/drain behavior, security-work account authorization, live credits display, and quota phase planning.

Current head: `3a8c639b41793a33bbb56b926ada4300d32d3fa3`
Base head: `47b0c3624faf9c40211cbf52957d91ddce4bb5cb`
Current CI: https://github.com/Soju06/codex-lb/actions/runs/26843390719

Folded here:
- https://github.com/Soju06/codex-lb/pull/659 - split sticky primary/secondary thresholds
- https://github.com/Soju06/codex-lb/pull/716 - opportunistic traffic class on main
- https://github.com/Soju06/codex-lb/pull/831 - fill-first routing
- https://github.com/Soju06/codex-lb/pull/855 - additional-quota bypass / gated-model routing
- https://github.com/Soju06/codex-lb/pull/863 - drain and single-account routing
- https://github.com/Soju06/codex-lb/pull/871 - reset-window routing preference
- https://github.com/Soju06/codex-lb/pull/872 - per-quota manual account routing policy
- https://github.com/Soju06/codex-lb/pull/762 - security-work authorized account routing
- https://github.com/Soju06/codex-lb/pull/767 - live credits display
- https://github.com/Soju06/codex-lb/pull/737 - quota phase planner

Issue relationships:
- Addresses part of #428 via the folded per-quota manual account routing policy from #872. The remaining arbitrary ordered-priority / drag-and-drop fallback semantics stay open for a follow-up.

Kept separate:
- https://github.com/Soju06/codex-lb/pull/832 - account display/API enhancement
- https://github.com/Soju06/codex-lb/pull/703 - dashboard guest read-only behavior
- https://github.com/Soju06/codex-lb/pull/865 - separate account/API surface work
- https://github.com/Soju06/codex-lb/pull/870 - separate quota/auth surface work
- https://github.com/Soju06/codex-lb/pull/873 - separate follow-up surface work

Latest rebase/follow-up fixes:
- Merged current `origin/main` into the PR branch and resolved backend/frontend/test conflicts.
[...]

PR #762, opened 2026-05-21
feat(proxy): route security work to authorized accounts

Komzpa:
## Summary
Revives #572 on a maintainer-owned branch because the original PR head has `maintainerCanModify=false`.

Credit: original implementation by @Komzpa in #572; this branch rebases and updates that work on current `main`.

- Adds security/cyber-flagged account routing with an account-level authorization toggle and proxy fallback advisory behavior.
- Removes old stack bleed, preserves additional quota flags for security routing, and fixes websocket gate/error handling around security replays.

## Validation
- `openspec validate add-security-work-account-routing --strict` passed
- `uv run ruff check ...` passed
- `uv run ty check app/modules/accounts app/modules/proxy` passed
- `uv run pytest ... -k 'security_work or security_retry'` passed
- `bun run test src/features/accounts/components/accounts-page.test.tsx` passed
- `uv run python -m app.db.migrate ... upgrade && ... check` passed: `migration_policy=ok`, `schema_drift=none`
- `git diff --check origin/main..HEAD` passed

## OpenSpec
- `add-security-work-account-routing` remains included and valid.

Revives #572


Soju06:
@codex review

KakatkarAkshay:
Maybe worth considering: for normal work, we could prefer regular accounts first and save the security-authorized accounts for requests that actually need them. That might help preserve usage on those accounts.

Komzpa:
@KakatkarAkshay I think it's best to manage it separately, see #761 - maybe it's authorized but you plan to get rid of it so burn first, or something like that.

KakatkarAkshay:
> @KakatkarAkshay I think it's best to manage it separately, see #761 - maybe it's authorized but you plan to get rid of it so burn first, or something like that.

Ah, I hadn’t seen #761 earlier. That makes sense, keeping this policy separate there sounds good.

Soju06:
@codex review

Komzpa:
@codex review

Soju06:
@codex review

Komzpa:
@codex review

Soju06:
@codex review

Soju06:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Soju06:
Hermes blocker review for current head `f944ecd57f5b710f9512e5e01c2300ecd66f301c`:

`app/modules/proxy/service.py` still lets direct HTTP streamed Responses security-work retry move a `previous_response_id` request to another account when owner resolution misses.

Evidence:
- `service.py:9876-9883` only sets `require_preferred_account=True` when `_resolve_websocket_previous_response_owner()` returns an account. If a `previous_response_id` is present but the owner lookup misses, the request is no longer treated as pinned.
- The security-work retry branches at `service.py:10301-10325`, `service.py:10419-10449`, and `service.py:10681-10704` only guard `require_preferred_account` / file-pinning, not `payload.previous_response_id is not None`.
- The retry then calls `_stream_once(account, payload, ...)` with the original payload at `service.py:10217-10223`, so the old `previous_response_id` is sent to the newly selected security-authorized account.
- The OpenSpec delta says security-work retry is allowed only when the request can be safely replayed, and that previous-response-pinned requests must not move accounts (`openspec/changes/add-security-work-account-routing/specs/responses-api-compat/spec.md:3-25`). WebSocket has the needed fresh-payload guard (`service.py:8758-8764`), but the direct HTTP stream path does not.

Please fail closed for any direct HTTP stream request with `payload.previous_response_id` unless the retry path has an explicitly safe fresh/full-resend payload and removes the anchor before selecting another account. Add a regression test for this direct stream path; the existing previous-response security tests cover HTTP-bridge/WebSocket helpers, not this `_stream_once(payload)` path.

Soju06:
@codex review

Soju06:
@codex review

Soju06:
@codex review

Komzpa:
Folded into #716.

#716 is the surviving `main`-based routing/quota branch for this bucket. It includes the routing/quota changes from this PR, has current-head CI green, and has a clean current-head Codex review, so this separate PR can be closed to keep the queue reviewable.

PR #772, opened 2026-05-21
fix(proxy): allow larger compressed responses bodies

Komzpa:
## Summary
- add a Responses-specific decompressed request body limit
- keep the existing generic decompression limit for non-Responses routes
- cover both the raised Responses limit and the unchanged generic-route rejection in middleware tests

## Why
Large Codex Responses requests can be sent compressed and legitimately decompress above the generic API body limit before the Responses route has a chance to validate or slim the `response.create` payload. This makes the proxy return a pre-route `413 payload_too_large` for requests that should continue into the normal Responses handling path.

This keeps the conservative default for ordinary API routes while allowing larger compressed bodies on `/backend-api/codex/responses` and `/v1/responses`.

## Tests
```bash
uv run python -m py_compile app/core/config/settings.py app/core/middleware/request_decompression.py
uv run pytest tests/unit/test_request_decompression_middleware.py tests/integration/test_request_decompression.py -q
git diff --check origin/main..HEAD
```


Soju06:
@codex review

Soju06:
@codex review

PR #774, opened 2026-05-22
fix(accounts): own DB session in detached token-refresh task

AnobleSCM:
## Summary
Fixes a background DB connection-pool leak: the detached, shielded token-refresh task can keep using a request-scoped `AsyncSession` after the originating request is cancelled (client disconnect), stranding a pooled connection. Accumulated, this exhausts the background pool and surfaces as `QueuePool limit ... connection timed out` 500s on every pool-backed path until restart.

## Root cause
`AuthManager.ensure_fresh` submits `refresh_account` to `_RefreshSingleflight`, which runs it as `asyncio.create_task(...)` then `await asyncio.shield(task)`. `refresh_account` writes via the manager's bound repo. In the proxy path that repo is bound to the request-scoped background session opened by `ProxyService._ensure_fresh` (`async with self._repo_factory()` -> `get_background_session()`).

On client disconnect mid-refresh, the request is cancelled at `await asyncio.shield(task)`; the shield keeps the refresh task alive, but the caller unwinds and the `async with` closes the session. The still-running task then calls `update_tokens` on the closed, concurrently-finalized `AsyncSession` (not safe for concurrent use), checking out a connection that is never returned to the pool.

Observed signature in a production multi-agent deployment: thousands of `QueuePool limit` 500s over hours, onset only after days of uptime (~1 leak per disconnect-interrupted refresh), correlated with client reconnect storms, and not per-request.

## Fix
- `AuthManager` gains an optional `refresh_repo_factory`. New `_run_refresh` (the singleflight body) opens its **own** session for the refresh write when the factory is set, decoupling the detached task from the caller's cancellation. Falls back to the bound repo when no factory is given (callers whose session is not client-cancellable, e.g. the usage refresh scheduler).
- `ProxyService._ensure_fresh` passes `refresh_repo_factory=self._accounts_refresh_scope` (a fresh accounts repo from the existing `_repo_factory`).

## Test
`tests/unit/test_auth_manager.py::test_ensure_fresh_detached_refresh_owns_session_on_caller_cancel`: refresh in flight -> cancel the caller -> assert the refresh wrote via its own session (opened + closed), never the request-scoped repo. Fails before the change, passes after.

## Verification
- `uv run pytest tests/unit/test_auth_manager.py -q` -> passes
- `uvx ruff check` / `uvx ruff format --check` (touched files) -> clean
- `uv run ty check` (touched files) -> clean

Includes an OpenSpec change under `database-backends` documenting the invariant (detached/shielded tasks own their DB session lifetime). No env/config/API/migration change.

### DIFF
--- a/app/modules/accounts/auth_manager.py
+++ b/app/modules/accounts/auth_manager.py
@@ -4,6 +4,7 @@ import asyncio
 import logging
 import time
 from collections.abc import Awaitable, Callable, Coroutine
+from contextlib import AbstractAsyncContextManager
 from datetime import datetime
 from hashlib import sha256
 from typing import Protocol, TypeAlias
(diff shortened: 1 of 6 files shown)


Soju06:
@codex review

Soju06:
@codex review

PR #793, opened 2026-05-24
fix(proxy): respect stream_idle_timeout_seconds in HTTP bridge keepalive backstop

Soju06:
## Problem

The HTTP bridge downstream event loop (`_stream_http_bridge_session_events`) fires `stream_idle_timeout` after a hardcoded `_STREAM_KEEPALIVE_MAX_COUNT` (6) × `sse_keepalive_interval_seconds` (10s) ≈ **60s**, regardless of the operator-configured `CODEX_LB_STREAM_IDLE_TIMEOUT_SECONDS` setting.

This causes **502 `stream_idle_timeout`** for slow-starting upstream responses:
- `input_image.image_url` — upstream must download and process the image before emitting `response.created` (typically 60-90s)
- `include_logprobs=true` on some models
- `file_input_id` with large files

Meanwhile, the upstream relay task (`_relay_http_bridge_upstream_messages`) correctly uses `stream_idle_timeout_seconds` (e.g. 300s) — so the upstream is still happily waiting while the downstream bridge has already declared timeout.

## Root cause

```python
# line 1547 — hardcoded 6, ignores settings
if keepalive_count > _STREAM_KEEPALIVE_MAX_COUNT:
    # emit stream_idle_timeout, break
```

## Fix

Derive `effective_max_keepalive` from `stream_idle_timeout_seconds / keepalive_interval`, using `_STREAM_KEEPALIVE_MAX_COUNT` as a minimum floor:

```python
stream_idle_timeout_seconds = getattr(get_settings(), "stream_idle_timeout_seconds", 600.0)
effective_max_keepalive = max(
    _STREAM_KEEPALIVE_MAX_COUNT,
    int(stream_idle_timeout_seconds / keepalive_interval),
)
if keepalive_count > effective_max_keepalive:
```

With `stream_idle_timeout_seconds=300` and `sse_keepalive_interval_seconds=10`:
- **Before**: max_count = 6 → 60s window → image requests timeout
- **After**: max_count = max(6, 300/10) = 30 → 300s window → image requests succeed

## Backward compatibility

- `_STREAM_KEEPALIVE_MAX_COUNT` (6) remains the minimum floor
- Default `stream_idle_timeout_seconds=600` → effective window = 600s (was 60s)
- Existing tests updated with explicit `stream_idle_timeout_seconds` matching old behavior
- New reproducer test confirms the fix

## Verification

- RED confirmed: new test fails on unpatched code (`got 3` events instead of `>3`)
- GREEN confirmed: new test passes on patched code (21 events)
- Full suite: **2808 passed, 7 skipped, 4 warnings**
- `ruff check`: clean

### DIFF
--- a/app/modules/proxy/service.py
+++ b/app/modules/proxy/service.py
@@ -1544,7 +1544,12 @@ class ProxyService:
                     except asyncio.TimeoutError:
                         keepalive_count += 1
                         downstream_response_id = _websocket_downstream_response_id(request_state)
-                        if keepalive_count > _STREAM_KEEPALIVE_MAX_COUNT:
+                        stream_idle_timeout_seconds = getattr(get_settings(), "stream_idle_timeout_seconds", 600.0)
+                        effective_max_keepalive = max(
+                            _STREAM_KEEPALIVE_MAX_COUNT,
+                            int(stream_idle_timeout_seconds / keepalive_interval),
+                        )
+                        if keepalive_count > effective_max_keepalive:
                             logger.info(
                                 "HTTP bridge stream idle timeout request_id=%s keepalive_count=%s",
                                 request_state.request_id,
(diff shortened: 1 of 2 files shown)


Soju06:
@codex review

PR #794, opened 2026-05-24
fix(proxy): inline external image URLs in HTTP bridge path

Soju06:
## Problem

The HTTP bridge (WebSocket pool) path was missing the image URL → data URL conversion that the HTTP direct-stream path already performs via `_inline_input_image_urls()`.

The upstream ChatGPT WebSocket only accepts `data:image/…` payloads. When an external `https://` image URL was sent through the HTTP bridge path, the upstream silently rejected the request — causing the proxy to hang for ~70 seconds until the keepalive backstop fired `stream_idle_timeout`.

Codex CLI doesn't hit this because it converts images to base64 data URLs client-side (`image.into_data_url()`). Third-party clients using the OpenAI-compatible `/v1/responses` API with `input_image.image_url` external URLs were affected.

## Root Cause

`_stream_via_http_bridge()` → `_prepare_http_bridge_request()` serialises the payload to JSON and sends it on the upstream WebSocket **without** applying `_inline_input_image_urls()`. The HTTP direct path (`_stream_responses_with_session`) does apply it.

## Fix

- Add `_inline_http_bridge_image_urls()` that applies the same inlining to the already-serialised `text_data` JSON in the HTTP bridge path.
- Quick `"input_image" not in text_data` pre-check to skip the parse/fetch cycle for payloads without image items.
- Updates `request_state.request_text` when inlining occurs so that WebSocket reconnect replays also use the inlined payload.
- Reverts #793 (keepalive max count change) which was a symptom-level fix that made the hang last longer instead of shorter.

## Tests

- 3 new unit tests: conversion, disabled-flag passthrough, already-inlined data URL passthrough.
- Full suite: 2810 passed, 7 skipped, 4 warnings.

Reverts #793

### DIFF
--- a/app/modules/proxy/service.py
+++ b/app/modules/proxy/service.py
@@ -1544,12 +1544,7 @@ class ProxyService:
                     except asyncio.TimeoutError:
                         keepalive_count += 1
                         downstream_response_id = _websocket_downstream_response_id(request_state)
-                        stream_idle_timeout_seconds = getattr(get_settings(), "stream_idle_timeout_seconds", 600.0)
-                        effective_max_keepalive = max(
-                            _STREAM_KEEPALIVE_MAX_COUNT,
-                            int(stream_idle_timeout_seconds / keepalive_interval),
-                        )
-                        if keepalive_count > effective_max_keepalive:
+                        if keepalive_count > _STREAM_KEEPALIVE_MAX_COUNT:
                             logger.info(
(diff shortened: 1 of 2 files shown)


Soju06:
@codex review

Soju06:
@codex review

Komzpa:
@codex review

Soju06:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

PR #799, opened 2026-05-25
fix(accounts): reuse existing row on reauth for same ChatGPT identity (closes #788)

### LINKED ISSUE #788: bug: reauth should update existing account instead of creating __copy duplicates
## Problem

When an existing account is deactivated because its refresh token was revoked, re-authenticating the same ChatGPT account through the dashboard can create a second local account row instead of updating the existing row.

Observed on `codex-lb` v1.17.0 with SQLite storage:

- Original row kept the long-term history and was `status=deactivated` with `deactivation_reason='Refresh token was revoked - re-login required'`.
- Reauth created a new active row with the same `email` and same `chatgpt_account_id`, but with a suffixed local id like `<deterministic-id>__copy2`.
- Dashboard then showed both rows: old/deactivated with historical usage and new/active with fresh auth.
- Historical tables are keyed by local `accounts.id`, so deleting either row naively loses either history or the fresh credentials/current status.

This is related to the broader duplicate/stale-account operator issue in #787, but this issue is the narrower data-integrity bug in the reauth/import path: same upstream identity should reconcile into one local account.

## Why it happens

From the installed source, both dashboard import and OAuth reauth build a deterministic local account id from the upstream ChatGPT account id + email, then call `AccountsRepository.upsert(account)`.

`AccountsRepository.upsert()` currently does this when merge-by-email is not active:

1. `existing = await self._session.get(Account, account.id)`
2. if an existing row is found and `merge_by_email` is false, it assigns `account.id = await self._next_available_account_id(account.id)`
3. the new row is inserted under `__copyN`

That is a reasonable behavior for intentional side-by-side imports, but it is wrong for reauth of the same upstream identity, especially when the existing row is deactivated because its refresh token was revoked.

## Suggested fix

Add an identity-aware reauth/upsert path that treats `chatgpt_account_id` as the durable upstream identity and reconciles duplicates before insert.

Possible implementation shape:

1. Add a repository method such as `upsert_by_chatgpt_identity(account)` or a `merge_by_identity=True` mode.
2. Acquire the same write/advisory lock, scoped to `chatgpt_account_id` when present.
3. Find candidate rows by `Account.chatgpt_account_id == account.chatgpt_account_id` and/or exact deterministic `Account.id == account.id`.
4. Choose a canonical row, preferably:
   - the existing deterministic id row if present, or
   - the oldest row with request/usage history, or
   - the oldest created row.
5. Copy fresh auth fields, plan, email, status, reset/block fields onto the canonical row using the same fields as `_apply_account_updates`.
6. Repoint dependent tables from non-canonical duplicate rows to the canonical id:
   - `usage_history`
   - `additional_usage_history`
   - `request_logs`
   - `sticky_sessions`
   - `http_bridge_sessions`
   - `api_key_accounts` (use conflict-safe dedupe because its primary key includes `account_id`)
7. Delete the duplicate `accounts` rows and invalidate account-selection/API-key caches.
8. Use this identity-aware path from dashboard OAuth reauth and auth import when the upstream `chatgpt_account_id` is known.

A follow-up migration could also add a unique partial index on `accounts(chatgpt_account_id)` where `chatgpt_account_id IS NOT NULL`, but only after a one-time reconciliation migration handles existing duplicate local rows.

## Expected behavior

Reauthenticating an existing ChatGPT account should:

- return the existing/canonical account id in the API response,
- reactivate the account with fresh tokens,
- preserve all historical usage/request/sticky-session data,
- leave only one dashboard row for that upstream identity,
- never create a `__copyN` row for the same `chatgpt_account_id` unless the user explicitly opts into a raw side-by-side import mode.

## Local recovery that worked

I repaired my local DB manually by backing up the SQLite database, then in one transaction:

1. copied fresh encrypted token fields/status/last_refresh from the `__copy2` row onto the historical canonical row,
2. moved dependent rows in `usage_history`, `additional_usage_history`, and `request_logs` from the duplicate id back to the canonical id,
3. checked the other FK tables for references,
4. deleted the duplicate `accounts` row.

Afterward there was one active account row for the upstream identity, no references to the deleted duplicate id, and the history range stayed intact.

ozpool:
## Why

OAuth reauth currently creates a second local account row with an `__copyN` suffix whenever the existing row's deterministic id collides under `merge_by_email=False` (issue [#788](https://github.com/Soju06/codex-lb/issues/788)). The original row — often deactivated because its refresh token was revoked — is left orphaned while the new row takes the active state. The dashboard then shows two rows for one upstream ChatGPT identity, and history-keyed tables become hard to reconcile.

## What

Add an identity-aware path on `AccountsRepository.upsert`:

- `merge_by_chatgpt_identity=True` (set by the OAuth reauth caller) finds the canonical row for the given `chatgpt_account_id` — preferring the oldest by `created_at` so long-term history stays attached — and applies the fresh tokens, plan, status, and deactivation-reason updates to that row.
- If duplicate rows already exist for the same upstream identity, reauth repoints dependent rows to the canonical account before deleting the duplicates: usage history, additional usage history, request logs, sticky sessions, HTTP bridge sessions, API key assignments, and limit-warmup state. Duplicate API key assignments and warmup window/reset collisions are de-duplicated instead of violating unique constraints.
- The path is intentionally independent of the dashboard `importWithoutOverwrite` setting. That setting governs the side-by-side import flow (two local rows for the same email when importing twice); reauth always reconciles to one local row per upstream identity.
- Falls through to the existing deterministic-id behavior when the incoming row has no `chatgpt_account_id`, so legacy local rows that pre-date the field keep working unchanged.
- Holds a chatgpt-id-scoped Postgres advisory lock on reauth so concurrent reauths for the same upstream identity serialize.

`OAuthService._persist_tokens` now passes `merge_by_chatgpt_identity=True` for both the direct-session and factory-session code paths.

## Tests

- `tests/integration/test_repositories.py` covers deactivated-row reuse, oldest-row preference when stray `__copy` rows already exist, dependent-row reconciliation, API key assignment de-dupe, warmup collision de-dupe, no-op create when no canonical match is found, and the no-upstream-id fall-through.
- `tests/integration/test_oauth_flow.py` pins the OAuth reauth contract: two reauths for the same upstream identity now land on one row, with the second reauth's plan applied in place. The duplicate-email conflict test was reworked so its precondition (two distinct local rows under one email) no longer relies on OAuth side-by-side, by using two distinct upstream ids.
- `openspec/changes/merge-reauth-by-chatgpt-identity/` documents the reauth identity merge and duplicate-row reconciliation contract.

Focused validation after the latest repair:

```bash
uv run pytest -q tests/integration/test_repositories.py tests/integration/test_oauth_flow.py tests/unit/test_accounts_repository_locks.py
uv run ruff check app/modules/accounts/repository.py app/modules/oauth/service.py tests/integration/test_repositories.py tests/integration/test_oauth_flow.py tests/unit/test_accounts_repository_locks.py openspec/changes/merge-reauth-by-chatgpt-identity/proposal.md openspec/changes/merge-reauth-by-chatgpt-identity/tasks.md
uv run ty check app/modules/accounts/repository.py app/modules/oauth/service.py tests/integration/test_repositories.py tests/integration/test_oauth_flow.py tests/unit/test_accounts_repository_locks.py
```

## Compatibility note for the renamed test

`test_device_oauth_flow_keeps_separate_accounts_when_import_without_overwrite_enabled` was pinning behavior that this issue calls out as the bug, so it has been renamed and its assertions inverted rather than kept alongside the new test. The `importWithoutOverwrite` setting still applies fully to the dashboard import path — only the OAuth reauth code path now reconciles by upstream identity.

Closes #788

### DIFF
--- a/app/modules/accounts/repository.py
+++ b/app/modules/accounts/repository.py
@@ -279,11 +279,7 @@ class AccountsRepository:
         await self._session.execute(
             update(RequestLog).where(RequestLog.account_id.in_(duplicate_ids)).values(account_id=canonical.id)
         )
-        await self._session.execute(
-            update(AccountLimitWarmup)
-            .where(AccountLimitWarmup.account_id.in_(duplicate_ids))
-            .values(account_id=canonical.id)
-        )
+        await self._reconcile_limit_warmups(canonical.id, duplicate_ids)
         await self._session.execute(
             update(StickySession).where(StickySession.account_id.in_(duplicate_ids)).values(account_id=canonical.id)
         )
@@ -293,6 +289,34 @@ class AccountsRepository:
             .values(account_id=canonical.id)
         )
         await self._session.execute(delete(Account).where(Account.id.in_(duplicate_ids)))
+
+    async def _reconcile_limit_warmups(self, canonical_account_id: str, duplicate_ids: list[str]) -> None:
+        existing_keys = {
+            (window, reset_at)
+            for window, reset_at in (
+                await self._session.execute(
+                    select(AccountLimitWarmup.window, AccountLimitWarmup.reset_at).where(
+                        AccountLimitWarmup.account_id == canonical_account_id
+                    )
+                )
+            ).all()
+        }
+        duplicate_warmups = (
+            (
+                await self._session.execute(
+                    select(AccountLimitWarmup).where(AccountLimitWarmup.account_id.in_(duplicate_ids))
+                )
+            )
+            .scalars()
+            .all()
+        )
+        for warmup in duplicate_warmups:
+            key = (warmup.window, warmup.reset_at)
+            if key in existing_keys:
+                await self._session.delete(warmup)
+            else:
+                warmup.account_id = canonical_account_id
+                existing_keys.add(key)
 
     async def update_status(
         self,
--- a/app/modules/accounts/repository.py
+++ b/app/modules/accounts/repository.py
@@ -9,7 +9,18 @@ from sqlalchemy.exc import OperationalError
 from sqlalchemy.ext.asyncio import AsyncSession
 
 from app.core.utils.time import utcnow
-from app.db.models import Account, AccountStatus, DashboardSettings, RequestLog, StickySession, UsageHistory
+from app.db.models import (
+    Account,
+    AccountLimitWarmup,
+    AccountStatus,
+    AdditionalUsageHistory,
+    ApiKeyAccountAssignment,
+    DashboardSettings,
+    HttpBridgeSessionRecord,
+    RequestLog,
+    StickySession,
+    UsageHistory,
+)
 
 _SETTINGS_ROW_ID = 1
 _DUPLICATE_ACCOUNT_SUFFIX = "__copy"
@@ -161,6 +172,10 @@ class AccountsRepository:
             canonical = await self._account_by_chatgpt_identity(account.chatgpt_account_id)
             if canonical is not None:
                 _apply_account_updates(canonical, account)
+                await self._reconcile_chatgpt_identity_duplicates(
+                    canonical=canonical,
+                    chatgpt_account_id=account.chatgpt_account_id,
+                )
                 await self._session.commit()
                 await self._session.refresh(canonical)
                 return canonical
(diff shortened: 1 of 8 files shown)


Soju06:
@codex review

ozpool:
Addressed the codex P2 in [33ab6cb](https://github.com/Soju06/codex-lb/pull/799/commits/33ab6cb).

When `merge_by_email=True` AND `merge_by_chatgpt_identity=True` with a `chatgpt_account_id` set, the postgres advisory lock was previously only keyed on `account.email`. Two concurrent reauths for the same upstream identity but different email claims would take different locks, both miss `_account_by_chatgpt_identity`, and both insert — recreating the duplicate-row pattern this PR is meant to prevent.

Fix takes the identity-keyed lock whenever identity reconciliation is in play, regardless of `merge_by_email`. Ordering identity-first, email-second across all callers gives a stable acquisition order so two reauths that overlap on either key serialize without deadlock.

Coverage in `tests/unit/test_accounts_repository_locks.py` monkeypatches the postgres dialect + the two lock methods and pins the lock sequence across four scenarios:

| `merge_by_email` | `merge_by_chatgpt_identity` | identity lock | email lock |
|---|---|---|---|
| True | True | `chatgpt:<id>` | `<email>` |
| False | True | `chatgpt:<id>` | — |
| False | False | `<account.id>` (fallback) | — |
| True | False | — | `<email>` |

`uv run pytest tests/integration/test_repositories.py tests/integration/test_accounts_repository.py tests/unit/test_accounts_repository_locks.py` clean (16 passed). `uv run ruff check` + `ruff format` clean.

@codex review

Komzpa:
@codex review

Soju06:
@codex review

Soju06:
@codex review

Soju06:
Blocking issue-coverage / merge-gate note: this PR still shouldn't auto-close #788 as written.

The current implementation is a good bleed-stopper, but the identity path only updates the chosen canonical row and returns:

- `app/modules/accounts/repository.py:160-166` finds the oldest row for `chatgpt_account_id`, applies fresh token/status metadata, commits, and returns it.
- It does not repoint dependent rows or delete/merge any other rows with the same `chatgpt_account_id`.

That means an already-affected database with `acc...` + `acc...__copyN` rows still keeps multiple dashboard rows for one upstream identity after reauth, and the stale duplicate can remain active/eligible. The new duplicate-existing regression (`tests/integration/test_repositories.py:187-207`) proves the oldest row is selected, but it does not assert that the duplicate row is removed or that only one row remains.

#788's expected behavior includes leaving one dashboard row / reconciling history-keyed tables. So please either:

1. implement the reconciliation/dedupe step for existing same-`chatgpt_account_id` rows (including FK repoint/dedupe for the dependent tables), or
2. keep this PR explicitly scoped as a bleed-stopper by changing `Closes #788` / the title to `Refs`/`Addresses`, adding or linking a follow-up issue for existing-row reconciliation, and making the OpenSpec/tasks state that pre-existing duplicates remain out of scope.

Code/lock direction looks reasonable for the bleed-stopper path; the blocker is the current "closes #788" claim versus the issue's full data-reconciliation acceptance.

Soju06:
@codex review

Komzpa:
@codex review

Soju06:
@codex review

Soju06:
Blocking follow-up: identity reconciliation updates/deletes hot-path account state but does not invalidate the in-process/cross-replica caches.

Evidence on the current head:

- `OAuthService._persist_tokens()` calls `repo.upsert(account, merge_by_chatgpt_identity=True)` directly (`app/modules/oauth/service.py:554-558`) and there is no cache invalidation in the OAuth path.
- That repository path can rewrite `api_key_accounts` (`app/modules/accounts/repository.py:245-270`), repoint usage/session rows, and delete duplicate `accounts` rows (`app/modules/accounts/repository.py:272-292`).
- The normal account delete path invalidates the same affected caches: account-selection cache, API-key cache, and `NAMESPACE_API_KEY` cross-instance invalidation (`app/modules/accounts/service.py:207-214`). Import/reactivate/pause also invalidate account-selection cache.
- The stale data is used by proxy hot paths: account-selection inputs are cached for 5s unless invalidated (`app/modules/proxy/account_cache.py:22-66`, consumed at `app/modules/proxy/load_balancer.py:410-412`), and API-key data is cached for 2s unless invalidated (`app/core/auth/api_key_cache.py:18-55`).

I reproduced the shape locally with an in-memory SQLite DB: after reauth, DB rows were reconciled to `acc_main`, but cached API-key/selection data still pointed at deleted `acc_dup`:

```text
saved_id= acc_main
db_accounts= ['acc_main']
db_assignments= ['acc_main']
cached_api= {'assigned_account_ids': ['acc_dup']}
cached_selection= {'account_ids': ['acc_dup']}
```

Please invalidate account-selection cache after successful OAuth identity merge, and clear/bump API-key cache invalidation when duplicate reconciliation touches API-key assignments or deletes duplicate accounts. This is part of #788's data-integrity acceptance because immediately after reauth the proxy can otherwise route with stale deleted-account / old-assignment state for the cache TTL window.

Komzpa:

Current #865 status:

PR #804, opened 2026-05-26
feat(accounts): add per-account OAuth proxy support

rupebac:
## Summary

  Adds per-account SOCKS5 proxy support for OAuth account import and account-bound upstream egress, so proxied accounts are validated before persistence and continue using their configured proxy for
  refreshes, usage, model fetches, files, HTTP responses, and WebSocket traffic.

  ## Type of change

  Linked issue: 

  ## OpenSpec

  Change directory: `openspec/changes/add-oauth-account-proxy/`

  ## Changes

  - Adds nullable per-account proxy columns to `accounts`, plus repository/service/API contracts for storing, validating, updating, and clearing account proxy config.
  - Adds account-bound HTTP/WebSocket client handling so proxied accounts use SOCKS5 egress, while direct accounts use isolated per-account direct sessions instead of the shared global client.
  - Updates OAuth add-account flow to defer persistence when proxy config is supplied, validate the proxy with upstream token refresh/account calls, and avoid creating an unproxied active account on
  failure.
  - Adds dashboard UI for proxy config during OAuth/import and account detail editing, including proxy error handling and password-preservation behavior.
  - Keeps browser OAuth `redirect_uri` tied to the registered configured callback and reuses that same URI for token exchange.
  - Adds OpenSpec requirements/context for account egress proxy behavior and outbound client isolation.

  ## Test plan

  ```bash
  uv run ruff check
  # All checks passed!

  uv run pyright app/
  # 0 errors, 1 warning, 0 informations

  uv run pytest tests/integration/test_migrations.py -q
  # 12 passed, 3 skipped

  uv run pytest tests/integration/test_oauth_flow.py::test_browser_oauth_redirect_uses_registered_uri_and_matches_token_exchange -q
  # 1 passed

  openspec validate add-oauth-account-proxy --strict
  # Change 'add-oauth-account-proxy' is valid

  openspec validate --specs
  # Totals: 22 passed, 0 failed (22 items)

  Earlier broader verification also passed:

  uv run pytest tests/integration/test_oauth_flow.py tests/integration/test_accounts_proxy_api.py tests/integration/test_account_proxy_repository.py tests/integration/test_migrations.py tests/unit/
  test_account_http_client.py tests/unit/test_account_proxy_failures.py tests/unit/test_account_proxy_probe.py tests/unit/test_account_proxy_schemas.py tests/unit/test_account_tls.py -q
  # 117 passed, 3 skipped

  cd frontend && npm run typecheck
  # passed

  cd frontend && npm test -- --run src/features/accounts/oauth-dialog.test.tsx src/features/accounts/hooks/use-oauth.test.ts src/features/accounts/proxy.test.tsx src/features/accounts/proxy-
  errors.test.ts src/features/accounts/components/import-dialog.test.tsx
  # 40 tests passed

  Live verification:

  - Deployed branch to the test host.
  - Added a proxied OAuth account successfully.
  - Confirmed the account persisted as active with proxy metadata and no tls_profile field.
  - Confirmed invalid proxy credentials fail before account persistence with Proxy validation failed: proxy_auth.

  ## Screenshots / output (optional)

  Live OAuth/proxy validation was exercised manually on the deployed test instance. No screenshots attached.

  ## Checklist

  - [x] Title is in Conventional Commits format (<type>(<scope>)?: <subject>).
  - [ ] Linked the related issue / discussion above.
  - [x] Ran uv run pre-commit run local-ci --hook-stage manual --all-files or the relevant make <target> subset locally.
  - [x] If touching specs: openspec validate --specs passes and /opsx:verify is clean.
  - [x] CHANGELOG is not edited by hand (release-please handles it).

### DIFF
--- a/app/core/auth/dependencies.py
+++ b/app/core/auth/dependencies.py
@@ -186,12 +186,17 @@ async def validate_codex_usage_identity(request: Request) -> ApiKeyData | None:
 
     async with get_background_session() as session:
         accounts_repo = AccountsRepository(session)
-        is_authorized = await accounts_repo.exists_active_chatgpt_account_id(account_id)
-    if not is_authorized:
+        matching_account_ids = await accounts_repo.list_active_account_ids_by_chatgpt_account_id(account_id, limit=2)
+    if not matching_account_ids:
         raise ProxyAuthError("Unknown or inactive chatgpt-account-id")
+    lease_account_id = matching_account_ids[0] if len(matching_account_ids) == 1 else None
 
     try:
-        await fetch_usage(access_token=token, account_id=account_id)
+        await fetch_usage(
+            access_token=token,
+            account_id=account_id,
+            lease_account_id=lease_account_id,
+        )
     except UsageFetchError as exc:
         if exc.status_code == 429:
             from app.core.exceptions import ProxyRateLimitError
--- a/app/core/auth/refresh.py
+++ b/app/core/auth/refresh.py
@@ -60,11 +60,11 @@ def should_refresh(
     """Return ``True`` when the access token should be refreshed.
 
     The effective interval is
-    ``interval_days ± jitter_hours``, where the offset is a stable
+    ``interval_days - jitter_offset``, where the offset is a stable
     deterministic function of ``account_id``. Two accounts onboarded on
-    the same day therefore land on different refresh days, which keeps
-    upstream OAuth from observing a clustered "fleet refresh" pattern
-    against the same ``oauth_client_id``.
+    the same day therefore land on different refresh points before the
+    configured max age, which keeps upstream OAuth from observing a
+    clustered "fleet refresh" pattern against the same ``oauth_client_id``.
 
     When ``account_id`` is omitted (legacy / defensive callers, tests
     that do not care about the specific account) the un-jittered
@@ -79,7 +79,7 @@ def should_refresh(
     if account_id:
         jitter_hours = float(settings.account_token_refresh_jitter_hours)
         offset_seconds = _refresh_jitter_offset_seconds(account_id, jitter_hours)
-        effective = base + timedelta(seconds=offset_seconds)
+        effective = base - timedelta(seconds=offset_seconds)
     else:
         effective = base
     # Defense in depth against a misconfigured ``jitter_hours`` (the
@@ -96,7 +96,7 @@ def should_refresh(
 
 
 def _refresh_jitter_offset_seconds(account_id: str, jitter_hours: float) -> float:
-    """Return a stable offset in ``[-jitter_hours*3600, +jitter_hours*3600]``.
+    """Return a stable offset in ``[0, jitter_hours*3600]``.
 
     The offset is derived from ``SHA-256(salt || account_id)`` so it is:
 
@@ -114,12 +114,11 @@ def _refresh_jitter_offset_seconds(account_id: str, jitter_hours: float) -> floa
         return 0.0
     digest = hashlib.sha256(_TOKEN_REFRESH_JITTER_SALT + account_id.encode("utf-8")).digest()
     # Take 8 bytes of the digest as an unsigned int and map it onto the
-    # symmetric window. ``2 * jitter_hours`` is the full window in
-    # hours, modulo'd into seconds.
+    # full offset window.
     raw = int.from_bytes(digest[:8], byteorder="big", signed=False)
-    full_window_seconds = 2.0 * jitter_hours * 3600.0
+    full_window_seconds = jitter_hours * 3600.0
     fraction = (raw % 1_000_000_007) / 1_000_000_007  # [0, 1)
-    return fraction * full_window_seconds - jitter_hours * 3600.0
+    return fraction * full_window_seconds
 
 
 def classify_refresh_error(code: str | None) -> bool:
@@ -161,10 +160,15 @@ async def refresh_access_token(
                     )
                     raise RefreshError("invalid_response", "Refresh response invalid", False) from exc
                 if resp.status >= 400:
+                    code = _extract_error_code(payload_data)
+                    message = _extract_error_message(payload_data)
                     logger.warning(
-                        "Token refresh failed request_id=%s status=%s",
+                        "Token refresh failed account_id=%s request_id=%s status=%s code=%s message=%s",
+                        account_id,
                         get_request_id(),
                         resp.status,
+                        code,
+                        message,
                     )
                     raise _refresh_error_from_payload(payload_data, resp.status)
     except RefreshError:
--- a/app/core/auth/refresh.py
+++ b/app/core/auth/refresh.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
-import asyncio
 import contextvars
+import hashlib
 import logging
 from dataclasses import dataclass
 from datetime import datetime, timedelta
@@ -12,13 +12,18 @@ from pydantic import ValidationError
 from app.core.auth import OpenAIAuthClaims, extract_id_token_claims
 from app.core.auth.models import OAuthTokenPayload
 from app.core.balancer import PERMANENT_FAILURE_CODES
-from app.core.clients.http import lease_http_session
+from app.core.clients.account_http import lease_account_http_session
 from app.core.config.settings import get_settings
 from app.core.types import JsonObject
 from app.core.utils.request_id import get_request_id
 from app.core.utils.time import to_utc_naive, utcnow
 
 TOKEN_REFRESH_INTERVAL_DAYS = 8
+# Salt the per-account jitter so the SHA-256 input cannot collide with
+# any other use of ``account_id`` we might add later (e.g. a separate
+# scheduling axis). The actual value is irrelevant; only its stability
+# matters.
+_TOKEN_REFRESH_JITTER_SALT = b"codex-lb:token-refresh-jitter:v1"
 
 logger = logging.getLogger(__name__)
 _TOKEN_REFRESH_TIMEOUT_OVERRIDE: contextvars.ContextVar[float | None] = contextvars.ContextVar(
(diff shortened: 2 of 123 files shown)


rupebac:
this has been purely vibed coded, but it has been working great for me

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Soju06:
@codex review

rupebac:
This started as a local spike, but I have now rebased it onto the latest PR head and prepared a focused follow-up commit (`947924a`, pending push to this PR branch).

Current follow-up scope:

- Token refresh jitter
  - Changed per-account refresh jitter from a signed offset (`interval ± jitter`) to an early-only offset (`interval - jitter`).
  - This preserves spreading across accounts while keeping `token_refresh_interval_days` as the hard maximum token age.
  - Added coverage that jitter stays within `[0, jitter]` and never delays accounts beyond the configured interval.

- Targeted re-authentication
  - The Accounts UI now passes the selected account id when clicking re-auth.
  - OAuth state carries that target account id through browser, manual callback, and device-code flows.
  - Successful OAuth writes the new tokens back into the selected existing account row instead of going through generic add-account/upsert behavior.
  - Existing proxy configuration is preserved.
  - If the selected account has a stored proxy, server-side OAuth bootstrap/token-exchange calls use that stored proxy during re-auth.
  - Re-auth rejects returned credentials for a different ChatGPT account id/email instead of overwriting the selected account.
  - If the stored proxy password cannot be decrypted, re-auth fails explicitly rather than silently falling back to no proxy.

- Browser OAuth redirect regression
  - Live re-auth testing exposed that browser OAuth was rewriting the configured callback from `http://localhost:1455/auth/callback` to the remote dashboard host, e.g. `http://192.168.x.x:1455/auth/callback`.
  - The previous working image did not rewrite this URI, and OpenSpec already says browser OAuth must use the configured `oauth_redirect_uri` unchanged.
  - Restored the registered callback behavior and updated the integration test that had drifted to expect the rewrite.

- Diagnostics
  - Added refresh-failure logging with account id, request id, HTTP status, upstream error code, and message.
  - Added logging before deactivating an account due to permanent token refresh failure.

- OpenSpec
  - Added `fix-token-refresh-jitter-max-age`.
  - Added `fix-oauth-reauth-preserve-proxy`.
  - Synced resulting requirements into the relevant main specs.

Validation run locally:

- `uv run ruff check ...`
- `uv run pytest tests/integration/test_oauth_flow.py -q` -> `40 passed`
- broader targeted backend pytest run -> `144 passed`
- `npm run test -- src/features/accounts/hooks/use-oauth.test.ts`
- `npm run typecheck`
- `npm run lint -- ...`
- `npm run build`
- `openspec validate --specs`
- `openspec validate fix-token-refresh-jitter-max-age --strict`
- `openspec validate fix-oauth-reauth-preserve-proxy --strict`

SHAREN:
Hi. Do you have any rough estimate of how much proxy bandwidth is used when per-account proxy is enabled?

For example, if an account consumes around 100 million tokens, how much traffic would go through the proxy approximately? And if usage is around 500 million tokens per week, what proxy bandwidth should I plan for?

I understand it depends on request/response size, streaming overhead, retries, files, etc. I only need an approximate order of magnitude.

rupebac:
for me working a lot, I use less than 1gb every week (with 2 accounts per proxy)... but I am having an issue with this right now... accounts get deactivated every 2 days or so and need to re-auth... not sure what yet, keep doing tests... I have deactivated the codex tls profile, still the same.. rotate usage fetch's so they do not match at the same time, and the same... keep doing more tests, will inform.

Soju06:
@codex review

Soju06:
@codex review

Soju06:
@codex review

Komzpa:
@codex review

gabrielnogueiracsc:
> for me working a lot, I use less than 1gb every week (with 2 accounts per proxy)... but I am having an issue with this right now... accounts get deactivated every 2 days or so and need to re-auth... not sure what yet, keep doing tests... I have deactivated the codex tls profile, still the same.. rotate usage fetch's so they do not match at the same time, and the same... keep doing more tests, will inform.

Did you find more something about this? I'm having the same here but without proxy. I started think if isolated every account in one proxy would avoid this to happen...

Soju06:
@codex review

Soju06:
Hermes owner-review found current-head blockers that should be addressed before merge:

1. **Credential leak on websocket proxy errors** — `app/core/clients/account_http.py` builds account websocket proxy URIs with `username:password@host`, and `app/core/clients/proxy_websocket.py` returns `str(InvalidProxy)` directly in the OpenAI error envelope. `websockets.exceptions.InvalidProxy("socks5h://user:secret@proxy.example.com:1080", "bad")` renders the full URI including credentials, so malformed/unsupported websocket proxy errors can leak the stored proxy password to the downstream caller. Please redact/sanitize proxy URIs before client-visible errors/logs and add a regression.

2. **Proxy edit UI/API password semantics are misleading** — the frontend/API schemas say omitted `password` keeps the existing password, and the edit dialog defaults to `Keep`, but `AccountsService.set_account_proxy()` only reuses the old password when host/port/username are unchanged. If an operator leaves `Keep` selected while changing the proxy host/port/user, the backend can persist the new proxy with `hasPassword=false` after a no-auth probe. Please either require `Replace` when the proxy identity changes or make the UI/API wording explicit and prevent silent password clearing.

3. **Stable OpenSpec contradiction** — `openspec/specs/account-egress-proxy/spec.md` requires import-with-proxy to validate and atomically persist proxy fields, while the later “Auth.json import/export does not carry proxy configuration” requirement says import/export endpoints MUST NOT read or write proxy configuration and proxy is managed only through dedicated proxy endpoints. Please narrow the latter to “proxy fields inside the auth.json file are ignored/export does not include proxy state,” or otherwise reconcile the normative requirements.

Additional note: the current non-outdated Codex thread about `reauthAccountId + expectProxy` looks addressed/unreachable because `start_oauth()` now rejects that combination before the `tokens_ready` finalization path.

Soju06:
@codex review

Soju06:
@codex review

rupebac:
> > for me working a lot, I use less than 1gb every week (with 2 accounts per proxy)... but I am having an issue with this right now... accounts get deactivated every 2 days or so and need to re-auth... not sure what yet, keep doing tests... I have deactivated the codex tls profile, still the same.. rotate usage fetch's so they do not match at the same time, and the same... keep doing more tests, will inform.
> 
> Did you find more something about this? I'm having the same here but without proxy. I started think if isolated every account in one proxy would avoid this to happen...

uhmm not yet... the thing about proxy per account is that you can´t reuse the connection for usage fetch across all accounts, and stuff like that, as that defeats the purpose of the proxy itself and would connect accounts... also, this was stripping out the installation id header from codex, as that would also connect accounts (even through proxy)... the test I am reproducing now is generating a random installation id per account, as I thought that sending requests without one may be the reason as well. Also, fetching usage of all accounts at the same time is sort of connecting accounts (also addressed locally in my latest changes). But only 12h since deployed, usually for me they were lasting 2-3 days... so I need to wait a bit more... I do have 1 personal account (without proxy) that is never expiring, so there is some sort of reputation taking place I think... I am testing with ISP Dedicated proxies from iproyal.... It could be that the reputation of these IPs is damaged already not sure... If there are more ideas, please let me know :)

Komzpa:
@codex review

rupebac:
same issue (accounts getting invalidated), but I just found out that the proxies I was using from iproyal had a really bad scamalytics score... I very much suspect of this now... I have got another 3 proxies from another provider, this time with a 0 fraud score... retrying... will keep you updated... @Soju06  my local branch I have a bunch of fixups, but let me test this ip thing, and I will push. We gotta find out why this is happening. If you don´t agree with something please let me know, lets keep it up together.... These I have locally done VS origin now (not yet pushed):

1. Token invalidation mitigation
      - Added per-account codex_installation_id.
      - Added migration: app/db/alembic/versions/20260531_000000_add_accounts_codex_installation_id.py.
      - Injects installation ID into main /codex/responses payload metadata.
      - Injects installation ID into websocket response.create.
      - Sends installation ID as a header for compact, while stripping unsupported client_metadata from compact payloads.
      - Strips inbound x-codex-installation-id so callers cannot cross-contaminate accounts.

  2. Usage refresh staggering
      - Default usage refresh interval changed 60s -> 120s.
      - Scheduler now refreshes one eligible account per slice instead of hammering all accounts together.
      - Skips deactivated accounts and invalidates dashboard cache at cycle boundaries.

  3. Expired-token handling
      - Treats app_session_terminated as permanent.
      - Usage refresh now deactivates accounts on that refresh failure path.
      - Prevents repeated routing/retry loops after an account is known bad.

  4. Stale HTTP bridge session guard
      - Added in-memory routing_unavailable account set.
      - Paused/deactivated/deleted accounts are immediately marked unavailable for routing.
      - Reactivated/re-imported/proxy-repaired accounts clear that flag.
      - HTTP bridge reuse now checks this flag, not just DB status captured in an old session object.

  5. OAuth callback behavior
      - OAuth redirect handling changed in app/modules/oauth/service.py; tests updated in tests/integration/test_oauth_flow.py.

  6. Edit proxy UI
      - “Edit egress proxy” now has quick paste: user:pass@host:port.
      - Quick paste also supports socks5://, socks5h://, and host/port without auth.
      - Removed bad native browser pattern validator from proxy port inputs; schema/parser validation remains.

Komzpa:
Superseded by #875, which consolidates this egress-proxy work on current main and is now the surviving merge-clean branch.

SHAREN:
As the saying goes, it’s no use waving your fists after the fight is over—but since I’ve already finished the report, I’ll go ahead and post it anyway.


Hi everyone! While testing this PR with residential/mobile proxies, we caught a critical bug related to SOCKS5 authentication for
  WebSocket connections.

  ### 🐛 The Bug
  If a proxy username contains a semicolon `;` (which is very common for residential proxies to dictate session TTL and country,
  e.g., `user__cr.my;anon.1;sessttl.5`), the account gets immediately deactivated with `failed to connect to SOCKS proxy`.
  Interestingly, HTTP requests work fine, but stream generation (which uses WebSockets) drops the connection.

  ### 🔍 The Cause
  In `codex-lb`, standard HTTP requests use `aiohttp` (which uses `yarl` to properly unquote URLs).
  However, WebSocket connections use `python_socks`. In `app/core/clients/account_http.py` inside
  `get_account_websocket_proxy_uri()`, the credentials are url-encoded via `urllib.parse.quote()`, turning `;` into `%3B`.
  When this URI is later passed to `python_socks` (via `Proxy.from_url()`), it **does not** unquote the credentials. It sends `%3B`
  literally to the SOCKS5 proxy, causing auth to fail and the connection to instantly drop.

  ### 🛠️ Our Local Fix
  We removed the `quote()` wrappers in `account_http.py` so raw characters are passed. Here is the exact diff that got our proxies
  working:

  ```diff
  --- a/app/core/clients/account_http.py
  +++ b/app/core/clients/account_http.py
  @@ -655,8 +655,8 @@ async def get_account_websocket_proxy_uri(account_id: str) -> str | None:
               if config.username:
                   from urllib.parse import quote  # noqa: PLC0415

  -                user = quote(config.username, safe="")
  -                password = quote(config.password or "", safe="")
  +                user = config.username
  +                password = (config.password or "")
                   userinfo = f"{user}:{password}@" if password else f"{user}@"
               else:
                   userinfo = ""

  💡 Note: While removing quote() fixes the immediate issue locally, a more robust upstream fix might be to pass credentials directly
  into python_socks via kwargs instead of serializing them into a URI string and parsing them back out via from_url().

SHAREN:
@rupebac could you please tell me whether you managed to properly set up a separate proxy for each account? I installed your build, but it still often asks me for SMS verification.

rupebac:
> @rupebac could you please tell me whether you managed to properly set up a separate proxy for each account? I installed your build, but it still often asks me for SMS verification.

yes, I think I have sorted it out... Since I changed my proxies from iproyal to seamlessproxies. I just got 1-2 deactivated accounts that I had to re-auth(out of my 8-9 test accounts) in like 4 days (and who knows if these deactivations were due to any signal from before the proxy change or something related to it)... You have to check the reputation of the ip's you are getting from your provider in ipqualityscore.com and scamalytics.com, cos a lot comes down to that I think. Also I noticed that refresh tokens get invalidated if the same account you're using it simulatenously in another IP (but this is not so common I think right?)

But I can´t be 100% sure if what fixed it in my case was just the proxy change, or my latest changes that I just pushed to this branch and documented in my previous comment. Maybe you can try with my latest changes and your current proxies? Because these changes I guess are not yet included in #875 PR.

Btw, if you are getting some http connections getting closed randomly by the proxy (I had this problem with seamlessproxies, but they just fixed it), these are mainly the /compact calls (they are not going through WS).... WS upgraded long connections are not being closed by seamlessproxies, only http ones. In my case it helped enabling remote_compaction_v2 hidden feature in Codex CLI, now the compactation is done over the WS.

rupebac:
fixed a problem where Import was requiring the refresh token to exist in auth.json when proxy is enabled.... @SHAREN  should I merge anything into another branch, or a separated PR ? what you reckon?

SHAREN:
@rupebac  Do it whichever way is more convenient for you. I’ll also try to push my changes to the repository.

SHAREN:
@rupebac https://github.com/SHAREN/codex-lb/commits/renat/proxy-working-state-20260605

rupebac:
I am continuing in my branch, differences just got too much, and anyway I think it is good to have 2 forks that can test multiple things in terms of account burns.... I have to say I don´t agree with the proxy implementation on #875, the use case should be simpler I think, 1 account -> 1 proxy... there is no need for pools, or failovers... it makes sense to be more like a kill switch... you don´t want to change the ip ever for one account.

In my last commit, I am injecting a real User Agent and "originator" in all usages of aiohttp, instead of the default one from the aiohttp session. This I think was the biggest red flag, they could easily identify 3rd party proxy with this.... Also I made sure the JA3 and JA4 fingerprints for TLS are exactly the same as Codex CLI.. And I also implemented a shifting of the refreshes, so they are not matching at the same time...

I have also implemented an egress verification, to make sure the public IP that OpenAI is seeing is not matching more than 1 account on our side (it warns)... Also a detection that the dns resolution is not leaking outside the proxy...

I have imported also your new routing strategies, I think they are good, however later I realized that better strategies are the one that uses more accounts simultaneously, not less, because I think we want to reduce the load per account... We need a option on the strategy to allow a max concurrency of 1 per account, and even a cold down period... Now what I am wondering is, if somehow they use the content of the context, to connect multiple accounts... 

I have also redesigned the accounts sections from the Dashboard, now look more compact, clear and nicer (to my liking :) ). And no scrollbar in that section.

All is done vibe coding, I haven´t analized the code... only tested it... we gotta move faster I think...

Also the warm-up thing, is very very suspicious, last week I got one account banned, the only one that was using warm-up.... A smarter and more randomized warm-up would be required if we want such feature... like the warm-up message being a real random query, not a static text... And also the time patterns to be random and let miss some.

Would be good if more people can try my branch and report....  Also I am running out of accounts, if anyone can help here...

Komzpa:
@rupebac @SHAREN please send pull requests if you have fixes or branches

rupebac:
> @rupebac @SHAREN please send pull requests if you have fixes or branches

I did rebase some of my changes on top of #875 and put a comment there but got no reply... I did try to communicate and sync up, but I ain´t wasting time for nothing... I will continue on my fork and my branch from this PR, and when you merge #875 to master I will try to rebase what makes sense. For now, I want to implement ton of stuff... working now on a hard cap on concurrency and load per account...

bjspi:
@rupebac May I ask you how much you pay for your proxies and if you really think / consider using Proxys per account to be safer w.r.t. account banning? ( I assume account-banning from OpenAI side is the only reason to use proxys right?)?

SHAREN:
> @rupebac May I ask you how much you pay for your proxies and if you really think / consider using Proxys per account to be safer w.r.t. account banning? ( I assume account-banning from OpenAI side is the only reason to use proxys right?)?

It took me 0.5-1 GB per day, it turned out to be a lot for me.

PR #801, opened 2026-05-25
fix(proxy): classify stale Codex websocket anchors

Soju06:
## Summary
- Adds an OpenSpec change that centralizes the source of truth for Codex-native WebSocket `previous_response_id` continuity failures.
- Implements a sanitized Codex-native direct WebSocket stale-anchor classifier: `codex_previous_response_stale`.
- Preserves raw upstream masking: downstream never receives raw `previous_response_not_found` or the missing upstream `resp_...` id.
- Keeps grouped `missing_tool_output` terminal failures on the generic `stream_incomplete` path, even when Codex stale-anchor classification is enabled.
- Keeps `/v1/responses` WebSocket compatibility on the existing generic `stream_incomplete` masking, while `/backend-api/codex/responses` gets the Codex-specific classifier that clients can use for soft-reset retry.
- Records stable context in `openspec/specs/responses-api-compat/context.md` and keeps proxy-side replay ledger out of the primary path.

## Context
Long background terminal waits can expire the upstream Responses anchor while the local Codex turn is still active. Retrying the same delta with the stale `previous_response_id` repeats the failure. The desired recovery is for a compatible Codex client to classify the stale-anchor failure, reset incremental WebSocket state, and retry once from full conversation history without `previous_response_id`.

## Implementation
- Adds `PREVIOUS_RESPONSE_STALE_CODE` / `PREVIOUS_RESPONSE_STALE_MESSAGE`.
- Marks Codex-native direct WebSocket request state as eligible to expose the stale-anchor classifier.
- Routes previous-response stale rewrites through a single helper so nested, top-level, grouped, connect-failure, and terminal-error paths share the same sanitized behavior.
- Preserves non-stale continuity errors, including grouped missing-tool-output failures, as generic `stream_incomplete` failures.
- Leaves OpenAI-compatible `/v1/responses` WebSocket behavior unchanged.

## Verification
```bash
[...]

PR #810, opened 2026-05-26
fix(proxy): parse multiline Codex websocket errors

kramarb:
## Summary
- Parse raw websocket JSON frames when SSE parsing fails, so pretty-printed upstream error frames still hit stale-anchor masking.
- Add regression coverage for multiline `previous_response_not_found` websocket errors.

## Test plan
- `uv run pytest tests/integration/test_proxy_websocket_responses.py -q`
- Live local websocket repro returned `codex_previous_response_stale` instead of raw `previous_response_not_found`.

### DIFF
--- a/app/modules/proxy/service.py
+++ b/app/modules/proxy/service.py
@@ -8141,6 +8141,14 @@ class ProxyService:
     ) -> str:
         event_block = f"data: {text}\n\n"
         payload = parse_sse_data_json(event_block)
+        if payload is None:
+            try:
+                raw_payload = json.loads(text)
+            except json.JSONDecodeError:
+                raw_payload = None
+            if isinstance(raw_payload, dict):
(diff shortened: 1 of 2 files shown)


kramarb:
Extra context on the failure mode:

The leak happened when the upstream websocket returned a raw JSON error frame formatted across multiple lines, e.g. `type: "error"` with `code: "previous_response_not_found"`.

The proxy path first wrapped websocket text as a single SSE `data:` block before parsing. That works for compact one-line JSON, but not for pretty-printed JSON because only the first line has the `data:` prefix. Parsing failed, so the previous-response masking path never saw the error and the raw upstream frame was forwarded to Codex Desktop.

This PR keeps the existing stale-anchor handling from #801 and only adds a fallback to parse the raw websocket frame as JSON before normalizing it to SSE.

Komzpa:
@codex review

PR #814, opened 2026-05-26
[codex] Fix TLS trust store for aiohttp clients

onlysdesign-ui:
## Summary
- build aiohttp TLS contexts from the default platform/OpenSSL trust roots and add certifi's CA bundle
- add certifi as a direct runtime dependency for installs with incomplete default CA discovery
- cover connector SSL wiring and certifi loading with unit tests

## Root cause
On macOS installs that use python.org Python in a background launchd process, the default OpenSSL trust store can miss system CAs. OAuth browser callbacks then reach codex-lb but fail while exchanging the authorization code with auth.openai.com, surfacing as a 500 callback page.

The TLS context must keep the normal platform/OpenSSL/env CA resolution path for private or intercepting proxy CAs, then load certifi as an additional CA source.

## Validation
- `uv run python -m pytest tests/unit/test_http_client.py tests/unit/test_oauth_client.py tests/unit/test_files_client.py tests/unit/test_proxy_utils.py::test_upstream_unavailable_certificate_connect_error_is_not_transient_retry`
- `uv run ruff check app/core/clients/http.py tests/unit/test_http_client.py`
- `uv run ty check`
- `git diff --check`


Komzpa:
@codex review

Issue #787, opened 2026-05-23
feat: surface stale + duplicate accounts after token-invalidation cascade (operator playbook + diagnostic suggestions)


thomasnormal:
## Problem

After a Codex/ChatGPT-side OAuth token-invalidation event (Settings → "Revoke all sessions", or upstream-side mass invalidation), re-adding the affected accounts via the dashboard leaves the load-balancer pool in a state that **looks healthy but isn't**:

- All accounts show `status=active` in `/api/accounts` with future `access_exp` timestamps
- But upstream codex returns `401 token_invalidated` for the vast majority of requests
- The dashboard "Reactivate" button is hidden when an account is in `Deactivated` state, so the operator's instinct is to re-add via OAuth, which causes the next problem...
- The re-add flow **creates a second account row with the same email but a new account ID** (the old row gets a `__copy2` suffix in some imports, in others it stays in the pool with its dead refresh token)
- Stale account rows keep getting hit by the load balancer's round-robin, generating endless 401s that look identical to the original cascade

Net effect: the operator spends time re-adding 18+ accounts only to discover the fleet is still 70%+ failing because:
1. Several "active" accounts have refresh tokens that were revoked >1 day ago and the lb's local view doesn't reflect that
2. Duplicate rows for the same email (one fresh, one stale) keep the dead token in rotation

## How the cascade looked

In my case (a tmux fleet of ~20 codex agents): the 9h fleet outage took these steps to resolve:

1. Restart codex-lb (no effect — lb auth state is persisted)
2. Re-add each affected account via dashboard Import flow (creates duplicates instead of replacing)
3. Manually identify duplicates: same email, two distinct `accountId` values
4. Manually identify stale accounts: `lastRefreshAt` older than the most recent reset event
5. `DELETE /api/accounts/{id}` for each via Bearer-token API (web UI doesn't expose Delete for `active` accounts cleanly)
6. `POST /api/sticky-sessions/delete` in batches (the existing `delete-filtered` endpoint 500's on non-empty filter — `staleOnly:false` + non-empty `accountQuery` or `keyQuery` raises an SQLAlchemy "max parameters exceeded" error)
7. After all that, lb 401:200 ratio went from ~7:1 to 0:N — recovery confirmed

## What's missing in the dashboard

A few things would have made this a 5-minute job:

### A. "Detect stale accounts" indicator

In `/accounts` table, a column or visual flag that highlights accounts where:
[...]

PR #828, opened 2026-05-27
fix(sticky-sessions): chunk delete_entries to avoid sqlite bind overflow

ozpool:
## Summary

Addresses #787 (D).

`POST /api/sticky-sessions/delete-filtered` with a non-empty filter that matches a large number of rows previously issued a single `DELETE ... WHERE OR(key=:k AND kind=:t)...` statement whose bind-parameter count grew with the match-set size. On large `sticky_sessions` tables (the reporter saw thousands of rows pinned to a single account email) the statement tripped SQLite's `SQLITE_LIMIT_VARIABLE_NUMBER` (999 on libsqlite < 3.32, 32766 on newer) and the endpoint surfaced a 500 `internal_error`.

The fix chunks the deduplicated target list inside `StickySessionsRepository.delete_entries` at `_DELETE_ENTRIES_CHUNK_SIZE = 250` pairs (500 bind parameters per chunk), which is safe on every libsqlite version that ships with current Python interpreters and well under Postgres's 65535 bind-parameter ceiling.

## Behavior

- `/api/sticky-sessions/delete-filtered` with any filter size now returns 200 with the full delete count.
- `/api/sticky-sessions/delete` already enforces `maxItems=500` at the schema layer, so the explicit-list endpoint never reached this path.
- Chunks commit independently (one `sqlite_writer_section` per chunk) to keep the writer lock window short — matches the pattern already used by `delete` / `upsert` on this repository.

## Tests

`tests/integration/test_sticky_sessions_api.py::test_sticky_sessions_api_deletes_filtered_chunks_large_match_set` (new) inserts 600 sticky_sessions for one account, calls `/delete-filtered`, and pins:

- response `200` with `{"deletedCount": 600}`
- zero `bulk-delete-%` rows remaining in `sticky_sessions`

This forces the chunk loop to execute more than once and would have surfaced the original 500 if the single-statement variant were still in place. Existing `test_sticky_sessions_api_deletes_filtered_rows` and the full `tests/integration/test_sticky_sessions_api.py` / `tests/integration/test_proxy_sticky_sessions.py` suites stay green (32 passed).

### DIFF
--- a/app/modules/proxy/sticky_repository.py
+++ b/app/modules/proxy/sticky_repository.py
@@ -15,6 +15,15 @@ from app.db.models import Account, StickySession, StickySessionKind
 from app.db.session import sqlite_writer_section
 from app.modules.sticky_sessions.schemas import StickySessionSortBy, StickySessionSortDir
 
+# Each (key, kind) pair in delete_entries contributes 2 bind parameters to
+# the underlying DELETE...OR (key=:k AND kind=:t)... statement. SQLite's
+# default SQLITE_LIMIT_VARIABLE_NUMBER is 999 on builds older than 3.32
+# and 32766 on newer builds, so chunking conservatively at 250 pairs
+# (500 bind parameters) keeps delete-filtered safe on any libsqlite that
+# ships with current Python interpreters. Postgres allows up to 65535
+# bind parameters, which this chunk size also respects.
+_DELETE_ENTRIES_CHUNK_SIZE = 250
+
 
 @dataclass(frozen=True, slots=True)
 class StickySessionListEntryRecord:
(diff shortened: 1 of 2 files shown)


Soju06:
@codex review

Soju06:
Owner review note: code/tests look merge-ready, but the repo review gate is not complete because `@codex review` hit the Codex usage limit and the PR has no fresh clean Codex evidence / `🤖 codex: ok` label.

I did not find a code-level blocker. Focused sticky-session delete tests pass, the chunk size is safely under SQLite/Postgres bind limits, and the partial `Addresses #787 (D)` wording looks appropriate because the larger #787 remains open.

Komzpa:
@codex review

Komzpa:
@codex review

PR #837, opened 2026-05-27
feat(config): flexible location of DATA_DIR by env-variable

### LINKED ISSUE #836: bug: DATA_DIR in interactive sandbox lands in `/var/lib` [HAS FIX]
### Pre-flight checklist

### codex-lb version

1.19.0

### Deployment method
- Local Python (uv)

### What happened?
When working in an interactive container as a sandbox, `.containerenv` exists and leads to `DATA DIR` in `/var/lib`. This isn't writeable by a non-root user.

There is also no way to point codex-lb at an arbitrary data directory via an environment variable.

- If `$HOME/.codex-lb` already exists, codex-lb should prefer it even when running inside a container.
- An environment variable (`CODEX_LB_DATA_DIR`) should allow operators to override the default data directory path.

### Steps to reproduce
1. Run codex-lb locally → `~/.codex-lb/` gets populated
2. Run the same binary inside an interactive container (e.g. podman run ...)
3. codex-lb ignores `~/.codex-lb` and tries to write to `/var/lib/codex-lb` → permission error

```yaml
# No env var available to override data dir
```

### Additional context

Current precedence in `_default_home_dir()`:
```python
if _in_container():
    return DOCKER_DATA_DIR
return Path.home() / ".codex-lb"
```

This means container detection always wins, even when the user has existing data.

tobwen:
## Summary

Running codex-lb inside an interactive container (e.g. `podman run …`) sets `/.containerenv`, which triggers `_in_container()` and forces the data directory to `/var/lib/codex-lb`. That path is typically not writable for a non-root user, causing a permission error on startup. Additionally, there is no environment variable to override the default data directory.

Fixes #836

## Type of change

- [x] `feat:` new feature (data directory precedence and env override)

## Root cause

`_default_home_dir()` returns `DOCKER_DATA_DIR` as soon as `_in_container()` is true, without checking whether `$HOME/.codex-lb` already exists. This orphans existing local data and breaks interactive containers where `/var/lib/codex-lb` is read-only or unwritable.

## Fix

1. Check `CODEX_LB_DATA_DIR` first - highest priority override.
2. Check whether `Path.home() / ".codex-lb"` exists. If yes, use it even inside a container.
3. Fall back to `/var/lib/codex-lb` only when running in a container **and** the home dir does not already exist.

## Test plan

```text
$ uv run pytest tests/unit/test_settings_home_dir.py -v
5 passed
```

Each regression case fails on `origin/main` (wrong precedence) and passes with the patch.

## Checklist

- [x] Ran `uv run pre-commit run local-ci --hook-stage manual --all-files` locally.

## Additional hint

This doesn't break containers since they normally shouldn't have `~/.codex-lb`.

### DIFF
--- a/app/core/config/settings.py
+++ b/app/core/config/settings.py
@@ -425,11 +425,15 @@ class Settings(BaseSettings):
     def _apply_data_dir_defaults(self) -> "Settings":
         if self.data_dir == DEFAULT_HOME_DIR:
             return self
-        if self.database_url == DEFAULT_DATABASE_URL:
+        explicitly_set = self.model_fields_set
+        if "database_url" not in explicitly_set and self.database_url == DEFAULT_DATABASE_URL:
             self.database_url = f"sqlite+aiosqlite:///{self.data_dir / 'store.db'}"
-        if self.encryption_key_file == DEFAULT_ENCRYPTION_KEY_FILE:
+        if "encryption_key_file" not in explicitly_set and self.encryption_key_file == DEFAULT_ENCRYPTION_KEY_FILE:
             self.encryption_key_file = self.data_dir / "encryption.key"
-        if self.conversation_archive_dir == DEFAULT_CONVERSATION_ARCHIVE_DIR:
+        if (
+            "conversation_archive_dir" not in explicitly_set
+            and self.conversation_archive_dir == DEFAULT_CONVERSATION_ARCHIVE_DIR
+        ):
             self.conversation_archive_dir = self.data_dir / "conversation-archive"
         return self
 
--- a/app/core/config/settings.py
+++ b/app/core/config/settings.py
@@ -27,8 +27,8 @@ def _in_container() -> bool:
 
 def _default_home_dir() -> Path:
     env_dir = os.getenv("CODEX_LB_DATA_DIR")
-    if env_dir:
-        return Path(env_dir)
+    if env_dir and env_dir.strip():
+        return Path(env_dir.strip())
     home_dir = Path.home() / ".codex-lb"
     if home_dir.exists():
         return home_dir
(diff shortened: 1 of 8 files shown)


Soju06:
@codex review

Komzpa:
@codex review

Soju06:
@codex review

Komzpa:
@codex review

Soju06:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Soju06:
Hermes owner-review blocker: the behavior change has an OpenSpec change folder, but the change has no parsed requirement deltas.

Evidence:
- `uv run openspec validate data-dir-env-var-home-exists --strict` fails with: `Change must have at least one delta. No deltas found.`
- `uv run openspec change show data-dir-env-var-home-exists --json --deltas-only` reports `"deltaCount": 0`.

Because this PR changes runtime configuration semantics (`CODEX_LB_DATA_DIR`, home-dir/container precedence, and derived paths), the OpenSpec gate should include a `specs/<capability>/spec.md` delta with `## ADDED`/`## MODIFIED Requirements` and scenarios. The code/tests look directionally correct, but please add the missing parsed OpenSpec delta before merge.

Soju06:
@codex review

Soju06:
Follow-up pushed in `a66de748e45e34d368155513b5fdd25deb050e74`.

Verification:
- `git diff --check`
- `uv run openspec validate data-dir-env-var-home-exists --strict`
- `uv run openspec change show data-dir-env-var-home-exists --json --deltas-only` → `deltaCount: 1`
- `uv run pytest tests/unit/test_settings_home_dir.py tests/unit/test_proxy_utils.py::test_response_create_dump_uses_configured_data_dir -q`

Live gates are now clean on the new head: CI is green, Codex returned "Didn't find any major issues", and unresolved review threads are zero.

Issue #708, opened 2026-05-19
feat: CPU spikes every minute caused by usage refresh with large `usage_history` table


Art4DDD:
### Pre-flight checklist

- [x] I searched existing issues and discussions for similar proposals.
- [x] This is a concrete proposal, not an open-ended question.

### Summary

I am seeing regular CPU spikes approximately once per minute, even when the dashboard/admin UI is closed.

After checking logs and database activity, the spikes seem to come from the backend usage refresh scheduler, not from frontend polling.

### Environment

- Deployment: Docker on VPS
- Database: SQLite
- Active accounts: 44
- `usage_refresh_enabled = true`
- `usage_refresh_interval_seconds = 60`

### Observed behavior

CPU periodically jumps to high values, sometimes above 100%.
This happens even when the dashboard/admin page is not open.
The `usage_history` table currently has about `1,077,674` rows.
Every minute, around `88` new rows are inserted into `usage_history`.

That roughly matches:

`44 active accounts * 2 usage windows = 88 rows`

### Suspected root cause

The scheduler in `app/core/usage/refresh_scheduler.py` runs every 60 seconds and refreshes usage for all accounts.

The expensive part may be in `app/modules/usage/repository.py`, specifically `UsageRepository.latest_by_account()`.

On SQLite, this uses a `row_number() over partition by account_id` query over the `usage_history` table. When the table grows past 1M rows, this becomes expensive.

Another possible issue is `UsageRepository.add_entry()`.

It appears to commit each inserted usage row individually. With around 44 active accounts and two windows per account, this can mean around 88 commits every minute.

### Expected behavior

The backend should not create large recurring CPU spikes every minute as `usage_history` grows.

Usage refresh should ideally scale mostly with the number of active accounts, not with the total size of the historical table.

### Possible improvements

- Avoid writing a new usage row if the usage values have not changed.
- Batch insert usage rows and commit once per refresh cycle.
- Optimize `latest_by_account()` for SQLite.
- Add a separate “latest usage snapshot” table, while keeping `usage_history` only for charts/history.
- Add retention cleanup for old `usage_history` rows.
- Aggregate old usage samples into minute/hour buckets instead of keeping all raw samples forever.

### Why this matters

With 40-50 accounts, the service can create thousands of `usage_history` rows per hour. After a few days, the table can grow beyond 1M rows.

At that point, the usage refresh job causes recurring CPU spikes even without dashboard traffic, which can affect latency on smaller VPS instances.

### Area

[...]

PR #861, opened 2026-05-31
fix(accounts): avoid SQLite window plan for usage trends

### LINKED ISSUE #859: bug(accounts): account trends endpoint is slow on SQLite due to row_number query plan
### Pre-flight checklist

- [x] I searched existing issues for similar reports.
- [x] I'm on the latest patch release of my minor version.

### codex-lb version

`ghcr.io/soju06/codex-lb:1.19.0`

### Affected account plan(s)

Not account-plan specific. Reproduced from the dashboard account trends endpoint for accounts with larger local SQLite `usage_history`.

### Symptom category

Dashboard performance / account trends latency

### What does the dashboard / logs show?

Opening account trends in the dashboard can take about **23–25 seconds** per request and causes the `codex-lb` container CPU to spike above 100%.

The slow endpoint appears to be:

```text
GET /api/accounts/{account_id}/trends
```

The related implementation seems to be:

```text
UsageRepository.trends_by_bucket()
```

Environment / local DB shape:

```text
Deployment: Docker on Synology NAS
Database: SQLite
DB path: /var/lib/codex-lb/store.db
DB size: ~150 MB
Accounts: 8
usage_history: ~180,255 rows
request_logs: ~16,862 rows
additional_usage_history: ~20,671 rows
```

Measured from inside the container/service layer, bypassing browser/network/auth overhead:

| Operation | Observed latency |
|---|---:|
| Accounts list / service accounts | ~0.72s |
| Dashboard overview 7d | ~2.8–3.3s |
| Account trends for smaller-history accounts | ~1.2s |
| Account trends for larger-history accounts | ~23–25s |

Example slow account-trends timings:

```text
service_trends_36fb5d4f avg ~23.95s
service_trends_9a99cfe1 avg ~24.30s
service_trends_a29215f8 avg ~23.45s
service_trends_9c9b7bec avg ~23.82s
```

Those slower accounts had roughly ~28k–36k `usage_history` rows each.

At idle after the slow request, host/container resources looked normal:

```text
codex-lb CPU ~0.76%
RAM ~432 MiB
NAS load average ~1.11, 1.44, 1.74
```

During the slow dashboard request, the `codex-lb` container CPU spiked above 100%.

### What did you expect?

Dashboard account trends should return in sub-second or low-single-digit seconds at this SQLite data size, without saturating CPU.

### Last observed quota reset time (UTC)

Not applicable. This does not appear to be a quota reset issue.

### When did it start?

After the local SQLite `usage_history` table grew to roughly ~180k rows. Accounts with smaller history are much faster; accounts with ~28k–36k rows each show the slow ~23–25s behavior.

### Additional context

I found a possibly related issue, #708, but this report is more focused on the dashboard account trends endpoint/query rather than the background usage refresh scheduler.

Root-cause hypothesis: the SQLite query generated by `trends_by_bucket()` uses an aggregate subquery plus a `row_number()` window subquery, then joins the two derived subqueries.

The slow generated SQL shape includes a planner-sensitive join condition like:

```sql
anon_2.row_number = ?
```

On the same DB/data, SQLite appears to choose a poor plan for the parameterized version, scanning both derived/temp subqueries:

```text
SCAN anon_2
SCAN anon_1
USE TEMP B-TREE FOR ORDER BY
```

That shape measured around **~23–25 seconds**.

When the equivalent SQL uses a literal constant in the join condition:

```sql
anon_2.row_number = 1
```

SQLite chooses a much better plan with automatic indexing / bloom filtering:

```text
BLOOM FILTER ON anon_1 (window=? AND account_id=? AND bucket_epoch=?)
SEARCH anon_1 USING AUTOMATIC COVERING INDEX (...)
```

That literal version measured around **~0.23–0.31 seconds** on the same DB/data.

Possible SQLite-specific fixes:

1. Use literal SQL for the `row_number = 1` join condition on SQLite.
2. Materialize the aggregate/latest bucket result differently.
3. Split into simpler queries and combine in Python.
4. Add/query with an index or expression that avoids temp subquery nested scans.
5. Use a SQLite-specific optimized path for `trends_by_bucket()`.

1llu5ion:
## Summary

Optimize `UsageRepository.trends_by_bucket()` on SQLite so the account trends endpoint no longer uses a `row_number()` window-function plan to fetch latest per-bucket metadata.

The first scalar-subquery attempt was tested against a live SQLite database and regressed. This revision uses a two-stage aggregate instead:

1. aggregate each `(bucket, account, window)` once, including `max(recorded_at)`
2. resolve the deterministic latest row with `max(id)` among rows at that max timestamp
3. join that latest id back to `usage_history` for metadata

This preserves the existing latest-metadata semantics, including the recorded-at/id tie-breaker, while avoiding SQLite's expensive window-function plan.

## Type of change

Linked issue: Fixes #859

## OpenSpec

- [ ] This PR includes / updates an OpenSpec change
- [ ] This PR touches a codex-faithful path (image pipeline, request/response

Change directory: N/A

## Changes

- Adds a SQLite-specific trends query path that aggregates buckets once, then resolves latest per-bucket metadata via `max(recorded_at)` + deterministic `max(id)` tie-break instead of a `row_number()` window query.
- Leaves the existing PostgreSQL query path unchanged.
- Adds an integration regression test that captures the emitted SQLite trends SQL and asserts the latest-metadata path does not use `row_number()` while preserving latest metadata values.

## Live SQLite benchmark

Benchmarked against a live SQLite `usage_history` table with 184,654 rows, using a read-only transaction snapshot and 5 runs per query shape.

All results matched the baseline rows/metadata.

```text
-- ALL ACCOUNTS --
baseline old row_number window:       median 1252.71 ms
rejected scalar-subquery attempt:     median 4533.12 ms
after this fix, max-time/max-id path: median  665.90 ms

-- HEAVIEST SINGLE ACCOUNT --
baseline old row_number window:       median 240.67 ms
rejected scalar-subquery attempt:     median 850.67 ms
after this fix, max-time/max-id path: median 137.43 ms

Confirmation rerun for fixed path:
all accounts: median 663.22 ms
one account:  median 132.58 ms
```

## Test plan

```text
uv run pytest tests/integration/test_usage_repository.py::test_trends_by_bucket_sqlite_avoids_window_function_for_latest_metadata -q
# 1 passed

uv run pytest tests/integration/test_usage_repository.py -q
# 8 passed, 1 skipped

uv run ruff check app/modules/usage/repository.py tests/integration/test_usage_repository.py
# All checks passed!

uv run ruff format --check app/modules/usage/repository.py tests/integration/test_usage_repository.py
# 2 files already formatted
```

### DIFF
--- a/app/modules/usage/repository.py
+++ b/app/modules/usage/repository.py
@@ -277,74 +277,153 @@ class UsageRepository:
             conditions.append(UsageHistory.account_id == account_id)
 
         window_expr = _normalized_window_expr()
-        base_rows = (
-            select(
-                bucket_col,
-                UsageHistory.id.label("usage_id"),
-                UsageHistory.account_id.label("account_id"),
-                window_expr.label("window"),
-                UsageHistory.used_percent.label("used_percent"),
-                UsageHistory.reset_at.label("reset_at"),
-                UsageHistory.window_minutes.label("window_minutes"),
-                UsageHistory.recorded_at.label("recorded_at"),
+        if dialect == "sqlite":
+            base_rows = (
+                select(
+                    bucket_col,
+                    UsageHistory.id.label("usage_id"),
+                    UsageHistory.account_id.label("account_id"),
+                    window_expr.label("window"),
+                    UsageHistory.used_percent.label("used_percent"),
+                    UsageHistory.recorded_at.label("recorded_at"),
+                )
+                .where(*conditions)
+                .subquery()
             )
-            .where(*conditions)
-            .subquery()
-        )
 
-        aggregate_rows = (
-            select(
+            aggregate_rows = (
+                select(
+                    base_rows.c.bucket_epoch,
+                    base_rows.c.account_id,
+                    base_rows.c.window,
+                    func.avg(base_rows.c.used_percent).label("avg_used_percent"),
+                    func.count(base_rows.c.usage_id).label("samples"),
+                    func.max(base_rows.c.recorded_at).label("max_recorded_at"),
+                )
+                .group_by(
+                    base_rows.c.bucket_epoch,
+                    base_rows.c.account_id,
+                    base_rows.c.window,
+                )
+                .subquery()
+            )
+
+            latest_ids = (
+                select(
+                    aggregate_rows.c.bucket_epoch,
+                    aggregate_rows.c.account_id,
+                    aggregate_rows.c.window,
+                    func.max(base_rows.c.usage_id).label("usage_id"),
+                )
+                .join(
+                    base_rows,
+                    and_(
+                        base_rows.c.bucket_epoch == aggregate_rows.c.bucket_epoch,
+                        base_rows.c.account_id == aggregate_rows.c.account_id,
+                        base_rows.c.window == aggregate_rows.c.window,
+                        base_rows.c.recorded_at == aggregate_rows.c.max_recorded_at,
+                    ),
+                )
+                .group_by(
+                    aggregate_rows.c.bucket_epoch,
+                    aggregate_rows.c.account_id,
+                    aggregate_rows.c.window,
+                )
+                .subquery()
+            )
+
+            stmt = (
+                select(
+                    aggregate_rows.c.bucket_epoch,
+                    aggregate_rows.c.account_id,
+                    aggregate_rows.c.window,
+                    aggregate_rows.c.avg_used_percent,
+                    aggregate_rows.c.samples,
+                    UsageHistory.reset_at,
+                    UsageHistory.window_minutes,
+                    UsageHistory.recorded_at,
+                )
+                .join(
+                    latest_ids,
+                    and_(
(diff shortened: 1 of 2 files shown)


Soju06:
@codex review

1llu5ion:
Small housekeeping note: I accidentally pushed two dashboard-performance follow-up commits to this branch while preparing separate work related to #708. I force-pushed this PR branch back to the reviewed `fix(accounts): avoid SQLite window plan for usage trends` commit, so this PR is scoped to the account trends SQLite window-plan fix again.

The dashboard overview work has been moved to a separate draft PR: #866.

Soju06 (approved):
Reviewed owner-side for #859 SQLite account-trends fix. Local focused tests, SQLite SQL/plan probe, OpenSpec validation, current-main merge probe, CI, Codex verdict, and review threads are clean.

PR #938, opened 2026-06-05
chore: release v1.20.0-beta.3

Soju06:
## Summary
- Prepares beta release v1.20.0-beta.3 for the 1.20.0 stable train.
- Updates release-managed version files only; release-please remains the stable release owner.
- Merging this PR will publish v1.20.0-beta.3 as a GitHub prerelease and trigger package/image/chart publishing.
- Synced automatically from release-please PR #806; no manual beta workflow dispatch is required.

## Changes since v1.20.0-beta.2
- fix(security): harden CodeQL alert surfaces (#935) (
2992b8)
- fix(accounts): dedupe request usage rows by request id (#904) (
86e235)
- feat(accounts): add account list sort controls (#897) (
0e413e)
- feat(accounts): add dashboard action for account force-probe (#895) (
72222c)
- feat(api-keys): add key overview usage stats (#900) (
ed8caa)
- feat(ui): log the User-Agent and store it in database (#882) (
1a8e11)
- feat(dashboard): support weekly pace working days (#901) (
7abbcf)
- fix(acc): create `monthly` window for the `free` account due to the policy change (#909) (
50b9ad)
- feat: add reports page with cost/token charts and CSV export (#854) (
f5fcfa)

## Install after publish
```bash
uvx --from "codex-lb==1.20.0b3" codex-lb
docker pull ghcr.io/Soju06/codex-lb:1.20.0-beta.3
helm install codex-lb oci://ghcr.io/soju06/charts/codex-lb --version 1.20.0-beta.3 --devel
```

## Promotion
Merge the normal release-please PR to promote this beta-tested train to 1.20.0.

PR #946, opened 2026-06-07
chore: release v1.20.0-beta.3

Soju06:
## Summary
- Replaces #938 with the same `v1.20.0-beta.3` release metadata plus CI unblock fixes
- Adds an Alembic merge node so current migration heads collapse to one release head
- Gives the account-list status filter an explicit accessible name and updates the test to target the correct combobox
- Keeps the free-account monthly migration regression pinned to the historical schema instead of using the current repository before later dashboard columns exist

## Why this branch
#938 has `maintainerCanModify=false`, so I couldn't push these CI fixes directly to the original release branch.

## Test Plan
- `make migration-check`
- `PYTHONDONTWRITEBYTECODE=1 uv run pytest tests/unit/test_db_migrate.py tests/integration/test_migrations.py -q`
- `uv run ruff check app/db/alembic/versions/20260607_000000_merge_weekly_monthly_useragent_heads.py tests/integration/test_migrations.py`
- `cd frontend && npx --yes bun@1.3.14 run test src/features/accounts/components/account-list.test.tsx`
- `cd frontend && npx --yes bun@1.3.14 run typecheck`
- `git diff --check`

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "1.20.0-beta.2"  # x-release-please-version
+__version__ = "1.20.0-beta.3"  # x-release-please-version
 __all__ = ["app", "__version__"]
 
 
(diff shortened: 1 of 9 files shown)


Soju06:
@codex review

PR #992, opened 2026-06-12
[codex] fix accounts add button placement

plastictaste:
## Summary

- move the Accounts page Add account trigger out of the scrollable account list and into the persistent list controls
- preserve the existing add-account chooser, OAuth, and import flows
- add a regression test plus OpenSpec delta for the visible placement requirement

## Why

With larger account pools, the Add account action was rendered as the last item inside the scrollable account list. Operators had to scroll through every existing account before they could add another one.

## Validation

- `npx vitest run src/features/accounts/components/account-list.test.tsx`
- `npx tsc -b --pretty false`
- `npx vite build`
- Playwright smoke test against `http://127.0.0.1:2455/accounts` confirmed the Add account button is visible above the scroll region and is not contained by `[data-testid="account-list-scroll-region"]`

Not run:

- `openspec validate fix-account-add-button-position --strict` because the `openspec` CLI is not installed in this shell.


plastictaste:
@codex review

plastictaste:
@codex review

Komzpa:
@codex review

PR #981, opened 2026-06-11
fix(ci): ignore stale and resolved Codex inline findings

Komzpa:
## Summary

- classify inline Codex review comments by `original_commit_id` instead of trusting a rebased `commit_id` alone
- filter unresolved Codex review threads with the same current-head/original-commit evidence rule, so stale rebased threads do not keep `🤖 codex: needs work`
- ignore resolved or outdated inline Codex findings when computing current-head review state
- classify duplicate check-run names by the latest run start/creation time, so superseded failed or cancelled runs do not override a newer rerun on the same head
- add OpenSpec coverage and regression tests for stale rebased comments, resolved inline findings, unresolved inline findings, and duplicate check-run recency

Absorbs #1002.

## Tests

```bash
uv run pytest tests/unit/test_sync_codex_ok_labels.py
uv run ruff format --check .github/scripts/sync_codex_ok_labels.py tests/unit/test_sync_codex_ok_labels.py
uv run ruff check .github/scripts/sync_codex_ok_labels.py tests/unit/test_sync_codex_ok_labels.py
uv run ty check .github/scripts/sync_codex_ok_labels.py tests/unit/test_sync_codex_ok_labels.py
uv run openspec validate fix-codex-label-current-head-evidence --strict
uv run openspec validate ignore-resolved-codex-inline-findings --strict
uv run openspec validate --specs
git diff --check origin/main...HEAD
```

PR #1007, opened 2026-06-15
fix(ci): ignore resolved Codex inline findings

Komzpa:
## Summary
- ignore resolved/outdated Codex inline finding comments when deriving current-head label state
- keep unresolved review threads as the blocking source of truth
- document the GitHub automation requirement and add unit coverage

## Tests
- uv run pytest tests/unit/test_sync_codex_ok_labels.py


Komzpa:
Closing as superseded by #981. #981 includes this resolved-inline-finding fix and also folds in the related current-head/original-commit and duplicate-check-run label-sync fixes, so it should be the surviving PR for this CI-label area.

PR #875, opened 2026-06-02
feat(proxy): repair account-bound egress routing

Komzpa:
## Summary

- replaces #873 because its source branch is `Soju06:repair/pr-869-proxy-pool-egress` with `maintainerCanModify=false`
- folds the original #869 account-bound egress repair scope into a current `main` branch
- keeps local codex-lb account slots separate from raw upstream ChatGPT account IDs across HTTP, compact, websocket, transcribe, file, model, and usage validation paths
- uses account-scoped HTTP leases and account-local cache invalidation for account-bound egress, without sending local account IDs as upstream `chatgpt-account-id`
- adds server-owned per-account Codex installation IDs and strips client-supplied installation metadata from account-bound upstream requests
- adds settings API/UI support for upstream proxy endpoints, pools, account bindings, endpoint tests, and default-pool routing
- treats `407 Proxy Authentication Required` as a failed endpoint test so bad proxy credentials are not reported as connected
- pauses proxy-required `auth.json` imports until an active account binding exists, then reactivates the imported account when that binding is saved
- reactivates `proxy_unreachable` deactivated accounts when an active binding is saved, while leaving session-ended/reauth accounts deactivated
- folds owner-quota full-resend replay, stale WebSocket anchor replay hardening, duplicate stale-anchor replay prevention, selected bridge-restart continuity anchors, HTTP bridge WebSocket header / 1011 affinity hardening, SSE CR/LF-only line boundary parsing, and HTTP bridge admission/startup/stream-contract handling
- keeps #991's newer compact-timeout/budget path instead of #993's older local compact fallback tail
- credits SHAREN, rupebac, and folded contributor authors in the branch history / contributor metadata where applicable

## Folded PRs and Follow-ups

- Surviving PR: #875
- Supersedes/replaces: #873
- Folds/includes:
  - the original #869 repair scope
  - selected #804/#875 proxy feedback that applies to the pool-based design
  - SHAREN's SOCKS credential report from #804
  - SHAREN's numeric-port submit fix as applied to the replacement endpoint dialog
  - rupebac's `port/token-invalidation-routing-pr875` branch and `auth.json` import/proxy-validation caveat
  - #979 owner quota full resend replay
  - #991 stale WebSocket anchors, compact budget, and prewarm accounting
  - #1009 duplicate/overlapping stale-anchor replay hardening
  - the useful bridge-restart continuity anchor part of #993
  - #1017 HTTP bridge WebSocket headers and 1011 hard affinity
[...]

PR #1011, opened 2026-06-15
fix(sse): split SSE event blocks only on CR/LF/CRLF (Closes #1010)

### LINKED ISSUE #1010: bug: SSE parsing uses str.splitlines(), corrupting payloads with U+2028/U+2029 and other Unicode line boundaries
## Summary

`_extract_sse_data_lines` in `app/core/utils/sse.py` splits SSE event blocks with `str.splitlines()`. Python's `splitlines()` breaks on a broad set of Unicode boundaries — including `\v` (0x0B), `\f` (0x0C), `\x1c`–`\x1e`, NEL (U+0085), Line Separator (U+2028) and Paragraph Separator (U+2029) — whereas the SSE specification delimits lines using **only** CR, LF, or CRLF.

U+2028 and U+2029 are valid **unescaped** characters inside JSON strings, and models routinely emit text containing them. When an upstream Responses event carries such a character in its JSON payload, `splitlines()` splits the line mid-string, so the reassembled `data:` value is truncated and `json.loads` fails. `parse_sse_event` then returns `None`.

## Why it matters

`parse_sse_event` is the central event parser on the upstream receive path (`app/core/clients/proxy.py`). A `None` parse means the event is invisible to per-event logic even though the raw block is still forwarded to the client:

- terminal-event detection (`_RESPONSE_STREAM_TERMINAL_EVENT_TYPES` → `seen_terminal`) can miss `response.completed`/`failed`/`incomplete` when the final payload contains U+2028/U+2029, so the stream loop fails to recognize completion;
- tool-call dedupe, usage accounting, and output-item routing silently skip the affected event.

## Reproduction

```python
import json

payload = {"type": "response.output_text.delta", "delta": "line1 line2"}
# Valid JSON; U+2028 is not required to be escaped (ensure_ascii=False)
data = json.dumps(payload, ensure_ascii=False)
event_block = "event: response.output_text.delta\ndata: " + data

# Mirrors _extract_sse_data_lines:
lines = []
for raw in event_block.splitlines():           # <-- over-splits on U+2028
    if not raw or raw.startswith(":"):
        continue
    field, _, value = raw.partition(":")
    if value.startswith(" "):
        value = value[1:]
    if field == "data":
        lines.append(value)

print(repr("\n".join(lines)))
# '{"type": "response.output_text.delta", "delta": "line1'  -> truncated
json.loads("\n".join(lines))
# json.decoder.JSONDecodeError: Unterminated string ...
```

## Expected behavior

SSE line splitting should follow the spec and break only on CR, LF, and CRLF, so any other character (including U+2028/U+2029) is preserved inside `data:` values and the JSON parses intact.

## Actual behavior

The event fails to parse and is dropped from all per-event handling.

## Suggested fix

Split `event_block` on `\r\n | \r | \n` only (e.g. via a small regex) instead of `str.splitlines()` in `_extract_sse_data_lines`.

## Version

Observed on `main` (`app/core/utils/sse.py`).

ozpool:
## Summary

Closes #1010

`_extract_sse_data_lines` in `app/core/utils/sse.py` split SSE event blocks with `str.splitlines()`. `splitlines()` breaks on a wide set of Unicode boundaries (`\v`, `\f`, `\x1c`-`\x1e`, NEL U+0085, U+2028, U+2029), whereas the SSE spec delimits lines only by CR, LF, or CRLF.

U+2028 / U+2029 are valid unescaped characters inside JSON strings, and models emit them in text. When an upstream Responses event carried one in its JSON payload, `splitlines()` split the `data:` line mid-string, the reassembled value failed `json.loads`, and `parse_sse_event` returned `None`. The raw block was still forwarded to the client, but the event became invisible to receive-path logic such as terminal-event detection, tool-call dedupe, and usage accounting.

## Fix

- split SSE data lines on CR, LF, or CRLF only instead of `str.splitlines()`
- treat CR-only blank lines as complete HTTP streaming SSE event separators
- preserve CR-only event terminators when normalizing legacy upstream event aliases such as `response.text.delta`
- document the parser/framer contract in OpenSpec

## Tests

- `uv run pytest tests/unit/test_sse.py tests/unit/test_proxy_utils.py -q`
- `uv run ruff check app/core/clients/proxy.py app/core/utils/sse.py tests/unit/test_proxy_utils.py tests/unit/test_sse.py`
- `openspec validate --specs`
- `git diff --check`

### DIFF
--- a/app/core/clients/proxy.py
+++ b/app/core/clients/proxy.py
@@ -779,7 +779,7 @@ def _remaining_total_timeout(timeout_seconds: float | None, started_at: float, n
 
 
 def _find_sse_separator(buffer: bytes | bytearray) -> tuple[int, int] | None:
-    separators = (b"\r\n\r\n", b"\n\n")
+    separators = (b"\r\n\r\n", b"\n\n", b"\r\r")
     positions = [(buffer.find(separator), len(separator)) for separator in separators]
     valid_positions = [position for position in positions if position[0] >= 0]
     if not valid_positions:
--- a/app/core/clients/proxy.py
+++ b/app/core/clients/proxy.py
@@ -9,6 +9,7 @@ import ipaddress
 import json
 import logging
 import os
+import re
 import socket
 import time
 from contextlib import asynccontextmanager
@@ -95,6 +96,7 @@ _SSE_EVENT_TYPE_ALIASES = {
     "response.audio.delta": "response.output_audio.delta",
     "response.audio_transcript.delta": "response.output_audio_transcript.delta",
 }
+_SSE_LINE_BOUNDARY_RE = re.compile(r"\r\n|\r|\n")
 _RESPONSE_STREAM_TERMINAL_EVENT_TYPES = frozenset(
     {
         "response.completed",
@@ -1009,12 +1011,16 @@ def _normalize_sse_event_block(event_block: str) -> str:
         line_separator = "\n"
         terminator = "\n\n"
         body = event_block[: -len(terminator)]
+    elif event_block.endswith("\r\r"):
+        line_separator = "\r"
+        terminator = "\r\r"
+        body = event_block[: -len(terminator)]
     else:
         line_separator = "\r\n" if "\r\n" in event_block else "\n"
         terminator = ""
         body = event_block
 
-    lines = body.splitlines()
+    lines = _SSE_LINE_BOUNDARY_RE.split(body)
     if not lines:
         return event_block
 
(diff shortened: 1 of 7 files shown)


Soju06:
@codex review

Komzpa:
@codex review

Soju06:
@codex review

Komzpa:
#1011 has been folded into #875.

Carried scope: SSE event parsing now splits blocks only on CR, LF, or CRLF so U+2028/U+2029 and other non-SSE line separators inside JSON strings stay visible to Responses receive-path logic. The original ozpool commit is preserved in #875's branch history.

Survivor: https://github.com/Soju06/codex-lb/pull/875

PR #1034, opened 2026-06-16
fix(quota-planner): normalize decision datetimes for persistence

Soju06:
## Summary
- Normalize quota planner decision datetimes before writing to timezone-naive DB columns
- Cover both `log_decision()` inserts and `update_decision_status()` updates
- Add regression coverage for aware UTC inputs and naive passthrough
- Add OpenSpec change for the quota planner persistence contract

## Root cause
113 prod logged `Quota planner tick failed` with asyncpg/Postgres:

```text
TypeError: can't subtract offset-naive and offset-aware datetimes
```

Planner actions are produced from timezone-aware UTC scheduler instants, but `quota_planner_decisions.scheduled_at` and `executed_at` are timezone-naive `DateTime` columns. Passing aware datetimes into those columns can fail at the asyncpg bind/encode layer.

## Test Plan
- `uv run pytest tests/unit/test_quota_planner.py tests/integration/test_quota_planner_api.py -q`
- `openspec validate fix-quota-planner-db-datetimes --strict`
- `uvx ruff check .`
- `uvx ruff format --check .`
- `uv run ty check`

## Notes
- No schema migration: this preserves the existing naive-UTC DB contract
- JSON audit snapshots may still retain ISO strings with timezone offsets for operator context

### DIFF
--- a/app/modules/quota_planner/repository.py
+++ b/app/modules/quota_planner/repository.py
@@ -7,7 +7,7 @@ from datetime import datetime, timedelta
 from sqlalchemy import Integer, and_, cast, func, literal, select, update
 from sqlalchemy.ext.asyncio import AsyncSession
 
-from app.core.utils.time import utcnow
+from app.core.utils.time import to_utc_naive, utcnow
 from app.db.models import (
     Account,
     QuotaPlannerDecision,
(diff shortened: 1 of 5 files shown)


Soju06:
@codex review

PR #1093, opened 2026-06-26
feat: smart HTTP→upstream transport routing (sticky vs single-shot)

Soju06:
## Why

Today every downstream HTTP/SSE request that resolves to an upstream `websocket` transport is **unconditionally pinned to upstream HTTP** in the streaming retry path:

```python
if request_transport == _REQUEST_TRANSPORT_HTTP and upstream_stream_transport == "websocket":
    upstream_stream_transport = "http"
```

That pin (#698) avoids the "half-rendered turn restart" hazard, but throws away the WebSocket path's main operational benefit: **upstream `server_is_overloaded` rejections are absorbable on the WS path** (that's where the overload-absorption retry ladder lives).

Production evidence (request_logs, 1790 `server_is_overloaded` samples):

- **1789 / 1790 (99.9%) of overload rejections happen before the first token is emitted** (`output_tokens == 0`, no `response.*` item event).
- Only **1 / 1790 (0.06%)** rejected after partial output.

So the hazard the pin guards against is statistically negligible in practice, while the pin's cost — denying **all** downstream-HTTP traffic the WS overload-absorption ladder — is paid on every overloaded turn.

A blanket "send all HTTP over WS" is wrong too: the WS upstream path pays a per-request handshake **plus** an admission-control slot (`response_create_gate`), and prewarm only amortizes that across a **multi-turn** session (prewarm fires only when `previous_response_id is None`). A genuinely **single-shot** caller would eat the full WS setup overhead on every call with nothing to amortize it against.

The right rule is **transport selection by request shape**: send session-continuation / cache-affine ("sticky") requests over WebSocket so they get prewarm amortization and overload absorption; send genuinely single-shot requests over HTTP where the lighter POST-and-reuse path wins.

## What changes

Replace the unconditional pin with a **configurable transport routing policy** decided per request:

1. **New global setting** `http_downstream_transport_policy`:
   - `smart` (**new default**) — route by sticky-session signal.
   - `always_http` — preserve today's pin behavior.
   - `always_websocket` — never downgrade; downstream HTTP may use upstream WS when the base transport resolves to WS.
   - `pinned` — alias of `always_http`.

2. **Policy `smart` (sticky-session signal).** A downstream-HTTP request whose base transport resolves to `websocket` keeps upstream WebSocket **iff** any sticky-continuation signal is present (reusing existing `affinity.py` helpers, no reimplementation):
   - `payload.previous_response_id` set, **OR**
   - `prompt_cache_key` present (`_prompt_cache_key_from_request_model`), **OR**
   - a Codex session header (`_sticky_key_from_session_header`), **OR**
   - an `x-codex-turn-state` header (`_sticky_key_from_turn_state_header`).

   Otherwise (single-shot) it falls back to upstream HTTP, exactly as today.

3. **Per-API-key override** `transport_policy_override` (nullable; `smart|always_http|always_websocket`). When set it wins over the global policy for requests authed by that key; null follows the global default. A one-off key (e.g. `soju-graphiti` with no session continuity) can be pinned to `always_http`.

### Precedence (preserved, sits ABOVE the policy)

- Explicit `upstream_stream_transport = "websocket" | "http"` config override wins outright.
- Oversized-payload → HTTP and image / image-generation → HTTP bypasses still force HTTP regardless of policy.
- Native Codex WebSocket clients (`request_transport == "websocket"`) are **unaffected** — they keep their dedicated WS path; this change only governs the downstream-HTTP branch.

The only behavioral delta: under the new `smart` default, sticky downstream-HTTP turns now ride upstream WebSocket (gaining the overload-absorption ladder) instead of being pinned to HTTP.

## Implementation notes

- Decision factored into a pure, unit-testable helper `_resolve_http_downstream_transport(policy, *, payload, headers) -> "http" | "websocket"` (`pinned` resolves like `always_http`).
- Effective policy = per-key override if non-null, else global setting.
- A structured log line records the policy in effect, whether the per-key override applied, the sticky verdict, and the final upstream transport for production triage.
- API-key input paths use strict enum validation (HTTP 400 on bad values); read paths use a lenient normalizer (DB drift → null), matching the established `traffic_class` / `service_tier` pattern.

## Migration

Additive Alembic revision on the current single head:
- `api_keys.transport_policy_override` — nullable, existing rows default null (follow global).
- `dashboard_settings.http_downstream_transport_policy` — NOT NULL, `server_default 'smart'`.

Idempotent guards + symmetric downgrade. Single upgrade head verified.

**Behavior note for release:** the default global policy ships as `smart`, which changes runtime behavior for sticky downstream-HTTP turns (they now ride upstream WS). Operators wanting the legacy guarantee are one setting away (`always_http`).

## Dashboard UI

- Global Settings: `http_downstream_transport_policy` dropdown (default smart), neutral operator-facing labels.
- API-key create/edit: `transportPolicyOverride` control with an explicit "Follow global default" (null) option plus the three values.

## OpenSpec

Change folder: `openspec/changes/smart-http-transport-routing/` (proposal + tasks + delta specs for `responses-api-compat` and `api-keys`). `openspec validate smart-http-transport-routing --strict` → valid.

## Verification

All run locally on this branch:

| Gate | Result |
|---|---|
| `ruff check` | All checks passed |
| `ruff format --check` | 638 files formatted |
| `ty check` | All checks passed |
| backend + integration `pytest` | 724 passed |
| frontend `vitest` (api-key dialogs + routing settings) | 40 passed |
| `openspec validate --strict` | valid |
| `openspec validate --specs` | 30 passed |
| Alembic | single head, symmetric downgrade |

Behavioral tests cover every scenario in the delta spec: smart single-shot→HTTP, smart sticky→WS (one test per signal), always_http/pinned pin, always_websocket keep, per-key override wins, null override follows global, explicit websocket override beats policy, oversized-payload bypass under always_websocket, native WS untouched; plus api-keys create/update/clear/invalid-400/read round-trip and migration default-null.

---

Note: this is a new enhancement driven by overload-absorption analysis, not an issue fix — no `Closes #N`.


Soju06:
@codex review

Soju06:
Fixed in `90d122d3`. Root cause confirmed: the smart-routing path eagerly resolved a base `auto` transport to a concrete `"websocket"` override for sticky requests that keep WS, which made downstream `transport_mode == "websocket"` (not `"auto"`), so `_should_fallback_to_http_after_websocket_status` returned False and a WS handshake rejection (e.g. 426) surfaced as an error instead of transparently retrying over HTTP — a regression vs `main` for the `auto` + sticky case.

Fix: the policy now only ever forces a concrete override when **downgrading to HTTP**. When the decision keeps WebSocket, it passes the base/configured transport through unchanged (`upstream_stream_transport = "http" if policy_transport == "http" else configured_transport`), so `auto` stays `auto` and the existing handshake→HTTP fallback is preserved. Same reasoning applied to `always_websocket`.

Regression coverage added:
- `test_http_downstream_sticky_smart_auto_mode_keeps_websocket_handshake_fallback` — sticky smart request over base `auto`, WS handshake raises 426, asserts the override stays `"auto"` and the turn transparently falls back to HTTP (`response.completed` received).
- `test_http_downstream_sticky_smart_policy_preserves_auto_transport_mode` — asserts the override handed downstream is `"auto"`, not a hard `"websocket"`.
- Existing per-key/no-override scenarios updated to expect `auto` preservation.

OpenSpec delta updated to make this normative (WS-keep MUST NOT convert base `auto` to a hard `websocket`; only downgrades may force a concrete override). `openspec validate --strict` green. ruff/ty/356 targeted tests pass.

Soju06:
@codex review

PR #1121, opened 2026-07-01
fix(proxy): stop local rate-limit selection retry loops

### LINKED ISSUE #1078: bug: single-account proxy hangs on every request after first upstream 429 (v1.20.0 #1000 regression)
## Summary

Since v1.20.0, a single-account deployment of codex-lb hangs on **all** requests after the first upstream 429, and only a process restart clears the stuck state. Direct connection to the same account (bypassing codex-lb) keeps working during the same window.

## Environment

- codex-lb version: `1.20.1` (regression appeared in `1.20.0`)
- Deployment method: Docker Compose (`ghcr.io/soju06/codex-lb:latest`)
- Number of accounts: **1** (single-account pool — most exposed to the bug)
- Routing strategy: default (`capacity_weighted`)
- Codex CLI as downstream client, Responses API over WebSocket

## Symptom

- Proxy runs fine for a while (minutes to tens of minutes).
- Eventually the Codex CLI reports `exceeded retry limit, last status: 429 Too Many Requests, request id: …`.
- After that, **every** subsequent request to the proxy is stuck — clients either see 429 propagated immediately, or hang until their own timeout.
- Dashboard shows the single account as `RATE_LIMITED`.
- `docker restart codex-lb` clears the in-memory state and the proxy is immediately healthy again, until the cycle repeats.
- Bypassing the proxy and hitting the same account directly during the same window works fine.

## Root cause

PR #1000 (`fix(proxy): keep streams alive while account capacity recovers`, commit `8d8061ba`) wraps `_select_account_with_budget_compatible` in a `while True:` retry loop in both `app/modules/proxy/_service/websocket/mixin.py` and `app/modules/proxy/_service/http_bridge/streaming.py`. The loop waits before retrying based on a regex match against the selection error message:

`app/modules/proxy/_service/support.py`
```python
_ACCOUNT_SELECTION_RETRY_HINT_RE = re.compile(r"try again in\s+([0-9]+(?:\.[0-9]+)?)s", re.IGNORECASE)

def _account_selection_recovery_sleep_seconds_from_message(message):
    ...
    retry_hint = _ACCOUNT_SELECTION_RETRY_HINT_RE.search(message)
    if retry_hint is not None:
        return min(max(float(retry_hint.group(1)), 1.0), 300.0)
```

But `select_account` itself produces exactly that message shape when no account is selectable, via `_format_retry_hint`:

`app/core/balancer/logic.py:1006`
```python
def _format_retry_hint(wait_seconds: float) -> str:
    capped = min(max(0.0, wait_seconds), float(SELECTOR_RETRY_HINT_MAX_SECONDS))  # 300s ceiling
    return f"Rate limit exceeded. Try again in {capped:.0f}s"
```

So the loop is self-sustaining:

```
select_account → "Rate limit exceeded. Try again in 300s"
       ↑                                          ↓
       └── #1000 matches the hint, waits 300s ←── wait done, retry selection
```

In a single-account pool, once the only account flips to `RATE_LIMITED`, every new request enters this loop and blocks for up to the request budget. Even after the wait elapses, the local `AccountState.reset_at` may still be in the future, so the next `select_account` returns the same message and the loop re-enters. Restarting the process clears the in-memory `RATE_LIMITED` flag, which is the only reason a bounce "fixes" it.

The `infer_status_from_usage=False` change from #1030 is *not* the cause here — the report is about a real upstream 429 that legitimately puts the account into `RATE_LIMITED`. The problem is that #1000 then refuses to fail fast on locally-generated retry hints.

## Reproduction

1. Deploy v1.20.1 with a single account.
2. Generate enough traffic to trigger one real upstream 429 (rate limit, not quota).
3. Observe that the account goes `RATE_LIMITED` and **all** subsequent requests hang or return 429, with no recovery until `docker restart`.
4. During the stuck window, hit the same account directly — it works.

## Proposed fix

The locally-generated retry hint should not feed back into the capacity-wait loop — `select_account` has already computed the wait, so re-looping just defers the same failure. Skipping the locally-prefixed message restores v1.19.0 fail-fast behavior while preserving #1000's handling of genuine upstream error messages (workspace spend cap, etc.).

`app/modules/proxy/_service/support.py`
```diff
     lowered = message.lower()
     if (
         "require re-authentication" in lowered
         or "all accounts are paused" in lowered
         or "no accounts with a plan" in lowered
         or "no accounts with available additional quota" in lowered
         or "no fresh additional quota data" in lowered
+        or lowered.startswith("rate limit exceeded. try again in")
     ):
         return None
```

I have this patch deployed on a single-account instance and can confirm the global-stall behavior is gone — the 429 is now propagated to the client immediately, which is what v1.19.0 did.

## Workaround

For anyone hitting this before a fix lands:

1. Patch `app/modules/proxy/_service/support.py` as shown above and mount it into the container:
   ```yaml
   volumes:
     - ./patches/support.py:/app/app/modules/proxy/_service/support.py:ro
   ```
2. Or roll back to v1.19.0 (note: schema migrations from v1.20.x are not auto-downgradable, so this requires a separate database).

## Notes

- The blast radius is largest for single-account pools because there is no alternative account to fail over to inside the wait loop.
- For multi-account pools, the bug is masked as long as at least one account is selectable, but the same self-sustaining loop kicks in once all accounts are simultaneously rate-limited.
- The `_format_retry_hint` ceiling of 300s means each loop iteration can hold a request for up to 5 minutes before failing.

Happy to open a PR with the one-line fix + the updated tests if that's useful.

Komzpa:
## Summary

- stop treating the local balancer `Rate limit exceeded. Try again in ...` no-account message as a recoverable account-capacity wait
- keep bounded recovery waits for non-local retry hints
- add an OpenSpec delta and regression coverage for the local-rate-limit loop

Closes Soju06/codex-lb#1078.

## Validation

- `uv run pytest tests/unit/test_proxy_utils.py -q`
- `uv run ruff check app/modules/proxy/_service/support.py tests/unit/test_proxy_utils.py`
- `openspec validate fix-local-rate-limit-selection-loop --strict`
- `git diff --check`

### DIFF
--- a/app/modules/proxy/_service/http_bridge/streaming.py
+++ b/app/modules/proxy/_service/http_bridge/streaming.py
@@ -215,7 +215,10 @@ def _http_bridge_account_capacity_wait_seconds(exc: ProxyResponseError) -> float
     code, message = _proxy_error_code_message(exc)
     if code in {"account_response_create_cap", "account_stream_cap", "capacity_exhausted_active_sessions"}:
         return None
-    return _account_selection_recovery_sleep_seconds_from_message(message)
+    return _account_selection_recovery_sleep_seconds_from_message(
+        message,
+        error_code=code,
+    )
 
 
 def _http_bridge_capacity_wait_plan(
--- a/app/modules/proxy/_service/support.py
+++ b/app/modules/proxy/_service/support.py
@@ -41,7 +41,12 @@ _ACCOUNT_SELECTION_RECOVERY_HEARTBEAT_SECONDS = 10.0
 _ACCOUNT_SELECTION_RETRY_HINT_RE = re.compile(r"try again in\s+([0-9]+(?:\.[0-9]+)?)s", re.IGNORECASE)
 
 
-def _account_selection_recovery_sleep_seconds_from_message(message: str | None) -> float | None:
+def _account_selection_recovery_sleep_seconds_from_message(
+    message: str | None,
+    *,
+    error_code: str | None = None,
+    suppress_uncoded_local_selector_hint: bool = False,
+) -> float | None:
     message = (message or "").strip()
     if not message:
         return None
@@ -53,6 +58,10 @@ def _account_selection_recovery_sleep_seconds_from_message(message: str | None) 
         or "no accounts with a plan" in lowered
         or "no accounts with available additional quota" in lowered
         or "no fresh additional quota data" in lowered
+        or (
+            (error_code == "no_accounts" or (suppress_uncoded_local_selector_hint and error_code is None))
+            and lowered.startswith("rate limit exceeded. try again in")
+        )
     ):
         return None
 
@@ -74,7 +83,11 @@ def _account_selection_recovery_sleep_seconds_from_message(message: str | None) 
 
 
 def _account_selection_recovery_sleep_seconds(selection: AccountSelection) -> float | None:
-    return _account_selection_recovery_sleep_seconds_from_message(selection.error_message)
+    return _account_selection_recovery_sleep_seconds_from_message(
+        selection.error_message,
+        error_code=selection.error_code,
+        suppress_uncoded_local_selector_hint=selection.account is None,
+    )
 
 
 def _account_capacity_wait_payload(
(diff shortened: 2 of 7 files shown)


Soju06:
@codex review

Soju06:
@codex review

PR #1138, opened 2026-07-05
fix: list Codex shell-only models in backend catalog

Komzpa:
## Summary

- keep Codex shell-command models in `/backend-api/codex/models` even when they are not OpenAI-compatible API models
- keep other `supported_in_api=false` models out of the Codex catalog, `/v1/models`, `/api/models`, and the OpenAI-style `data` payload
- restore the live snapshot as authoritative after model-registry refresh instead of extending it with bootstrap-only models
- update the API key OpenSpec model-list contract for the Codex-native shell-model exception

## Tests

- `uv run pytest tests/unit/test_model_registry.py tests/integration/test_v1_models.py -q`


Komzpa:
@codex review

PR #1154, opened 2026-07-09
fix(proxy): hide blank reasoning summary comments

Komzpa:
## Summary
- strip standalone blank `<!-- -->` placeholders from reasoning `summary_text` items before forwarding Responses output
- limit the cleanup to reasoning summaries so assistant-visible message content and non-empty comments are not rewritten
- add OpenSpec coverage and regressions for Codex CLI streaming and non-streaming response reconstruction

## Tests
- `uv run pytest tests/integration/test_proxy_responses.py -k 'reasoning_output or reasoning_summary' -q`
- `openspec validate strip-reasoning-summary-comments --strict`
- `uv run ruff check app/modules/proxy/api.py tests/integration/test_proxy_responses.py`
- `uv run ruff format --check app/modules/proxy/api.py tests/integration/test_proxy_responses.py`
- `uv run ty check app/modules/proxy/api.py tests/integration/test_proxy_responses.py`

### DIFF
--- a/app/modules/proxy/api.py
+++ b/app/modules/proxy/api.py
@@ -5820,6 +5820,10 @@ def _could_be_blank_html_comment_line(text: str) -> bool:
     return not remainder[3:].strip(" \t\r")
 
 
+def _is_reasoning_summary_interleavable_event(event_type: JsonValue | None) -> bool:
+    return isinstance(event_type, str) and (event_type == "response.in_progress" or event_type.startswith("codex."))
+
+
 async def _normalize_reasoning_summary_stream(stream: AsyncIterator[str]) -> AsyncIterator[str]:
     pending: dict[tuple[str | None, int | None, int | None], list[dict[str, JsonValue]]] = {}
 
@@ -5842,8 +5846,12 @@ async def _normalize_reasoning_summary_stream(stream: AsyncIterator[str]) -> Asy
             continue
         event_type = payload.get("type")
         event_key = _reasoning_summary_delta_key(payload)
-        if pending and not (
-            event_type in _REASONING_SUMMARY_DELTA_TYPES | _REASONING_SUMMARY_DONE_TYPES and event_key in pending
+        if (
+            pending
+            and not _is_reasoning_summary_interleavable_event(event_type)
+            and not (
+                event_type in _REASONING_SUMMARY_DELTA_TYPES | _REASONING_SUMMARY_DONE_TYPES and event_key in pending
+            )
         ):
             for pending_key in tuple(pending):
                 for buffered in flush(pending_key):
--- a/app/modules/proxy/api.py
+++ b/app/modules/proxy/api.py
@@ -205,7 +205,14 @@ from app.modules.usage.updater import UsageUpdater
 
 logger = logging.getLogger(__name__)
 
-_REASONING_SUMMARY_BLANK_HTML_COMMENT_RE = re.compile(r"(?m)^[ \t]*<!--\s*-->[ \t]*(?:\r?\n)?")
+_REASONING_SUMMARY_BLANK_HTML_COMMENT_RE = re.compile(r"(?m)^[ \t]*<!--\s*-->[ \t]*(?:\r?\n|\Z)")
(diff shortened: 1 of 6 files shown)


Soju06:
@codex review

ozpool (commented):
I went through this one carefully and it looks correct and well-scoped.

A few things I checked:

- The interception is behavior-preserving. `reasoning` is already in `_PUBLIC_RESPONSE_OUTPUT_ITEM_TYPES`, so today a reasoning item is returned verbatim via `dict(item)`. The new `if item_type == "reasoning"` branch pre-empts that, and for an unchanged summary `_normalize_reasoning_output_item` still returns `dict(item)`, so nothing changes except the blank-comment strip. Non-`summary_text` parts and non-`reasoning` items are untouched.

- Both proxy paths are covered, which matches the spec's "streamed and collected" claim. `_normalize_public_output_item` is called from the streaming `response.output_item.added/done` handler and from the collected reconstruction loop after `_merge_collected_output_items`, so the cleanup lands on both `/backend-api/codex/responses` and the `/v1/responses` non-streaming path. Nice that both are actually exercised: the new test covers the backend-api stream, and the updated `test_v1_responses_non_streaming_reconstructs_reasoning_output` feeds `"Need more steps\n\n<!-- -->"` and asserts the reconstructed summary is `"Need more steps"`.

- The regex is tight in the right way: `<!--\s*-->` only matches when the interior is whitespace, so a meaningful comment like `<!-- keep -->` is left alone, while `<!-- -->` (and a two-line `<!--\n-->`) is removed. Anchoring with `(?m)^[ \t]*...[ \t]*(?:\r?\n)?` keeps it to standalone lines.

One small edge worth a thought, not a blocker: `_strip_blank_html_comment_lines` only `rstrip()`s, so a blank marker that lands in the *middle* of a summary (rather than trailing after the heading) removes the comment line but leaves the surrounding blank lines, e.g. `"A\n\n<!-- -->\n\nB"` -> `"A\n\n\nB"`. In practice the marker shows up trailing after the heading so this may never occur, but if you want to be defensive you could collapse the resulting blank run. Happy either way.

Soju06 (commented):
Thanks — the placeholder symptom is real and limiting the cleanup to reasoning `summary_text` is the right scope. Three things to fix before this is ready:

1. **Unconditional `.rstrip()` rewrites placeholder-free summaries.** `_strip_blank_html_comment_lines` calls `.rstrip()` even when the regex removed nothing, so a summary like `"Need more steps\n"` (or Markdown that relies on trailing whitespace) gets rewritten despite containing no placeholder. This matches the unresolved Codex thread (r3549786849) — I reproduced it from the code. Only strip trailing whitespace when the regex actually removed a placeholder, and add a byte-identical passthrough test for a summary with trailing whitespace and no `<!-- -->`.

2. **The streaming delta path still leaks the marker.** Only `response.output_item.added`/`.done` events go through `_normalize_public_output_item`; `response.reasoning_summary_text.delta` and `response.reasoning_summary_part.*` events are forwarded untouched. A live Codex TUI that renders summaries from deltas will still show `<!-- -->`, so the original symptom persists on the streamed path. Either clean the delta path too (careful: the marker can be split across delta boundaries, so a per-delta regex is not sufficient) or demonstrate that the TUI renders summaries only from `output_item.done` and record that in the change notes. Either way, extend the fake-stream test to include delta events.

3. **Regex vs spec wording.** The spec delta says "standalone blank HTML comment placeholder lines", but `_REASONING_SUMMARY_BLANK_HTML_COMMENT_RE` has no end-of-line anchor after the optional trailing whitespace, so `<!-- -->some text` at line start also has the marker stripped even though that line is not standalone. Anchor the regex so it only matches when the comment is the entire line, or adjust the spec text to describe what the code does.

Once those land, re-run `@codex review` on the new head and undraft.

Soju06:
@codex review

Soju06:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

PR #1159, opened 2026-07-09
fix(proxy): preserve non-message developer input items

kvz:
## Summary

Codex responses-lite clients (gpt-5.6 models) deliver tool definitions inside `input` as an `additional_tools` item; the instruction-hoisting normalizer silently dropped that item, so every gpt-5.6 session through codex-lb ran without any tools.

## Type of change

## OpenSpec

Change directory: `openspec/changes/preserve-non-message-developer-input-items/`

## Changes

- `_normalize_responses_input_instructions()` now only hoists `system`/`developer` input items that are actual messages (`type` omitted or `"message"`); other items pass through to upstream unchanged.
- Regression tests for `ResponsesRequest` and `ResponsesCompactRequest` with an `additional_tools` input item.
- OpenSpec requirement recording that non-message developer input items must be preserved.

## Details

gpt-5.6 models advertise `use_responses_lite: true` and `tool_mode: code_mode_only`. In this mode the Codex CLI (0.144.0) sends **no top-level `tools` field** — tool definitions ride in the first input item:

```json
{"type": "additional_tools", "role": "developer", "tools": [{"type": "custom", "name": "exec"}, ...]}
```

The hoisting normalizer treats every `system`/`developer`-role input item as an instruction message. `additional_tools` has no `content`, so it contributed no instruction text, produced no preserved item, and was removed from `input` entirely. Upstream then received a well-formed request with no tools anywhere and the model would answer that no terminal/filesystem tool is exposed (or emit `to=functions.exec` as plain text). gpt-5.5 and earlier (classic top-level `tools`) were unaffected, which made this look like a model regression rather than a proxy bug.

Typeless role items (`{"role": "system", "content": "..."}` from OpenAI-compatible clients) keep the existing hoisting behavior.

## Test plan

Verified against a live proxy by asking gpt-5.6-sol to read a file with unguessable random content via shell (bluff-proof probe):

- before: model answers `TOOLLESS` (request on the wire confirmed to have lost the `additional_tools` item)
- after: model reads the file over ws and HTTP transports, at xhigh and ultra efforts; gpt-5.5 classic path unaffected

```
uv run pytest tests/unit/test_openai_requests.py -q
104 passed in 0.37s
```

### DIFF
--- a/app/core/openai/requests.py
+++ b/app/core/openai/requests.py
@@ -297,6 +297,14 @@ def _normalize_responses_input_instructions(data: JsonValue) -> JsonValue:
         if role not in ("system", "developer"):
             input_items.append(item)
             continue
+        # Only hoist actual message items. Codex responses-lite clients (gpt-5.6+ models with
+        # use_responses_lite/tool_mode=code_mode_only) send non-message developer items such as
+        # {"type": "additional_tools", "role": "developer", "tools": [...]} inside input; those
+        # have no content, so hoisting used to drop them entirely and the model lost all tools.
+        item_type = item_mapping.get("type")
+        if item_type is not None and item_type != "message":
+            input_items.append(item)
+            continue
         instruction_text, preserved_content = _split_responses_instruction_item_content(item_mapping)
         if instruction_text:
             instruction_parts.append(instruction_text)
--- a/openspec/changes/preserve-non-message-developer-input-items/.openspec.yaml
+++ b/openspec/changes/preserve-non-message-developer-input-items/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-07-09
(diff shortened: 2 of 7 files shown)


CapCap:
Can confirm this fix works

receiver1 (approved):

Soju06:
Thanks @kvz — this was a tight, well-tested diagnosis of exactly the right defect, and it was a close call in triage. We merged #1161 for the same underlying bug because it also reconstructs the canonical Responses-Lite signaling (HTTP header + websocket marker) across the bridge/websocket/replay paths, which the normalization fix alone didn't cover. Your generalized guard — never folding any non-message typed developer item — is better defense-in-depth than what merged, so I've filed #1171 to bring that in with credit to this PR. A follow-up PR from you for #1171 would be very welcome. Closing as superseded by #1161.

PR #1158, opened 2026-07-09
fix(proxy): preserve Codex Responses Lite tools for GPT-5.6 models

Aminkbi:
## Summary

Fixes #1157.

GPT-5.6 lite models (`gpt-5.6-sol` / `-terra` / `-luna`, `use_responses_lite: true`, `tool_mode: code_mode_only`) do not send tools in the top-level `tools` field. Codex ≥ 0.144.0 prepends an input item `{"type": "additional_tools", "role": "developer", "tools": [...]}` and marks the request with the internal `x-openai-internal-codex-responses-lite` header (HTTP) or the `ws_request_header_x_openai_internal_codex_responses_lite` `client_metadata` key (websocket) — see `codex-rs/core/src/client.rs`.

codex-lb broke this contract in two places:

1. **`_normalize_responses_input_instructions()`** treated every `developer`/`system` input item as instruction text. The `additional_tools` item has no `content`, so it was silently dropped — the upstream model received no shell/filesystem tools and answered *"I can't inspect the repository because no shell/filesystem tool is available in this session."*
2. **`strip-internal-responses-lite-header`** removes the Lite header from all upstream requests, so even an intact tool bundle would never be processed in lite mode upstream.

## What changes

- **Lite-shaped requests pass through byte-identical**: when `input` contains an `additional_tools` item, the instruction lift is skipped entirely, preserving the tool bundle, the developer instructions message, and `custom_tool_call` / `custom_tool_call_output` items exactly as Codex constructed them.
- **The lite signal is forwarded for native Codex clients, per transport**: `x-openai-internal-codex-responses-lite: true` on upstream HTTP responses/compact requests (initial and retry attempts), and the `ws_request_header_x_openai_internal_codex_responses_lite` key in `client_metadata` on upstream websocket `response.create` payloads — mirroring the Codex client's own transport split.
- **Non-native clients keep the existing strip behavior**, preserving the protection that motivated `strip-internal-responses-lite-header` (upstream rejects the header on models without lite support).

## OpenSpec

- New change: `openspec/changes/preserve-native-responses-lite-passthrough/` (proposal, design, tasks, delta spec). Validates with `openspec validate --strict`.
- `openspec/specs/responses-api-compat/spec.md` gains the requirement from `strip-internal-responses-lite-header` **verbatim** — that change was implemented but never synced to the main spec, and the sync is needed to anchor this change's `MODIFIED` delta. `openspec validate --specs` passes 30/30.

## Verification

[...]

Issue #1157, opened 2026-07-09
bug: GPT-5.6 Sol loses Responses Lite additional_tools and reports no shell/filesystem access


Navid2zp:
### Pre-flight checklist

- [x] I searched existing issues and confirmed this is not a duplicate.
- [x] I am running a supported codex-lb version (see Releases).
- [x] I can reproduce the problem with the steps below.

### codex-lb version

1.20.1

### Deployment method

Docker Compose

### Client used against codex-lb

Codex app (desktop / web)

### ChatGPT account plan(s) involved

Pro

### Model(s) involved

gpt-5.6-sol

### What happened?

When GPT-5.6 Sol is used through codex-lb, Codex receives a normal text response but cannot call local tools. For repository tasks it responds:

```
I can’t inspect the repository because no shell/filesystem tool is available in this session.
```

This appears to be a GPT-5.6 Responses Lite compatibility issue, not a filesystem permission issue.

OpenAI added `gpt-5.6-sol` with:

```json
{
  "tool_mode": "code_mode_only",
  "use_responses_lite": true,
  "shell_type": "shell_command",
  "minimal_client_version": "0.144.0"
}
```

In Responses Lite mode, Codex does not send the tool bundle through the ordinary top-level `tools` field. Instead it prepends this kind of input item:

```json
{
  "type": "additional_tools",
  "role": "developer",
  "tools": [{"type": "custom", "name": "exec", "format": {"type": "grammar", "syntax": "lark"}}]
}
```

However, `app/core/openai/requests.py::_normalize_responses_input_instructions()` treats every `developer`/system input item as a text instruction. Since `additional_tools` has no `content`, it is not appended to `input_items` and is removed when the normalized input replaces the original input.

Because Responses Lite also omits top-level `tools`, the upstream request reaches GPT-5.6 Sol with no shell/filesystem tool definitions.

The same normalization is used by the HTTP, `/v1/responses`, and Responses WebSocket paths.

### What did you expect to happen?

codex-lb should preserve the Responses Lite `additional_tools` input item unchanged so GPT-5.6 Sol receives the grammar-based `exec` tool and can invoke the nested shell/filesystem tools.

The `x-openai-internal-codex-responses-lite: true` header should also be preserved or safely reconstructed for authenticated native Codex requests.

### Steps to reproduce

```markdown
1. Run a supported codex-lb build containing the current Responses request normalization.
2. Configure Codex 0.144.0+ with a custom Responses provider pointing to codex-lb:


[...]

PR #1163, opened 2026-07-10
fix(models): serve the Codex catalog on /v1/models for Codex clients

kevinWangSheng:
## Summary

`GET /v1/models` only speaks the OpenAI-compatible list shape, but Codex clients configured via `openai_base_url` fetch their model catalog from that exact endpoint (always with a `client_version` query parameter). The catalog refresh fails to parse client-side and Codex silently falls back to its bundled model metadata. This PR content-negotiates on `client_version` and serves the Codex catalog to Codex clients, keeping the OpenAI-compatible shape for everyone else.

## Type of change

Linked issue: complements #1157 (the silent metadata half; the request-forwarding half is addressed by #1158/#1159/#1161)

## OpenSpec

- [x] This PR touches a codex-faithful path (request/response shape) and preserves upstream-equivalent behavior

Change directory: `openspec/changes/serve-codex-catalog-on-v1-models/`

## Changes

- `app/modules/proxy/api.py`: `v1_models` returns `_build_codex_models_response` when the request carries a non-empty `client_version` query parameter (codex-rs models-manager appends it unconditionally; OpenAI SDKs never send it), `_build_models_response` otherwise. `response_model=None` since the endpoint now serves two documented shapes.
- Integration tests: negotiated payload equals `GET /backend-api/codex/models`; bare and empty-parameter requests keep the `{"object": "list", "data": [...]}` shape.
- OpenSpec change with a `model-catalog-compat` delta.

## Why this matters (real-world failure)

With GPT-5.6 (`use_responses_lite: true`, `tool_mode: "code_mode_only"` in the client's bundled metadata), a Codex CLI pointed at codex-lb via `openai_base_url` builds responses-lite requests from **stale bundled metadata** no matter what the proxy serves, because its `/v1/models?client_version=…` refresh silently fails to parse. Every session runs without tools and the model reports it has no shell/filesystem access — the same symptom as #1157, but on the catalog layer, so it persists even after the request-forwarding fixes land. Reproduced with codex-cli 0.144.1 against codex-lb 1.20.1: the identical request works once the client receives catalog metadata it can parse.

`CodexModelsResponse` already carries both `models` and the OpenAI-compatible `object`/`data` fields (the dual-shape precedent in the opposite direction), so mixed consumers are unaffected in both directions.

## Test plan

```
uv run pytest tests/integration/test_v1_models.py -q                      # 33 passed
uv run pytest tests/integration/test_v1_models.py \
  tests/unit/test_path_rewrite_middleware.py \
  tests/integration/test_path_rewrite_alias.py \
  tests/integration/test_codex_client_compat.py -q                        # 55 passed
uv run ruff check / ruff format --check on touched files                  # clean
```

## Checklist

- [x] Ran the relevant test subset locally.

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -898,6 +898,15 @@
       "contributions": [
         "code"
       ]
+    },
+    {
+      "login": "kevinWangSheng",
+      "name": "Shawn",
+      "avatar_url": "https://avatars.githubusercontent.com/u/118158941?v=4",
+      "profile": "https://github.com/kevinWangSheng",
+      "contributions": [
+        "code"
+      ]
     }
   ],
   "contributorsPerLine": 7,
--- a/app/modules/proxy/api.py
+++ b/app/modules/proxy/api.py
@@ -823,10 +823,20 @@ async def models(
     return await _build_codex_models_response(api_key)
 
 
-@v1_router.get("/models", response_model=ModelListResponse)
+@v1_router.get("/models", response_model=None)
 async def v1_models(
+    request: Request,
     api_key: ApiKeyData | None = Security(validate_proxy_api_key),
 ) -> Response:
+    # Codex clients pointed at this proxy via `openai_base_url` fetch their model
+    # catalog from `<base_url>/models` and always append a `client_version` query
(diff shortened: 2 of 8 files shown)


Soju06:
@codex review

PR #1173, opened 2026-07-10
feat(models): add GPT-5.6 Codex support

Cortezjohannes:
## Summary
- bootstrap GPT-5.6 Sol, Terra, and Luna in the model registry
- expose low, medium, high, xhigh, max, and ultra reasoning efforts
- retain bootstrap models when live upstream snapshots lag behind
- cover the new metadata and fallback behavior with unit tests

## Verification
- `.venv/bin/pytest tests/unit/test_model_registry.py -q` (18 passed)
- live `gpt-5.6-sol` request at `max` through codex-lb
- app, CLI, and VS Code model catalogs list all three GPT-5.6 variants and all six efforts

### DIFF
--- a/app/core/openai/model_registry.py
+++ b/app/core/openai/model_registry.py
@@ -48,7 +48,14 @@ class ModelRegistrySnapshot:
     fetched_at: float
 
 
-_BOOTSTRAP_WEBSOCKET_PREFERRED_MODEL_PATTERNS = ("gpt-5.5", "gpt-5.5-*", "gpt-5.4", "gpt-5.4-*")
+_BOOTSTRAP_WEBSOCKET_PREFERRED_MODEL_PATTERNS = (
+    "gpt-5.6",
+    "gpt-5.6-*",
+    "gpt-5.5",
+    "gpt-5.5-*",
+    "gpt-5.4",
+    "gpt-5.4-*",
+)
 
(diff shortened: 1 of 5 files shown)


Soju06:
Thanks @Cortezjohannes — you were early on GPT-5.6, and the live verification notes were useful.

Closing as superseded: #1176 (merged today as 64bb2b89) shipped the GPT-5.6 Sol/Terra/Luna bootstrap catalog with metadata mirrored field-for-field from the upstream codex-rs `models.json` (rust-v0.144.1), including all six reasoning efforts up to `ultra`.

One part of this PR we can't take as-is: the bootstrap-retention change (`get_models_with_fallback` returning `{**bootstrap, **snapshot.models}`, and the `plan_types_for_model` bootstrap fallback once a snapshot exists) conflicts with the `model-catalog-compat` spec — "Refreshed snapshot replaces bootstrap catalog: bootstrap-only entries are not added to the response". After a successful refresh, bootstrap-only slugs must not be re-advertised.

There is a piece here worth salvaging, though: the request-log `ultra` observability in `request_submit.py` (deriving `ultra` from `x-codex-turn-metadata.reasoning_effort` when the wire effort is `max`) is not on current `main`. If you'd like, a small focused follow-up PR on current `main` with just that piece — plus its OpenSpec delta and tests — would be very welcome.

PR #1177, opened 2026-07-10
fix(proxy): only hoist message items into instructions

### LINKED ISSUE #1171: proxy: generalize input normalization to preserve all non-message system/developer input items
## Summary

#1161 fixed issue #1157 by preserving developer-role `additional_tools` input items through `_normalize_responses_input_instructions`. But the underlying hazard is broader: the instruction-hoisting normalizer folds **any** system/developer input item into instructions, so a future non-message typed developer item (any new item type upstream introduces with `role: developer/system` and no `content`) in a non-Lite request would still be silently dropped, reproducing the same class of bug.

Two of the duplicate PRs for #1157 (#1159 by @kvz and #1158 by @Aminkbi) carried a more general guard — never fold a system/developer input item whose `type` is present and not `"message"` — which is better defense-in-depth than the merged `additional_tools`-only early return.

## Suggested scope

- Generalize the preservation guard in `app/core/openai/requests.py` (`_normalize_responses_input_instructions`): only lift items whose type is absent or `"message"`; pass everything else through untouched, regardless of role.
- Keep the merged #1161 behavior (including `to_payload()` round-trip) intact; extend the existing regression tests with a synthetic non-message developer item type.
- Credit: approach from #1159 / #1158.

Refs #1157, #1161.

kvz:
## Summary

Implements #1171: only hoist message-shaped items into `instructions`; every other typed `system`/`developer` input item passes through untouched, regardless of request shape.

## Type of change

Linked issue: Closes #1171

## OpenSpec

Change directory: `openspec/changes/generalize-input-normalization-guard/`

## Changes

- `_normalize_responses_input_instructions()`: the hoisting loop now skips items whose `type` is present and not `"message"`, so a future non-message developer item (no `content`, outside the Lite `additional_tools` early return) keeps its position and wire shape instead of being silently dropped — the hazard class behind #1157.
- The merged #1161 behavior is untouched: the Lite early return and `to_payload()` round-trip semantics stay exactly as merged, and typeless role items (`{"role": "system", "content": ...}`) keep the existing hoisting behavior for OpenAI-compatible clients.
- Regression test with a synthetic `developer_state` item type, parametrized over `ResponsesRequest` and `ResponsesCompactRequest`.
- OpenSpec change recording the message-only hoisting requirement.

Guard carried over from #1159 / #1158, per the maintainer's scope in #1171.

## Test plan

```
uv run pytest tests/unit/test_openai_requests.py -q
108 passed

uv run pytest tests/unit/test_proxy_utils.py tests/unit/test_proxy_http_bridge.py -q
790 passed
```

### DIFF
--- a/app/core/openai/requests.py
+++ b/app/core/openai/requests.py
@@ -305,6 +305,15 @@ def _normalize_responses_input_instructions(data: JsonValue) -> JsonValue:
         if role not in ("system", "developer"):
             input_items.append(item)
             continue
+        # Only lift actual instruction messages. A typed non-message item that
+        # carries a system/developer role (like the Lite ``additional_tools``
+        # bundle, or any future upstream item type without ``content``) must
+        # keep its position and wire shape — folding it into ``instructions``
+        # silently drops it, the hazard class behind #1157.
+        item_type = item_mapping.get("type")
+        if item_type is not None and item_type != "message":
+            input_items.append(item)
+            continue
         instruction_text, preserved_content = _split_responses_instruction_item_content(item_mapping)
         if instruction_text:
             instruction_parts.append(instruction_text)
--- a/openspec/changes/generalize-input-normalization-guard/.openspec.yaml
+++ b/openspec/changes/generalize-input-normalization-guard/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-07-10
--- a/openspec/changes/generalize-input-normalization-guard/proposal.md
+++ b/openspec/changes/generalize-input-normalization-guard/proposal.md
@@ -1,0 +1,22 @@
+# Generalize Input Normalization Guard
+
+## Summary
+
+Only lift input items that are actual instruction messages into `instructions`; pass every other typed `system`/`developer` input item through untouched, regardless of request shape.
+
+## Motivation
+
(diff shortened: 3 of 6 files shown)


kvz:
Superseded by #1172, which cherry-picked the #1159 guard onto current main — exactly the right resolution, thanks. Closing.

Soju06:
Thanks @kvz — this correctly targets the hazard class from #1171, and the synthetic `developer_state` regression test is a nice touch.

Closing as superseded: #1171 was closed earlier today by #1172 (merged as af6b515b), which cherry-picked your #1159 commit with authorship credit (`Co-authored-by` plus an all-contributors entry) and extended the same guard:

- a shared `_is_preserved_non_message_directive` predicate applied consistently across instruction hoisting, interleaved-reasoning sanitization, and compact-trim anchor selection (`_compact_state_anchor_indices`), so preserved directives survive all three passes rather than only the hoisting loop;
- a directive-only `instructions` default: when the only system/developer items are preserved non-message directives, normalization still counts as a change so `instructions` gets defaulted and validation accepts the request. For reference, this PR's variant still has that gap — the new guard branch appends the item without setting `changed = True`, so a directive-only input returns the original payload and then fails required-`instructions` validation. Codex flagged exactly this on #1172 (discussion r3557394902) and the merged version fixes it;
- route-level integration tests on top of the unit coverage.

Appreciate the follow-through on this series — what merged is your #1159 approach, generalized.

PR #1178, opened 2026-07-10
fix(proxy): preserve model_messages in catalog response

zepx:
## Summary

codex-lb stripped the `model_messages` field from upstream model catalog entries via `_FILTERED_FIELDS = {"model_messages"}` in `app/core/clients/model_fetcher.py`. The Codex client's model picker requires this field to display newer models (gpt-5.6-sol/terra/luna); without it the client silently drops models absent from its bundled metadata, so gpt-5.6 models never appear in the picker even though codex-lb is serving them.

PR #331 already removed `base_instructions` from the same filter for an identical reason but left `model_messages` behind by oversight. The `add-gpt56-bootstrap-models` change (#1176) states that `model_messages` is "left to the live registry refresh" — this PR makes that contract true by removing the filter that prevented it.

## What changes

- Remove the dead `_FILTERED_FIELDS` abstraction from `app/core/clients/model_fetcher.py`; replace the filter comprehension with `raw = dict(data)` to preserve all upstream fields verbatim.
- Add regression tests:
  - `tests/unit/test_model_fetcher.py`: asserts `model_messages` survives in `UpstreamModel.raw` after fetch parsing.
  - `tests/integration/test_v1_models.py`: asserts `model_messages` appears in the `/backend-api/codex/models` response after registry → HTTP serialization.
- Add OpenSpec change folder `preserve-model-messages-in-catalog` with proposal, tasks, and `model-catalog-compat` spec delta.

## Validation

- `uv run ruff check` — ✅ all checks passed
- `uv run ruff format --check` — ✅ already formatted
- `uv run pytest tests/unit/test_model_fetcher.py tests/integration/test_v1_models.py` — ✅ 36 passed
- `openspec validate preserve-model-messages-in-catalog --strict` — ✅ Change is valid
- Regression test confirmed: fails on `origin/main` (`KeyError: 'model_messages'`), passes with fix.

## Impact

- `GET /backend-api/codex/models` and `GET /v1/models?client_version=<v>` now include `model_messages` on each model entry, matching the upstream ChatGPT backend response field-for-field.
- Catalog responses are larger (~16 KB per model for the instruction template). This is inherent to the compatibility fix — the client needs the field.
- No database migration; the registry is in-memory only.
- No new security exposure: `model_messages` is upstream catalog content the Codex client already receives when talking directly to ChatGPT's backend.

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -947,6 +947,16 @@
       "contributions": [
         "code"
       ]
+    },
+    {
+      "login": "zepx",
+      "name": "Choong Jun Jin",
+      "avatar_url": "https://avatars.githubusercontent.com/u/1377772?v=4",
+      "profile": "https://github.com/zepx",
+      "contributions": [
+        "code",
+        "test"
+      ]
     }
   ],
   "contributorsPerLine": 7,
--- a/app/core/clients/model_fetcher.py
+++ b/app/core/clients/model_fetcher.py
@@ -22,7 +22,6 @@ from app.core.upstream_proxy import ResolvedUpstreamRoute
 logger = logging.getLogger(__name__)
 
 _FETCH_TIMEOUT_SECONDS = 15.0
-_FILTERED_FIELDS = {"model_messages"}
 
 
 class ModelFetchError(Exception):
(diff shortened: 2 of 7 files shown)


Soju06:
@codex review

PR #1180, opened 2026-07-10
fix(proxy): synthesize interrupted outputs for custom and apply-patch calls

### LINKED ISSUE #1168: bug: interrupted custom_tool_call outputs are not synthesized, causing upstream 400 'No tool output found for custom tool call'
## Summary

The interrupted-tool-output machinery only handles `function_call` items. When a turn ends with an unresolved `custom_tool_call` (e.g. the user interrupts Codex before the shell output is sent) and the next request references that response — via client-sent `previous_response_id` or a proxy-injected continuity anchor — upstream rejects it with:

```
{"type":"error","error":{"type":"invalid_request_error","message":"No tool output found for custom tool call call_xxx","param":"input"},"status":400}
```

First observed in the field by @CapCap while testing #1161 (https://github.com/Soju06/codex-lb/pull/1161#issuecomment-4932248918-region). Root-caused during that PR's review: the gap is **pre-existing on main** (all three sites below are byte-identical on `origin/main`); #1161 merely makes it routinely reachable because Responses Lite custom shell tools now actually work through the proxy, so models emit `custom_tool_call` where they previously never received the tools at all. It was already reachable on main via standard `tools: [{"type": "custom"}]`.

## Mechanism (three `function_call`-only sites)

1. `app/modules/proxy/_service/response_create.py:314` — `_response_output_item_done_function_call_id` records pending calls only for `item.type == "function_call"`; `custom_tool_call` (and `apply_patch_call`) completions never enter `pending_function_call_ids` / `continuity_state.last_pending_function_call_ids`.
2. `app/modules/proxy/_service/websocket/mixin.py` (~1361–1386, and the HTTP-bridge equivalent) — `_inject_missing_interrupted_function_call_outputs` synthesizes only `function_call_output` items from that function-only pending list, so an interrupted custom call gets no synthetic output.
3. `app/modules/proxy/service.py:2050` — `_is_missing_tool_output_error` matches only `"no tool output found for function call call_"`, so the custom variant bypasses the existing masking/retry recovery and the raw upstream 400 leaks to the client.

## Suggested fix scope

Extend the pending-call tracker, synthetic interrupted-output injection, and the `_is_missing_tool_output_error` classifier to cover `custom_tool_call` / `apply_patch_call` and their output item types, with regression tests at the websocket and HTTP-bridge surfaces (interrupted custom tool turn → next request with `previous_response_id` → no upstream 400).

Refs #1157, #1161.

kvz:
## Summary

Implements #1168: extend the interrupted-tool-output machinery from `function_call` only to also cover `custom_tool_call` and `apply_patch_call`, at all three sites identified in the issue.

## Type of change

Linked issue: Closes #1168

## OpenSpec

Change directory: `openspec/changes/synthesize-interrupted-custom-tool-outputs/`

## Changes

- **Site 1 — pending-call tracker**: `_response_output_item_done_function_call_id` → `_response_output_item_done_tool_call`, returning `(call_id, item_type)` for `function_call` / `custom_tool_call` / `apply_patch_call`. State fields `pending_function_call_ids: list[str]` / `last_pending_function_call_ids` become `pending_tool_calls: dict[str, str]` / `last_pending_tool_calls` (call_id → item type, insertion-ordered), recorded at both the websocket and HTTP-bridge upstream event handlers.
- **Site 2 — synthetic injection**: `_synthetic_interrupted_tool_call_output(call_id, call_item_type)` emits the matching output item type (`custom_tool_call_output` / `apply_patch_call_output` / `function_call_output`); the websocket injection site passes the typed pending map through.
- **Site 3 — classifier**: `_is_missing_tool_output_error` now matches the custom and apply-patch variants of "No tool output found for … call call_", so the existing masking/retry recovery engages for them.
- Regression tests: websocket surface (interrupted custom + apply-patch turn → next request with `previous_response_id` → typed synthetic outputs injected ahead of the client input), HTTP-bridge surface (custom tool call completion recorded as pending), and classifier variants.
- OpenSpec change recording the requirement.

## Notes

- The HTTP bridge records pendings but has no injection site today; adding one felt out of scope for "extend" — recovery on that surface comes via the extended classifier. Happy to follow up if you'd rather have bridge-side injection too.
- Field observation matching the issue: a gpt-5.6 session whose turn was severed near an `exec` call wedged permanently on the custom-variant 400 (every full-resend retry got re-anchored onto the same response), which is exactly the classifier bypass this fixes.

## Test plan

```
uv run pytest tests/unit/test_proxy_utils.py tests/unit/test_proxy_http_bridge.py tests/unit/test_openai_requests.py -q
899 passed

uv run pytest tests/unit -q --ignore=tests/unit/test_helm_external_secrets.py
3180 passed, 3 skipped
```
(helm tests fail in my environment on unmodified main; excluded as pre-existing.)

### DIFF
--- a/app/modules/proxy/_service/http_bridge/service_stubs.py
+++ b/app/modules/proxy/_service/http_bridge/service_stubs.py
@@ -403,8 +403,8 @@ def _service_tier_from_event_payload(*args: Any, **kwargs: Any) -> Any:
     return _service_global("_service_tier_from_event_payload")(*args, **kwargs)
 
 
-def _response_output_item_done_function_call_id(*args: Any, **kwargs: Any) -> Any:
-    return _service_global("_response_output_item_done_function_call_id")(*args, **kwargs)
+def _response_output_item_done_tool_call(*args: Any, **kwargs: Any) -> Any:
+    return _service_global("_response_output_item_done_tool_call")(*args, **kwargs)
 
 
 def _rewrite_websocket_downstream_response_id(*args: Any, **kwargs: Any) -> Any:
--- a/app/modules/proxy/_service/http_bridge/upstream_events.py
+++ b/app/modules/proxy/_service/http_bridge/upstream_events.py
@@ -61,7 +61,7 @@ from app.modules.proxy._service.http_bridge.service_stubs import (
     _pop_terminal_websocket_request_state,
     _previous_response_id_from_not_found_message,
     _release_websocket_response_create_gate,
-    _response_output_item_done_function_call_id,
+    _response_output_item_done_tool_call,
     _rewrite_websocket_continuity_corruption_event,
     _rewrite_websocket_downstream_response_id,
     _rewrite_websocket_previous_response_owner_unavailable_event,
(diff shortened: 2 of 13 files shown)


kvz:
Superseded by #1175, which covers everything here plus the HTTP-bridge injection path I had scoped out and flagged as an open question — strictly better. Confirmed the merged fix resolves the field wedging we saw (gpt-5.6 sessions permanently stuck on the custom-variant 400 after interrupts). Closing.

PR #1062, opened 2026-06-21
fix(ui): stabilize account layout and API key dialog

qdzsh:
## Summary

Fixes the Accounts page responsive layout and keeps the API key edit dialog open while selecting items from portalled model/account menus — including when the dropdown is dismissed by a mouse click on the backdrop.

## Type of change

- [x] `fix:` bug fix

Linked issue: None

## OpenSpec

- [x] This PR includes or updates an OpenSpec change

Change directory: `openspec/changes/fix-dashboard-responsive-select-dismissal`

## Changes

- Keep a modal open when an "outside" interaction is really the user dismissing a nested floating layer (Radix Select / DropdownMenu / Popover):
  - exempt interactions whose target is inside a floating layer;
  - also exempt the case where a floating layer is **open** and the user clicks the backdrop to dismiss it — captured at `pointerdown` on the capture phase, so it survives Radix closing the dropdown (its `data-state` flips to `closed`) before the dialog's own outside handler runs (the race that made mouse clicks still close the dialog).
- Extract the guard into a shared `useFloatingLayerDismissGuard` hook consumed by `DialogContent` and `SheetContent`. `AlertDialog` is intentionally left alone — it is non-dismissible on outside interaction by design.
- Make Accounts page panels and account controls fit narrow viewports without horizontal overflow.
- Add regression tests for both cases.

## Test plan

```
git diff --check
npm run lint
npm run build
npm run test
openspec validate --specs
```

Results:

- `git diff --check` passed.
- `npm run lint` passed.
- `npm run build` passed.
- `npm run test` passed: 101 files, 632 tests.
- `openspec validate --specs` blocked locally: `openspec` is not installed in this shell.

## Checklist

- [x] Title is in Conventional Commits format.
- [x] No related issue.
- [x] Added or updated tests covering the change.
- [x] Ran the relevant frontend checks locally.
- [ ] OpenSpec validation passes. Blocked locally: `openspec` is not installed in this shell.
- [x] CHANGELOG is not edited by hand.

PR #1198, opened 2026-07-10
fix(ui): keep API key dialogs open for dropdown interactions

jnsong:
## Summary

- keep API-key create/edit dialogs open when a Radix dropdown portal receives an interaction
- preserve ordinary outside-click and keyboard dismissal
- add focused regression coverage and OpenSpec artifacts

## Exact 1.20.1 reproduction

On the deployed v1.20.1 dashboard, open Settings, edit the API key named ubuntu-hermes-agent-gpt55-active, open Allowed models, and select a model. Before this patch, the portalled dropdown interaction was treated as outside the parent dialog and dismissed the editor before Save.

The submitted branch is rebased onto current upstream main; the separate live deployment remains the exact verified v1.20.1-based build.

## Current-head regression evidence

Commands were run after refreshing this branch onto current upstream main:

- bun run test src/features/api-keys/components/api-key-dialog-interactions.test.ts src/features/api-keys/components/api-key-edit-dialog.test.tsx src/features/api-keys/components/api-key-create-dialog.test.tsx — 3 files, 28 tests passed
- bun run test src/features/api-keys/components — 12 files, 81 tests passed
- bun run typecheck — passed
- bun run lint — passed
- bun run build — passed; 1,764 modules transformed
- npx --yes @fission-ai/openspec@1.6.0 validate --specs — 30 specs passed, 0 failed

The first test attempt after switching to current main failed before test collection because the existing dependency install lacked current-main's new i18next dependency. bun install --frozen-lockfile aligned node_modules to the committed lockfile; every gate above then passed.

## Live deployment verification

Built and deployed the exact v1.20.1-based patch as codex-lb:gpt56-dropdown-fix (image sha256:51d5aaf496711337488cb798e82d6db75c8f2b0dccbad0cd73139e3521a1efd9) while preserving the codex-lb-data volume and runtime configuration.

At http://192.168.50.10:2455/settings:

- edited ubuntu-hermes-agent-gpt55-active
- toggled gpt-5.6-sol off and back on; the edit dialog remained open after both interactions
- added gpt-5.5, closed the dropdown with Escape, and saved
- reloaded and verified the four-model allowlist: gpt-5.5, gpt-5.6-sol, gpt-5.6-terra, gpt-5.6-luna
- verified enforced_model remains blank/null in the persisted api_keys row
- separately reopened the editor and verified an ordinary overlay click closes it
- external health returned {"status":"ok"}

## Issue

No matching issue was found or filed. This PR remains a draft pending maintainer guidance.

### DIFF
--- a/frontend/src/features/api-keys/components/account-multi-select.tsx
+++ b/frontend/src/features/api-keys/components/account-multi-select.tsx
@@ -191,7 +191,7 @@ export function AccountMultiSelect({
 
   return (
     <div className="space-y-1.5">
-      <DropdownMenu>
+      <DropdownMenu modal={false}>
         <DropdownMenuTrigger asChild>
           <Button
             type="button"
--- a/frontend/src/features/api-keys/components/api-key-create-dialog.tsx
+++ b/frontend/src/features/api-keys/components/api-key-create-dialog.tsx
@@ -44,6 +44,8 @@ const TRANSPORT_POLICY_LABELS = {
   always_websocket: "Prefer persistent sessions",
 } as const;
 
+import { preventApiKeyDialogDropdownDismiss } from "./api-key-dialog-interactions";
+
 const formSchema = z.object({
   name: z.string().min(1, "Name is required"),
 });
@@ -303,7 +305,10 @@ export function ApiKeyCreateDialog({ open, busy, onOpenChange, onSubmit }: ApiKe
   return (
     <Dialog open={open} onOpenChange={onOpenChange}>
       {open ? (
-        <DialogContent className="sm:max-w-3xl">
+        <DialogContent
+          className="sm:max-w-3xl"
+          onInteractOutside={preventApiKeyDialogDropdownDismiss}
+        >
           <DialogHeader>
             <DialogTitle>Create API key</DialogTitle>
             <DialogDescription>Set restrictions and expiration for this key.</DialogDescription>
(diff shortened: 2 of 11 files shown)


Soju06:
Thank you for the thorough work here, @jnsong — the careful reproduction, the live-deployment verification notes, and the regression tests are more diligence than most first PRs bring, and it is genuinely appreciated.

After verifying against current `main`, we are closing this as **superseded**:

- The bug you reproduced is real in **v1.20.1**, but it was already fixed on `main` by #1062 (merged 2026-07-01, included in the v1.21.x line). That fix introduced `useFloatingLayerDismissGuard`, wired into the shared `DialogContent` (`frontend/src/components/ui/dialog.tsx`), which guards dismissal for floating-layer interactions across **all** dialogs — a strict superset of the guard this PR adds to the two API-key dialogs.
- To confirm, we ran this PR's own regression test additions (`api-key-edit-dialog.test.tsx`) against **unpatched current main**: **18/18 pass** without the patch.

**Remediation for your deployment:** upgrade to v1.21.x — `v1.21.0-beta.2` is available now, with stable v1.21.0 coming shortly. If dialog dismissal still reproduces on a current-main build after upgrading, please file an issue with browser specifics and a test that fails at base; at that point `modal={false}` can be evaluated properly, with coverage for all three consumers (create dialog, edit dialog, and the multi-selects).

One small housekeeping note for future contributions: `openspec/changes/fix-api-key-dialog-dropdown-dismiss/implementation-plan.md` embeds private infrastructure details (ssh user/host IPs, container layout). OpenSpec docs are checked-in repo SSOT, so deployment-specific details like these should stay out of the repo — worth scrubbing from your local planning templates.

Thanks again, and sorry this one collided with an already-landed fix — we hope to see more contributions from you.

PR #1189, opened 2026-07-10
fix(proxy): preserve multi-agent v2 passthrough

luawl:
## Summary

Preserves Codex multi-agent v2 metadata, compact affinity, and namespaced tool-call identity across Responses proxy boundaries. This fixes the deterministic passthrough gaps reported in #1185 without changing legacy flat-tool replay protection.

## Type of change

- [x] `fix:` - bug fix (no behavior change beyond the bug)
- [ ] `feat:` - new user-facing feature or capability
- [ ] `refactor:` - internal refactor (no behavior change, no API change)
- [ ] `docs:` - documentation only
- [ ] `chore:` / `ci:` / `build:` - tooling, CI, packaging
- [ ] `test:` - test-only change
- [ ] **Breaking change**

Linked issue: Fixes #1185

## OpenSpec

- [x] This PR includes / updates an OpenSpec change
- [ ] Not applicable - bug fix that matches the existing spec
- [ ] Not applicable - docs / CI / chore only
- [x] This PR touches a codex-faithful path and preserves upstream-equivalent behavior

Change directory: `openspec/changes/preserve-multi-agent-v2-passthrough/`

## Changes

- Projects per-request turn, subagent, parent-thread, and window metadata into every bridged `response.create` frame, including reused upstream sockets.
- Gives `x-codex-turn-state` compact affinity precedence over session and prompt-cache keys.
- Includes namespace and stable call ID in namespaced side-effect replay identity while retaining legacy argument-based protection for flat tools.
- Keeps stable namespaced identities across intervening side-effect, read-only, message, and unknown-output items until a real user/system/developer input boundary.
- Adds externally visible bridge and compact-routing regressions plus focused downstream/history dedupe coverage.

## Test plan

```
uv run pytest tests/unit/test_proxy_tool_call_dedupe.py tests/unit/test_proxy_utils.py -q
# 644 passed, 5 pre-existing AsyncMock warnings

uv run pytest tests/integration/test_http_responses_bridge.py::test_v1_responses_http_bridge_reuses_upstream_websocket_and_preserves_previous_response_id tests/integration/test_proxy_sticky_sessions.py::test_proxy_codex_turn_state_pins_responses_and_compact_without_sticky_threads -q
# 2 passed

uv run ruff check .
uv run ruff format --check .
uv run ty check
npx --yes @fission-ai/openspec@latest validate preserve-multi-agent-v2-passthrough --strict
npx --yes @fission-ai/openspec@latest validate --specs
git diff --check
```

[...]

PR #1188, opened 2026-07-10
fix(proxy): reserve stream capacity for recovery

Komzpa:
## Summary

Reserve one account-local stream slot from ordinary new work so an established Responses turn can reattach after a transient upstream disconnect even during heavy fan-out.

## Type of change

- [x] `fix:` — bug fix (no behavior change beyond the bug)

Linked issue: none; reproduced from live request `d1d6296a-807d-4ed9-8445-f238d1e363e8`.

## OpenSpec

- [x] This PR includes / updates an OpenSpec change
- [x] This PR touches a codex-faithful path and preserves upstream-equivalent behavior

Change directory: `openspec/changes/reserve-stream-capacity-for-recovery/`

## Changes

- Ordinary first-turn and follow-up selection use a configurable one-slot recovery reserve.
- Reattach selection may use the full existing per-account stream cap; the hard cap is unchanged.
- Add load-balancer and service-path regression coverage.

## Test plan

```
uv run pytest -q tests/unit/test_load_balancer_concurrency.py -k "stream_cap or stream_recovery_reserve" tests/unit/test_proxy_utils.py -k "stream_cap or stream_recovery_reserve or reserves_stream_slot_for_reattach"
uv run ruff check app/core/config/settings.py app/modules/proxy/load_balancer.py app/modules/proxy/service.py tests/unit/test_load_balancer_concurrency.py tests/unit/test_proxy_utils.py
uv run ty check app/modules/proxy/load_balancer.py app/modules/proxy/service.py
openspec validate reserve-stream-capacity-for-recovery --strict
git diff --check
```

Result: 7 passed; ruff, ty, OpenSpec, and diff checks passed.

## Checklist

- [x] Title is in Conventional Commits format.
- [x] Added tests covering the change.
- [x] Ran the relevant local validation subset.
- [x] OpenSpec strict validation passes.
- [x] CHANGELOG is not edited by hand.

PR #1169, opened 2026-07-10
fix(proxy): isolate unanchored parallel bridge requests

Komzpa:
## Summary

- isolate concurrent unanchored Responses requests that share one process-level Codex session header
- give colliding requests server-owned request-scoped bridge lanes, including duplicate client request IDs and model-class changes
- publish the unanchored handoff reservation before durable refresh or session registration can yield
- release pre-submit reservations on refresh cancellation, payload-preparation failure, and every request-scope exit before visible submission
- preserve anchored legacy forwarding while binding original-unanchored state and client-IP presence into the v2 owner-forward signature
- fence durable alias registration by owner epoch and per-alias generation, so a late stale rejection cannot erase a newer successful registration

## Relationship to the Lite work

This is an independent sibling of #1164. Current `main` contains #1161 and #1172 for canonical typed Responses input transport and #1176 for the upstream-verified GPT-5.6 catalog; none is duplicated here. This PR stays separate because it addresses HTTP bridge concurrency, mixed-version owner forwarding, and durable alias ownership.

The issue was exposed while running GPT-5.6 Sol plus concurrent built-in mini work through one Codex process session header: separate unanchored requests could otherwise reuse one response-create lane.

## Validation

- final rebased PR head: `c5110dbaebb4b340b442fc144ae23e355decf6a5`
- final live composite: 4515 passed, 45 skipped
- post-`main` #1172 focused integration: 416 passed
- current-head HTTP bridge suite: 240 passed
- architecture checks, repository-wide Ruff check/format, `ty`, and strict OpenSpec validation passed
- GPT-5.6 Sol current-head read-only review: no findings; additional replacement/cleanup ABA probes passed for both alias kinds
- live composite tool-call smoke passed through codex-lb for GPT-5.6 Sol, Terra, and Luna, including a resumed Sol session

The PR remains draft while current-head GitHub CI and public Codex review evidence complete.

PR #687, opened 2026-05-17
fix(proxy): trim durable full-resend continuations

Komzpa:
## Summary
- persist completed HTTP bridge input-prefix metadata beside the durable latest response id
- use that metadata to reattach a full resend only when its stored prefix fingerprint matches, then trim the resend down to the new tail
- fail closed for same-websocket full resends while an earlier response is already visible, so clients do not see the answer restart as another upstream turn
- close stale HTTP bridge sessions before returning a recovered `previous_response_id` session, so old upstream websocket state cannot leak after recovery
- limit same-websocket full-resend rejection to codex-session replay-sized requests, avoiding false positives for normal fresh multi-item turns
- mask `responses/compact` `previous_response_not_found` misses as public `stream_incomplete`, without penalizing the selected account or leaking raw `resp_*` ids
- mask raw websocket `previous_response_not_found` terminal errors even when local pending state has lost the previous-response anchor, so clients receive the public `stream_incomplete` fail-closed event instead of a leaked `resp_*` id

## Evidence
- reported websocket continuity failures could surface raw upstream `previous_response_not_found` terminal errors repeatedly before a later turn recovered without `previous_response_id`
- reported retries could resend a full `response.create` with `previous_response_id=None` after visible text had already started; codex-lb forwarded it as a fresh upstream turn
- `responses/compact` continuity failures could surface upstream `invalid_request_error` / `previous_response_id` misses as raw public errors instead of the existing terminal `stream_incomplete` contract
- `.venv/bin/python -m pytest tests/unit/test_proxy_utils.py -k 'compact_previous_response_not_found or compact_responses_records_transient_error_for_generic_upstream_failure or stream_previous_response_not_found_proxy_error_is_masked_to_stream_incomplete'`
- `.venv/bin/python -m pytest tests/integration/test_proxy_compact.py -k 'masks_previous_response_not_found or proxy_compact_success_preserves_compaction_payload'`
- `.venv/bin/python -m pytest tests/unit/test_proxy_api_websocket_auth.py -k 'previous_response'`
- `.venv/bin/python -m ruff check app/modules/proxy/service.py tests/unit/test_proxy_utils.py tests/integration/test_proxy_compact.py`
- `.venv/bin/python -m pytest tests/unit/test_db_migrate.py::test_run_upgrade_auto_remaps_legacy_revision_ids tests/integration/test_migrations.py::test_run_startup_migrations_auto_remaps_legacy_alembic_revision_ids`
[...]

PR #1199, opened 2026-07-11
fix(proxy): preserve native capability routing across failover

dmdfami:
## What changed

- retain per-account model and service-tier capabilities while exposing the merged catalog union;
- fail closed when no refreshed account catalog supports the requested model or non-default tier;
- revalidate account, model, tier, and continuity ownership before every unsent direct-WebSocket turn;
- rotate retry-safe fresh requests and verified self-contained full resends after pre-visible quota failures;
- keep delta-only `previous_response_id` chains and account-scoped files pinned to their owner;
- apply the same replay-safety rule to direct WebSocket, HTTP bridge, and HTTP streaming/image bypass;
- normalize selected Codex installation metadata across HTTP, bridge, WebSocket, and retry paths;
- add OpenSpec requirements and regression coverage for capability routing, quota failover, continuity, and metadata.

## Why

The merged upstream model catalog can advertise a model or Fast tier that only a subset of pooled accounts supports. Plan-only routing may select an incompatible account. Separately, blindly changing accounts for every request can break upstream-owned `previous_response_id` continuity, orphan tool outputs, and misroute account-scoped files.

This change separates discovery from eligibility and permits account rotation only when the request is independently replayable. Short owner-dependent continuations still fail closed.

Fast intent and actual execution also remain distinct: forwarding `service_tier = "priority"` does not claim Fast unless the completed upstream response reports the priority tier.

No linked issue; this was reproduced through live Codex App/CLI compatibility testing against the native backend route.

## Validation

- `npx --yes @fission-ai/openspec@latest validate harden-native-codex-capability-routing --strict`
- `npx --yes @fission-ai/openspec@latest validate --specs --strict` — 30 specs passed
- `uv run ruff check ...`
- `uv run ty check`
- 694 focused backend/unit tests passed, 3 skipped before deployment
- 196 Responses/bridge/files/compact/transient integration tests passed after final changes
- live sequence passed: WebSocket tool call → HTTP image full resend → WebSocket continuation → HTTP compact → WebSocket continuation from compact output

## Notes

- One unrelated WebSocket Lite test can hang when run independently in this local environment; it is not reported as passing.
- The pull request is intentionally draft pending upstream CI and review.

### DIFF
--- a/app/core/clients/proxy.py
+++ b/app/core/clients/proxy.py
@@ -81,6 +81,7 @@ from app.core.utils.request_id import get_request_id
 from app.core.utils.sse import format_sse_event, parse_sse_data_json
 
 CODEX_INSTALLATION_ID_HEADER = "x-codex-installation-id"
+CODEX_TURN_METADATA_HEADER = "x-codex-turn-metadata"
 CODEX_RESPONSES_LITE_HEADER = "x-openai-internal-codex-responses-lite"
 CODEX_RESPONSES_LITE_WEBSOCKET_METADATA_KEY = "ws_request_header_x_openai_internal_codex_responses_lite"
 
@@ -472,19 +473,58 @@ def filter_inbound_headers(headers: Mapping[str, str]) -> dict[str, str]:
     return {key: value for key, value in headers.items() if not _should_drop_inbound_header(key)}
 
 
+def _rewrite_turn_metadata_installation_id(value: JsonValue, codex_installation_id: str | None) -> JsonValue:
+    if not codex_installation_id or not isinstance(value, str):
+        return value
+    try:
+        metadata = json.loads(value)
+    except json.JSONDecodeError:
+        return value
+    if not isinstance(metadata, dict):
+        return value
+    if "installation_id" not in metadata:
+        return value
+    metadata["installation_id"] = codex_installation_id
+    return json.dumps(metadata, ensure_ascii=True, separators=(",", ":"))
+
+
 def apply_codex_installation_metadata(payload: dict[str, JsonValue], codex_installation_id: str | None) -> None:
     raw_metadata = payload.get("client_metadata")
     client_metadata: dict[str, JsonValue] = {}
     if is_json_mapping(raw_metadata):
         for key, value in raw_metadata.items():
-            if isinstance(key, str) and key.lower() != CODEX_INSTALLATION_ID_HEADER:
-                client_metadata[key] = value
+            if not isinstance(key, str) or key.lower() == CODEX_INSTALLATION_ID_HEADER:
+                continue
+            client_metadata[key] = (
+                _rewrite_turn_metadata_installation_id(value, codex_installation_id)
+                if key.lower() == CODEX_TURN_METADATA_HEADER
+                else value
+            )
     if codex_installation_id:
         client_metadata[CODEX_INSTALLATION_ID_HEADER] = codex_installation_id
     if client_metadata:
         payload["client_metadata"] = client_metadata
     else:
         payload.pop("client_metadata", None)
+
+
+def apply_codex_installation_headers(
+    headers: Mapping[str, str],
+    codex_installation_id: str | None,
+) -> dict[str, str]:
+    """Keep canonical turn metadata consistent with the selected account."""
+    updated = dict(headers)
+    if not codex_installation_id:
+        return updated
+    for key, value in list(updated.items()):
+        lowered = key.lower()
+        if lowered == CODEX_INSTALLATION_ID_HEADER:
+            updated[key] = codex_installation_id
+        elif lowered == CODEX_TURN_METADATA_HEADER:
+            rewritten = _rewrite_turn_metadata_installation_id(value, codex_installation_id)
+            if isinstance(rewritten, str):
+                updated[key] = rewritten
+    return updated
 
 
 _NATIVE_CODEX_USER_AGENT_PREFIXES: tuple[str, ...] = (
@@ -2465,6 +2505,7 @@ async def _stream_responses_with_session(
     enforce_openai_sdk_contract: bool = True,
 ) -> AsyncIterator[str]:
     settings = get_settings()
+    headers = apply_codex_installation_headers(headers, codex_installation_id)
     upstream_base = (base_url or settings.upstream_base_url).rstrip("/")
     url = f"{upstream_base}/codex/responses"
     require_route_or_direct_egress_opt_in(
--- a/app/core/openai/model_refresh_scheduler.py
+++ b/app/core/openai/model_refresh_scheduler.py
@@ -91,7 +91,8 @@ class ModelRefreshScheduler:
                 detach_session_objects(session)
             grouped = _group_by_plan(accounts)
             if not grouped:
-                logger.debug("No active accounts for model registry refresh")
+                await get_model_registry().clear()
+                logger.info("Model registry cleared because no active accounts remain")
                 return
 
             encryptor = TokenEncryptor()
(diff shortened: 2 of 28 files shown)


dmdfami:
@codex review

This is ready for maintainer review. The change includes strict OpenSpec validation, focused regression coverage, and live native Codex transport/continuity verification. Please pay particular attention to the safe full-resend boundary versus owner-bound previous-response deltas.

Soju06 (changes requested):
Thank you for this substantial piece of work, @dmdfami — the capability-index idea and the OpenSpec artifacts show real engagement with how this proxy routes. That said, after verifying the current head (`80c4d06`) against `main`, there are confirmed blockers that need to be addressed before this can move forward.

### Confirmed blockers

**(a) Pre-existing durable-anchor regression tests fail at this PR's own head.** CI Required is red; `Tests (pytest, unit)` fails on:
- `tests/unit/test_proxy_http_bridge.py::test_stream_via_http_bridge_injects_durable_anchor_for_trimmable_full_resend`
- `tests/unit/test_proxy_http_bridge.py::test_stream_via_http_bridge_replays_durable_full_resend_when_owner_is_unavailable`

These encode the durable full-resend contract from #687/#698. A PR that reworks failover must reconcile that contract, not break it.

**(b) The relaxed replay gate can silently strip a client-provided `previous_response_id`.** In `request_submit.py` the retry gate no longer requires `proxy_injected_previous_response_id`; any payload that `_websocket_client_previous_response_full_resend_is_retry_safe` accepts gets `fresh_replay_payload = effective_payload.model_copy(update={"previous_response_id": None})` and is replayed as a fresh conversation. The helper cannot prove an input list is a full transcript — for standard SDK delta usage (`previous_response_id` + only the new items), stripping the anchor is silent context loss, not an equivalent retry.

**(c) `account_catalogs_authoritative` conflates catalog-fetch failure with capability absence.** The flag is set from `per_account_results is not None`, so after a restart (no previous snapshot to carry over) an account whose catalog fetch transiently failed is absent from `model_accounts` and `account_ids_for_model` authoritatively excludes it from **all** routing. Similarly, `_mapped_model_has_registry_entry` now hard-fails operator-mapped slugs that are absent from account catalogs with `NO_PLAN_SUPPORT_FOR_MODEL`. Fail-closed designs like this need explicit maintainer sign-off and a documented degradation path before they can land.

**(d) Per-turn websocket revalidation fail-closes healthy owner-pinned continuations.** `_revalidate_open_websocket_account` runs selection on every unsent turn and raises `previous_response_owner_unavailable` when it comes back empty — a transient exclusion (rate-limit blip, temporary health mark) now kills an otherwise healthy owner-pinned continuation mid-session.

**(e) The branch forks pre-#1189/#1200.** Fork point is `65dc4b75` (2026-07-10), before #1189 and #1200 merged (2026-07-11); this PR rewrites the same ws-metadata/canonicalization surfaces (`response_create.py`, `websocket/mixin.py`, streaming helpers). A rebase onto current `main` is required before review can continue.

**(f) Scope: ~1267 net lines across roughly four concerns.** Please split this up:
- The capability-index (`account_ids_for_model` / registry plumbing) and the installation-metadata normalization are separable, lower-risk PRs that could land first on their own merits.
- The failover/replay rework needs the durable-anchor contract reconciled **in OpenSpec** (MODIFIED requirements in `openspec/changes/<slug>/specs/...`) rather than silently broken in tests.

Happy to review focused follow-ups — the underlying direction has merit, it just needs to arrive in reviewable, contract-preserving pieces.

dmdfami:
Thank you for the detailed review. I have started replacing this broad branch with focused branches from current `main`:

- #1204 extracts only selected-account installation metadata normalization.
- #1205 extracts account capability routing and explicitly distinguishes authoritative omission from unknown catalog state. A first-refresh partial failure now degrades to plan-level routing, retained last-known data is used only for still-active accounts, and operator-mapped unknown slugs are not hard-failed.

Both exclude the replay/WebSocket revalidation changes. I am reconciling that remaining work separately against the existing durable-anchor contract and will not remove a client-owned `previous_response_id` unless the proxy has independently retained or verified an equivalent fresh request body.

dmdfami:
The replay/WebSocket portion is now split into #1207 from current `main`. It keeps client-owned `previous_response_id` requests owner-bound for quota/account switching, skips per-turn selector revalidation when a continuation is already on its required owner socket, and permits cross-account replay only for proxy-injected anchors backed by verified retained continuity plus an equivalent fresh body. The two durable-anchor regressions named in this review pass unchanged.

The three focused replacements are now #1204, #1205, and #1207.

Soju06 (changes requested):
Thanks again for splitting this into #1204 / #1205 / #1207 as requested — that was the right call. #1204 (installation metadata) and #1205 (capability routing) look good and are proceeding to merge.

To be clear on where this PR now sits: it is **not** superseded by the split. The three split PRs cover only three of the four original concerns; the failover/replay "fresh-resend" rework is unique to this branch, so #1199 stays open as needs-work. Two blockers remain at the current head (`80c4d06`).

## Blocker 1: breaks the #687 / #698 durable-anchor contract (CI red)

This breaks two pre-existing regression tests that encode the durable-anchor full-resend contract:

- `tests/unit/test_proxy_http_bridge.py::test_stream_via_http_bridge_injects_durable_anchor_for_trimmable_full_resend`
- `tests/unit/test_proxy_http_bridge.py::test_stream_via_http_bridge_replays_durable_full_resend_when_owner_is_unavailable`

Both pass on `main` and fail at this head (e.g. `assert second_call.kwargs["previous_response_id"] is None` now sees `'resp_completed_anchor'`). That contract is normative today, so either keep these green, or if the new behavior is intended, **change the requirement explicitly in the OpenSpec change's delta specs** (`openspec/changes/.../specs/responses-api-compat/spec.md`) rather than silently breaking it. Right now the code changes the behavior while the spec/tests still assert the old one.

## Blocker 2: the fresh-resend gate strips `previous_response_id` on a client-provided delta

In `app/modules/proxy/_service/http_bridge/streaming.py`, the owner-unavailable path calls `_websocket_client_previous_response_full_resend_is_retry_safe(..., continuity_state=None)` and, when it returns true, replays the turn with `previous_response_id=None` (a fresh conversation). But with `continuity_state=None` that helper (`websocket/helpers.py`) returns `True` on nothing more than: `previous_response_id` present, `input` is a list with `len(input) > 1`, and the items are self-contained (no orphaned tool outputs). There is no proxy-injected-anchor provenance check and no verified full-transcript prefix match.

That means a legitimate **client-provided delta continuation** on owner-unavailable failover (multi-item input, no orphaned tool calls) is silently replayed as a fresh conversation — dropping all prior context — instead of failing closed. Please require proxy-injected-anchor provenance (or a verified full-transcript prefix match) before stripping `previous_response_id`, and add a regression proving a client delta continuation is preserved / fails closed on failover rather than being silently replayed.

## Also

- **Please rebase onto current `main`.** This branch forks from `v1.21.0-beta.1` (`65dc4b75`), before #1189 / #1200, and now overlaps merged #1188, #1169, and #1203 in the files it touches (`response_create.py`, `websocket/mixin.py`, and the http_bridge streaming helpers). Rebasing first will avoid a painful merge and make the real diff reviewable.
- Once reworked, consider whether this can split further: the anchor-contract change and the capacity/failover work may separate cleanly into two focused PRs.

Requesting changes. The direction is reasonable; it mainly needs to stop treating a client delta as a full resend, preserve (or explicitly re-spec) the durable-anchor contract, and land on current `main`.

Komzpa:
Superseded by the focused replacements Soju06/codex-lb#1204, Soju06/codex-lb#1205, and Soju06/codex-lb#1207. Soju06/codex-lb#1204 has merged; Soju06/codex-lb#1205 and Soju06/codex-lb#1207 are green, mergeable, and current-head Codex-clean. Together they preserve this PR's installation-metadata, capability-routing, and safe replay scope without carrying the older conflicting aggregate branch. Closing this carrier in favor of those focused PRs. Thank you for the original implementation and the careful split.

PR #228, opened 2026-03-19
fix(proxy): restore token cache affinity routing

Soju06:
## Summary
- restore `/v1/*` prompt-cache bounded affinity instead of durable `codex_session` routing
- derive and forward `prompt_cache_key` before backend Codex `CODEX_SESSION` early return
- add session header alias handling, affinity observability, TTL migration coverage, and websocket parity logging

## Testing
- `openspec validate --specs`
- `.venv/bin/ruff check app/core/config/settings.py app/modules/proxy/api.py app/modules/proxy/service.py tests/unit/test_proxy_utils.py tests/unit/test_openai_requests.py tests/integration/test_proxy_sticky_sessions.py tests/integration/test_openai_compat_features.py app/db/alembic/versions/20260319_111500_merge_prompt_cache_ttl_and_upstream_transport_heads.py`
- `.venv/bin/pytest -q tests/unit/test_proxy_utils.py tests/unit/test_openai_requests.py tests/unit/test_chat_request_mapping.py tests/integration/test_proxy_sticky_sessions.py tests/integration/test_openai_compat_features.py tests/integration/test_proxy_websocket_responses.py tests/integration/test_migrations.py`

PR #1200, opened 2026-07-11
fix(proxy): stop synthesizing top-level tools for upstream Codex requests

### LINKED ISSUE #1184: bug: proxy synthesizes top-level tools:[] breaking GPT-5.6 reserved tools (collaboration.spawn_agent 400)
## Symptom (field report on v1.21.0-beta.1)

Codex + gpt-5.6 through codex-lb intermittently fails with:

```
{"type":"error","error":{"type":"invalid_request_error","message":"Invalid Value: 'tools'. Function 'collaboration.spawn_agent' is reserved for use by this model and must match the configured schema.","param":"tools"},"status":400}
```

## Root cause

Under Responses-Lite the Codex client omits top-level `tools` entirely (codex-rs `build_responses_request`: `(instructions, tools) = (String::new(), None)`), shipping the bundle in the `additional_tools` input item. codex-lb's `ResponsesRequest.tools` is declared with `default_factory=list` (app/core/openai/requests.py:618) and `to_payload()` never pops the unset field on the native path — so **every Lite request reaches upstream with an explicit `"tools": []` the client never sent**. gpt-5.6 Sol/Terra (`multi_agent_version: v2`) configure `collaboration.spawn_agent` as a reserved model tool; an explicit `tools` param triggers reserved-tool schema reconciliation, and `[]` cannot match — hence `param: "tools"`. The chat-source path already pops this deliberately (request_policy.py:246-250); the native path lacks the equivalent.

Secondary hazard found during the same trace: `_canonicalize_tools` (requests.py:831-863, from #228 for prompt-cache affinity) re-sorts every key of every tool **in the wire payload**, which would break any byte/structural-equality check on a client-sent reserved tool. Canonicalization should feed the cache-affinity hash only.

Verified by byte-diff repro: `additional_tools` passthrough is byte-identical (#1161/#1172 hold); compact strips `tools` entirely; the injected empty array is the only top-level `tools` upstream can see from a Lite client.

## Fix scope

1. `to_payload()`: omit `tools` when the client did not send the field (`"tools" not in model_fields_set`); preserve an explicit client-sent `[]`. Audit sibling default-injected fields (`tool_choice`, `parallel_tool_calls`) for the same pattern.
2. Make `_canonicalize_tools` non-mutating on the wire (hash-only), forwarding client tool entries byte-preserved.
3. Regression tests at the product path: Lite websocket frame and HTTP-bridge body contain no top-level `tools` key; a client-sent reserved function tool reaches upstream byte-identical.

Refs #1157, #1176. Found on v1.21.0-beta.1; fix should ship in beta.2.

Soju06:
Fixes #1184

## Root cause

Under Responses Lite the Codex client omits top-level `tools` entirely (codex-rs `build_responses_request`: `(instructions, tools) = (String::new(), None)`), shipping the tool bundle in the `additional_tools` input item. `ResponsesRequest.tools` is declared with `default_factory=list`, so `to_payload()` synthesized an explicit `"tools": []` on every Lite request across all upstream transports (websocket `response.create` frames via `_response_create_text`, HTTP-bridge bodies via `_prepare_http_bridge_request`, and the direct HTTP stream path). gpt-5.6 Sol/Terra (`multi_agent_version: v2`) configure `collaboration.spawn_agent` as a reserved model tool; any explicit `tools` param triggers reserved-tool schema reconciliation, and `[]` cannot match — hence the intermittent `400 invalid_request_error` with `param: "tools"`.

Secondary hazard from the same trace: `_canonicalize_tools` (#228, prompt-cache affinity) re-sorted every tool and every object key **in the wire payload**, which would break any byte/structural-equality check upstream performs on a client-sent reserved tool entry.

## Changes

- `ResponsesRequest.to_payload()` pops `tools` when `"tools" not in model_fields_set`; an explicit client-sent `[]` is still forwarded. Websocket/bridge/HTTP paths inherit the fix because they all serialize via `to_payload()`.
- `V1ResponsesRequest.to_responses_request()` drops the synthesized default before constructing the native request, so `/v1/responses` propagates client omission (its own `model_dump` would otherwise mark `tools` as set).
- `_canonicalize_tools` is replaced by a non-mutating `canonicalized_tools()` used only by the `_tools_hash` affinity/observability consumer; the wire payload keeps the client's tool entries byte-preserved (array order, object key order, unknown keys, array-value order).
- OpenSpec change `omit-synthesized-top-level-tools` with normative deltas on `responses-api-compat` (strict validation passes).

## Sibling-field audit

- `tool_choice` / `parallel_tool_calls`: default `None`, already dropped by `model_dump(exclude_none=True)` when unset; explicit values — including Lite's explicit `parallel_tool_calls: false` — keep forwarding. No code change needed; behavior locked by a new unit test.
- `store: false` is a deliberate proxy invariant (validator-coerced) and keeps being emitted.
- `include` shares the `default_factory=list` pattern, but `"include": []` is semantically identical to omission upstream and not a reserved-schema hazard; left unchanged to keep this PR one concern wide.
- `validate_tool_types` passes `{"type": "namespace"}` entries through untouched (no alias, not an unsupported type), verified by the byte-identity tests which run through the field validator.

## Byte-diff evidence

- Lite request (input contains `additional_tools`, no top-level `tools`): before — upstream websocket frame and HTTP-bridge body both contained `"tools":[]`; after — no `tools` key on either transport (`additional_tools` passthrough stays byte-identical per #1161/#1172).
- Client-sent reserved namespace tool (`collaboration` namespace wrapping `spawn_agent` with `strict: false`, non-standard `"encrypted": true` marker, non-alphabetical `required` order, leading description whitespace): before — array re-sorted and every key alphabetized on the wire; after — the upstream frame contains the tools array byte-identical to the client serialization (asserted as an exact `"tools":<json>` substring of the raw frame).
- Explicit client-sent `"tools": []` — forwarded as `[]` before and after (unchanged).

## Test plan

All new regression tests verified fail-before/pass-after by stashing the `app/` changes:

- `tests/unit/test_openai_requests.py` — unset tools omitted; explicit `[]` preserved; reserved namespace tool byte-identical; tool order/key order preserved; `canonicalized_tools` order-insensitive and non-mutating (never reorders array values like `required`); V1 conversion propagates omission and preserves explicit `[]`; unset `tool_choice`/`parallel_tool_calls` absent while explicit values forward.
- `tests/unit/test_proxy_utils.py` — `_tools_hash` identical across tool-array reorder and key reorder, different for different tools.
- `tests/integration/test_proxy_websocket_responses.py` — Lite frame exact-payload assertion no longer contains `tools` (fails pre-fix); new byte-identity test on the raw upstream `response.create` frame with the reserved namespace fixture.
- `tests/integration/test_http_responses_bridge.py` — new Lite bridge test: bridge body has no `tools` key while the Lite client-metadata marker survives.
- `tests/integration/test_proxy_responses.py` — Lite HTTP path payload has no `tools` key; new explicit-`[]` forwarding test.

Quality gate: `uv run ruff check .` + `uv run ruff format --check .` + `uv run ty check` all clean; `tests/unit` 3276 passed; touched integration files (`test_proxy_responses.py`, `test_http_responses_bridge.py`, `test_proxy_websocket_responses.py`, `test_proxy_compact.py`) 212 passed, plus `test_openai_compat_features.py` / `test_proxy_sticky_sessions.py` green; `openspec validate omit-synthesized-top-level-tools --strict` valid.

Refs #1157, #1176, #228, #1161.

### DIFF
--- a/app/core/openai/requests.py
+++ b/app/core/openai/requests.py
@@ -701,6 +701,18 @@ class ResponsesRequest(BaseModel):
 
     def to_payload(self) -> JsonObject:
         payload: MutableJsonObject = self.model_dump(mode="json", exclude_none=True)
+        if "tools" not in self.model_fields_set:
+            # ``tools`` is declared with ``default_factory=list``, so
+            # ``model_dump(exclude_none=True)`` synthesizes an explicit
+            # ``"tools": []`` even when the client omitted the field. Codex
+            # Responses-Lite clients omit top-level ``tools`` entirely (the
+            # bundle rides in the ``additional_tools`` input item), and models
+            # with reserved model tools (e.g. ``collaboration.spawn_agent`` on
+            # gpt-5.6 ``multi_agent_version: v2``) reject any explicit
+            # ``tools`` param that cannot match the reserved schema. Only
+            # forward the field when the client actually sent it — including
+            # an explicit client-sent ``[]``. See issue #1184.
+            payload.pop("tools", None)
         return _strip_unsupported_fields(payload)
 
 
@@ -777,7 +789,11 @@ def _strip_unsupported_fields(payload: MutableJsonObject) -> MutableJsonObject:
     _normalize_service_tier_aliases(payload)
     _sanitize_interleaved_reasoning_input(payload)
     _strip_poisoned_local_compact_fallback_items(payload)
-    _canonicalize_tools(payload)
+    # ``tools`` is deliberately NOT canonicalized here: the wire payload must
+    # forward client tool entries byte-preserved (array order, key order, and
+    # unknown keys untouched) so reserved model tools survive upstream
+    # byte/structural-equality checks. Order-insensitive canonicalization is
+    # cache-affinity/observability-only; see ``canonicalized_tools``.
     for key in _UNSUPPORTED_UPSTREAM_FIELDS:
         payload.pop(key, None)
     return payload
@@ -828,15 +844,18 @@ def _is_poisoned_local_compact_fallback_message(item: JsonValue) -> bool:
     return False
 
 
-def _canonicalize_tools(payload: MutableJsonObject) -> None:
-    tools = payload.get("tools")
-    if not is_json_list(tools):
-        return
-    tool_list = tools
-    if not tool_list:
-        return
-    sorted_tools = sorted(tool_list, key=_tool_sort_key)
-    payload["tools"] = [_sort_keys_recursive(t) for t in sorted_tools]
+def canonicalized_tools(tools: list[JsonValue]) -> list[JsonValue]:
+    """Return an order- and key-order-insensitive canonical form of ``tools``.
+
+    Used only for prompt-cache affinity/observability hashing (change #228):
+    two requests that differ solely in tool array order or object key order
+    hash identically. The result MUST NOT feed the upstream wire payload —
+    outgoing requests forward the client's tool entries byte-preserved (see
+    issue #1184). Array values (e.g. ``parameters.required``) are never
+    reordered; only mapping keys are sorted.
+    """
+    sorted_tools = sorted(tools, key=_tool_sort_key)
+    return [_sort_keys_recursive(t) for t in sorted_tools]
 
 
 def _tool_sort_key(tool: JsonValue) -> str:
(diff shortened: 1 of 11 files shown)


Soju06:
@codex review

PR #1207, opened 2026-07-11
fix(proxy): reselect only safe websocket turns

dmdfami:
## Summary

- revalidate an already-open direct Responses WebSocket only for an unsent turn that can safely move accounts
- reconnect an owner-pinned request to a different resolved owner without removing its client anchor
- skip per-turn selector revalidation when a continuation is already on its required owner socket
- allow pre-visible quota/account failover for previous-response turns only when the anchor was proxy-injected from verified local continuity and an equivalent fresh body was retained
- keep client-owned `previous_response_id` requests owner-bound even when their input resembles a full resend

This is the third focused replacement for #1199. It reconciles the replay work with the existing durable-anchor contract instead of relaxing the client-owned replay gate. HTTP bridge durable behavior is unchanged.

## OpenSpec

Change directory: `openspec/changes/reselect-safe-websocket-turns/`

The change uses MODIFIED requirements for the existing WebSocket full-resend and hard-owner contracts.

## Validation

- `uv run pytest -q tests/unit/test_proxy_utils.py` (637 passed)
- both durable-anchor regressions named in the #1199 review pass
- `uv run pytest -q tests/integration/test_proxy_websocket_responses.py`
- `uv run ruff check ...`
- `uv run ruff format --check ...`
- `uv run ty check`
- `npx --yes @fission-ai/openspec validate reselect-safe-websocket-turns --strict`
- `npx --yes @fission-ai/openspec validate --specs --strict` (30 passed)
- `git diff --check`

Related to #1199.


Soju06 (changes requested):
Thanks for splitting #1199 into #1204 / #1205 / #1207 exactly as we asked — that made this much easier to review. #1204 (installation metadata) and #1205 (capability routing) both look good and are proceeding to merge. This piece and #1199 still need more work.

I re-reviewed against the current head (`2977b10c`, "verify cross-transport failover context"), not the earlier `df31a01c`. Thank you for the follow-up commit — but note it reworks the **HTTP** streaming replay path in `streaming/retry.py`; it does not touch the WebSocket revalidation, so the blocker below still stands at the current head.

## Blocker (P1): revalidation runs a full capacity-consuming reselection against the socket's own owner

`_revalidate_open_websocket_account` (`app/modules/proxy/_service/websocket/mixin.py`) calls `_select_account_with_budget_compatible(..., preferred_account_id=account.id)` to decide whether the already-open socket may serve the next movable turn. That is a fresh budget/capacity account selection, and it acquires a *new* `AccountLease` for the account that already holds the live WebSocket lease. So the owner's own in-flight lease is counted as consumed capacity: on a single-capacity account the selection concludes "No available accounts", `selected_account` comes back `None`, and the caller then either drops a perfectly fresh (or proxy-verified) turn or retires the healthy socket.

This is not hypothetical — it breaks three pre-existing regression tests that pass on `main` and fail at this head:

- `tests/integration/test_proxy_websocket_responses.py::test_v1_responses_websocket_reuses_upstream_for_sequential_requests`
- `tests/integration/test_proxy_websocket_responses.py::test_backend_responses_websocket_keeps_session_alive_after_anonymous_prev_nf_created_followup`
- `tests/integration/test_proxy_websocket_responses.py::test_backend_responses_websocket_matches_previous_response_error_to_anchor_with_two_followups`

On `main` all three pass; at `2977b10c` all three fail. The captured logs show the mechanism directly:

```
No account selected strategy=capacity_weighted sticky=False model=gpt-5.4 error=No available accounts
Proxy preferred account unavailable request_id=... kind=websocket_revalidate ... preferred_account_id=... error=No available accounts. Service is operating in degraded mode
```

i.e. the `websocket_revalidate` selection self-reports the owner as unavailable, and the second `response.create` never forwards upstream (`created_2` returns the wrong response id).

## What's needed

1. **Don't do a capacity-consuming reselection for an owner-pinned continuation.** Check the current owner's model/tier eligibility *without* acquiring a second lease against it — exclude the self-held lease from the capacity math, reuse the existing lease, or use a capacity-free eligibility check. An owner-pinned continuation must not be treated as needing full reselection.
2. **Get those three integration tests green** — the second `response.create` must actually forward upstream, with no lease leaked when revalidation declines.
3. **Add an end-to-end regression** driving a second turn through the new client-loop block, covering: (a) a healthy client-owned continuation stays on its owner without revalidation; (b) a fresh second turn reuses the healthy socket; (c) an owner-mismatch retires and reconnects to the resolved owner.

The concept — "reselect only safe turns" — is right, and the surrounding structure is close. It's the "revalidate = full reselection against the live owner" step that needs to change. Requesting changes; happy to look again once the three tests are green and the regression is in.

Soju06:
@codex review

Komzpa:
@codex review

Soju06:
@codex review

dmdfami:
Follow-up `60001d06` is ready for review. It is rebased on `438d15e7` and keeps that hardening intact.

Additional coverage comes from a live native Codex failure observed on request `4153a3d3-2b80-41fc-b957-928d4dab5d89`: the HTTP bridge had proven a durable full resend (`original_items=254`, `trimmed_to=4`, retained prefix `stored_items=250`) but still returned owner-unavailable when the pinned owner exhausted quota. The new path releases the old account-local create lease, strips only the proxy-verified anchor, excludes the failed owner, reallocates sticky selection, and retries the retained full body. Client-owned deltas and account-scoped files remain fail-closed.

It also closes the remaining owner-replay edges: post-refresh HTTP failover, direct-WS failed-owner exclusion/sticky reallocation, owner-mismatch turn-state removal, and the same security guard on the HTTP bridge.

Validation: 905 focused unit tests, 204 relevant integration tests, Ruff/format, `ty check`, strict change validation, and all 31 main specs passed.

@codex review

dmdfami:
CI architecture fix is in `5fdb3e70` (supersedes `60001d06`). The HTTP bridge now reaches the shared WebSocket replay helper through the existing `service.py` facade / `http_bridge.service_stubs` boundary; `scripts/check_proxy_architecture.py`, Ruff/format and the focused quota/owner tests pass locally.

dmdfami:
Head `8e5b9da9` is now fully green, including all three integration-core shards and the required CI gate. I also fast-forwarded locally and re-ran the focused verified-owner/model-transition tests plus `scripts/check_proxy_architecture.py`; all passed. The remaining `CHANGES_REQUESTED` status points to the older `2977b10c` review, whose WebSocket self-capacity blocker was addressed by `a4996184` and is covered by the now-green integration suite. Ready for maintainer re-review/merge when convenient.

Soju06:
@codex review

Soju06:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Soju06:
@codex review

Soju06:
@codex review

Soju06:
@codex review

Soju06:
@codex review

PR #1218, opened 2026-07-11
fix(proxy): keep native capacity waits alive

Komzpa:
## Summary

- keep native Codex SSE requests alive while they wait for recoverable local account capacity
- preserve structured startup errors for OpenAI SDK-compatible streams
- cover the image-capable path that bypasses the HTTP responses bridge

## Problem

Image-capable native Codex requests use the direct streaming path. When local
account capacity was exhausted, that path suppressed capacity progress because
HTTP error propagation was enabled. The response startup probe therefore had no
SSE bytes to return, so a client could hit its idle timeout while the proxy was
still waiting and retrying normally.

## Behavior

Native streams with the OpenAI SDK contract disabled now receive the existing
nonterminal `codex.keepalive` event during a capacity wait. The event does not
claim that an upstream response exists, consume a stream lease, or synthesize a
`response.*` lifecycle event. SDK-compatible streams continue to suppress this
extension before startup so they can return structured local-capacity errors.

## Verification

- new image-bypass regression failed before the fix with no initial SSE byte
- `12` focused tests passed
- `641` proxy utility tests passed
- `51` OpenAI compatibility integration tests passed
- Ruff, ty, strict OpenSpec validation, and the main OpenSpec suite passed

### DIFF
--- a/app/modules/proxy/_service/streaming/retry.py
+++ b/app/modules/proxy/_service/streaming/retry.py
@@ -363,7 +363,7 @@ class _StreamingRetryMixin:
                         error_message=error.message if error else None,
                         recovery_sleep_seconds=recovery_sleep_seconds,
                         deadline=deadline,
-                        emit_keepalives=not propagate_http_errors,
+                        emit_keepalives=not propagate_http_errors or not enforce_openai_sdk_contract,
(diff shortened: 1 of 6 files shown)


Soju06:
@codex review

PR #1228, opened 2026-07-11
test(usage): use UTC clock for freshness fixture

luawl:
## Summary

- make the monthly quota-freshness fixture use the same naive-UTC clock as `UsageUpdater`
- fix a deterministic failure for contributors running the suite outside UTC
- no runtime behavior changes

## Root cause

The test inserted `recorded_at=datetime.now()`, which is naive local time. Production computes freshness with `app.core.utils.time.utcnow()`, a naive UTC timestamp. In `America/New_York`, the inserted row therefore appeared roughly four hours old and the updater incorrectly performed the mocked upstream refresh instead of treating the row as fresh.

Using `usage_updater_module.utcnow()` matches the exact clock and timestamp convention consumed by the code under test, including future clock monkeypatching.

## Validation

- targeted regression: 1 passed
- complete `tests/unit/test_usage_updater.py`: 82 passed
- Ruff passed
- `git diff --check` passed

This is a test-only portability correction, so no OpenSpec change is required.

### DIFF
--- a/tests/unit/test_usage_updater.py
+++ b/tests/unit/test_usage_updater.py
@@ -1525,7 +1525,7 @@ async def test_usage_refresh_uses_fresh_monthly_row_for_quota_freshness(monkeypa
         account.id,
         100.0,
         window="monthly",
-        recorded_at=datetime.now(),
+        recorded_at=usage_updater_module.utcnow(),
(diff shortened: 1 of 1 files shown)


Soju06:
@codex review

PR #1236, opened 2026-07-12
fix(proxy): accept large websocket response.create and fail fast with 400

Soju06:
## Why

Production incident (2026-07-12): a Codex Desktop remote session reconnected its Responses websocket and resent the full conversation history (visual-QA screenshots inline) as one websocket text message >16 MiB. uvicorn's default `ws_max_size` closed each connection at the protocol layer (`1009`) before the application-level slimming guard could run — no app log, no `request_logs` row. The official client treats that close as a generic retryable stream error: it burned 5 full-payload retries, permanently downgraded the session to HTTP transport, and the fallback POST then failed at the front proxy as an opaque nginx `413`.

## What

- **`--ws-max-size` / `UVICORN_WS_MAX_SIZE`** (default **128 MiB**, parity with `max_decompressed_responses_body_bytes`) wired to `uvicorn.run`, so oversized `response.create` messages reach the existing historical-slimming guard instead of dying with `1009`. `permessage-deflate` stays negotiated; the budget applies to decompressed size.
- **Local oversized rejection is now `400`, not `413`** (envelope unchanged: `payload_too_large` / `invalid_request_error` / `param="input"`), in both guards (`response_create.py`, `core/clients/proxy.py`). Verified against codex-rs (main, 2026-07-11): a wrapped websocket error event with `status: 400` is surfaced immediately as non-retryable and the session stays on websocket; `413` triggers 5 full-payload resends followed by a sticky session-wide websocket→HTTP downgrade (`core/src/responses_retry.rs`, `codex-api/src/endpoint/responses_websocket.rs`).
- OpenSpec change **`raise-ws-ingress-message-limit`** (proposal/design/specs/tasks complete, strict validation green) with the full official-client evidence in `context.md`; reverse-proxy sizing guidance added to `responses-api-compat/ops.md`. The `400` supersedes the `413` picked by the archived `2026-04-14-guard-oversized-response-create` change, whose delta was never synced into the main spec.

## Verification

- `uv run ruff check app tests` clean; targeted suites green: `tests/unit/test_cli.py` (15), `tests/unit/test_proxy_utils.py` (692), `tests/unit/test_proxy_http_bridge.py` (256), `tests/integration/test_proxy_websocket_responses.py` + `tests/integration/test_http_responses_bridge.py` (149).
- E2E against a locally booted server: default flags → a 20 MiB websocket `response.create` survives the protocol layer and receives `{"type":"error","status":400,"error":{"code":"payload_too_large"}}`; with `--ws-max-size 1048576` → a 2 MiB message is closed with `1009`, proving the knob is wired through.
- `openspec validate raise-ws-ingress-message-limit --strict` and `openspec validate --specs` (31/31) pass.


Soju06:
@codex review

Soju06:
Addressed in 4838d4ab: `scripts/distroless-entrypoint.py` now execs `python -m app.cli` (matching `docker-entrypoint.sh`), and the dev `docker-compose.yml` launches uvicorn directly with `--reload --ws-max-size 134217728` so dev keeps reload while matching the production websocket ingress budget. Verified locally: a 20 MiB websocket `response.create` against the uvicorn-CLI launcher shape reaches the app guard and returns the status-400 `payload_too_large` event instead of a 1009 close.

@codex review

PR #1239, opened 2026-07-12
perf(request-logs): skip-scan unfiltered filter-option facets

Soju06:
## Summary

`GET /api/request-logs/options` runs four unbounded `SELECT DISTINCT` statements over `request_logs` when the filter panel loads with no active filters (the default state) — account ids, `(model, reasoning_effort)`, api-key ids, `(status, error_code)`. PostgreSQL has no loose index scan, so each is a full pass over the corresponding index, re-run by the dashboard's 30s options poll while the logs page is open, and growing with table size forever.

Unfiltered requests now emulate the loose index scan with recursive CTEs (`WITH RECURSIVE facet_skip`): seed `min(col)`, then repeatedly probe `min(col) WHERE col > previous` — one btree probe per distinct value against the existing composite facet indexes. Pair facets iterate the leading column by skip scan and per value resolve the second column (nested skip scan + a `LIMIT 1` NULL-presence probe).

Result parity is exact, per backend:
- NULL pair placement follows each backend's `ORDER BY ASC` NULL ordering (SQLite first, PostgreSQL last), matching the legacy `DISTINCT` output.
- Empty-string values keep the legacy per-facet falsy post-filter semantics (leading values dropped, second columns preserved).
- `deleted_at IS NULL` and the status-facet self-filter exemption are unchanged.
- **Filtered** requests keep the legacy `DISTINCT` shape — user filters already bound those scans, and per-value probes can degenerate under selective residual filters.

## OpenSpec

`openspec/changes/optimize-request-log-filter-options/` — delta on `query-caching`. `openspec validate --specs`: 31/31 pass.

## Tests

- Pair facets with NULL + non-NULL efforts for one model; soft-delete exclusion; empty table; unfiltered↔filtered parity oracle; empty-string second-column parity; unfiltered api-key facet.
- A statement-capture test pins the contract: the unfiltered options call issues **no** `SELECT DISTINCT` against `request_logs` and does use the recursive skip scan.
- Suite added to `POSTGRES_PYTEST_TARGETS`; all 12 tests pass on SQLite and PostgreSQL 16 locally. `ruff` / `ty` clean.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,8 @@ POSTGRES_PYTEST_TARGETS := \
 	tests/integration/test_repositories.py::test_accounts_upsert_with_merge_enabled_serializes_concurrent_same_email \
 	tests/integration/test_sticky_sessions_api.py::test_durable_bridge_owned_alias_registration_is_epoch_fenced \
 	tests/integration/test_proxy_api_extended.py::test_proxy_stream_usage_limit_returns_http_error \
-	tests/integration/test_repositories.py::test_accounts_upsert_with_merge_disabled_uses_identity_lock_on_postgresql
+	tests/integration/test_repositories.py::test_accounts_upsert_with_merge_disabled_uses_identity_lock_on_postgresql \
+	tests/test_request_logs_options_api.py
 SHELL := /bin/bash
 
 .PHONY: help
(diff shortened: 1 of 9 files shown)


Soju06:
@codex review

PR #1240, opened 2026-07-12
feat(retention): opt-in pruning for request logs and usage history

Soju06:
> **Stacked on #1238** (`feat/account-usage-rollup`) — request-log pruning is gated on the fold watermark that PR introduces. Retarget to `main` after #1238 merges.

## Summary

`request_logs`, `usage_history`, and `additional_usage_history` grow without bound (usage history alone ~1,440 rows/day/account/window); the only deletes anywhere are per-account cleanup. Unbounded growth is what turns every windowed dashboard aggregate and facet query slowly worse forever.

- **Opt-in env settings** (default `0` = off, so zero behavior change unless enabled): `CODEX_LB_REQUEST_LOG_RETENTION_DAYS` (≥ 30), `CODEX_LB_USAGE_HISTORY_RETENTION_DAYS` (≥ 45, covers both usage tables), both capped at 3650 — floors keep every in-product consumer window (7-day, monthly ~31d, default report ranges) inside retained data, and the cap prevents `timedelta` overflow from silently disabling the job.
- **Hourly leader-gated job**, 10k-row batches, one transaction per batch; first opt-in pass drains the backlog incrementally.
- **Safety invariants**:
  - Request-log pruning deletes only rows at or below the rollup watermark (`min(cutoff, folded_through)`; skipped entirely before the first fold) — lifetime account totals survive pruning **by construction**, including under a stalled fold (mutation-hardened regression test).
  - Usage-history pruning always retains each identity's newest row (`(account, coalesce(window,'primary'))` / `(account, quota_key, window)`), so idle/paused accounts keep their last-known usage. Protected id sets are materialized once per pass, not per batch.
- **Per-API-key rollup** (`api_key_usage_rollups`, migration `20260712_020000`): adversarial review caught that per-key lifetime summaries aggregate `request_logs` unbounded and would erode silently under pruning. They now fold in the same slice under the same watermark with their exact semantics (no dedupe, soft-deleted rows counted, warmup excluded); key deletion drops the rollup row. Bonus: removes another unbounded dashboard scan from `GET /api/keys`.

## OpenSpec

`openspec/changes/add-log-retention/` — new `data-retention` capability + `query-caching` delta; `openspec/specs/data-retention/context.md` documents the operator contract (what truncates, floors rationale, `previous_response_id` caveat, folded key sums persisting across account hard-deletes). `openspec validate --specs`: 31/31.

## Multi-agent adversarial review

7 confirmed findings, all addressed: per-key totals erosion (→ key rollup), zero coverage of the watermark guard (mutation test verified the suite passed with the guard deleted → lagging-watermark test added), per-batch GROUP-BY rescans under the SQLite writer lock (→ materialized protected sets), missing validation ceiling (→ `le=3650`), NULL-window/multi-identity and multi-batch test gaps (→ tests added).

## Validation

- SQLite: 3,658 passed (full unit + retention/rollup/api-keys/accounts integration)
- PostgreSQL 16: retention + rollup suites (20) and api-keys paths green; suites added to `POSTGRES_PYTEST_TARGETS`
- Migration `upgrade head`/`check` (drift none)/`downgrade -1`/re-upgrade on both backends
- `ruff` / `ty` clean


Soju06:
@codex review

Soju06:
Addressed the P1 in the latest commit: the `20260712_020000` migration now resets the fold state (deletes `account_usage_rollups` rows **and** rewinds `folded_through` to epoch in the same upgrade — the documented all-or-nothing escape hatch) whenever prior fold state exists, so the next fold pass re-backfills account and key rollups together from raw `request_logs`. Reads stay correct during the re-backfill via the full live-tail fallback. Added a migration regression test that seeds an advanced watermark + rollup rows at `20260712_010000` and asserts the reset after upgrading.

@codex review

Soju06:
@codex review

Soju06:
Addressed both P2s in the latest commit:

- **Snapshot-safe deletes**: request-log pruning now runs only while the fold is current (watermark ≥ now − 2·FOLD_LAG) and deletes only rows a full fold lag below the watermark. A live reader's loaded watermark trails the current one by at most one steady-state fold advance (≪ the 24h lag), so the deleted window is unreachable by any in-flight read; during initial backfill or a stalled scheduler, pruning suspends entirely (regression test updated to pin the skip).
- **Reader-ordered latest protection**: protected usage-history rows are now every row tied at each identity's max `recorded_at` (the readers' ordering) instead of `max(id)`; new regression test inserts an out-of-chronology backfill row and asserts the true latest sample survives.

Green on SQLite and PostgreSQL 16.

@codex review

Soju06:
Rebased on the updated base (#1238's fold-start fix) and extended it here: the earliest-row probe now takes the least of the account-countable and key-countable minimums, since per-key sums also count soft-deleted rows that the account aggregate excludes. Regression test seeds a soft-deleted, key-attributed oldest row and asserts it folds into `api_key_usage_rollups` instead of being skipped below the watermark. All retention + rollup suites green on SQLite and PostgreSQL 16.

@codex review

Soju06:
@codex review

PR #1241, opened 2026-07-12
perf(proxy): cut pre-upstream DB round trips on interactive turns

Soju06:
## Summary

Hot-path audit findings between request arrival and the first upstream byte, all behavior-preserving:

1. **API-key auth cache TTL 2s → 60s.** The 2s TTL is shorter than typical Codex CLI think-time, so interactive turns effectively always missed the cache and re-read the key row (plus lazy-reset writes) before every upstream call. Key mutations already bump the `api_key` invalidation namespace and every instance's poller (0.5s) clears the cache (`app/main.py` registration), so revocation latency is poller-bound and unchanged; the TTL is only the backstop for a broken poller. Per-key `expires_at` is still enforced on every cache hit.
2. **Sticky upsert 4 round trips → 1 + commit.** `INSERT ... ON CONFLICT ... RETURNING` replaces upsert + re-SELECT + `session.refresh`. Row contents and the per-request `updated_at` bump (which feeds the TTL purge cutoff) are unchanged; `expire_on_commit=False` keeps the returned entity readable. Sticky requests pay these round trips inline pre-upstream, so this is direct TTFT.
3. **Same-session `asyncio.gather` removed** in selection-input loading (three `latest_by_account` reads on one `AsyncSession`) — the async/session-ownership rules forbid concurrent statements on a shared session (asyncpg hazard), and the connection serializes statements anyway. Sequential awaits, same results.

## OpenSpec

`openspec/changes/optimize-proxy-preupstream-reads/` — `query-caching` delta (invalidation-driven auth caching with ≥60s TTL backstop, single-statement sticky upsert, no shared-session concurrency). `openspec validate --specs`: 31/31.

## Tests

- New statement-capture test: sticky upsert issues exactly one `INSERT ... RETURNING` per call (insert and conflict-update arms), no follow-up selects.
- New cache tests: poller-style `clear()` blocks stale repopulation via the existing `if_version` guard; TTL floor pinned.
- Sticky + load-balancer + proxy suites green: 3,519 unit; sticky/proxy integration on SQLite and PostgreSQL 16. `ruff`/`ty` clean.

### DIFF
--- a/app/core/auth/api_key_cache.py
+++ b/app/core/auth/api_key_cache.py
@@ -52,7 +52,13 @@ class ApiKeyCache(Generic[_CacheValueT]):
         self._version += 1
 
 
-_api_key_cache: ApiKeyCache[object] = ApiKeyCache(ttl_seconds=2)
+# Key mutations bump the api_key cache-invalidation namespace, and every
+# instance's poller (0.5s interval) clears this cache, so revocation latency
+# is poller-bound, not TTL-bound. The TTL only limits staleness if the poller
+# is broken; keeping it short instead forced interactive Codex turns (which
+# are usually more than a few seconds apart) to re-read the key row from the
+# database before every upstream call.
+_api_key_cache: ApiKeyCache[object] = ApiKeyCache(ttl_seconds=60)
 
 
 def get_api_key_cache() -> ApiKeyCache[object]:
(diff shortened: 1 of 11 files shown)


Soju06:
@codex review

Soju06:
Addressed the P2 in the latest commit: the returning upsert now executes with `populate_existing=True`, so a session that already holds the sticky row in its identity map (e.g. rebinding a key to a different account within one request) gets the persisted attributes back instead of the stale instance. Added a same-session rebind regression test, green on SQLite and PostgreSQL 16.

@codex review

PR #1242, opened 2026-07-12
perf(proxy): make SSE event framing linear instead of quadratic

Soju06:
## Summary

`_iter_sse_events` (`app/core/clients/proxy.py`) read upstream SSE in 1 KiB chunks and, after **every** read, rescanned the entire accumulated buffer from offset 0 for the three separator forms. While a large event accumulates — image outputs and large reasoning deltas can approach the 16 MiB `max_sse_event_bytes` cap — that is O(n²) byte scanning running synchronously on the event loop shared by every in-flight stream.

**Measured** (1 KiB reads, single event):

| Event size | Before | After |
|---|---|---|
| 2 MiB | 4.3 s | 0.038 s |
| 8 MiB | 70.5 s | 0.177 s |

The fix tracks a scan cursor: each read scans only the new bytes plus a 3-byte overlap (the longest separator, `\r\n\r\n`, can straddle a read boundary by at most 3 bytes); the cursor resets after each popped event. Reads are also raised from 1 KiB to 16 KiB (`iter_chunked` caps — it does not pad — so small deltas still flush immediately).

Framing semantics are byte-identical: all three separator forms, separators straddling read boundaries, event-size limit enforcement, and idle-timeout handling are unchanged.

## OpenSpec

`openspec/changes/fix-sse-buffer-scan-quadratic/` — `outbound-http-clients` delta: SSE framing MUST scan each byte a bounded number of times. `openspec validate --specs` green.

## Tests

- New regressions: separator split mid-`\r\n\r\n` across reads; multiple events popped after a large partial (cursor reset).
- `tests/unit/test_proxy_utils.py` full suite (694), proxy integration + e2e suites green; `ruff`/`ty` clean.

### DIFF
--- a/app/core/clients/proxy.py
+++ b/app/core/clients/proxy.py
@@ -129,7 +129,13 @@ def _is_response_stream_terminal_event_type(event_type: str, *, enforce_openai_s
     return event_type == "error" and not enforce_openai_sdk_contract
 
 
-_SSE_READ_CHUNK_SIZE = 1 * 1024
+# 16 KiB reads: 1 KiB reads made a multi-MB SSE event cost thousands of
+# iterations, and each iteration's separator rescan froze the event loop
+# (see the scan cursor in _iter_sse_events).
+_SSE_READ_CHUNK_SIZE = 16 * 1024
+# Longest separator is 4 bytes; a separator can straddle a chunk boundary by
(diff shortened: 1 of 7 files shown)


Soju06:
@codex review

PR #1243, opened 2026-07-12
perf(dashboard): compress responses, cache assets, lazy-load charts

Soju06:
## Summary

First uncached dashboard load shipped **~1.7 MB of uncompressed JS** on the critical path. Three fixes, no visual or API change:

1. **Response compression** — new `DashboardGZipMiddleware` gzips `/api/` and `/assets/` responses only. Proxy paths (`/backend-api`, `/v1`, websockets) are never wrapped in a compressor; verified live that they carry no `content-encoding`.
2. **Immutable asset caching** — Vite content-hashes everything under `/assets/`, so those responses now carry `Cache-Control: public, max-age=31536000, immutable`; `index.html` keeps its existing `no-cache` so deploys swap hashes. Also hoists the per-request `StaticFiles(...)` construction behind an `lru_cache`.
3. **Charts off the critical path** — every chart component was already `React.lazy`, but the forced `vendor-charts` manual chunk swallowed shared helper modules (CJS interop, `use-sync-external-store`, `es-toolkit`, `clsx`), which made the 572 KB recharts chunk a **static** import of the entry chunk and modulepreloaded before first paint; `lazy-recharts.ts` also still statically imported `Cell`. The manual chunk is removed and `Cell` is lazy like the rest — recharts v3 resolves cells via `type.displayName`, which the lazy wrapper carries.

**Measured on the build:** critical path 1.70 MB → ~1.13 MB raw (charts chunk is async-only: absent from `modulepreload` and from entry static imports — verified in the built output and dep maps), ~307 KB transferred with gzip; repeat loads are served from browser cache.

## Verification

- Live curl checks: asset responses gzip + immutable; `index.html` `no-cache`; `/backend-api/*` untouched.
- Frontend: 839 tests green (incl. donut/cell rendering paths). Backend: static/spa unit tests green; `ruff`/`ty` clean.
- OpenSpec: `openspec/changes/optimize-dashboard-serving/` (`frontend-architecture` delta), `openspec validate --specs` green.

### DIFF
--- a/app/core/middleware/dashboard_gzip.py
+++ b/app/core/middleware/dashboard_gzip.py
@@ -18,10 +18,21 @@ class DashboardGZipMiddleware:
         self._gzip = GZipMiddleware(app, minimum_size=minimum_size)
 
     async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
-        if scope["type"] == "http" and scope.get("path", "").startswith(_COMPRESSED_PATH_PREFIXES):
+        if (
+            scope["type"] == "http"
+            and scope.get("path", "").startswith(_COMPRESSED_PATH_PREFIXES)
+            and not _has_range_header(scope)
+        ):
             await self._gzip(scope, receive, send)
             return
         await self._plain(scope, receive, send)
+
+
+def _has_range_header(scope: Scope) -> bool:
+    """Ranged requests bypass gzip: FileResponse builds the 206 and its
+    Content-Range against the uncompressed file, so compressing the body
+    afterwards would describe offsets the encoded bytes no longer match."""
+    return any(name == b"range" for name, _ in scope.get("headers", ()))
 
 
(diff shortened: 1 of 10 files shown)


Soju06:
@codex review

Soju06:
@codex review

Soju06:
Addressed the P3: requests carrying a `Range` header now bypass the compressing wrapper entirely (the 206/`Content-Range` FileResponse builds describes unencoded bytes). Added integration tests for the ranged bypass, asset gzip+immutable headers, and the proxy-path exclusion.

@codex review

Soju06:
@codex review

Soju06:
Addressed the P2: the Range detection now lower-cases header names before comparing (not every ASGI server normalizes casing), with a unit test covering `Range`/`RANGE`/`range`.

@codex review

Soju06:
@codex review

PR #1244, opened 2026-07-12
perf(usage): aggregate the summary window in SQL, drop dead log refresh

Soju06:
## Summary

`GET /api/usage/summary` fetched **every** `RequestLog` ORM row in the secondary window (typically 7 days — 10^5+ full rows with Text columns on a busy proxy) and reduced them in Python on every dashboard poll.

- New `aggregate_usage_metrics_since` computes the window in three small indexed statements: totals (count / errors / tokens / cached), the existing top-error query, and per-model cost. Values flow through the pre-existing `metrics_override` / `cost_override` seam of `build_usage_summary_response`.
- **Helper-exact semantics**, pinned by an equivalence test on both backends: output-token reasoning fallback, per-row `cached ≤ input` clamp (dialect least/greatest; SQLite's 2-arg `min`/`max`), NULL-cost rows excluded from per-model cost, warmup exclusion. Disclosed nuance: top-error **ties** now resolve deterministically (count desc, code asc — the same rule `DashboardService.get_overview` already uses) instead of Python dict insertion order.
- Removes the dead post-commit `session.refresh(log)` in `add_log` — all columns are assigned pre-insert and sessions run `expire_on_commit=False`, so it was a guaranteed extra round trip on **every** proxied request's log write.

## OpenSpec

`openspec/changes/optimize-usage-summary-aggregation/` — `query-caching` delta. `openspec validate --specs` green.

## Tests

- New equivalence test seeding reasoning-fallback rows, cached>input, negative cached, NULL input, NULL-cost models, warmup rows — SQL aggregate == legacy Python summation (SQLite + PostgreSQL 16).
- 3,516 unit + usage/request-log/dashboard integration suites green; `ruff`/`ty` clean.

### DIFF
--- a/app/core/usage/types.py
+++ b/app/core/usage/types.py
@@ -133,3 +133,22 @@ class RequestActivityAggregate:
     output_tokens: int
     cached_input_tokens: int
     cost_usd: float
+
+
+@dataclass(frozen=True)
+class UsageSummaryLogsAggregate:
+    """SQL-side replacement for summing a window of RequestLog ORM rows in
+    Python; field semantics mirror the log helpers exactly (reasoning-token
+    fallback for output, per-row cached<=input clamp, None-cost rows excluded
+    from per-model cost)."""
+
+    request_count: int
+    error_count: int
+    total_tokens: int
+    cached_input_tokens: int
+    top_error: str | None
+    cost_by_model: list[tuple[str, float]]
+
+    @property
+    def cost_total_usd(self) -> float:
+        return sum(cost for _, cost in self.cost_by_model)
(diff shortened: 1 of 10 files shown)


Soju06:
@codex review

Soju06:
Addressed the P2 in the latest commit: the window now aggregates in **one** grouped statement (`GROUP BY model, is_error, error_code` — cardinality stays tiny) and every metric (counts, error rate, top error, per-model cost) derives from that single result set in Python, restoring the internal consistency the legacy single-SELECT path had. Bonus parity: empty-string error codes are skipped for top-error exactly like the legacy helper. Equivalence test green on SQLite and PostgreSQL 16.

@codex review

Soju06:
Addressed the P3: the delta scenario now states the deterministic tied top-error rule (highest count, then code ascending) as the sole exception to legacy equality, so the spec matches the implemented contract.

@codex review

PR #1245, opened 2026-07-12
perf(clients): keep upstream connections and DNS across interactive turns

Soju06:
## Summary

The shared upstream connectors ran on aiohttp defaults: **15 s connection keepalive and 10 s DNS TTL**. Interactive Codex turns are usually further apart than that, so nearly every turn paid a fresh DNS lookup + TCP/TLS handshake to chatgpt.com (~100–300 ms of time-to-first-token) even though the `ClientSession` itself is pooled and shared.

Both direct `TCPConnector`s (HTTP + websocket handshake) now set `keepalive_timeout=90` and `ttl_dns_cache=300`. SOCKS-proxied connectors are untouched. No other behavior changes.

## OpenSpec

`openspec/changes/tune-upstream-connection-reuse/` — `outbound-http-clients` delta pinning the reuse floors. `openspec validate --specs` green.

## Tests

Connector-construction unit assertions updated to pin the new kwargs; `tests/unit/test_http_client.py` (16) green; `ruff`/`ty` clean.

### DIFF
--- a/app/core/clients/http.py
+++ b/app/core/clients/http.py
@@ -136,6 +136,13 @@ async def _build_http_client() -> HttpClient:
             limit=settings.http_connector_limit,
             limit_per_host=settings.http_connector_limit_per_host,
             ssl=ssl_context,
+            # aiohttp defaults (15s keepalive, 10s DNS TTL) are shorter than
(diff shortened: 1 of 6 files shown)


Soju06:
@codex review

PR #1256, opened 2026-07-13
fix(db): serialize startup migrations across replicas and report schema-ahead correctly

Soju06:
## Why

Every replica runs Alembic `upgrade head` at boot with no cross-process mutual exclusion. Two replicas starting concurrently both read the same current revision and race DDL: the loser dies on duplicate-object errors — including the non-idempotent `CREATE TABLE alembic_version` in the capacity helper and the DELETE+INSERT legacy remap — and crash-loops under the entrypoint's `set -eu`. Separately, migration state inspection is direction-blind (`needs_upgrade = current != head`), so an old binary restarting against a schema migrated by a newer build reports a misleading "schema is behind Alembic head" error, sending operators the wrong way during rollouts.

## What Changes

- New `app/db/migration_lock.py`: a sync `migration_lock(...)` context manager providing a cross-process mutex — PostgreSQL: session-level `pg_try_advisory_lock(hashtext('codex_lb:migrations'))` polled on a dedicated connection held for the whole upgrade; file-backed SQLite: an exclusive `BEGIN IMMEDIATE` transaction on a sentinel `<db_path>.migrate-lock` file; in-memory SQLite/unknown dialects: no-op.
- `run_upgrade()` acquires the lock around its entire sequence; after acquiring it re-inspects migration state and, when the schema is already at head, logs and returns — the losing replica's startup succeeds. Covers both app startup and the container-entrypoint CLI path.
- `stamp_revision()` takes the same lock; the alembic_version capacity helper switches to `CREATE TABLE IF NOT EXISTS` as defense-in-depth.
- New setting `database_migration_lock_timeout_seconds` (default 300.0, env `CODEX_LB_DATABASE_MIGRATION_LOCK_TIMEOUT_SECONDS`); on timeout, a `TimeoutError` names the lock and the setting.
- `MigrationState` gains `unknown_revisions` and `is_ahead`; the schema-ahead case now raises a direction-correct "schema revision(s) are not known to this build" message instead of claiming the schema is behind head.
- Spec deltas in `database-migrations` (two ADDED requirements) plus `context.md` ops notes (per-backend lock mechanism, sentinel-file lifecycle, PgBouncer/NFS caveats).
- No new DB table and no new Alembic revision (the mechanism is a lock, not schema).

## Testing

- `uv run pytest tests/integration/test_migration_serialization.py tests/integration/test_migrations.py tests/unit/test_db_migrate.py tests/unit/test_db_session.py -q` → 90 passed, 5 skipped (PG-only under SQLite).
- PostgreSQL run (`CODEX_LB_TEST_DATABASE_URL=postgresql+asyncpg://...`): `tests/integration/test_migration_serialization.py tests/integration/test_migrations.py -q` → 25 passed, 0 skipped — includes the PG concurrent-race and advisory-lock timeout tests.
- `uv run pytest tests/unit/test_otel.py tests/unit/test_settings_cache.py tests/unit/test_settings_multi_replica.py tests/unit/test_settings_service.py -q` → 62 passed. One pre-existing test failure (capacity CREATE TABLE assertion, missed by an interrupted earlier run) was found and fixed.
- `ruff check` / `ruff format --check` clean; `openspec validate serialize-startup-migrations` valid; `openspec validate --specs` → 31 passed.

## Merge ordering / notes

- Depends on #1252 (OpenSpec SSOT sync) — MODIFIED spec deltas target the synced spec headers.
- No Alembic revision added; safe to land in any order relative to the sibling migration-bearing PRs (and makes their concurrent-startup application safer).
- Part of the 12-change multi-replica reliability effort; resolves no filed issues.

### DIFF
--- a/app/core/config/settings.py
+++ b/app/core/config/settings.py
@@ -153,6 +153,7 @@ class Settings(BaseSettings):
     database_sqlite_pre_migrate_backup_max_files: int = Field(default=5, ge=1)
     database_sqlite_startup_check_mode: Literal["quick", "full", "off"] = "quick"
     database_alembic_auto_remap_enabled: bool = True
+    database_migration_lock_timeout_seconds: float = Field(default=300.0, gt=0)
     upstream_base_url: str = "https://chatgpt.com/backend-api"
     upstream_stream_transport: Literal["http", "websocket", "auto"] = "auto"
     http_downstream_transport_policy: Literal["smart", "always_http", "always_websocket", "pinned"] = "smart"
--- a/app/db/migrate.py
+++ b/app/db/migrate.py
@@ -21,6 +21,7 @@ from sqlalchemy.engine import Connection
 
 from app.core.config.settings import get_settings
 from app.db.alembic.revision_ids import LEGACY_MIGRATION_TO_NEW_REVISION, OLD_TO_NEW_REVISION_MAP, REVISION_ID_PATTERN
+from app.db.migration_lock import migration_lock
 from app.db.migration_url import to_sync_database_url
 from app.db.models import Base
 
@@ -127,6 +128,8 @@ class MigrationState:
     has_alembic_version_table: bool
     has_legacy_migrations_table: bool
     needs_upgrade: bool
+    unknown_revisions: tuple[str, ...]
+    is_ahead: bool
 
 
 class MigrationBootstrapError(RuntimeError):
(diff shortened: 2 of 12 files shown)


Soju06:
@codex review

Soju06:
@codex review

Soju06:
@codex review

PR #1270, opened 2026-07-13
perf(proxy): parse each SSE event once per layer, skip no-op reserialization

Soju06:
## Summary

Follow-up to the perf-audit streaming findings (the O(n²) framing fix landed in #1242; this addresses the redundant per-event CPU). Each streamed SSE event was parsed 5–7× and re-serialized 2–3× across layers, all on the shared event loop:

| Layer | Before | After |
|---|---|---|
| Streaming mixin | `parse_sse_data_json` + `parse_sse_event` (internally re-parses) per event | one `json.loads`, validate from the payload (`parse_sse_event_payload`) |
| Tool-call rewrite | full re-parse even when nothing changed; re-parse of its own serialized output when changed | reuses the caller's parsed event; validates the rewritten payload directly |
| Bridge upstream reader | same double parse + rewrite re-parse | same single-parse treatment |
| `/v1` normalizers | parse + unconditional re-serialize per event | identity-gated raw pass-through (every mutating branch copies the dict, so `normalized is parsed` proves no mutation); reasoning normalizer replays buffered raw blocks on clean flushes |

The mixin's final canonical `format_sse_event` is **intentionally kept** — it is the single serialization guaranteeing stable downstream framing, and the `/v1` pass-through is byte-identical precisely because its input is that canonical output.

**Measured**: mixin layer 21.4 → 12.1 µs per 400-byte delta event (the removed rewrite re-parse and `/v1` savings come on top).

## OpenSpec

`openspec/changes/optimize-sse-single-parse/` — `responses-api-compat` delta (parse-once + no-op-pass-through contract). `openspec validate --specs` green.

## Validation

- Unit 3,562; integration-core 1,150; bridge/ws/e2e 176 — all green (the suites' byte-level SSE assertions are the parity oracle).
- Proxy architecture check, `ruff`, `ty` clean.

### DIFF
--- a/app/core/openai/parsing.py
+++ b/app/core/openai/parsing.py
@@ -19,7 +19,15 @@ _COMPACT_RESPONSE_ADAPTER = TypeAdapter(CompactResponsePayload)
 
 
 def parse_sse_event(line: str) -> OpenAIEvent | None:
-    payload = parse_sse_data_json(line)
+    return parse_sse_event_payload(parse_sse_data_json(line))
+
+
+def parse_sse_event_payload(payload: JsonValue | None) -> OpenAIEvent | None:
+    """Validate an already-parsed SSE data payload.
+
+    Hot streaming paths parse each event's JSON exactly once and reuse the
+    payload here instead of re-parsing the raw line per consumer.
+    """
     if payload is None:
         return None
     try:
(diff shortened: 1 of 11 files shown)


Soju06:
@codex review

Soju06:
Addressed the P2: the raw pass-through now also requires the block to already carry the canonical `event: <type>` framing that `format_sse_event` would emit — data-only blocks (e.g. bridge-rewritten terminal events) are reformatted exactly as before, so named-event clients keep the `event:` line. The reasoning-normalizer raw replays are upstream of this normalizer, so they flow through the same gate. Regression tests cover both directions.

@codex review

Soju06:
@codex review

PR #1272, opened 2026-07-13
fix(automation): fall back to github.token when label sync hits rate limits

Soju06:
## Why

`CODEX_LABEL_SYNC_TOKEN` is a user PAT whose 5,000/hr REST quota is shared with every other consumer of that user's token. During the 2026-07-13 adaptive-windows review cycle the quota exhausted repeatedly, failing every "Sync Codex labels" run (HTTP 403) and painting spurious CI failures on #1266–#1268 for up to an hour at a time.

## What

- The gh wrapper in `sync_codex_ok_labels.py` detects `API rate limit exceeded` failures, switches once to `GH_FALLBACK_TOKEN`, and retries the failed call; subsequent calls keep the fallback.
- Both workflow jobs provide `github.token` as `GH_FALLBACK_TOKEN` — it carries a separate per-repository Actions quota (1,000/hr), ample for label sync.
- No usable distinct fallback (absent, identical, or also exhausted) keeps the existing fail-on-read contract.

Longer-term the PAT should move to a machine account or GitHub App to fully decouple quotas; this PR removes the recurring failure mode either way.

## OpenSpec

`openspec/changes/label-sync-rate-limit-fallback/` — modifies the `github-automation` write-token fallback requirement with runtime rate-limit scenarios. Strict validation passes.

## Testing

- New unit tests: fallback activates once and retries; identical fallback fails; exhausted fallback fails (`tests/unit/test_sync_codex_ok_labels.py`, 28 passed).

### DIFF
--- a/.github/scripts/sync_codex_ok_labels.py
+++ b/.github/scripts/sync_codex_ok_labels.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 
 import argparse
 import json
+import os
 import re
 import subprocess
 import sys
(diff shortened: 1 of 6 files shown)


Soju06:
@codex review

PR #1277, opened 2026-07-13
perf(request-logs): cache the listing total per filter signature

Soju06:
## Summary

The last remaining dashboard full-scan from the perf audit: every request-log page load **and every pagination click** ran an exact `COUNT(*)` over the filtered set — effectively the whole table under default filters, re-executed by the dashboard's 30s poll.

The exact total only feeds the display-only "X–Y of N" figure and the last-page jump (verified in `pagination-controls.tsx`), which tolerate short staleness — so instead of changing the API contract to `has_more` (a UX regression), the total is now cached **per filter signature** (all filters except limit/offset):

- `CODEX_LB_REQUEST_LOG_COUNT_CACHE_TTL_SECONDS` — default 30 s, `0` disables. The test suite disables it globally so in-test totals stay exact; a dedicated statement-capture test re-enables it and pins one count per signature (page 2 reuses page 1's count; a different filter counts separately).
- Bounded at 256 entries, per-instance, display-only; page contents remain exact and newest-first, so fresh rows always appear on page 1 regardless of a ≤30 s-stale total.

## OpenSpec

`openspec/changes/cache-request-log-count/` — `query-caching` delta. Validation green.

## Tests

3,694 unit + request-log suites green; `ruff`/`ty` clean.

### DIFF
--- a/app/core/config/settings.py
+++ b/app/core/config/settings.py
@@ -226,6 +226,8 @@ class Settings(BaseSettings):
     sticky_session_cleanup_interval_seconds: int = Field(default=300, gt=0)
     # Data retention (0 = disabled). Non-zero values have safety floors so
     # every in-product consumer window stays inside retained data.
+    # Display-only pagination total for the request-log listing; 0 disables.
+    request_log_count_cache_ttl_seconds: float = Field(default=30.0, ge=0)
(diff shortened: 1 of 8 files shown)


Soju06:
@codex review

PR #1278, opened 2026-07-13
perf(frontend): code-split dashboard routes

Soju06:
## Summary

Follow-up to #1243 (which took charts off the critical path): the entry chunk still carried **all six pages'** code because `App.tsx` imported them statically. Each route now loads via `React.lazy` behind a single `<Suspense fallback={null}>` in the app layout.

**Measured on the build:**

| | Before | After |
|---|---|---|
| Entry chunk (raw) | 807 KB | **360 KB** |
| Entry chunk (gzip) | 214 KB | **112 KB** |

Per-page chunks (dashboard 71 KB, settings 133 KB, accounts 52 KB, automations 54 KB, …) fetch on first visit. Verified in the built output that the entry neither statically imports nor modulepreloads any page chunk, and the recharts chunk stays async-only.

No visual change — the fallback is null and page data loading already gates rendering.

## OpenSpec

`openspec/changes/split-dashboard-routes/` — `frontend-architecture` delta. Validation green.

## Tests

Frontend suite: 841 passed (122 files).

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -1,3 +1,4 @@
+import { lazy, Suspense } from "react";
 import { Navigate, Outlet, Route, Routes } from "react-router-dom";
 
 import { AppHeader } from "@/components/layout/app-header";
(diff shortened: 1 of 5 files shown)


Soju06:
@codex review

PR #1279, opened 2026-07-13
perf(frontend): self-host JetBrains Mono, drop Google Fonts

Soju06:
## Summary

The last external request on the dashboard's critical path: `index.html` pulled JetBrains Mono via a **render-blocking** Google Fonts stylesheet, so every first paint waited on `fonts.googleapis.com` — and on air-gapped/egress-restricted deployments (a common shape for this product) first paint stalled until the request timed out. Geist Sans was already self-hosted.

- JetBrains Mono variable (weights 400–500 from one file per subset; latin + latin-ext, **~43 KB total**) now ships in `frontend/public/fonts/` with `@font-face` + `font-display: swap` + original `unicode-range`s, mirroring the existing Geist Sans pattern.
- The Google Fonts `<link>` and both preconnects are removed; the built shell has **zero** external origins (verified by grep over the build output).
- Non-latin glyphs (cyrillic/greek/vietnamese, previously extra remote subsets) fall back to the `ui-monospace, monospace` stack.

Assets under `/assets` and `/fonts` also benefit from #1243's gzip + immutable caching.

## OpenSpec

`openspec/changes/self-host-mono-font/` — `frontend-architecture` delta (fully self-hosted shell). Validation green.

## Tests

Frontend suite: 841 passed (122 files).


Soju06:
@codex review

PR #1295, opened 2026-07-13
fix(db): handle percent-encoded SQLite URLs on Windows

alchemistkiv:
## Summary

On Windows the default SQLite database URL contains a drive letter and backslashes (e.g. `sqlite+aiosqlite:///C:\Users\...\store.db`). SQLAlchemy's `URL.render_as_string()` percent-encodes that path into `sqlite:///C%3A%5CUsers%5C...%5Cstore.db`. Two code paths break on the resulting bare `%` characters, preventing the server from starting or serving dashboard/usage endpoints.

This PR fixes both paths so codex-lb runs on Windows with the default `~/.codex-lb/store.db` SQLite database.

### 1. Alembic startup crash (`app/db/migrate.py`)

`_build_alembic_config()` stores the encoded URL via `config.set_main_option("sqlalchemy.url", ...)`. Alembic's `Config` is backed by `configparser` with `BasicInterpolation`, which treats a lone `%` as interpolation syntax. On a fresh startup (or when `inspect_migration_state()` is exercised) this raises:

```
ValueError: invalid interpolation syntax in 'sqlite:///C%3A%5CUsers%5CKIVANC%5C.codex-lb%5Cstore.db' at position 11
ERROR:    uvicorn.error Application startup failed. Exiting.
```

The server never reaches "Application startup complete".

### 2. Runtime usage queries hit a stray database (`app/db/sqlite_utils.py`)

`sqlite_db_path_from_url()` returns the literal percent-escaped string without decoding it. `sqlite3.connect()` then either:

- fails with `unable to open database file` (when the encoded path can't be created), or
- **silently creates a 0-byte database** next to the current working directory (e.g. `C:\Users\KIVANC\C%3A%5CUsers%5CKIVANC%5C.codex-lb%5Cstore.db`).

The usage repository reads accounts/usage from that empty database, so every dashboard endpoint that joins account data returns:

```
sqlite3.OperationalError: no such table: accounts
GET /api/accounts -> 500 Internal Server Error
GET /api/dashboard/overview -> 500 Internal Server Error
```

while `/health` and `/v1/models` still return 200, which makes the failure easy to miss.

### `app/db/migrate.py`

Escape `%` -> `%%` before handing the URL to the Alembic `Config`. `configparser` stores `%%` verbatim and `get_main_option()` decodes `%%` back to `%`, so the URL handed to SQLAlchemy is unchanged — the escaping is transparent to every consumer of the config.

```python
sync_database_url = to_sync_database_url(database_url)
config.set_main_option("sqlalchemy.url", sync_database_url.replace("%", "%%"))
```

### `app/db/sqlite_utils.py`

`urllib.parse.unquote()` the path before turning it into a `Path`, so `sqlite3.connect()` receives the real filesystem path on every platform.

```python
path = urllib.parse.unquote(path)
```

## Reproduction (Windows, default SQLite)

1. `uv tool install --pre codex-lb` (any 1.21.0-beta.x; also reproduces on 1.20.x once a migration check runs against a Windows path).
2. Start the server: `codex-lb` (dashboard at `http://127.0.0.1:2455`).
3. Observe startup failure with the `invalid interpolation syntax` traceback, or — once that is bypassed — `GET /api/accounts` returning `500` with `no such table: accounts` and a stray `C%3A%5C...store.db` file appearing next to the CWD.

After this PR the server reaches `Application startup complete`, the stray file is no longer created, and `/api/accounts`, `/api/dashboard/overview`, `/v1/models` all return `200`.

## Why this is a Windows-specific regression risk

The default `DEFAULT_DATABASE_URL` is built from `Path.home() / ".codex-lb" / "store.db"`, which on Windows is an absolute path starting with a drive letter and using backslashes. `make_url(...).render_as_string()` percent-encodes both `:` (`%3A`) and `\` (`%5C`), so any Windows SQLite install hits both branches above. Non-Windows paths (forward slashes, no drive letter) produce no `%` and are unaffected.

## Checklist

- [x] Targets the default SQLite backend shipped with codex-lb (no config changes required by users).
- [x] No behaviour change for non-Windows / non-SQLite (PostgreSQL) installs — `unquote` is a no-op on already-decoded paths, and the `%%` escaping only matters when a `%` is present.
- [x] Changes are minimal and localized to the two failing code paths.


ozpool (commented):
Verified both changes against `main`; they hold up and are Unix-safe. LGTM.

- The encoder is confirmed: `to_sync_database_url` returns `make_url(...).render_as_string(hide_password=False)` (`app/db/migration_url.py:15`), which percent-encodes a Windows path (`C:\Users\...` -> `C%3A%5CUsers%5C...`).
- `sqlite_db_path_from_url` (`app/db/sqlite_utils.py`) slices the path and hands it straight to `Path(path).expanduser()` with no decode, so the `%3A%5C` literal reaches `sqlite3.connect()` — matches the "unable to open database file" / stray 0-byte db symptom. `urllib.parse.unquote(...)` placed after the `?`/`#` partitions is the right spot.
- The `%` -> `%%` escape in `_build_alembic_config` is the correct configparser idiom: `set_main_option`/`get_main_option` round-trip through `BasicInterpolation`, so `%%` decodes back to `%` and the URL SQLAlchemy receives is unchanged.

Round-trip stays correct even when the value already contains an encoded `%`: `render_as_string` emits `%25`, the escape makes `%%25`, `get_main_option` decodes to `%25`, SQLAlchemy decodes to `%`. On POSIX both paths are no-ops (no `%` to escape, nothing to unquote), so there's no behavior change off Windows.

Soju06:
Thanks for the fix. Two gate items before this can merge:

1. **Regression test** at the failing surface: a test that feeds a Windows-style `sqlite+aiosqlite:///C:\Users\...` URL through the same code path and fails without this change (bug fixes need coverage at the externally failing path, not helper-only).
2. **OpenSpec**: Windows URL compatibility is a documented-behavior change, so please add a small `openspec/changes/<slug>/` entry.

I've updated the branch from `main` so the new required checks run.

Soju06:
@codex review

alchemistkiv:
Thanks @Soju06 — both gate items addressed in e520e32f.

1. **Regression test** (	ests/unit/test_sqlite_url_windows.py): feeds a percent-encoded Windows SQLite URL (sqlite:///C%3A%5CUsers%5Cme%5C.codex-lb%5Cstore.db) directly into sqlite_db_path_from_url (decoded filesystem path returned, not the literal escaped string) and _build_alembic_config (no ValueError; URL round-trips through get_main_option). Verified the tests **fail without the fix** with the exact production symptom (ValueError: invalid interpolation syntax in ...C%3A%5C... at position 11, plus the literal escaped path reaching sqlite3) and **pass with it**. Locally green on 
uff, 	y, and pytest (4 passed).

2. **OpenSpec** (openspec/changes/windows-sqlite-url-encoding/): proposal + tasks + database-backends and database-migrations spec deltas.

The encoded URL is fed directly (rather than going through 	o_sync_database_url) so the failing surface reproduces on any platform: on Linux CI the encoded form keeps its %, so the %% escape is actually exercised; on Windows the round-trip is a no-op there.

Soju06:
@codex review

CI runs approved on e520e32f. Re-requesting Codex on this head — the two open P1 threads (missing OpenSpec change + regression tests at the failing product path) look addressed by this commit; please resolve them with a pointer once the fresh review confirms.

Soju06:
Update from the maintainer side: I approved the CI runs on e520e32f, but the run came back red on **Lint (ruff)** — please run `uv run ruff check --fix` / `ruff format` locally and push. Codex's fresh review on this head also left three P1 threads that match the 2026-07-16 gate asks: (1) an OpenSpec change folder for the DB-URL behavior change, (2) regression coverage at the externally failing product path, and (3) a test that exercises the failing application path (not just `sqlite_db_path_from_url` directly). It looks like e520e32f didn't include those pieces — once they land with lint green, ping here and I'll re-approve CI and take it from there.

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
The source fix at `d4041dfb261f441d39dd43a43aaf82ea8b9651ac` has green CI and a clean exact-head Codex review, but the branch is now 43 commits behind `main`.

I attempted the maintainer `update-branch` path. GitHub refused it because the current OAuth credential cannot update `.github/workflows/codex-review-labels.yml`; the contributor branch was not changed.

Please merge or rebase current `main`, then rerun CI and `@codex review`. I added `needs rebase` so the queue reflects the remaining gate.

Soju06:
Thanks for sticking with this through all the review rounds — the core fix is right (the `%` -> `%%` escape for Alembic's `ConfigParser` in `_build_alembic_config`, plus decoding only the recognizable SQLAlchemy-rendered Windows forms in `normalize_sqlite_url`), CI is fully green at d4041df, and Codex's code review at that head came back clean. Three threads from the 07-23 Codex pass are still open and are the last gate:

1. `openspec/changes/windows-sqlite-url-encoding/specs/database-migrations/spec.md` — the scenario still describes the percent-encoded URL reaching the Alembic config, but since `to_sync_database_url()` now routes through `normalize_sqlite_url()` first, the encoded Windows form is decoded before the config is built. Please align the scenario with the implemented behavior (the `%%` escape now primarily protects literal `%` characters, e.g. in passwords).
2. `openspec/changes/windows-sqlite-url-encoding/specs/database-backends/spec.md` — the requirement says every SQLite URL converted for filesystem use MUST be percent-decoded, but the implementation deliberately decodes only the recognized SQLAlchemy-rendered Windows drive/UNC forms and preserves literal `%xx` elsewhere (correctly so). Please narrow the wording to match.
3. `app/db/sqlite_utils.py` — UNC edge: `normalize_sqlite_url()` decodes `%23` to a literal `#` (e.g. `sqlite:///%5C%5Cserver%5Cshare%23x%5Cstore.db`), and `sqlite_db_path_from_url()` then truncates the path at that `#` because only the raw-drive branch skips the fragment split. Either skip the `#` partition for decoded UNC (`\\`) paths as well, or keep `%23` encoded during normalization.

On the `needs rebase` label: no rebase is actually required — I checked and this branch merges cleanly into current main; the earlier third-party update-branch attempt failed only on workflow-file token permissions. Once the three items above land, I'll update the branch from the maintainer side, re-approve CI, and take it from there. This is fix-class, so it is exempt from the current beta-soak train.

Soju06:
Maintainer takeover: this change has been rebased onto current `main` and continued in #1597, which carries over all commits here with original authorship and resolves the three open review threads (UNC `#` truncation fix + OpenSpec scenario/requirement corrections). @alchemistkiv thank you for the fix and the investigation — you remain the author of record on the carried-over commits. Follow-up review will happen on #1597.

Soju06:
Closing: the maintainer takeover #1597 (which carried these commits forward with original authorship and resolved the open review threads) merged on 2026-08-06, so the Windows percent-encoded SQLite URL fix has shipped from that branch. Nothing left to land from this head — thanks again for the fix, the authorship is preserved in #1597.

PR #1298, opened 2026-07-14
feat(retention): enforce seven-day request-log rollups

RyeBelen:
## What changed

- set raw request-log retention to seven UTC days
- roll pruned rows into daily aggregates before deletion
- preserve API-key lifetime usage, account-deduplicated usage, dashboard activity, reports, and API-key limit projections across the raw/rollup boundary
- fail closed when grouped and deleted row counts differ
- add and verify the aggregate schema migration and regression coverage

## Why

The production SQLite database retained raw request logs longer than needed. Existing coarse rollups did not preserve every downstream projection required to safely shorten that window, especially effective output tokens, exact cost accounting, account request deduplication, dashboard history, and report filters.

## Impact

Operators can now dry-run and apply a seven-day prune while retaining long-lived metrics. Recent detailed logs remain raw; older complete UTC days remain available through aggregates. API-key admission accounting remains conservative for rolled-up partial boundary days.

## Validation

- comprehensive repository set: 278 passed, 2 skipped
- focused final set: 21 passed
- Ruff and `ty check`: passed on changed paths
- migration schema/remap coverage: passed
- production Railway deployment: `a84956b8-0c0b-4a7a-b750-805b95ffb802`
- production prune: 54,551 raw rows -> 2,740 aggregate groups, exact delete parity
- repeated production dry-run: 0 eligible rows
- API-key, account, and general metric projections: identical before/after
- SQLite `quick_check`: `ok` before prune, after prune, and after compaction
- live health and proxied Responses API traffic: HTTP 200

## Production backup

The verified pre-prune SQLite snapshot and gzip copy remain on the production volume. Their paths and SHA-256 hashes are recorded in the OpenSpec change context.

## Known validation constraint

`openspec validate --specs` could not be run locally because the OpenSpec executable is not installed in this environment. The OpenSpec artifacts are included and all available code, migration, and production checks passed.


RyeBelen:
Superseded: we are dropping the fork-specific seven-day retention implementation and adopting upstream's native retention pipeline instead. Closing this conflicting PR without merge.

PR #1312, opened 2026-07-14
fix(api-keys): reject duplicate limit rules on create

ShiroKSH:
## Summary

- Reject duplicate API-key limit rule identities during creation.
- Share the existing duplicate-rule validation with the update path.
- Add an API regression test that verifies the failed create persists no key.

## Root cause

The create path stored every submitted limit rule, while the update path already rejected duplicate `(limit_type, limit_window, model_filter)` identities. Duplicate persisted rules were enforced independently and could silently make an API key stricter than configured.

## Validation

- `uv run pytest tests/integration/test_api_keys_api.py::test_api_key_create_rejects_duplicate_limit_rules tests/integration/test_api_keys_api.py::test_api_key_update_does_not_persist_assignments_when_limits_are_invalid tests/unit/test_api_keys_validation_error.py -q`
- `uv run ruff check app/modules/api_keys/service.py tests/integration/test_api_keys_api.py`
- `uv run ruff format --check app/modules/api_keys/service.py tests/integration/test_api_keys_api.py`
- `uv run ty check`
- `npx --yes @fission-ai/openspec@latest validate --specs`


ozpool (commented):
Verified against `main` (`43aa3942`). The asymmetry is real: `create_key` only ran `_validate_model_enforcement`, while `_build_limit_rows_for_update` already rejected duplicate identities, so the same payload was accepted on create and refused on edit. Sharing one validator across both paths is the right shape.

Checks I ran on the diff:

- **The refactor drops nothing.** `submitted_by_key` in `_build_limit_rows_for_update` was only ever used for the `len(...) != len(submitted_limits)` comparison (no later reference in the function), so replacing it with `_validate_unique_limit_rule_identities` is behavior-preserving. The set-of-identities check is equivalent to the old dict-vs-length check.

- **Placement is right for the spec's "before it persists" clause.** The new call sits ahead of `_generate_plain_key()` and ahead of the `_resolve_assigned_account_ids` / `_resolve_assigned_source_ids` round-trips, so a duplicate payload fails before any key material is generated and before those lookups run, not just before the insert.

- **No None hazard from the unguarded call.** `create_key` previously only touched limits under `if payload.limits:`, and `ApiKeyCreateRequest.limits` is `list[LimitRuleCreate] | None` at the schema layer, so validating `payload.limits` unguarded at the top looked worth checking. It is fine: the route builds limits via `_build_limit_inputs(payload)` (`api.py:128`), which always returns a list, and `ApiKeyCreateData.limits` is typed `list[LimitRuleInput]`. The update path's `... if limits_set else None` (`api.py:185`) feeds `ApiKeyUpdateData`, not this call.

- **Test matches the spec scenario.** It asserts the typed `400` / `invalid_api_key_payload` envelope *and* that nothing was persisted via the follow-up `GET /api/api-keys/` returning `[]`, which is the "AND no API key or limit row is persisted" clause rather than just the status code. It fails on `main`, where the create returns 200.

One optional suggestion, non-blocking: `ApiKeyValidationError("Duplicate limit rules are not allowed")` does not say which rule collided. Since the identity tuple is already computed, including the offending `(limit_type, limit_window, model_filter)` would make this easier to act on when a payload carries many rules.

LGTM otherwise.

Soju06:
@codex review

PR #1315, opened 2026-07-14
fix(proxy): retry server_is_overloaded responses

### LINKED ISSUE #1313: bug: server_is_overloaded is not treated as retryable transient
### Pre-flight checklist

- [x] I searched open and closed issues; this is a remaining error-code gap related to #565, not a duplicate of the already-fixed `overloaded_error` case.
- [x] Reproduced against current `main` (`13d3a321`).
- [x] The behavior is covered by a deterministic unit-level reproduction.

### codex-lb version

`main` at `13d3a321` (also present in `v1.21.0-beta.3`).

### Deployment method

From source / Docker.

Codex CLI, OpenCode, and direct streaming HTTP clients can all observe the same Responses failure envelope.

Mixed pool.

### Model(s) involved

Any Responses model when upstream emits `error.code = "server_is_overloaded"`.

### What happened?

Upstream can now emit a terminal streamed error with:

```json
{"code":"server_is_overloaded","message":"Our servers are currently overloaded. Please try again later."}
```

`app/modules/proxy/helpers.py` classifies only `server_error`, `upstream_error`, `stream_incomplete`, and `overloaded_error` as transient without relying on an HTTP 5xx. The streaming retry set in `app/modules/proxy/service.py` also omits `server_is_overloaded`.

When this envelope arrives after the SSE connection already returned HTTP 200, the failure is not treated like the existing `overloaded_error` case and can be surfaced instead of retried/failing over.

`server_is_overloaded` should be classified as `retryable_transient` and use the same bounded pre-visible retry/failover path as `overloaded_error`.

### Steps to reproduce

1. Build a streamed Responses fixture whose first terminal event has `error.code = "server_is_overloaded"` and no 5xx status.
2. Pass the code through `classify_upstream_failure` or the streaming retry predicate.
3. Observe that current `main` does not classify it as a transient retry code.

### Relevant logs

```shell
response.failed: code=server_is_overloaded message="Our servers are currently overloaded. Please try again later."
```

```yaml
upstream_stream_transport: http-or-websocket
multiple_accounts: true
```

### Additional context

Related to #565. The existing OpenSpec requirement covers `overloaded_error`; this report proposes extending that contract to the newer `server_is_overloaded` code.

mereyabdenbekuly-ctrl:
## Summary

Treat the newer upstream `server_is_overloaded` Responses error code as the
same retryable transient condition as `overloaded_error`, including bounded
pre-visible stream retry.

## Type of change

- [ ] **Breaking change**

Linked issue: Fixes Soju06/codex-lb#1313. Related to
Soju06/codex-lb#565.

## OpenSpec

Change directory: `openspec/changes/retry-server-is-overloaded/`

## Changes

- Add `server_is_overloaded` to the canonical retryable-transient classifier.
- Put both known overload codes in the existing bounded stream retry set
  without changing retry budgets or replay safety.
- Extend the Responses compatibility requirement for both known overload codes.
- Parameterize the routed `/backend-api/codex/responses` regression across
  `overloaded_error` and `server_is_overloaded`.
- Preserve SSY's original implementation commit and add the alias-completeness
  correction as a separate maintainer tail commit.

## Test plan

```text
PATH="$PWD/.venv/bin:$PATH" make lint
# proxy architecture checks passed
# Ruff check/format passed across the repository

uv run ty check
# All checks passed

uv run pytest -q tests/unit/test_failover_foundation.py tests/integration/test_proxy_transient_retry.py
# 71 passed in 10.14s

npx --yes @fission-ai/openspec@1.6.0 validate --specs
# 43 passed, 0 failed

npx --yes @fission-ai/openspec@1.6.0 validate retry-server-is-overloaded --strict
# Change 'retry-server-is-overloaded' is valid
```

## Screenshots / output

The integration regression sends each known overload code as the first
`response.failed` SSE event and verifies that the public Responses route retries
and returns only `response.completed` from the second attempt.

## Checklist

- [x] Ran the relevant local CI subsets.
- [x] OpenSpec validation passes and all change artifacts are complete.


ozpool (commented):
Verified the core classification change against `main` (`43aa3942`). Adding `server_is_overloaded` to `_TRANSIENT_CODES` is correct and mirrors the existing `overloaded_error` precedent: `classify_upstream_failure` only reaches `retryable_transient` through that set or an HTTP 5xx, so a terminal SSE envelope arriving on an already-200 stream had no path to it. The integration test drives the real `/backend-api/codex/responses` route and asserts same-account retry with no `response.failed` leaking downstream, which is the right shape for this class of fix.

Three things on the `service.py` side that I think need a decision before this lands.

**1. The diff adds `overloaded_error` to `_TRANSIENT_RETRY_CODES`, but nothing declares it.**

`tasks.md` says "Add `server_is_overloaded` to the bounded streaming transient retry set", `proposal.md`'s What Changes scopes the retry path to `server_is_overloaded`, and the spec delta only MODIFIES the *classification* requirement. The diff adds both codes, and the parametrized test asserts the new `overloaded_error` behavior:

```
_should_retry_transient_stream_error("overloaded_error", ...)
  main:      False
  this PR:   True
```

That is a behavior change to already-shipped code rather than coverage for the new code. Given the OpenSpec gate, this looks like it would be cleaner either declared (proposal entry plus a spec scenario covering the retry set) or dropped, keeping the PR to `server_is_overloaded` only.

**2. `_TRANSIENT_RETRY_CODES` also feeds account penalization, which the design does not mention.**

`design.md` describes the set as driving `_should_retry_transient_stream_error`, but it has a second consumer:

```python
# app/modules/proxy/_service/streaming/helpers.py:422
def _should_penalize_stream_error(code: str | None) -> bool:
    if code is None:
        return False
    return code in _facade()._ACCOUNT_RECOVERY_RETRY_CODES or code in _facade()._TRANSIENT_RETRY_CODES
```

So both codes flip penalization as well:

```
_should_penalize_stream_error("overloaded_error")       main: False   this PR: True
_should_penalize_stream_error("server_is_overloaded")   main: False   this PR: True
```

This one seems worth a deliberate call rather than a side effect: an upstream-wide overload is not really an account-health signal, so penalizing the selected account for it may be the wrong outcome even in the cases where the retry itself is right. If it is intended, the design's Risks section seems like the place to record it.

**3. `upstream_error` still splits the two sets.**

After this PR the classifier and the retry set still disagree:

```
_TRANSIENT_CODES - _TRANSIENT_RETRY_CODES == {"upstream_error"}
```

If the intent is for the two to converge, `upstream_error` needs the same decision. If they are deliberately distinct (classifier drives account failover, retry set drives bounded same-account replay), then that is the argument for dropping `overloaded_error` in point 1 rather than adding to the set.

None of this affects the core `server_is_overloaded` classification fix, which I believe is correct.

Komzpa:
@codex review

Komzpa:
@codex review

Issue #608, opened 2026-05-13
BUG: The 'gpt-5.3-codex-spark' model is not supported when using Codex with a ChatGPT account.


ehendrix23:
I just updated to 1.17.0 (from 1.15.0 I believe but not 100% sure). Now when I use get-5.3-codex-spark on my Plus account I get:
```
The 'gpt-5.3-codex-spark' model is not supported when using Codex with a ChatGPT account.
```

I have load-balancing between 2 accounts, when it is going to my Pro account it works, but to my Plus account it does not. This is new as before I didn't have this issue.

PR #1321, opened 2026-07-14
fix(proxy): retry stale per-account model rejections

luawl:
## Summary

- classify only the exact ChatGPT account/model rejection for the requested model as `account_model_unsupported`
- retry once on another model-advertising account before upstream acceptance across native WebSocket, the HTTP responses bridge, and raw HTTP/SSE
- keep the rejection request-local: exclude only that account, do not penalize global account health, and preserve the original upstream 400 when no replacement exists
- retain fail-closed behavior for uploaded-file pins and owner continuations unless a proxy-verified self-contained fresh body can move safely

## Why

Local request logs exposed a short model-registry race: an account remained stored as `pro` while its authoritative usage payload reported `free`, so the last-known per-account catalog briefly routed Sol to it. Upstream rejected those turns before `response.created`; the next catalog refresh corrected the route. A pre-acceptance, request-local retry absorbs that race without globally hiding the model or mutating identity-sensitive account metadata.

Related to the multi-account rejection reported in #608. This is recurrence hardening for a transient stale snapshot, not a change to global catalog visibility.

## Replay safety

- exact `invalid_request_error` message and quoted model must match the requested model
- no retry after a response id, nonterminal `response.*` event, downstream sequence/output, another pending request, or an earlier replay
- rejected account is added only to the current request's exclusion set
- replacement selection still applies the model/service-tier catalog filters
- account response-create and stream leases are released before reselection
- file-pinned and unverified owner-bound requests never cross accounts
- no compatible replacement surfaces the original 400, not `no_accounts`, `stream_incomplete`, or 502

## Review size

The PR is one proxy-routing concern. Its **957 net lines** break down as:

- **305 net production-code lines** (306 additions, 1 deletion)
- **539 regression-test lines**
- **113 OpenSpec proposal/design/spec/task lines**

## Validation

- `uv run pytest -q tests/unit/test_proxy_utils.py` - 700 passed
- `uv run pytest -q tests/integration/test_proxy_websocket_responses.py` - 63 passed
- `uv run pytest -q tests/integration/test_http_responses_bridge.py` - 90 passed
- focused new/adjacent regression selection - 27 passed
- `uv run ruff format --check ...` - passed
- `uv run ruff check ...` - passed
- `uv run ty check` - passed
- `uv run python scripts/check_proxy_architecture.py` - passed
- `npx -y @fission-ai/openspec@latest validate retry-stale-account-model-rejection --strict` - passed
- `npx -y @fission-ai/openspec@latest validate --specs --strict` - 43 passed

## OpenSpec

- [x] Behavior change documented under `openspec/changes/retry-stale-account-model-rejection/`
- [x] Strict change and all-spec validation passed

### DIFF
--- a/app/modules/proxy/_service/http_bridge/helpers.py
+++ b/app/modules/proxy/_service/http_bridge/helpers.py
@@ -445,16 +445,24 @@ def _log_http_bridge_startup_wait_timeout(
     )
 
 
-def _http_bridge_precreated_retry_failure_error(exc: BaseException) -> tuple[str, str]:
+def _http_bridge_precreated_retry_failure_error(exc: BaseException) -> tuple[int, str, str, str, str | None]:
     if isinstance(exc, ProxyResponseError):
         parsed = _parse_openai_error(exc.payload)
         code = _normalize_error_code(parsed.code if parsed else None, parsed.type if parsed else None)
         message = parsed.message if parsed and parsed.message else "HTTP bridge pre-created retry failed"
-        return code, message
+        error_type = parsed.type if parsed and parsed.type else "server_error"
+        error_param = parsed.param if parsed else None
+        return exc.status_code, code, message, error_type, error_param
     if isinstance(exc, TimeoutError):
-        return "upstream_unavailable", "HTTP bridge pre-created retry failed: upstream websocket reconnect timed out"
+        return (
+            502,
+            "upstream_unavailable",
+            "HTTP bridge pre-created retry failed: upstream websocket reconnect timed out",
+            "server_error",
+            None,
+        )
     message = str(exc).strip() or "HTTP bridge pre-created retry failed"
-    return "upstream_unavailable", message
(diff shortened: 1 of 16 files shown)


Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Soju06:
@codex review

Soju06:
Codex review at the current head is clean; the remaining blocker is the required **Lint (ruff)** job, which fails on the proxy architecture check:

> `http_bridge/mixin.py has 2403 lines; limit is 2400`

Your additions push the file over the 2,400-line ceiling. Please move the new retry logic into a separate module (or extract an existing cohesive chunk of the mixin) so the check passes.

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

PR #1327, opened 2026-07-15
fix(balancer): decorrelate round-robin tie-breaking across replicas

Soju06:
## Why

The `round_robin` account selector orders candidates by planner cost, then least-recently-selected time, then a stable `account_id` tie-break. In a multi-replica deployment every replica shares the same account set and computes the identical sort key, so on an *exact* tie of the primary keys (e.g. a cold start where every account is never-selected, or several accounts whose `last_selected_at` reset to `0.0`) every replica breaks the tie toward the same lexicographically-first `account_id`. The result is herding / thundering-herd: all replicas pick one account first instead of spreading equal load across equally-good accounts. This is the deferred multi-replica split from the propagate-balancer-health-signals design.

## What Changes

- Add a stable per-replica salt (defaulting to the HTTP responses-session bridge instance id, else the host identity) mixed into the **final** round-robin tie-break via a keyed hash.
- Primary ordering (planner cost, then least-recently-selected) is unchanged; only genuine ties are decorrelated so peers spread across equally-good accounts.
- The salt is process-stable (not random per call), so single-replica selection stays deterministic and unchanged in aggregate.
- Wire `configure_replica_salt(...)` once at proxy start-up from the bridge instance identity.
- No schema/migration change: the salt is in-memory routing state only.

OpenSpec change: `openspec/changes/replica-decorrelated-round-robin/` (account-routing delta, strict-validated).

## Testing

Commands run (all green, in the worktree):

- `uv run ruff check app tests` — All checks passed
- `uv run ruff format --check app tests` — 752 files already formatted
- `uv run ty check` — All checks passed
- `scripts/check_proxy_architecture.py` — proxy architecture checks passed
- `openspec validate replica-decorrelated-round-robin --strict` — valid
- `openspec validate --specs` — 43 passed, 0 failed
- `uv run pytest tests/unit/test_load_balancer.py tests/unit/test_load_balancer_concurrency.py` — 239 passed

New regression tests (in `tests/unit/test_load_balancer.py`): two replicas with distinct salts break an exact tie toward different accounts; many salts spread across the pool; primary ordering (least-recently-selected) and planner-cost precedence unaffected by salt; single-replica deterministic; salt precedence (explicit > configured > host default). These 3 decorrelation tests were verified to **fail on the pre-change behavior** — with the old `account_id`-only final tie-break every replica herds onto `acct-1`, so `winner_a != winner_b`, `len(winners) > 1`, and the precedence assertion all fail.

Deferred follow-up from the multi-replica reliability effort (post #1252–#1264).


Soju06:
@codex review

PR #1348, opened 2026-07-15
test(balancer): make sticky-selection test deterministic under decorrelated tie-breaking

### LINKED ISSUE #1342: flaky test: test_fresh_sticky_mapping_uses_normal_budget_gate (assert 'b' == 'a')
`tests/unit/test_select_with_stickiness.py::test_fresh_sticky_mapping_uses_normal_budget_gate` fails intermittently on CI with:

```
tests/unit/test_select_with_stickiness.py:979: AssertionError
E       AssertionError: assert 'b' == 'a'
```

Observed 3 times on 2026-07-15 alone, on unrelated docs/CI-only PRs (#1337, #1338/#1341 — e.g. runs 29393265584, 29394139208, 29397991473), and previously during the v1.21.0 release cycle. Passes on re-run every time, which suggests ordering/tie-break nondeterminism in the balancer selection under the fresh-sticky-mapping path rather than a real regression — possibly related to the decorrelated round-robin tie-breaking (#1327).

Worth either pinning the tie-break seed/order in the fixture or relaxing the assertion to the actual invariant (selected account satisfies the budget gate) instead of a specific account id.

Soju06:
## Summary

Fixes the flaky `test_fresh_sticky_mapping_uses_normal_budget_gate` (5 CI failures on 2026-07-15 alone, always green on re-run).

**Root cause:** with a fresh sticky mapping both candidate accounts tie on the round-robin primary keys (equal planner cost, `last_selected_at=None`), so selection falls to the #1327 decorrelated tie-break `sha256(replica_salt + account_id)` — and since tests never configure a salt, it defaults to `socket.gethostname()`. CI runners get random hostnames; for ~half of them the hash orders `'b'` before `'a'`, failing the assertion. Any given host is process-stable, which is why re-runs (fresh runner) pass and local loops never reproduced. Reproduced deterministically by pinning an Azure-runner-style salt (`fv-az123`).

**Fix (test-only):** give the competitor account a recent `last_selected_at` so the intended account wins on the round-robin *primary* key and selection never reaches the salted tie-break. This sharpens the test's original intent — the intended account is secondary-pressured, so if the secondary budget gate were wrongly applied to fresh mappings it would be excluded and the test would fail.

**Proof:** unpatched test fails under the pinned salt; patched test passes 100/100 separate pytest runs and 100/100 in-process runs under 100 random salts; related selection/balancer files all green (324 passed). Sibling audit found no other tests depending on tie-break randomness (details in the issue).

Fixes #1342

## Type of change

## OpenSpec

Test-only; no behavior/contract change, no OpenSpec folder required.

## Simplicity

- [x] No new settings, docs, or surfaces.

### DIFF
--- a/tests/unit/test_select_with_stickiness.py
+++ b/tests/unit/test_select_with_stickiness.py
@@ -960,8 +960,18 @@ async def test_sticky_thread_below_threshold_does_not_reallocate():
 
 @pytest.mark.asyncio
 async def test_fresh_sticky_mapping_uses_normal_budget_gate():
+    # "a" is secondary-pressured (99% > secondary threshold 98%) but must stay
+    # eligible for a FRESH mapping: the secondary budget gate applies only to
+    # sticky reallocation, not to normal selection. Mark "b" as recently
+    # selected so round_robin's least-recently-selected primary key
+    # deterministically prefers "a" whenever "a" is eligible — the selection
+    # must not fall through to the final tie-break, which is decorrelated per
+    # replica (keyed on the host identity) and therefore host-dependent.
+    # If the secondary gate were wrongly applied here, "a" would be excluded
+    # and "b" selected, failing the assertion below.
     acc_a = _active("a", used_percent=10.0, secondary_used_percent=99.0)
     acc_b = _active("b", used_percent=20.0, secondary_used_percent=10.0)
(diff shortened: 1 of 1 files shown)


Soju06:
@codex review

PR #1352, opened 2026-07-15
fix(db): pin asyncpg session time zone to UTC

Daltonganger:
## Problem

The application persists **naive UTC** datetimes into `timestamptz` columns (`app/core/utils/time.py::utcnow` returns `datetime.now(timezone.utc).replace(tzinfo=None)`).

asyncpg binds a naive datetime using the **connection's session time zone**. In production `_postgres_async_connect_args` does not set a session time zone, so it follows the container's `TZ`. When the container runs e.g. `TZ=Europe/Amsterdam`, PostgreSQL interprets those naive-UTC values as local time and shifts **every** written `timestamptz` ~2h away from real UTC.

The skew is silent but corrupts every wall-clock comparison the coordinator relies on:

- `bridge_ring_members` heartbeats are written ~2h in the past, so replicas look perpetually **stale** → leader election / the sticky- and bridge-session cleanup scheduler stop doing their work.
- account/stream **lease expiry** is mis-evaluated.

The practical failure mode: abandoned HTTP bridge sessions (`http_bridge_sessions`, `state=ACTIVE` with an expired lease from an aborted client) are never reclaimed and accumulate until the per-account stream cap is exhausted. New streams then block in `account_stream_cap` retry loops and **never return response headers**, so OpenCode/Codex clients fail with `response headers timed out`. A container restart temporarily clears it, which matches the "works for a while after redeploy" symptom.

### Reproduction

Run the container with `TZ=Europe/Amsterdam` (any non-UTC zone) against PostgreSQL, then:

```sql
select instance_id, now() - last_heartbeat_at as age from bridge_ring_members;
```

`age` sits at a constant ~2h (the UTC↔local offset) even though the heartbeat loop is running.

## Fix

Force the asyncpg session time zone to UTC via `server_settings={"timezone": "UTC"}` so stored timestamps stay correct regardless of the container time zone. Verified locally: with the setting, `SHOW timezone` is `UTC` and heartbeat age drops to ~0.

## Tests

Added unit tests to `tests/unit/test_db_session.py` covering the UTC pin (prod + `CODEX_LB_TEST_DATABASE_URL` paths, non-postgres passthrough, and that the connect args reach the engine kwargs). Full file passes (32 tests).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,7 @@ POSTGRES_PYTEST_TARGETS := \
 	tests/integration/test_sticky_sessions_api.py::test_durable_bridge_owned_alias_registration_is_epoch_fenced \
 	tests/integration/test_proxy_api_extended.py::test_proxy_stream_usage_limit_returns_http_error \
 	tests/integration/test_repositories.py::test_accounts_upsert_with_merge_disabled_uses_identity_lock_on_postgresql \
+	tests/integration/test_db_session_timezone.py \
 	tests/test_request_logs_options_api.py \
 	tests/integration/test_account_usage_rollup.py \
 	tests/integration/test_data_retention.py
--- a/app/db/session.py
+++ b/app/db/session.py
@@ -45,12 +45,23 @@ def _is_sqlite_memory_url(url: str) -> bool:
     return _is_sqlite_url(url) and ":memory:" in url
 
 
-def _postgres_async_connect_args(url: str) -> dict[str, int] | None:
+def _postgres_async_connect_args(url: str) -> dict[str, object] | None:
     if not url.startswith("postgresql+asyncpg://"):
         return None
-    if not os.environ.get("CODEX_LB_TEST_DATABASE_URL"):
-        return None
-    return {"prepared_statement_cache_size": 0}
+    # Pin the asyncpg session time zone to UTC. The application persists naive
+    # UTC datetimes (see app.core.utils.time.utcnow) into timestamptz columns.
+    # asyncpg binds a naive datetime using the connection's session time zone,
+    # so if that time zone follows the container's TZ (e.g. Europe/Amsterdam)
+    # PostgreSQL shifts every written timestamp away from real UTC. That skew is
+    # silent but corrupts every wall-clock comparison the coordinator relies on:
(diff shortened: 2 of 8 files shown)


Soju06:
Thanks — the tz-pinning approach looks sound. Two gate items before this can merge (see [PR readiness](https://github.com/Soju06/codex-lb/blob/main/CLAUDE.md)):

1. **OpenSpec**: this changes DB session semantics (a compatibility/behavior contract), so it needs an `openspec/changes/<slug>/` folder with the MUST-style requirement (asyncpg sessions pin to UTC) and rationale in `context.md`.
2. Please confirm the PostgreSQL CI job exercises the changed path (a regression test that would fail without the pin would be ideal).

Soju06:
@codex review

Soju06:
Thanks for the quick turnaround — the OpenSpec change folder and the Postgres integration regression test (forcing the DB default to Europe/Amsterdam and asserting SHOW TIME ZONE = UTC through the app's engine kwargs) are exactly what we asked for, and the core fix looks merge-worthy.

Two things are blocking now:

1. **Please drop the CI-scripts refactor from this PR.** The latest commit bundles a rework of `.github/` infrastructure (new `github_api.py`, `detect_changed_areas.py` replacing the pinned `dorny/paths-filter` action in `ci.yml`, `fetch_pr_labels.py` in `simplicity-budgets.yml`, the `sync_codex_ok_labels.py` rewrite, plus their tests). That's a separate concern from the timezone fix, it isn't mentioned in the PR body or covered by the OpenSpec change, and swapping a pinned third-party action for custom scripts needs its own review. Per our one-concern-per-PR rule, please split it into its own PR (with its own rationale) if you want to pursue it — this PR should only touch `app/db/session.py`, the openspec folder, the two test files, and the Makefile target line.

2. **Rebase onto current `main`** — the branch is showing merge conflicts (`mergeable=CONFLICTING`), likely from the `.github/` files above, which the split should mostly resolve.

Once that's pushed I'll approve the CI run and re-trigger `@codex review` on the new head. The tz pin itself is good to go after that.

Komzpa:
@codex review

PR #1366, opened 2026-07-16
fix(proxy): dedup and cap oversized response.create dumps

### LINKED ISSUE #1345: debug: response-create dump directory has no retention or dedup
During the 2026-07-15 payload_too_large incident, every retry of the same oversized `response.create` wrote a fresh ~15MB gzipped dump to `debug/response-create-dumps`, even though `request_text_sha256` was identical across retries. Combined with dumps accumulating since 2026-05-13, the directory reached **154 files / 1.1GB** on a 28G host (76% full) before manual cleanup.

Suggestions (either or both):
- skip dumping when a dump with the same `request_text_sha256` already exists (dedup)
- cap the directory by count/bytes/age (e.g. keep newest N pairs, prune on write)

Related fix for the incident itself: #1344

ozpool:
## Summary

`<data-dir>/debug/response-create-dumps` is an unbounded write path on the operator's data volume: identical payloads always produce a new dump pair, and nothing ever removes an old one. This deduplicates dumps by payload fingerprint and prunes the oldest pairs on write, so the directory stays bounded with no configuration.

## Type of change

Linked issue: Closes #1345

## OpenSpec

Change directory: `openspec/changes/dedup-and-cap-response-create-dumps/`

ADDED requirement on `deployment-installation` (the capability that already owns the dump directory as an operator/disk contract): dump-directory bounding is normative and needs no setting. This does not touch a codex-faithful path — the dumps are a local debug artifact and no upstream wire format is involved.

## Changes

- `_write_response_create_dump` appends a 16-hex prefix of the existing `request_text_sha256` to the dump id, so duplicate detection is a `glob` rather than a read + JSON-parse of every stored meta file on an already-degraded path.
- A dump whose fingerprint is already stored is skipped, keeping the first capture. The suppression is logged at `warning` with the fingerprint and the existing dump path, so the recurrence signal from the incident survives — only the bytes are suppressed, not the evidence.
- After a successful write, the oldest pairs beyond `_RESPONSE_CREATE_DUMP_MAX_PAIRS` (20) are pruned. Dump ids are timestamp-prefixed with microseconds, so lexicographic order is chronological and pruning needs no `stat` calls. Each pruned payload file takes its meta sibling with it.
- Both helpers swallow `OSError` and `unlink` uses `missing_ok=True`, so a concurrent prune between replicas cannot fail a request.

Deduplication removes the amplifier that produced 154 near-identical files; the count cap bounds the residual tail of distinct payloads. I left out byte and age caps deliberately — they would add a second budget and constant without addressing anything the count cap and dedup miss. Happy to add one if you'd rather bound bytes directly.

## Simplicity

- [x] New feature defaults to **off** or works with **zero config**
- [x] No new required setup step (or maintainer approval via `simplicity-budget-approved` label)
- New setting(s) and why each can't be a default: **none.** `_RESPONSE_CREATE_DUMP_MAX_PAIRS = 20` is a module constant reachable through the existing `_service_global_or` indirection, matching `_OVERSIZED_RESPONSE_CREATE_LARGEST_ITEMS`. It is an internals-only debug bound, so per P2 it stays out of `.env.example` and needs no `CODEX_LB_*` setting.
- [x] README sections / `.env.example` / dashboard nav within budget (or `simplicity-budget-approved` label requested)

## Test plan

Three unit tests in `tests/unit/test_proxy_utils.py`: duplicate suppression, distinct-payload retention (guards against over-eager dedup), and prune-on-write including the meta sibling.

```
$ .venv/bin/python -m pytest tests/unit/test_proxy_utils.py -k "dump" -q
4 passed, 733 deselected in 0.17s

$ .venv/bin/python -m pytest tests/unit/test_proxy_utils.py -q
737 passed in 21.59s
```

Fail-on-main check — reverting only the two app files to `upstream/main` (fix confirmed absent by grep) and re-running:

```
$ .venv/bin/python -m pytest tests/unit/test_proxy_utils.py -k "dump" -q
FAILED tests/unit/test_proxy_utils.py::test_response_create_dump_skips_duplicate_payload
FAILED tests/unit/test_proxy_utils.py::test_response_create_dump_prunes_oldest_beyond_cap
2 failed, 2 passed, 733 deselected
```

`test_response_create_dump_keeps_distinct_payloads` passes on both sides, as it should.

Other gates:

```
$ .venv/bin/python -m ruff check <changed files>          All checks passed!
$ .venv/bin/python -m ruff format --check <changed files>  3 files left unchanged
$ uv run ty check app/modules/proxy/_service/response_create.py app/modules/proxy/service.py
                                                           All checks passed!
$ .venv/bin/python scripts/check_proxy_architecture.py     proxy architecture checks passed
$ .venv/bin/python .github/scripts/check_simplicity_budgets.py
  README lines: 128/200 OK | headings: 9/10 OK | env example: 45/60 OK | core nav: 5/5 OK
$ npx -p @fission-ai/openspec openspec validate dedup-and-cap-response-create-dumps --strict
  Change 'dedup-and-cap-response-create-dumps' is valid
$ npx -p @fission-ai/openspec openspec validate --specs --strict
  Totals: 47 passed, 0 failed (47 items)
```

## Output

The fail-on-main run captures the reported bug directly — the same 52-byte payload written to two separate pairs:

```
WARNING Saved unit response.create dump request_id=req_dedup_first request_log_id=None
        dump_path=.../20260716T060855.043104Z-websocket-gpt-5.5-req_dedup_first.response-create.json.gz bytes=52
WARNING Saved unit response.create dump request_id=req_dedup_retry request_log_id=None
        dump_path=.../20260716T060855.043546Z-websocket-gpt-5.5-req_dedup_retry.response-create.json.gz bytes=52
```

With this patch the first capture is written with the fingerprint in its id, and the retry is suppressed against it:

```
WARNING Saved unit response.create dump request_id=req_dedup_first request_log_id=None
        dump_path=.../20260716T061537.461824Z-websocket-gpt-5.5-req_dedup_first-907828523c335b87.response-create.json.gz bytes=52
WARNING Skipped duplicate unit response.create dump request_id=req_dedup_retry request_log_id=None
        request_text_sha256=907828523c335b876ca0c6296ee2901fa0e7dc7540b8ab24b86a06d745b24aa2
        existing_dump_path=.../20260716T061537.461824Z-websocket-gpt-5.5-req_dedup_first-907828523c335b87.response-create.json.gz bytes=52
```

Not verified here: no long-running gateway was used to reproduce the 1.1GB fill and watch it stay bounded across a real incident. The bound is demonstrated at the directory level.

## Checklist

- [x] Ran the relevant `make` subset locally (lint, typecheck, architecture-check, simplicity budgets, unit suite — pasted above).
- [x] If touching specs: `openspec validate --specs` passes (47/47).

### DIFF
--- a/app/modules/proxy/_service/response_create.py
+++ b/app/modules/proxy/_service/response_create.py
@@ -99,11 +99,24 @@ def _response_create_dump_max_pairs() -> int:
 
 
 def _existing_response_create_dump(dump_dir: Path, sha_slug: str) -> Path | None:
-    """Return an existing dump for the same payload fingerprint, if any."""
+    """Return an existing *complete* dump for the same payload fingerprint.
+
+    A dump only counts as existing when both the payload and its ``.meta.json``
+    sibling are present. A lone payload file (e.g. a crash or disk-full failure
+    between the two writes) is treated as absent so a retry can recreate the
+    missing meta instead of permanently suppressing the capture operators are
+    trying to diagnose.
+    """
     try:
-        return next(iter(dump_dir.glob(f"*-{sha_slug}{_RESPONSE_CREATE_DUMP_SUFFIX}")), None)
+        matches = sorted(dump_dir.glob(f"*-{sha_slug}{_RESPONSE_CREATE_DUMP_SUFFIX}"))
     except OSError:
         return None
+    for dump_path in matches:
+        dump_id = dump_path.name[: -len(_RESPONSE_CREATE_DUMP_SUFFIX)]
+        meta_path = dump_dir / f"{dump_id}{_RESPONSE_CREATE_META_SUFFIX}"
+        if meta_path.exists():
+            return dump_path
+    return None
 
 
 def _prune_response_create_dumps(dump_dir: Path, *, max_pairs: int) -> None:
@@ -728,6 +741,13 @@ def _write_response_create_dump(
         else:
             meta["summary"] = {"payload_type": type(parsed_payload).__name__}
 
+    max_pairs = _response_create_dump_max_pairs()
+    # Trim any pre-existing over-cap backlog before the new write. If the volume
+    # is already full, the write below raises and returns without pruning, so a
+    # directory left above the bound (e.g. after an upgrade that lowers the cap)
+    # would otherwise stay full on the exact failure the cap is meant to bound.
+    _prune_response_create_dumps(dump_dir, max_pairs=max_pairs)
+
     try:
         dump_dir.mkdir(parents=True, exist_ok=True)
         with gzip.open(dump_path, "wt", encoding="utf-8") as handle:
@@ -745,7 +765,7 @@ def _write_response_create_dump(
         )
         return False
 
-    _prune_response_create_dumps(dump_dir, max_pairs=_response_create_dump_max_pairs())
+    _prune_response_create_dumps(dump_dir, max_pairs=max_pairs)
 
     logger.warning(
         "Saved %s response.create dump request_id=%s request_log_id=%s dump_path=%s meta_path=%s bytes=%s",
--- a/app/modules/proxy/_service/response_create.py
+++ b/app/modules/proxy/_service/response_create.py
@@ -44,6 +44,10 @@ _RESPONSE_CREATE_TOOL_OUTPUT_OMISSION_NOTICE = (
 )
 _RESPONSE_CREATE_IMAGE_OMISSION_NOTICE = "[codex-lb omitted historical inline image to fit upstream websocket budget]"
 _OVERSIZED_RESPONSE_CREATE_DUMP_DIR: Path | None = None
+_RESPONSE_CREATE_DUMP_SUFFIX = ".response-create.json.gz"
+_RESPONSE_CREATE_META_SUFFIX = ".meta.json"
+_RESPONSE_CREATE_DUMP_SHA_SLUG_LEN = 16
+_RESPONSE_CREATE_DUMP_MAX_PAIRS = 20
 _RESPONSE_CREATE_COMPATIBILITY_METADATA_HEADERS = (
     "x-codex-turn-metadata",
     "x-openai-subagent",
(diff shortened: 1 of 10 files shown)


Komzpa:
@codex review

Soju06:
Thanks — this is a well-scoped fix with exactly the right OpenSpec and simplicity treatment, and the fail-on-main test evidence in the body is appreciated. Before merge I'd like the three Codex P2s addressed; I agree with all of them:

1. **Prune before the write, not only after.** The upgrade-onto-a-full-volume case is the incident itself: with more than 20 existing pairs and no room for the next gzip, the write raises and returns before the prune ever runs, so the new bound never applies where it matters most. Prune the excess first, and keep the post-write prune for the new entry.

2. **Require a complete pair before suppressing a retry.** If the payload file lands but the meta write fails, the filename-only glob will suppress every retry of exactly the payload an operator is trying to diagnose, and the broken pair is never repaired. Checking that the meta sibling exists (or unlinking the orphan payload on write failure) closes this.

3. **One product-path regression test.** Our merge gates ask for coverage at the externally failing surface, not just the helper — a test that drives an oversized response.create through the ws/bridge path and asserts dedup + prune apply would satisfy that.

Everything else (CI, mergeability, OpenSpec, Closes #1345) is green — happy to merge once these land.

Komzpa:
@codex review

ozpool:
All three of these are in now.

The prune ordering and the partial-pair case both went in with `d9fb67d9`. `_existing_response_create_dump` now only counts a fingerprint as already captured when the `.meta.json` sibling is actually on disk, so a payload left behind by a failed meta write stops suppressing retries and gets repaired on the next one instead. The prune also runs before the gzip write as well as after it, so a directory that's already over the cap gets trimmed even when the write itself is what fails on a full volume — which as you said is the case that actually matters here.

Product-path coverage is `bbdc0040`, which @Komzpa wrote and pushed. It extends both the `/v1/responses` bridge test and the websocket one to send the same oversized payload twice and assert only one pair ends up on disk, then removes the meta and checks the retry rebuilds a complete pair rather than being silently swallowed.

One thing I should flag rather than let you find it: those two cover dedup and the orphan repair at the product surface, but the 20-pair cap itself is still only asserted in the unit tests. If you'd rather have all three of them at the integration level I'm happy to push the prune assertion down there too — just say the word and I'll add it.

PR #1374, opened 2026-07-16
fix(balancer): reject implausible rate-limit resets

leventov:
## Summary

Reject malformed or wrong-unit OpenAI service reset metadata before it can persist an account as rate limited for an implausible duration. Existing poisoned rate-limited rows now recover through fresh quota evidence while preserving the 30-second post-block floor and normal quota-exceeded semantics.

## Type of change

- [x] fix: bug fix (no behavior change beyond the bug)
- [ ] Breaking change

Linked issue: Resolves no filed issue. A repository issue search found no active report for this malformed-reset persistence bug.

## OpenSpec

- [x] This PR includes an OpenSpec change
- [x] This PR preserves valid OpenAI service reset metadata behavior and rejects only invalid metadata

Change directory: openspec/changes/bound-rate-limit-reset-metadata/

## Changes

- Bound absolute and relative reset metadata to a finite 366-day horizon, with a one-second tolerance only for persisted whole-second rounding.
- Fall back from invalid absolute metadata to valid relative metadata, then to the existing Retry-After or backoff path.
- Repair implausible persisted rate-limited rows only from recent available evidence in every applicable quota window, with post-block evidence and the minimum floor where applicable.
- Add focused unit, scheduler, and multi-replica regressions plus production comments documenting the subtle invariants.

## Simplicity

- [x] No new setting, required setup step, migration, README section, environment surface, dashboard navigation, or default change.
- [x] Simplicity gates P1-P5 reviewed; no budget exception is required.

## Test plan

    uv run pytest tests/unit/test_load_balancer.py tests/unit/test_usage_updater.py tests/unit/test_usage_refresh_scheduler_recovery.py tests/integration/test_load_balancer_multi_replica.py
    # 354 passed
    uvx ruff check .
    uvx ruff format --check .
    uv run ty check
    npx --yes @fission-ai/openspec@1.6.0 validate --specs --strict
    # 47 specs passed
    git diff --check origin/main...HEAD

The pre-commit Codex review loop addressed all findings and terminated cleanly with no discrete actionable bugs.

## Screenshots / output

No dashboard surface changes. For the reproduced failure, an absolute reset timestamp resolving to year 2446 is now rejected and replaced by the bounded fallback; a previously poisoned row can return to active after fresh quota evidence and the post-block floor.

## Checklist

- [x] Title is in Conventional Commits format.
- [x] No related filed issue exists after repository search.
- [x] Added tests covering the external scheduler and multi-replica recovery paths.
[...]

PR #1247, opened 2026-07-12
fix(proxy): report eligible pool usage exhaustion as 429

glopyglerky:
## Summary

- preserve structured account-selection failure codes through the balancer and proxy layers
- classify total usage exhaustion only after normal account eligibility filtering
- return Codex-compatible HTTP `429` with `error.type = usage_limit_reached` on pre-stream failure paths
- preserve internal account failover, continuity-owner errors, and local concurrency-cap semantics
- omit `resets_at` until CLB can distinguish authoritative upstream timestamps from synthesized reset times

Closes #1246.

## Behavior

- one exhausted account plus one usable account still fails over internally
- every eligible account usage-limited returns `429 usage_limit_reached`
- paused, deactivated, reauth-required, unsupported, assignment-filtered, or security-filtered accounts do not incorrectly turn an ordinary availability error into usage exhaustion
- local stream/response-create caps retain their existing stable error code
- continuity failures remain `502` when the wider eligible pool is not exhausted

## Testing

Run on an x86_64 AerynOS build VM:

- `uv run ruff check app tests`
- `uv run ty check`
- targeted selection, response-mapping, Codex-control, HTTP-bridge, and concurrency tests: passed
- full suite: `4765 passed, 45 skipped, 2 failed`

The two full-suite failures are existing account-recovery failures and reproduce unchanged on pristine upstream commit `02f1784c53c53e54d2d82ad6e6ce257d3b02df13`:

- `test_accounts_list_recovers_zero_capacity_rate_limited_status`
- `test_accounts_list_recovers_quota_exceeded_status_from_secondary_usage`

PR #1382, opened 2026-07-16
fix(proxy): spill bare session affinity under account caps

leventov:
## Summary

Fixes the account-cap usability failure where a bare Codex process-session mapping could keep a request pinned to a locally saturated account and return 429 even while another eligible account remained unused.

The fix treats bare process-session affinity as request-local soft locality under account caps. It does **not** move the stored mapping. Previous responses, files, conversations, explicit turn state, live/durable bridges, replay, and reattach remain hard owner-bound and fail closed.

This is the second of the two requested changes; #1376 is the first.

## Type of change

- [ ] **Breaking change**

Linked issue: none; this PR does not close a GitHub issue.

## OpenSpec

- [x] This PR touches Codex-faithful request/transport paths and preserves upstream-equivalent request and event shapes

Change directory: `openspec/changes/archive/2026-07-17-spill-bare-affinity-under-account-caps/`

Main capabilities:

- `openspec/specs/proxy-admission-control/`
- `openspec/specs/sticky-session-operations/`

## Motivation and architecture

The observed failure was not an upstream account-selection opportunity being exhausted: one account remained pristine while requests tied to a different process-session mapping returned 429 under local response-create/stream pressure.

An earlier prototype persistently rebound the session mapping. That design was rejected and preserved separately because admission settles at different points in compact, SSE, direct WebSocket, and HTTP-bridge flows. Correct rebind would require a distributed transaction across sticky rows, durable bridge rows, in-memory registries, replay state, and possibly a WebSocket shared by unrelated requests.

This PR instead makes spillover request-local:

- a self-contained, pre-visible request may use another eligible account when only its bare process-session owner is capped;
- the stored soft mapping is neither updated nor deleted, so locality returns when the owner recovers;
- a response/file/bridge produced by the alternate establishes ownership through its existing hard index;
- late admission races retain the existing bounded local-cap failure and never retire a shared socket or publish a replacement bridge.

The source distinction is rolling-upgrade safe. Bare session headers use a SHA-256-derived key in an internal LF-prefixed namespace that normalized HTTP headers cannot reproduce. Explicit turn state keeps the raw historical key. Legacy raw rows are always checked and win conservatively as possible hard ownership.

## Changes

- Classify bare process-session locality separately from explicit turn-state ownership without adding a database enum or migration.
- Let account selection filter a soft mapped owner through response-create/stream caps while preserving the mapping on spillover and no-alternate failure.
- Centralize hard-owner conflict resolution across previous responses, files, conversations, turn state, live/durable bridge aliases, and legacy raw rows.
- Preserve the full model/API-key/security ownership pool before transient quota, status, exclusion, health, budget, cap, preferred-owner, or single-account routing filters.
- Resolve live and durable turn-state aliases on compact, plain HTTP streaming, direct WebSocket, and HTTP-bridge paths.
- Revalidate nonblank `conversation` ownership on every reused direct-WebSocket frame.
- Prevent retired-socket turn-state tokens from crossing to another account on a movable reconnect.
- Carry origin-resolved file ownership across replica owner forwarding in the authenticated full-context signature; legacy forwards without this additive proof remain compatible.
- Add comments at provenance, owner-precedence, non-persistence, signature, compatibility, and transport-handoff boundaries.
- Sync and archive the OpenSpec change with stable capability context.

## Sequencing and dependencies

This PR is **independent and not stacked on #1376**; its base is current `main`, and it has no Git dependency on the first PR.

Recommended merge order:

1. #1374
2. #1376
3. this PR
4. rebase #1247

The order is operational/review sequencing rather than a hard dependency. #1247 mechanically overlaps several proxy hot-path files touched here, so rebasing it after the continuity invariants land is safer than resolving those overlaps in parallel.

## Review history

The uncommitted diff went through the maximum three-iteration adversarial review loop. Each iteration used a fresh blind native reviewer with `gpt-5.6-sol`, `xhigh` reasoning, and no inherited context:

1. source precedence, previous-response fallback, raw/namespaced coexistence, hard-row mutation, conversation pool ordering, and file-owner product coverage;
2. derived-key aliasing, legacy/preferred conflicts, status-filtered ambiguity, missing/cross-account file pins, spec/error consistency, and stale comments;
3. cross-transport turn-state ownership, live/durable alias conflicts, reused-socket validation, single-account ownership scope, stale WebSocket tokens, and signed remote file-owner forwarding.

All findings were addressed with focused regressions. No fourth reviewer was started because the review-loop limit is three iterations.

## Scope note

This is a large diff, but it is one proxy-continuity capability rather than unrelated concerns. Splitting the shared selector from compact/stream/WebSocket/bridge propagation would create mergeable intermediate states that weaken ownership on one transport—the exact class of drift found during review. Most of the size is fail-closed regression coverage and the archived OpenSpec record.

## Simplicity

- [x] Works with **zero config** and is the safe default
- New settings: none
- Database/schema migrations: none
- [x] README sections / `.env.example` / dashboard nav unchanged and within budget

Why no setting: bare process-session affinity is only a locality hint when no owner-bearing input exists. Request-local spillover mutates no ownership state, so conservative hard continuity remains the invariant without adding operator surface.

## Test plan

```
15 focused final-review regressions passed

uv run pytest -q \
  tests/unit/test_proxy_utils.py \
  tests/unit/test_load_balancer_concurrency.py \
  tests/unit/test_proxy_load_balancer_refresh.py \
  tests/unit/test_proxy_http_bridge.py \
  tests/unit/test_http_bridge_forwarding.py \
  tests/unit/test_durable_bridge_sessions.py
# 1234 passed, 3 skipped

uv run pytest -q \
  tests/integration/test_proxy_files.py \
  tests/integration/test_proxy_sticky_sessions.py \
  tests/integration/test_proxy_compact.py
# 233 passed

uv run ruff check .
uv run ruff format --check .
# 791 files formatted; all checks passed

uv run ty check
uv run python scripts/check_proxy_architecture.py
uv run python .github/scripts/check_simplicity_budgets.py
npx --yes @fission-ai/openspec@1.6.0 validate --all --strict --no-interactive
# typing, architecture, simplicity, and 63/63 active OpenSpec items/specs passed
```

Known non-failing test output: FastAPI's TestClient deprecation warning, three existing load-balancer-refresh skips, and occasional aiosqlite worker shutdown warnings after the integration event loop closes.

## Screenshots / output

No dashboard-visible changes.

Behavioral example:

- Before: session S maps to capped account A; request returns local 429 while eligible account B remains unused.
- After: a self-contained request may run on B; S still maps to A; any response/file/bridge created on B becomes hard owner evidence for follow-up work.

## Checklist

- [x] No issue is closed; motivation is documented above.
- [x] Added unit and integration regressions at externally failing paths.
- [x] Ran the relevant local CI subsets and static gates.
- [x] OpenSpec strict validation and verification are clean; the change is archived.
- [x] Simplicity gates P1-P5 reviewed.
- [x] `CHANGELOG.md` was not edited.


Komzpa:
@codex review

Komzpa:
@codex review

Soju06:
Thanks for the follow-up to #1376 — request-local spillover with no mapping mutation and hard owners staying fail-closed is the right shape, and the fail-closed regression coverage plus the archived OpenSpec change are appreciated. Three things block this from being merge-ready right now:

1. **e2e regression on head**: `tests/e2e/test_proxy_flow.py::test_proxy_chat_completions_flow` fails on `c5a6f82` — `usageSummary` comes back `None` after an API-keyed chat completion (TypeError at line 59). This test is green on current main, so it looks like the selection-pool/reservation rework drops API-key usage attribution on the plain chat-completions path. Please trace where the reservation settles rather than re-running — this is exactly the reservation-settlement invariant we gate on.

2. **Unresolved Codex P-threads (3)**: the P1 about the owner-switch reconnect stripping the client-supplied `x-codex-turn-state` from `filtered_headers`, and the two P2s (soft preferred account bypassing `single_account` routing; live/durable bridge aliases being deduped by account id instead of failing closed on distinct sessions). Fix each or reply on the thread with why it doesn't apply, and resolve them.

3. **Stale Codex verdict**: the latest review covered `786b8bd6`, not the current head. Once 1–2 are addressed, re-trigger `@codex review` so we have a clean-on-head verdict.

One more note: at roughly +3,300 net lines this is well over the usual ~800-line PR ceiling. I've read the scope note and the single-capability argument is reasonable, but keeping it unsplit is a maintainer call — don't be surprised if we ask to carve out the shared selector vs. transport propagation if review stalls.

Komzpa:
@codex review

PR #1391, opened 2026-07-17
fix(usage): make weekly-primary remap tiebreak data-aware

zepx:
## What

Fixes the dashboard weekly (7d) usage chart repeatedly jumping to 100% remaining.

## Why

The weekly window is returned by upstream `/wham/usage` in the `primary_window` slot (`limit_window_seconds == 604800`) while `secondary_window` carries an empty no-data placeholder (`used_percent=0`, no window duration, no `reset_at`). The usage updater persists both rows ~10 ms apart in the same refresh cycle.

codex-lb remaps a weekly `primary` row into the `secondary` slot via `should_use_weekly_primary` → `_should_prefer_primary_row`. That tiebreak decided the winner by which row's `recorded_at` was microseconds newer, so the no-data placeholder (written ~10 ms later) won on most cycles and the dashboard read it as "0% used = 100% remaining". On a live account this caused the weekly remaining to jump to 100% at least 11 times over 6 hours while the real weekly usage climbed from ~34% to ~76% used.

Reproduced against the live DB with the app's own `_effective_usage_windows`: the latest primary row (`used_percent=74.0`, `window_minutes=10080`, real `reset_at`) and the latest secondary placeholder (`used_percent=0.0`, `window_minutes=0`, `reset_at=None`) → `should_use_weekly_primary` returned `False`, so the dashboard reported 100% instead of the true 26%.

## How

- **Data-aware tiebreak (D1):** in `_should_prefer_primary_row`, a row carrying real quota metadata (positive `window_minutes` AND non-null `reset_at`) now always wins over a no-data placeholder, regardless of `recorded_at` ordering — a placeholder is the absence of a measurement, not 0% used.
- **Same-fetch margin (D2):** rows whose `recorded_at` differs by less than `SIBLING_FETCH_MARGIN_SECONDS` (5.0 s, already used by the updater for sibling-row detection) are treated as same-fetch and resolved by reset-at precedence instead of a sub-second `recorded_at` coin flip. Only a genuinely different fetch (>5 s) decides on `recorded_at`.
- **Single source of truth (D3):** the canonical `SIBLING_FETCH_MARGIN_SECONDS` constant moves to `app/core/usage/__init__.py`; `app/modules/usage/updater.py` now imports `usage_core.SIBLING_FETCH_MARGIN_SECONDS` instead of keeping its own copy.
- **Shared path:** all four consumers (account-summary `_effective_usage_windows`, trend `_effective_usage_trend_buckets`, dashboard `normalize_weekly_only_rows`, dashboard projection `_should_use_weekly_primary_history`) already call `should_use_weekly_primary`, so the fix applies everywhere from one change.

## Verification

- `pytest tests/unit/test_usage.py tests/unit/test_account_mappers.py tests/unit/test_usage_mappers.py tests/unit/test_usage_updater.py tests/unit/test_rate_limit_reset_credits_mapper.py tests/integration/test_dashboard_overview.py` → 144 passed
- `ruff check` on touched files → clean (E/F/I, line-length 120)
- `openspec validate fix-weekly-primary-placeholder-race --strict` → valid; `openspec validate --specs` → 47/47 passed
- Live-data repro now returns the real weekly usage (24% remaining at 76% used) instead of 100%

## OpenSpec

- `openspec/changes/fix-weekly-primary-placeholder-race/` (spec-driven): proposal, context, design, tasks, and an `ADDED` requirement in `usage-refresh-policy` with 4 scenarios. All 10 tasks complete.

## Known follow-up (from gpt-5.6-sol review)

This PR was reviewed with `gpt-5.6-sol` (medium thinking), which raised a valid Major finding: the data-aware precedence is currently unconditional, so a stale real weekly primary from an older fetch can permanently beat a fresh no-data placeholder, freezing stale weekly usage when upstream later drops the window. The metadata precedence should apply only within the same-fetch margin (or when timestamps are absent); for a genuinely later fetch, the newer fetch should win and a newer placeholder should be represented as unavailable/omitted rather than 0% used. I plan to address this + the nits (spec/impl boundary wording, constant name, elapsed-row state) + the missing cross-fetch tests in a follow-up to keep this PR one-concern-wide.

## Simplicity

- No new `CODEX_LB_*` setting, no README/\`.env.example\`/dashboard-nav growth, no migration, no new default-on behavior. Internals-only change to an existing shared tiebreak.


Komzpa:
@codex review

Issue #1167, opened 2026-07-10
bug: durable session state diverges from disconnected clients (Request blocked; gate starvation; dangling custom tool call 400s)


kvz:
## Environment

- codex-lb built from `main` @ `bdd571f0` (plus the #1159 patch)
- codex-cli 0.144.0, model `gpt-5.6-sol` (`use_responses_lite=true`, `tool_mode=code_mode_only`)
- 5 concurrent codex TUI sessions, multiple accounts, `capacity_weighted` routing with `sticky_kind=codex_session`
- Default proxy limits (`proxy_account_response_create_limit=4`, `proxy_account_stream_limit=8`, `proxy_account_lease_ttl_seconds=900`, `proxy_admission_wait_timeout_seconds=10`)

## Summary

codex-lb keeps per-session server-side state that outlives the client: sticky account assignment, in-flight upstream turns, and injected session anchors (`websocket_session_anchor_injected` … `trimmed_to=2`). When a client disconnects mid-turn or the container restarts under live sessions, that state diverges from the client's transcript. One evening of session churn produced three distinct, reproducible failure modes. Related: #881 (gate acquire-timeouts), #857 / #735 (bridge pendings) cover adjacent symptoms but not these mechanisms.

## Mode A — cross-account encrypted reasoning → upstream "Request blocked."

Codex persists reasoning items with `encrypted_content` (requests include `reasoning.encrypted_content`). Those blobs are only decryptable by the account that minted them. When a session's turns later route to a *different* account (sticky state reset after a container replace, or reallocation), upstream hard-rejects every request that replays those items:

```
■ Request blocked.
```

Turn dies in ~2.5s, zero tokens billed, nothing useful logged. `continuity_owner_resolution … outcome=hit` shows the LB knows the owning session, but selection had moved to a different account.

Evidence it is the reasoning items and nothing else:

- retry / TUI resume / `codex exec resume` on the session → blocked
- full transcript cloned under a fresh session id → still blocked
- same clone with all `reasoning` response items stripped → works immediately

Suggested direction: when a continuation cannot be routed to the account that minted the chain, drop the (undecryptable-anyway) reasoning items from resends instead of forwarding them to certain rejection — or refuse reallocation for sessions that carry encrypted reasoning.

## Mode B — orphaned turn starves `response_create_gate`; ws pendings are never stuck-retired

[...]

PR #1408, opened 2026-07-20
fix(proxy): recover unavailable previous-response owners via durable fresh replay

softkleenex:
## Summary

A `previous_response_id` continuation is hard-owner-bound: only the account
that produced the original response can safely resolve it. When that owner
becomes unavailable later (quota exhausted, paused, deactivated), the proxy
already has an escape hatch — if it can prove the client's resend is an exact
verified replay of the input the owner actually completed, it strips the
anchor and continues fresh on any healthy account.

That proof lived only in a process-local, LRU-bounded, in-memory index. It's
empty after a restart, an eviction, or on a different replica than the one
that completed the response, so a perfectly resumable conversation instead
failed closed with `previous_response_owner_unavailable` ("Hard affinity
owner account is unavailable") — even with `stickyThreadsEnabled` off, since
that's an unrelated, separate mechanism from this hard-ownership check.

Separately, the HTTP responses bridge (`http_responses_session_bridge_enabled`,
on by default) had **no** recovery at all for this case when the client
itself supplied `previous_response_id`. Its only existing fresh-replay
recovery is scoped to a different, proxy-injected reattach scenario, so a
client-anchored continuation whose bridge owner becomes unavailable always
failed closed, independent of the in-memory cache.

- Persist the input-item-count/fingerprint pair (already computed for
  in-memory continuity) onto `RequestLog`, so it survives restarts and is
  visible to every replica.
- `_resolve_websocket_previous_response_owner` (shared by the direct retry
  path, the HTTP bridge, and native WebSocket) warms the in-memory continuity
  cache from that durable row — without ever overwriting a live, more-current
  entry.
- The direct retry path (`_stream_with_retry`) re-checks the
  verified-fresh-replay condition after owner resolution, so a cache warmed
  from durable storage unlocks the existing (previously cache-only-gated)
  recovery branches later in the same request.
- The HTTP bridge (`_stream_via_http_bridge`) gets the missing recovery
  branch for a client-supplied `previous_response_id`, disjoint from its
  existing proxy-injected-reattach branch.
- No existing fail-closed guarantee is weakened: absent a verified exact
  fingerprint match (or when the input isn't a self-contained fresh replay —
  file references, dangling tool calls), the request still fails closed
  exactly as before.

Full proposal, design rationale, and spec delta:
`openspec/changes/recover-unavailable-continuation-owner/`.

## Test plan

- [x] New regression: durable fallback recovers a cold-continuity-cache
      direct-path request (`tests/unit/test_proxy_utils.py::test_stream_durable_fingerprint_recovers_after_cold_continuity_cache`)
- [x] New regression: durable fallback recovers a client-anchored HTTP-bridge
      request whose owner is unavailable
      (`tests/unit/test_proxy_http_bridge.py::test_stream_via_http_bridge_recovers_client_previous_response_id_via_verified_fresh_replay`)
- [x] Verified both new tests fail without the fix (reverted the relevant
      file locally and confirmed the failure)
- [x] Updated pre-existing tests broken by the new `request_logs` columns
      (repository unit tests, websocket observability dict-equality tests)
- [x] `uv run pytest` — full suite green except one pre-existing, unrelated
      flaky integration test (`test_quota_planner_warm_now_keeps_bootstrap_for_metadata_less_primary_rows`,
- [x] Migration verified against `test_schema_migration_contract_matches_after_upgrade`


softkleenex:
Heads up on the failing \`Lint (ruff)\` / \`CI Required\` checks: this is a pre-existing architecture-check failure, unrelated to this PR's diff — \`service.py\` is already 4 lines over the \`check_proxy_architecture.py\` line-count gate on current \`main\` itself (confirmed by checking out unmodified \`main\` directly). Opened #1416 as a small, standalone fix for it.

softkleenex:
Closing this — #1437 (merged) already covers this exact case (issue #1167, Mode A: continuity-owner loss with a durable-fingerprint-verified fresh replay), and does it more thoroughly than this PR: typed selection-time owner classification instead of a post-hoc string match, plus stripping encrypted reasoning/server item IDs from the resend body before submitting to the replacement account, which this PR did not handle.

Given the overlap (this PR is now in actual merge conflict with #1437's changes to \`http_bridge/streaming.py\`), rebasing and reconciling would just reintroduce a less complete duplicate of what's already on \`main\`. Thanks for the review pointer toward #1437 — closing in favor of it rather than asking for another look at a superseded approach.

PR #1416, opened 2026-07-20
refactor(proxy): trim service.py under the architecture line-count gate

softkleenex:
## Summary

`service.py` currently sits at 2604 lines, 4 over the
`check_proxy_architecture.py` line-count gate (2600). Since that check
runs as part of `Lint (ruff)` in CI, this fails **every** PR that touches
this repo right now, regardless of what the PR actually changes (confirmed
on current `origin/main` directly, and on two unrelated open PRs of mine —
#1408, #1401 — both fail this same check with no diff to `service.py` at
all).

This is a small, purely mechanical fix: move four self-contained,
dependency-light helper functions out of `service.py` into
`_service/support.py` (a module `service.py` is already allowed to import
from per the architecture rules) and re-export them so every existing
`_facade()`-based caller elsewhere in the codebase is unaffected:

- `_is_account_neutral_error_code`
- `_is_local_account_cap_code`
- `_http_error_status_from_payload`
- `_openai_error_envelope_from_response_failed_payload`

No behavior change. `service.py` is now 2542 lines.

Note: `load_balancer.py` is *also* over its own architecture line-count
limit on current `main` (3260 vs. 3021) — a much larger gap that needs a
real method-level refactor, not a small move. Left out of this PR
intentionally to keep it single-concern; happy to open a follow-up if
useful.

## Test plan

- [x] `uv run python scripts/check_proxy_architecture.py` — service.py
      check now passes (the only remaining failure is the pre-existing,
      out-of-scope `load_balancer.py` line count, also present on
      unmodified `main`)
- [x] `uv run pytest` — full suite green except one pre-existing,
      unrelated flaky integration test
      (`test_quota_planner_warm_now_keeps_bootstrap_for_metadata_less_primary_rows`,

### DIFF
--- a/app/modules/proxy/_service/support.py
+++ b/app/modules/proxy/_service/support.py
@@ -18,10 +18,12 @@ from app.core.auth.refresh import RefreshError, is_transient_refresh_contention,
 from app.core.balancer.types import UpstreamError
 from app.core.clients.proxy import ProxyResponseError
 from app.core.clients.proxy_websocket import UpstreamResponsesWebSocket
-from app.core.errors import openai_error
+from app.core.errors import OpenAIErrorEnvelope, openai_error
 from app.core.openai.model_registry import get_model_registry
 from app.core.openai.models import OpenAIEvent
 from app.core.plan_types import account_plan_matches_allowed
+from app.core.resilience.network_recovery import PROCESS_NETWORK_UNAVAILABLE_CODE
+from app.core.resilience.overload import is_local_overload_error_code
 from app.core.types import JsonValue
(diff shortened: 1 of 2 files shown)


Soju06:
Thanks for this — it correctly diagnosed the ratchet breakage and the extraction was sound. #1430 (merged as 1c7ee48) landed with your complete `service.py` → `_service/support.py` extraction incorporated verbatim (all four helpers are on `main`), plus the `load_balancer.py` ratchet restoration and checker aggregation. Closing as superseded; the work here is preserved in the merged change. Appreciate the contribution.

PR #1438, opened 2026-07-21
fix(proxy): ignore API-key-enforced service tier a model never advertises

### LINKED ISSUE #1409: bug: enforced Priority service tier makes gpt-5.4-mini and codex-auto-review unavailable
### Pre-flight checklist

### codex-lb version

1.21.0

### Deployment method

Docker Compose

Codex app (desktop / web)

Plus, Pro

### Model(s) involved

gpt-5.4-mini, codex-auto-review

### What happened?

On codex-lb 1.21.0, setting `Enforced service tier` to `Priority` on an API key makes `gpt-5.4-mini` and `codex-auto-review` unavailable in the Codex app.

Requests for these models are rejected with:

`No accounts with a plan supporting model 'gpt-5.4-mini'`

The request log/error code is `no_plan_support_for_model`. The equivalent failure occurs for `codex-auto-review`. No request is sent to the upstream model.

Using the same API key and the same Plus/Pro accounts, models that advertise Priority/Fast support, such as `gpt-5.5` and `gpt-5.6-sol`, remain available.

Clearing the enforced service tier makes both affected models available again.

`gpt-5.4-mini` and `codex-auto-review` should remain routable through accounts that support these models at the default service tier.

When Priority is enforced globally but a model does not advertise Priority support, codex-lb should omit or fall back from the Priority tier for that model instead of excluding every account and reporting that no account plan supports the model.

### Steps to reproduce

```markdown
1. Deploy codex-lb 1.21.0 using Docker Compose.
2. Add one or more ChatGPT Plus or Pro accounts.
3. Create or update an API key and set `Enforced service tier` to `Priority`.
4. Configure the Codex app to use codex-lb with this API key.
5. Select `gpt-5.4-mini` and send any prompt.
6. Separately, trigger an automatic review that uses `codex-auto-review`.
7. Observe that both requests fail with `no_plan_support_for_model`.
8. Using the same API key and accounts, send a request with `gpt-5.5` or `gpt-5.6-sol`. These models work.
9. Clear the enforced service tier and retry `gpt-5.4-mini` or `codex-auto-review`. Both models become available again.
```

ozpool:
Closes #1409

## The bug

An API key's `enforced_service_tier` is an operator-wide default, but account selection treats it exactly like a tier the client asked for.

When the enforced tier is `priority` and a model never advertises `priority`, `account_ids_for_model_service_tier` answers authoritatively with an **empty** set (`model_registry.py`, the `tier_accounts is None -> frozenset()` branch). Empty is a real answer, not "unknown", so `_filter_accounts_for_model_with_catalog_evidence` filters every account out and selection fails with `no_plan_support_for_model`. The accounts do support the model; they support it at its default tier.

That matches the report exactly: `gpt-5.4-mini` and `codex-auto-review` went unavailable while `gpt-5.5` and `gpt-5.6-sol` kept working, and clearing the enforced tier fixed it.

The error was also pointing the wrong way. `No accounts with a plan supporting model '<model>'` sends an operator after a plan problem that doesn't exist.

## The fix

Distinguish an enforced tier from a requested one, and only fall back for the enforced case.

- `_service_tier_is_api_key_enforced` derives the signal from the `ApiKeyData` that `_select_account_with_budget` already receives, so nothing new crosses the proxy boundary.
- `ModelRegistry.model_advertises_service_tier` answers whether the catalog lists the tier for the model *at all*, which is deliberately different from "no account carries it". It returns `True` whenever the answer is unknown (no snapshot, non-authoritative catalogs), so an incomplete catalog never triggers the fallback.
- The tier is dropped only when it is enforced **and** the model doesn't advertise it.
- The selection-inputs cache key gains the flag, since the same tier string now resolves different pools depending on where it came from.

## What I deliberately did not change

An explicitly requested tier still filters accounts, and is still rejected when the model doesn't advertise it. `test_select_account_rejects_quota_override_for_unadvertised_service_tier` from #1248 stays green untouched, and there's a new assertion covering the same distinction at this seam.

That test is the reason this isn't the simpler "fall back for any unadvertised tier" patch. A caller who names a tier should be told it's unavailable rather than quietly served a different one; an operator-wide default shouldn't take unrelated models offline. Those are different situations and the code couldn't tell them apart, because `apply_api_key_enforcement` writes the enforced value into `payload.service_tier` before selection ever sees it.

Worth noting the existing requirement is already written in terms of an *explicit* tier ("route a model or **explicit** non-default service tier", and the Fast scenario is scoped to "**WHEN** a request explicitly asks for priority"). The enforced case fell under that rule by accident rather than by decision, so the spec delta is a MODIFIED requirement making the distinction explicit.

## Verification

Three regression tests in `tests/unit/test_proxy_load_balancer_refresh.py`, all **proven failing before the fix**. Reverting only the two source files and re-running reproduces the reported symptom verbatim:

```
E       assert 'no_plan_support_for_model' is None
E        +  where 'no_plan_support_for_model' = AccountSelection(account=None,
           error_message="No accounts with a plan supporting model 'gpt-5.4-mini'", ...)
```

- enforced `priority` on a model with no advertised tiers now routes, and an explicit `priority` request for the same model still doesn't
- the same configuration driven from an `ApiKeyData` with `enforced_service_tier="priority"` through `apply_api_key_enforcement`, so the regression starts at the operator-facing surface rather than a hand-passed tier string
- advertised-but-unheld tier still fails, with the tier named in the message

Full unit suite 4233 passed / 0 failed. `ruff check`, `ruff format --check`, whole-project `ty check`, `check_proxy_architecture.py` and `check_simplicity_budgets.py` all clean. OpenSpec `enforced-service-tier-model-fallback` validates strict, and `validate --specs --strict` is 47/47.

No new settings, no new error code. I did consider a distinct code for the tier-excluded case but that widens the external error envelope for something the corrected message already explains — happy to add one if you'd rather have it machine-readable.

### DIFF
--- a/app/core/openai/model_registry.py
+++ b/app/core/openai/model_registry.py
@@ -615,6 +615,35 @@ class ModelRegistry:
             return None
         return tier_accounts.get(normalized_service_tier, frozenset())
 
+    def model_advertises_service_tier(self, slug: str, service_tier: str | None) -> bool:
+        """Report whether the catalog lists ``service_tier`` for ``slug`` at all.
+
+        This is deliberately distinct from "no account carries the tier": a model
+        that never advertises the tier cannot be routed at it by any account, so
+        an enforced tier must not be treated as an account-eligibility filter for
+        that model. Returns ``True`` whenever the answer is unknown (no snapshot,
+        non-authoritative catalogs, unusable slug/tier) so callers keep their
+        existing behavior until the catalog can actually answer.
+        """
+        if service_tier is None or self._snapshot is None:
+            return True
+        if not self._snapshot.account_catalogs_authoritative:
+            return True
+        normalized_slug = slug.strip().lower()
+        normalized_service_tier = canonical_service_tier_value(service_tier)
+        if not normalized_slug or not normalized_service_tier:
+            return True
+
+        tier_accounts = self._snapshot.model_service_tier_accounts.get(
+            slug
+        ) or self._snapshot.model_service_tier_accounts.get(normalized_slug)
+        tier_plans = self._snapshot.model_service_tier_plans.get(slug) or self._snapshot.model_service_tier_plans.get(
+            normalized_slug
+        )
+        if tier_accounts is None and tier_plans is None:
+            return False
+        return normalized_service_tier in (tier_accounts or {}) or normalized_service_tier in (tier_plans or {})
+
     def account_ids_for_model(self, slug: str) -> frozenset[str] | None:
         """Return exact supporting accounts, or ``None`` while coverage is incomplete."""
         if self._snapshot is None or not self._snapshot.account_catalogs_authoritative:
--- a/app/core/openai/model_registry.py
+++ b/app/core/openai/model_registry.py
@@ -634,6 +634,19 @@ class ModelRegistry:
         if not normalized_slug or not normalized_service_tier:
             return True
 
+        model_is_known = (
+            slug in self._snapshot.model_plans
+            or normalized_slug in self._snapshot.model_plans
+            or slug in self._snapshot.model_accounts
+            or normalized_slug in self._snapshot.model_accounts
+        )
+        if not model_is_known:
+            # An authoritative subscription snapshot cannot describe the tier
+            # capabilities of an operator-mapped or source-routed model that it
+            # does not contain. Keep the enforced tier when model identity is
+            # unknown instead of broadening fallback beyond catalog evidence.
+            return True
+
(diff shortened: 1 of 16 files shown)


Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
Addressed the three current-head findings in 87cf3f21f: omit-equivalent tier provenance, origin-effective tier preservation across signed owner forwarding, and the missing responses-api-compat delta. Local verification: 188 passed, 3 documented skips; Ruff, ty, strict change validation, and all 47 current specs passed. @codex review

PR #1418, opened 2026-07-20
fix(compact): preserve historical side-effect anchors

Komzpa:
## Summary

Keep complete historical side-effect call/output pairs during Responses compaction without making the entire tool tail permanently untrimmable.

The compact selector now uses the canonical downstream side-effect classifier, reserves budget for complete side-effect pairs before optional ordinary context, and still fails closed when required state alone exceeds the compact budget.

Current head: `1a7dab66abce86862fd421b3cf7b0c31706b953b`.

## Type of change

- [x] `fix:` — bug fix
- [ ] `feat:` — new user-facing feature or capability
- [ ] `refactor:` — internal refactor
- [ ] `docs:` — documentation only
- [x] `chore:` / `ci:` / `build:` — architecture-ratchet maintenance
- [ ] `test:` — test-only change
- [ ] **Breaking change**

Linked issue: none.

## OpenSpec

- [x] This PR includes / updates an OpenSpec change
- [ ] Not applicable — bug fix that matches the existing spec
- [ ] Not applicable — docs / CI / chore only
- [x] This PR touches a codex-faithful request-shape path and preserves upstream-equivalent behavior

Change directory: `openspec/changes/harden-compact-tool-tail-anchors`.

## Root cause

Historical side-effect pairs were treated as permanently required compact state. Long conversations could therefore retain every completed shell/edit operation until the request exceeded the compact input limit. The earlier helper also maintained a second tool classifier that missed code-mode names such as `exec` and `collaboration`.

## Changes

- Use the canonical downstream side-effect classifier shared with tool-call safety and deduplication.
- Keep developer/tool prelude, goal/plan context, and state-tool pairs as required anchors.
- Reserve space for complete historical side-effect call/output pairs before optional ordinary context.
- Prefer the newest historical pairs while allowing older pairs to be trimmed.
- Cover the real `/backend-api/codex/responses/compact` product path.
- Specify the retention and fail-closed contract in OpenSpec.
- Refresh the proxy architecture ratchets to the current `main` baseline without adding new growth.

## Simplicity

- [x] Works with zero configuration.
- [x] Adds no setting or setup step.
- New settings: none.
- [x] Adds no README, `.env.example`, or dashboard navigation surface.

## Test plan

```bash
uv run pytest tests/unit/test_openai_requests.py tests/unit/test_tool_call_dedupe.py -q
uv run pytest tests/integration/test_proxy_compact.py -q
[...]

PR #1457, opened 2026-07-22
fix(compact): elide observed inline images at wire cap

Komzpa:
## Summary

Prevent eligible image-heavy Codex tool results from poisoning terminal
compaction. When an oversized required function, custom, or apply-patch tool
output contains an already-observed inline data-URL image, compact preparation
keeps the call/output identity and surrounding text while replacing only those
image bytes with an explicit omission marker.

Current head: `e3dde57931a8074e86e51a22c4b6fda628025e51`.

## Root cause

The required latest call/output pair can exceed the compact endpoint's wire
budget solely because it carries hundreds of kilobytes of image data. The
previous fail-closed guard then returned `responses_compact_input_too_large`,
permanently wedging an otherwise recoverable thread.

## Changes

- try lossless required-state selection first;
- elide data-URL images only inside eligible required tool-output items;
- handle structured `input_image` parts plus base64 and percent-encoded data
  URLs in string outputs;
- preserve call IDs, surrounding text, all otherwise-fitting context, and
  account-pinned nested `input_file` references;
- leave latest user images and normal Responses requests unchanged;
- keep oversized text and hosted `computer_call_output` screenshots
  fail-closed until a schema-valid computer screenshot placeholder exists;
- keep the image-elision constants isolated so the rebased change composes
  cleanly with current request-normalization work.

## Validation

- focused compact request suite: 151 passed, including the terminal
  compaction-trigger route and retained-file-routing regressions;
- Ruff and `ty` passed for the touched Python;
- strict OpenSpec validation previously passed for
  `allow-compact-inline-image-elision`.

Hosted checks on the current head remain authoritative for merge.

### DIFF
--- a/app/core/openai/requests.py
+++ b/app/core/openai/requests.py
@@ -46,7 +46,7 @@ _COMPACT_TOOL_CALL_ITEM_TYPES = frozenset({"function_call", "custom_tool_call", 
 _COMPACT_TOOL_CALL_OUTPUT_ITEM_TYPES = frozenset(
     {"function_call_output", "custom_tool_call_output", "apply_patch_call_output"}
 )
-_COMPACT_INLINE_IMAGE_DATA_URL_RE = re.compile(r"data:image/[^,\s]+,[A-Za-z0-9+/=_-]+")
+_COMPACT_INLINE_IMAGE_DATA_URL_RE = re.compile(r"""data:image/[^,\s]+,[^\s"'<>]+""")
 _GOAL_CONTINUATION_CONTEXT_PREFIX = '<codex_internal_context source="goal">'
 _PLAN_MODE_CONTEXT_PREFIX = "<collaboration_mode># Plan Mode"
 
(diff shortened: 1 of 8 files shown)


Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Soju06:
Reviewed at 6e7e059 — the latest P1 (https://github.com/Soju06/codex-lb/pull/1457#discussion_r3636524729) is still unresolved and applies to the current head: `_resolve_file_account_for_responses()` (app/core/openai/requests.py:189) collects `input_file` IDs from pre-transform `payload.input`, so a stale file pin from an optional tool output that `to_payload()` later drops can still become a hard account owner and fail compaction. Please derive the pins from the transformed compact input (or restrict collection to required retained tool outputs), and add a regression test at that surface — an oversized compact request where the only `input_file` lives in trimmed optional history and must not pin routing. Heads up: #1418 lands first in the compact series, so expect a rebase onto main after it merges. Once the P1 fix + test are in and the rebase is done, this is next in line.

Komzpa:
@codex review

Komzpa:
@codex review

PR #1466, opened 2026-07-24
feat(proxy): add account-bound realtime live sideband

McHersheys:
## Summary

- bind each successful realtime call `Location` to the exact account that completed call creation, including refresh/failover success paths
- require a valid proxy API key for realtime call creation and `/v1/live/{call_id}` even when ordinary proxy routes allow trusted anonymous access
- add `/v1/live/{call_id}` WebSocket forwarding that resolves and leases only the immutable bound owner and preserves allowed Frameless query/header metadata
- fail closed for malformed, unknown, expired, cross-API-key, unassigned, unavailable, or ownership-conflicting calls; never refresh or replay a definitive sideband handshake denial through another route/account
- hash opaque call IDs in affinity storage, expire mappings after two hours, and use throttled bounded cleanup in a reserved namespace
- disable frame archiving, suppress SDP from opt-in upstream payload traces, redact live IDs from application WebSocket handshake logs, and replace client authorization/account/installation identity with the bound account identity
- preserve bounded WebSocket close codes/reasons and enable ping/pong liveness for live sideband transport without changing existing Responses WebSocket behavior

## Scope and safety

This is best-effort compatibility with a source-derived private Codex flow, not a public Realtime API contract. The bridge does not proxy WebRTC media, mint entitlement, bypass upstream authorization/attestation, interpret Frameless payloads, or log/persist SDP and sideband frame payloads.

No new setting, migration, dashboard surface, or default-install step is introduced.

## Independent-review follow-up

Post-publication independent reviews surfaced affinity, diagnostic-retention, close-semantics, routed-handshake, and immediate post-refresh credential-snapshot issues. The PR was returned to draft whenever a blocker was confirmed.

The current head now provides mandatory per-key isolation, immutable ownership, SDP trace suppression, bounded cleanup, canonical installation identity, definitive-denial no-replay behavior, live transport liveness, close propagation, application-path redaction, and account-bound/cross-key lifecycle coverage. At live attachment, exact-owner policy selection and stream leasing still use the normal bounded routing cache, but the credential-use boundary fresh-loads that same leased owner from persistent storage before decrypting its token, applying account/installation identity, or resolving its route. This prevents an immediate attach from reusing a rejected pre-refresh snapshot while retaining the no-refresh/no-failover sideband contract.

The stale-snapshot regression uses a nonzero production-style selection cache, makes call creation reject token A, commits token B plus updated identities during forced refresh, successfully binds the call, and proves the immediate sideband handshake uses token B and the updated persisted identities. A newly missing or unavailable fresh owner fails closed before decryption.

The branch incorporates current upstream `main` without rewriting the published history.

## Verification

- `make lint`
- `make typecheck`
- `make architecture-check`
- `python .github/scripts/check_simplicity_budgets.py`
- focused unit, API, sticky-affinity, immediate post-refresh lifecycle, and existing Responses WebSocket regression suites — **222 passed**
- `npx --yes @fission-ai/openspec@1.6.0 validate add-realtime-live-sideband --strict --no-interactive`
- `npx --yes @fission-ai/openspec@1.6.0 validate --specs --strict --no-interactive` — **47 specs passed**
- `git diff --check`

The passing integration run emitted the existing Starlette TestClient deprecation plus intermittent aiosqlite event-loop teardown warnings; no test failed.

## Not validated

No live private-upstream protocol probe was run: doing that safely requires a maintainer-controlled, isolated, entitled account rather than a pooled/shared production credential. No pooled token was accessed, no refresh endpoint or refresh manager was invoked for a probe, and no other account was touched. Local connector, route, affinity, fail-closed, relay, cancellation, cross-key, stale-cache refresh, and regression paths are covered with hermetic tests.


McHersheys:
@codex review

Soju06:
Closing this as superseded by #1492, which merged to main on 2026-07-27 as aa642f33 — but I want to be clear that a large part of what merged traces directly back to this PR, and thank you for that. Concretely, the merged implementation carried your design over nearly intact: the required-key auth for call creation and `/v1/live/{call_id}` (`app/core/auth/dependencies.py` landed byte-identical to this branch), the SHA-256 per-API-key affinity key in the reserved `\ncodex_live_call:` namespace with the 2-hour expiry and throttled bounded cleanup, the immutable exact-owner binding with no-fallback selection and fail-closed 404/503 semantics, and the credential-use-boundary fresh reload of the leased owner before decryption — your comment explaining that boundary survives verbatim in `app/modules/proxy/_service/realtime_live.py`. #1492 extended the same architecture with v3/legacy ingress rewriting, subprotocol negotiation, OTEL path redaction, typed `realtime_live` request rows for the dashboard, and docs, plus two post-review maintainer fixes (5f422af5: reserved-prefix scoping so the pre-existing selection-affinity namespace isn't captured, and per-app OTEL instrumentation for factory-created apps). Having compared both diffs end to end, I don't see anything remaining here that isn't already on main, and the branch now conflicts with the merged module — so rather than ask for a rebase that would reduce to an empty diff, I'm closing it. If you spot a behavior from this branch that the merged version regressed, please open a targeted fix PR and reference both this and #1492.

Issue #1393, opened 2026-07-17
HTTP bridge can loop or time out after silent and clean upstream closes


shaqman:
## Summary

The HTTP Responses bridge can leave Codex CLI reconnecting until its request times out when an upstream WebSocket either stays silent before response.created or closes cleanly before response.completed.

Observed client errors:

- Stream disconnected before completion: HTTP bridge session closed before response.completed
- Request timed out after repeated reconnect attempts

## Reproduction

1. Run Codex CLI through codex-lb with the HTTP Responses session bridge enabled.
2. Start a request that obtains a bridge session.
3. Have the upstream socket remain silent before response.created, or close with code 1000 while the request is still pending.
4. Observe the client reconnect repeatedly; stale pre-created state may retain the response-create gate, and repeated clean closes can re-enter recovery without a durable cooldown.

## Expected behavior

- Silent pre-response requests terminate or recover within a bounded interval.
- A clean upstream close with a pending request gets one bounded recovery replay.
- Hard-affinity retries remain on the required account.
- Repeated failures back off across replicas instead of looping.
- Clean closes do not penalize otherwise healthy accounts.

## Related work

Issue #987 is the closest prior report, but it was closed as likely fixed and the failure remains reproducible. Issue #1304 / PR #1305 addresses a different native downstream WebSocket receive race.

A tested implementation and OpenSpec change are prepared at shaqman/codex-lb branch fix/http-bridge-clean-close-penalty.

yjx-git001:
We are seeing the same failure pattern in production and would like to add an operator report.

Normal TTFT is often around 0.3-1s, but after the service has been running for a while we intermittently see 60-70s TTFT and, in the worst cases, requests with no recorded first token that finish only after roughly 500-600s. The client can remain in a Working/reasoning state without visible output. Increasing the account pool, using round-robin routing, and raising per-account concurrency did not eliminate the stalls.

The strongest correlation is the HTTP Responses Session Bridge path: disabling the bridge and using `upstream_stream_transport=http` greatly reduces the incidents. This matches the silent owner / stale `response_create_gate` behavior described here and in #1410 much more closely than ordinary account-capacity pressure.

Stable, low TTFT is critical for our production use. Could the maintainers please prioritize a robust fix for silent upstream owners and watchdog-starved stale gates, and include it in the next stable release? In particular, the #1410 follow-up appears relevant because a stale pending request should be retired even when admission bookkeeping flags were cleared by a re-enqueue path.

[...]

PR #1483, opened 2026-07-24
fix(proxy): retry missing response.created once

mhooooo:
## Summary

Recover Codex Responses streams without exposing the reconnect loop when either:

- an HTTP bridge `response.create` receives no `response.created`; or
- a reused HTTP/direct-WebSocket upstream is already closed before the next send begins.

Every replay stays bounded to one attempt, preserves the original request budget and required account, and remains fail-closed whenever dispatch may have begun.

## Type of change

- [x] `fix:` — bug fix
- [ ] `feat:` — new user-facing capability
- [ ] `chore:` / `ci:` / `build:` — tooling
- [ ] **Breaking change**

Linked issue: Refs #1393

## OpenSpec

- [x] This PR touches a codex-faithful path and preserves upstream-equivalent request and response framing

Change directory: `openspec/changes/retry-missing-response-created-once/`

## Changes

- Cap the eventless pre-`response.created` acknowledgement window at 30 seconds instead of 240 seconds.
- Cancel the stale HTTP receive and invoke the existing safe pre-created replay path only for a sole, pre-visible request.
- Add a sealed transport result proving a reused socket was closed before the send primitive was invoked.
- Reconnect and resend the exact HTTP or direct-WebSocket request once on the same required account when that proof exists, including compacted continuations.
- Keep mid-send failures terminal because their dispatch state is ambiguous.
- Refuse direct-WebSocket replay if the retired reader does not confirm cancellation; terminally settle the turn, close downstream, and retain cleanup ownership.

## Root cause and impact

The existing bridge watchdog always failed after a missing acknowledgement, even when its guarded replay path could safely recover. Separately, both HTTP and direct-WebSocket clients could reuse a normally closed warm upstream and surface a client reconnect before the next send.

The proxy now distinguishes a provably undispatched request from an ambiguous send. Only the former receives one exact same-account resend. Existing lifecycle, visibility, sibling-request, continuation, file-ownership, admission, and one-replay guards remain authoritative.

## Simplicity

- [x] Works with zero configuration
- [x] Adds no setting or required setup step
- [x] Adds no dependency, endpoint, schema, migration, dashboard surface, or README section

## Validation

```text
Focused HTTP bridge + direct WebSocket suite
631 passed

Ruff check and format
passed (849 files)

Ty type checking
passed

Proxy architecture fitness checks
passed

Strict OpenSpec validation
48 specs passed
71 changes passed

Local Codex review at current head
No actionable correctness regressions identified
```

Live exact-revision verification also completed authenticated `/models`, HTTP Responses, and direct-WebSocket Responses requests through the locally deployed combined image.

## Checklist

- [x] Title uses Conventional Commits format
- [x] Related issue referenced
- [x] Externally failing HTTP and direct-WebSocket paths have regressions
- [x] Relevant local gates passed
- [x] Strict OpenSpec validation passed
- [x] Simplicity gates reviewed
- [x] `CHANGELOG.md` is unchanged

### DIFF
--- a/app/modules/proxy/_service/http_bridge/helpers.py
+++ b/app/modules/proxy/_service/http_bridge/helpers.py
@@ -188,7 +188,7 @@ from app.modules.proxy.ring_membership import (
 
 logger = logging.getLogger("app.modules.proxy.service")
 _HTTP_BRIDGE_BACKGROUND_CLOSE_TIMEOUT_SECONDS = 5.0
-_HTTP_BRIDGE_EVENTLESS_RESPONSE_CREATED_MAX_SECONDS = 240.0
+_HTTP_BRIDGE_EVENTLESS_RESPONSE_CREATED_MAX_SECONDS = 30.0
 _HTTP_BRIDGE_MISSING_RESPONSE_CREATED_TIMEOUT_DETAIL = "missing_response_created_timeout"
 T = TypeVar("T")
 
--- a/app/modules/proxy/_service/http_bridge/upstream_events.py
+++ b/app/modules/proxy/_service/http_bridge/upstream_events.py
@@ -423,6 +423,27 @@ class _HTTPBridgeUpstreamEventsMixin:
                                 if not expired_owner:
                                     continue
                                 pending_count = len(session.pending_requests)
+                                can_retry_eventless_owner = pending_count == 1
+                            receive_cancelled = True
+                            if receive_task is not None:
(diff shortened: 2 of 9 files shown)


mhooooo:
Superseded by #1484, stacked on #1481. The retry-only approach here shortens one acknowledgement wait but does not converge when durable bridge ownership and the broad legacy session row remain persistently contradictory. #1484 fixes that pre-dispatch ownership loop while preserving the durable owner and fail-closed specific ownership boundaries.

PR #1490, opened 2026-07-25
fix(compact): recover previous-response-pinned compaction when the owner account is unavailable

### LINKED ISSUE #998: bug: compaction fails
### Pre-flight checklist

### codex-lb version

codex-lb:latest (Release v1.19.0)

### Deployment method

Docker Compose

Codex app (desktop / web)

Business

### Model(s) involved

gpt-5.5

### What happened?

50% of the time when compacting, I get this error:
`Error running remote compact task: Fatal error: remote compaction v2 expected exactly one compaction output item, got 0 from 0 output items`

The compaction to work normally. It was working perfectly 2 days ago, but has now stopped working.

### Steps to reproduce

```markdown
1. Start codex-lb with the following docker compose:

services:
  codex-lb:
    image: 'ghcr.io/soju06/codex-lb:latest'
    restart: unless-stopped
    depends_on:
      postgres:
        condition: service_healthy
    expose:
      - '2455'
      - '1455'
    volumes:
      - 'codex-lb-data:/var/lib/codex-lb'
    environment:
      CODEX_LB_DATABASE_URL: 'postgresql+asyncpg://codex_lb:[HIDDEN]@postgres:5432/codex_lb'
      CODEX_LB_ENCRYPTION_KEY_FILE: /var/lib/codex-lb/encryption.key
      CODEX_LB_DATABASE_MIGRATE_ON_STARTUP: 'true'
      CODEX_LB_OAUTH_CALLBACK_HOST: 0.0.0.0
      CODEX_LB_OAUTH_CALLBACK_PORT: '1455'
      CODEX_LB_OAUTH_REDIRECT_URI: 'https://[HIDDEN].com/auth/callback'
      CODEX_LB_DASHBOARD_BOOTSTRAP_TOKEN: [HIDDEN]
      CODEX_LB_USAGE_REFRESH_ENABLED: 'true'
  postgres:
    image: 'postgres:16-alpine'
    restart: unless-stopped
    exclude_from_hc: true
    environment:
      POSTGRES_USER: codex_lb
      POSTGRES_PASSWORD: [HIDDEN]
      POSTGRES_DB: codex_lb
    volumes:
      - 'postgres-data-v3:/var/lib/postgresql/data'
    healthcheck:
      test:
        - CMD-SHELL
        - 'pg_isready -U codex_lb -d codex_lb'
      interval: 10s
      timeout: 5s
      retries: 10
volumes:
  codex-lb-data: null
  postgres-data-v3: null

2. Configure the following in the Codex config.toml:

[model_providers.codex-lb]
name = "OpenAI"  # required — enables remote /responses/compact
base_url = "https://[HIDDEN].com/backend-api/codex"
wire_api = "responses"
experimental_bearer_token = "[HIDDEN]"

3. Run a Codex prompt using extra high reasoning (I have not tested with other reasoning levels
4. Wait for it to try compacting and watch it fail 50% of compactions.
```

### Relevant logs

```shell

```

```yaml

```

### Additional context

I am not the only one experiencing this. Users of CCX have experienced a similar issue. Heavy emphasis on similar, as they were getting `0 from 2 output items`, while I am getting `0 from 0 output items`. See this: https://github.com/BenedictKing/ccx/issues/179

Iweisc:
## Summary

When a long conversation's previous-response owner account runs out of quota mid-session and the next client action is a compaction, the compact request fails and the session wedges: the pin hard-requires the exhausted owner at selection time, the failure surfaces straight to the client, and every retry re-resolves the same dead owner until its quota window resets. Normal turns already escape this exact state via account-neutral fresh replay; this PR gives the compact surface the same selection-time recovery.

## Type of change

Linked issue: none found for this exact wedge (closed #998 is the unrelated compact output-shape bug). Repro: pin a compact to an account, mark it `RATE_LIMITED` with an unelapsed reset (or let the compact itself hit a 429 `usage_limit_reached`), and send a compaction with a self-contained history — before this fix the route returns `503 no_accounts` / re-raises the owner's 429 forever.

## OpenSpec

Change directory: `openspec/changes/fix-compact-previous-response-failover/` (one ADDED requirement in `responses-api-compat`; validated with `--strict`, `openspec validate --specs` passes)

Composition with existing contracts, verified against current specs and the active `classify-required-http-bridge-owner-unavailable` change:

- Recovery activates **only** when the pin is exclusively `previous_response_id` and only at **selection time** (owner unselectable, or excluded after an in-request pre-visible quota failover). Post-selection refresh/auth/transport/timeout failures stay owner-bound ("post-selection owner failure remains owner-bound").
- Turn-state-pinned compacts stay fail-closed ("Compact requests preserve scoped turn-state ownership"), file-pinned compacts stay owner-bound ("File-pinned compact refresh/connect failures fail closed"), and an owner-lookup **miss** stays fail-closed (a missing owner is not a selector result).
- API-key reservation settlement is untouched: recovery adds no terminal raise; every exit still passes the existing settle-before-raise sites.

## Changes

- `app/modules/proxy/_service/compact.py`
  - New `_compact_account_neutral_replay_payload` helper: returns the anchor-free `ResponsesCompactRequest` only when the request carries `previous_response_id`, a list `input` with more than one item, and the upstream-bound payload without the anchor passes the shared `responses_payload_is_account_neutral_fresh_replay` gate (rejects encrypted/compaction state, server-assigned item ids, account-scoped file/container handles, conversation/prompt handles, hosted/MCP state, unknown shapes).
  - In the `compact_responses` selection loop, when selection returns no account and the request is pinned only by the previous-response owner: for a verified payload, exclude the owner, strip the anchor, drop downstream session/turn affinity aliases from upstream-bound headers (`without_http_bridge_session_affinity_headers`), blank sticky affinity, and reselect with fallback enabled; for an unverified payload, keep today's failure and record it on the existing `continuity_fail_closed` counter (surface `compact`, reason `owner_account_unavailable`).
- Regression tests at the externally failing surface (`POST /backend-api/codex/responses/compact`) plus unit coverage (details below).

## Test plan

New integration tests reproduce the wedge and fail without the fix (`503 no_accounts` on a pinned compact; verified by stashing the fix):

- `test_proxy_compact_pinned_owner_quota_429_fails_over_with_account_neutral_full_resend` — owner 429s mid-request; asserts upstream call order owner→replacement, anchor present on the owner call and stripped on the replay, `session_id` header stripped on the replay, owner persisted `RATE_LIMITED`.
- `test_proxy_compact_pinned_owner_rate_limited_at_selection_fails_over_with_account_neutral_full_resend` — owner already `RATE_LIMITED` at selection; asserts exactly one upstream call on the replacement without `previous_response_id`.
- `test_proxy_compact_pinned_owner_unavailable_non_neutral_payload_stays_fail_closed` — encrypted compaction item in the history; asserts 503 and zero upstream calls.
- Unit: helper eligibility/rejection matrix (7 cases), service-level recovery (reselect kwargs: pin dropped, owner excluded, blank affinity; upstream payload/headers sanitized), non-neutral fail-closed with `continuity_fail_closed` recorded.

```
uv run pytest tests/unit tests/test_request_logs_options_api.py -q
# 4880 passed, 60 skipped
uv run pytest tests/integration --ignore=tests/integration/test_http_responses_bridge.py \
  --ignore=tests/integration/test_proxy_websocket_responses.py -q
# 1488 passed, 9 skipped, 1 failed (pre-existing, unrelated: test_quota_planner_api::
#   test_quota_planner_warm_now_keeps_bootstrap_for_metadata_less_primary_rows fails
#   identically on a clean checkout of main in this environment)
uv run pytest tests/integration/test_http_responses_bridge.py tests/integration/test_proxy_websocket_responses.py -q
# 182 passed
make lint && uv run ty check
# All checks passed
```

## Screenshots / output

Before (pinned compact, owner rate-limited — request wedges):

```
WARNING app.modules.proxy.service: Proxy preferred account unavailable request_id=... kind=compact
    preferred_account_id=acc_wedged_owner_ac88c3c7 error_code=None error=No available accounts
ERROR app.modules.proxy.api: proxy_error_response method=POST path=/backend-api/codex/responses/compact
    status=503 code="no_accounts" message="No available accounts"
```

After (same state — compact recovers on the replacement account):

```
WARNING app.modules.proxy.service: Compact previous-response owner unavailable; replaying verified
    account-neutral full resend request_id=... owner_account_id=acc_wedged_owner_ac88c3c7 selection_error_code=None
→ 200 {"object": "response.compaction", ...}
```

## Checklist

- [x] Linked the related issue / discussion above (none exists; bug described inline).
- [x] Ran the relevant `make <target>` subset locally (`make lint`, `ty check`, unit + integration slices).
- [x] If touching specs: `openspec validate --specs` passes and the change validates with `--strict`.
- [x] Simplicity gates reviewed: no new settings, endpoints, README sections, or dashboard surfaces; recovery is zero-config and only activates where the request previously failed.


Iweisc:
Addressed the P2 finding in `7f63c22c`.

You were right that the branch treated every excluded-owner selection miss alike. A pinned compact that gets a second 401 after the forced refresh also lands in `excluded_account_ids`, so the next outer iteration reached the recovery branch and replayed the history on another account — exactly the case the added requirement says must stay owner-bound.

Recovery is now gated on quota-caused owner loss:

- `preferred_account_id not in excluded_account_ids` — the owner was never used for this request (selection-time loss, e.g. already `RATE_LIMITED` with an unelapsed reset), **or**
- `owner_quota_failover_eligible` — set only where `action == "failover_next"` and the failing account is the pinned owner with `classified["failure_class"] in ("rate_limit", "quota")`, i.e. an explicitly pre-visible quota / rate-limit error.

Every other post-selection exclusion of the pinned owner (repeated 401 after refresh, refresh contention/transport, forced-refresh failures, exhausted 500 transient retries, `account_response_create_cap`, security-work re-selection) no longer reaches the branch and keeps its prior owner-bound surface.

Also added:

- `openspec/changes/fix-compact-previous-response-failover/specs/responses-api-compat/spec.md` — the eligibility condition is now normative, plus a scenario "Post-selection authentication failure on the pinned owner stays owner-bound".
- `tests/integration/test_proxy_compact.py::test_proxy_compact_pinned_owner_post_selection_401_stays_owner_bound` — asserts the client sees `401` and upstream is called only twice, both on the owner. Verified it fails without the gate (it replayed on the alt account, logging `Compact previous-response owner unavailable; replaying verified account-neutral full resend`).

Local: `ruff check` + `ruff format --check` + `ty check` clean, 4868 unit tests pass, 1671 integration tests pass, `openspec validate --strict` clean. The one integration failure (`test_quota_planner_api::test_quota_planner_warm_now_keeps_bootstrap_for_metadata_less_primary_rows`) reproduces on a clean checkout of `main` and is unrelated to this PR.

@codex review

Iweisc:
Addressed the P2 wire-length finding in `b9f45812`.

Confirmed the failure mode: `ResponsesCompactRequest.to_payload()` → `_strip_compact_unsupported_fields` → `_strip_poisoned_local_compact_fallback_items` drops a poisoned local-compact fallback message together with its trailing encrypted compaction item. So the `tests/unit/test_openai_requests.py::test_compact_strips_poisoned_local_compact_fallback_items` shape (3 items) serializes to exactly `[{"role": "user", "content": "continue"}]`. My `len(payload.input) > 1` check ran on the request model, so it passed; the neutrality validator then never saw the stripped compaction state either, and the replacement account would have compacted only the latest message.

`_compact_account_neutral_replay_payload` now derives the serialized upstream-bound payload once and runs **both** gates on it:

```python
replay_wire_payload = replay_payload.to_payload()
replay_wire_input = replay_wire_payload.get("input")
if not isinstance(replay_wire_input, list) or len(replay_wire_input) <= 1:
    return None
if not responses_payload_is_account_neutral_fresh_replay(replay_wire_payload):
    return None
```

The model-level check is now only `isinstance(payload.input, list)`; the item count is authoritative on the wire.

Also updated:

- `specs/responses-api-compat/spec.md` — verification is now normatively defined as running against the exact upstream-bound payload "after every wire transformation the compact serializer applies", with a new scenario "History that serializes to a single wire item stays fail-closed".
- `tests/unit/test_proxy_utils.py` — added the poisoned-fallback case to the rejection matrix ("multi-item input that serializes to a single wire item").
- `tests/integration/test_proxy_compact.py::test_proxy_compact_pinned_owner_unavailable_wire_truncated_history_stays_fail_closed` — the route returns `503` and upstream is never called.

Both new tests were verified to fail against the previous model-level check.

Local: `ruff check` + `ruff format` + `ty check` clean, 4869 unit tests pass, 1672 integration tests pass, `openspec validate --strict` clean. The only integration failure remains `test_quota_planner_api::test_quota_planner_warm_now_keeps_bootstrap_for_metadata_less_primary_rows`, which reproduces on a clean `main` and is unrelated.

@codex review

Iweisc:
Addressed the P1 in `a207fee9`.

You are right that `len(...) > 1` on the wire input is not proof of a full resend. `_trim_compact_input_for_upstream` keeps a head, an account-neutral trim marker, and a tail, and poisoned-fallback stripping can drop a pair while leaving two or more items — both would have passed. Only the owner can resolve the omitted context from the anchor I strip, so a shortened history must never leave the owner.

Verification is now an identity check against the pre-transformation input:

```python
request_input = replay_source.get("input")
if not isinstance(request_input, list) or len(request_input) <= 1:
    return None
try:
    replay_payload = ResponsesCompactRequest.model_validate(replay_source)
    replay_wire_payload = replay_payload.to_payload()
except (ValidationError, ClientPayloadError):
    return None
replay_wire_input = replay_wire_payload.get("input")
if not isinstance(replay_wire_input, list) or len(replay_wire_input) <= 1:
    return None
if replay_wire_input != request_input:
    return None
if not responses_payload_is_account_neutral_fresh_replay(replay_wire_payload):
    return None
```

Any serializer-dropped, trimmed, or otherwise rewritten history now fails closed on the owner. I also wrapped the speculative serialization: `_trim_compact_input_for_upstream` can raise `ClientPayloadError` (`responses_compact_input_too_large`) for an untrimmable payload, and that must not escape from the recovery probe — the request keeps its existing failure instead.

Added coverage:

- `tests/unit/test_proxy_utils.py::test_compact_account_neutral_replay_payload_rejects_wire_trimmed_history` — asserts the 60-item oversized history really does trim (`1 < len(wire) < 60`) and that the helper still returns `None`.
- `tests/integration/test_proxy_compact.py::test_proxy_compact_pinned_owner_unavailable_wire_trimmed_history_stays_fail_closed` — the route returns `503` and upstream is never called.

Both fail without the identity check. Spec updated: verification now normatively requires the serialized `input` to be "item-for-item identical to the validated request `input`", and the scenario is generalized to "History the wire serializer shortens stays fail-closed".

Local: `ruff check` + `ruff format` + `ty check` clean, 4870 unit tests pass, 1673 integration tests pass, `openspec validate --strict` clean. Only the unrelated pre-existing `test_quota_planner_api::test_quota_planner_warm_now_keeps_bootstrap_for_metadata_less_primary_rows` fails, and it reproduces on a clean `main`.

@codex review

Iweisc:
Addressed the P1 on `compact.py` (require independently trusted proof of the anchored history) in `1366e05c`.

**Why the previous check was insufficient.** The serializer-identity check only proved `to_payload()` did not rewrite or drop items. It said nothing about *which* conversation the list holds: a client may legitimately send only the turns after `previous_response_id` and rely on the owner account to hold everything before it. Two or more account-neutral items is exactly what that looks like on the wire, so recovery could drop the anchor and compact a partial history on the replacement account.

**Fix.** Recovery now additionally requires the proxy's own durable continuity record for that anchor — the same evidence the HTTP-bridge full-resend replay path already trusts (`durable_full_resend_anchor_count` / `..._fingerprint` in `_service/http_bridge/streaming.py`):

- new `DurableBridgeSessionCoordinator.lookup_previous_response_target` (alias-kind-scoped `previous_response_id` lookup, falling back to `find_session_by_latest_response_id`, API-key scoped);
- new `_CompactMixin._compact_previous_response_carries_anchored_history`, which requires the record to exist, to not name an account other than the pinned owner, to carry a recorded input prefix count + fingerprint, and the request `input` to strictly extend that recorded prefix via the shared `_input_prefix_matches_stored_context`;
- a missing record, a missing recorded prefix, a durable owner mismatch, a prefix fingerprint mismatch, or a failed durable lookup all keep the request owner-bound (fail-closed, with the `continuity_fail_closed` compact-surface counter still recorded).

The request-log path could not supply this: `find_latest_owner_record_for_response_id` returns only `(account_id, requested_at, session_id)`, no prefix count or fingerprint. So the durable bridge row is the only trusted source, and recovery is now scoped to conversations that have one.

**Coverage** (each new test verified to fail with the proof gate stubbed to `True`):

- `tests/unit/test_proxy_utils.py::test_compact_previous_response_owner_unavailable_without_anchor_proof_fails_closed` — parametrized over no record / no recorded prefix / durable owner mismatch / fingerprint mismatch / input that adds nothing after the recorded prefix.
- `...::test_compact_previous_response_anchor_lookup_failure_fails_closed` — durable lookup raising keeps the owner pin (asserts selection is not retried).
- `...::test_compact_previous_response_owner_unavailable_recovers_with_account_neutral_replay` now stubs the durable evidence and asserts the lookup is made with the anchor id and the request's API-key scope.
- `tests/integration/test_proxy_compact.py::test_proxy_compact_pinned_owner_unavailable_without_anchor_evidence_stays_fail_closed` — neutral full resend, no durable row → 503, zero upstream calls.
- `...::test_proxy_compact_pinned_owner_unavailable_diverging_history_stays_fail_closed` — durable row present but the request does not open with the recorded prefix → 503, zero upstream calls.
- The two recovery integration tests now persist a real durable anchor row (`claim_live_session` + `register_previous_response_id` with the prefix count/fingerprint) so they exercise the full evidence path end to end.

OpenSpec `spec.md` gained the normative requirement for trusted anchored-prefix evidence plus an "Unproven anchored history stays fail-closed" scenario; `tasks.md` item 9 records the work.

**Gates:** `uv run ruff check .` clean, `uv run ruff format` clean, `uv run ty check` clean, `uv run pytest tests/unit` 4876 passed, `uv run pytest tests/integration` 1675 passed with the one pre-existing unrelated failure (`test_quota_planner_api.py::test_quota_planner_warm_now_keeps_bootstrap_for_metadata_less_primary_rows`, which fails identically on a clean `main` checkout), OpenSpec strict validation passes.

@codex review

Iweisc:
Addressed the P1 on `compact.py:572` (require proof that the anchored response output is retained) in `7e8957ed`. The scenario you described was real: with the durable prefix being the input that produced the anchored response, `[user: hello]` + `[user: hello, user: please compact]` matched the fingerprint, strictly extended the prefix, and passed the account-neutral gate — while the assistant output `previous_response_id` stands for was absent.

**Fix.** `_compact_previous_response_carries_anchored_history` now applies `responses_input_suffix_retains_prior_output(payload.input, stored_count=...)` after the prefix fingerprint match, i.e. exactly the retained-output proof `owner_unavailable_allows_account_neutral_replay` uses in `_service/http_bridge/streaming.py`. The items after the recorded prefix must retain the anchored response's completed assistant output with all tool calls settled, and only then the new client input.

No projection step is applied here on purpose: compact recovery already requires the wire `input` to be item-for-item identical to the request `input`, so projecting reasoning items or server ids away would contradict that gate. Running the retained-output rule on the raw list is the stricter (fail-closed) direction.

**Coverage** (both verified to fail with the retained-output call stubbed to `True`):

- `tests/unit/test_proxy_utils.py::test_compact_previous_response_owner_unavailable_without_retained_output_fails_closed` — your exact example: durable prefix `[user: hello]`, request `[user: hello, user: please compact]` → 503, no upstream call, selection not retried.
- `tests/integration/test_proxy_compact.py::test_proxy_compact_pinned_owner_unavailable_without_retained_output_stays_fail_closed` — same shape at `POST /backend-api/codex/responses/compact` with a real durable anchor row → 503, zero upstream calls.
- The `..._without_anchor_proof_fails_closed` parametrization gained an "anchored response output is not retained after the recorded prefix" case.
- The two recovery tests now record the anchored turn's *request* input as the durable prefix (`_NEUTRAL_FULL_RESEND_INPUT[:1]`), so they exercise prefix match → retained assistant output → fresh user input end to end, which is the semantics `latest_input_item_count` actually carries.

OpenSpec `spec.md` now requires the retained-output proof in the normative paragraph and in the "Unproven anchored history stays fail-closed" and quota-failover scenarios; `tasks.md` item 10 records the work.

**Gates:** `uv run ruff check .` clean, `uv run ruff format` clean, `uv run ty check` clean, `uv run pytest tests/unit` 4878 passed, `uv run pytest tests/integration` 1676 passed with the one pre-existing unrelated failure (`test_quota_planner_api.py::test_quota_planner_warm_now_keeps_bootstrap_for_metadata_less_primary_rows`, identical on a clean `main`), OpenSpec strict validation passes.

@codex review

Iweisc:
Addressed the P1 on `compact.py:577` (bind prefix proof to the requested response alias) in `12d7b590`. Confirmed the defect: `register_owned_alias` writes `latest_input_item_count` / `latest_input_full_fingerprint` onto the **session** row, so each later response registration overwrites them, while an older `previous_response_id` alias still resolves to that same session. The proof was therefore being evaluated against whatever turn the session had most recently recorded.

**Fix.** `_compact_previous_response_carries_anchored_history` now requires `durable_lookup.latest_response_id == previous_response_id` before the prefix/retained-output checks run. A snapshot that has moved on to a later response describes a different turn, so it proves nothing about the anchored history and the request stays owner-bound. I chose the equality gate over per-alias prefix metadata deliberately: adding new persisted columns for a fail-closed recovery path is a schema change with a migration and backfill for existing rows, and it would widen this PR beyond one concern. The equality gate narrows recovery to the case where the anchor *is* the session's current head, which is the only case the stored evidence can actually attest to.

**Coverage** (both verified to fail with the new gate disabled):

- `tests/unit/test_proxy_utils.py::test_compact_previous_response_owner_unavailable_without_anchor_proof_fails_closed[...recorded prefix describes a newer response than the requested anchor]` — snapshot whose `latest_response_id` is a newer response, with a prefix count/fingerprint that would otherwise match.
- `tests/integration/test_proxy_compact.py::test_proxy_compact_pinned_owner_unavailable_stale_anchor_alias_stays_fail_closed` — your exact `r1` → `r2` shape at the route: `resp_stale_anchor` is registered with an earlier prefix the request no longer sends, then `resp_stale_newer` is registered on the same durable session with a prefix that *does* match the request. The alias for `resp_stale_anchor` still resolves, and the compact stays 503 with zero upstream calls.

`spec.md` now requires the durable record to report the requested `previous_response_id` as its own latest response (normative paragraph plus both affected scenarios); `tasks.md` item 11 records the work.

**Gates:** `uv run ruff check .` clean, `uv run ruff format` clean, `uv run ty check` clean, `uv run pytest tests/unit` 4879 passed, `uv run pytest tests/integration` 1677 passed with the one pre-existing unrelated failure (`test_quota_planner_api.py::test_quota_planner_warm_now_keeps_bootstrap_for_metadata_less_primary_rows`, identical on a clean `main`), OpenSpec strict validation passes.

@codex review

Iweisc:
Head `3f2641b9` addresses the two P1s from the review of `12d7b590`.

### 1. "Rebind continuity to the replacement compact owner" — fixed

Confirmed and real. Recovery blanked `affinity` only for its own retried selection, so nothing rewrote the client's continuity records: the sticky mapping for the session key and the durable session that proved the anchored history both still named the lost owner. The next turn, carrying the replacement account's `encrypted_content`, routed straight back to that owner.

`_compact_rebind_recovered_continuity_owner` (`app/modules/proxy/_service/compact.py`) now runs after the upstream compaction succeeds on the account that actually served it:

- **Sticky mapping** — `repos.sticky_sessions.upsert(selection_key, account_id, kind=...)`, the same rebind the bridge performs at `_service/http_bridge/request_submit.py:1745` when a session moves account.
- **Durable session** — `claim_live_session(..., account_id=<serving account>, latest_turn_state=None, latest_response_id=None, allow_takeover=True)` on the proving snapshot's canonical key, then `release_live_session(draining=True)`. The account change in `claim_session` also clears that session's aliases and recorded turn state, which describe the lost owner's upstream state. The lease is handed straight back because the compact surface runs no live bridge session and must not draw owner-forwarded traffic.
- **Hard turn-state keys are never rebound.** That key is the lost owner's own opaque state; pointing it at another account would authorize exactly the cross-account send this rebind exists to prevent. Only `session_header`-sourced Codex-session keys and soft prompt-cache / sticky-thread keys rebind, matching the bridge precedent.
- Rebinding happens after the compaction already succeeded upstream, and a retry would take the same recovery path, so a rebind failure is logged and the compaction is still returned rather than thrown away.

To make the gate observable, `_compact_previous_response_carries_anchored_history` now returns the proving `DurableBridgeLookup` (or `None`) instead of a bool, so recovery rebinds exactly the session that proved the history.

Coverage:
- `test_compact_previous_response_owner_unavailable_recovers_with_account_neutral_replay` — asserts the sticky upsert target/kind and the claim/release kwargs (`session_key_kind/value` from the proving snapshot, `account_id` = replacement, `allow_takeover=True`, `latest_response_id`/`latest_turn_state` cleared, `draining=True`).
- `test_compact_recovery_never_rebinds_a_hard_turn_state_sticky_row` — synthesized turn state, owner resolves to nothing, recovery runs: durable session rebinds, sticky row does not.
- `test_proxy_compact_pinned_owner_quota_429_fails_over_with_account_neutral_full_resend` (compact route) — after the 200, the persisted `StickySession` row and `HttpBridgeSessionRecord` both name the replacement account, the durable row is `DRAINING` with `latest_response_id IS NULL`, and the alias table is empty.

Both gates verified load-bearing: stubbing the rebind call fails the unit and route tests; widening the turn-state exclusion fails the turn-state test.

OpenSpec: new normative paragraph plus the "Recovered compaction moves continuity ownership to the serving account" and "Continuity rebind failure keeps the successful compaction" scenarios, and `tasks.md` item 12.

### 2. "Authenticate the retained output against the anchor" — parity with the shipped path, no manifest exists to authenticate against

This one I'd like to push back on, with evidence.

`responses_input_suffix_retains_prior_output` is exactly the check the shipped HTTP-bridge account-neutral fresh replay runs, on exactly the same durable evidence: `app/modules/proxy/_service/http_bridge/streaming.py:1132` calls the same helper with `stored_count` from the same `latest_input_item_count` / `latest_input_full_fingerprint` pair, and that path also compares no output content. Compact is now at parity with the mechanism already in production for normal turns; holding compact to a stricter bar would leave the far more common bridge path as the weaker one.

There is also nothing to authenticate against: the durable session row persists the anchored *request* prefix (count + fingerprint) and no representation of the response output. Adding an output fingerprint or manifest means a new column, a bridge write path that populates it, and a backfill for historical rows — a schema-and-bridge change well outside this PR's one concern, and it would have to land on the bridge path too or it buys nothing.

Finally, the threat it would close is already open by a shorter route: a client that fabricates assistant output can send that history as an **anchor-free** compact request, which carries no previous-response pin and is never owner-bound. What the anchor actually guarantees is that owner-resident history is not silently dropped, and that is precisely what the durable prefix match plus the retained-output shape proof enforce. Content-level output authentication is a real hardening idea for both surfaces — happy to file it as a follow-up issue if you want it tracked.

### Gates

`uv run ruff check .`, `uv run ruff format`, `uv run ty check`, `uv run pytest tests/unit` (4880 passed), `uv run pytest tests/integration` (1677 passed), `openspec validate fix-compact-previous-response-failover --strict` all clean. The single integration failure, `test_quota_planner_api.py::test_quota_planner_warm_now_keeps_bootstrap_for_metadata_less_primary_rows`, reproduces identically on clean `main` and is unrelated to this change.

@codex review

Iweisc:
Addressed the P1 on the post-success continuity rebind (`68551622`).

**Guard the durable rebind against a concurrent newer turn** — confirmed. The rebind reopened the durable row by session key with `claim_live_session(..., allow_takeover=True)`, which does not look at the snapshot that proved the anchored history. `claim_session` handles an account change by clearing the row's aliases and its recorded `latest_response_id` / `latest_turn_state`, so an ordinary bridge turn that completed for the same canonical session between the proof lookup and this rebind could have its continuity erased after it had already returned to its client.

The move is now a compare-and-set:

- `DurableBridgeRepository.rebind_session_account_if_current` (`app/modules/proxy/durable_bridge_repository.py:368`) is a single `UPDATE ... RETURNING` whose `WHERE` pins `id`, `owner_epoch`, `account_id`, and `latest_response_id` to the values the proof observed. It sets the new account/model/tier, bumps `owner_epoch` to fence any writer still holding the observed one, drops the lease (`owner_instance_id=None`, `state=DRAINING`), clears the recorded anchor/turn state and input prefix, and clears the aliases. No matching row → rollback and `None`, so the newer owner keeps the row untouched.
- `_compact_rebind_recovered_continuity_owner` calls that instead of claim/release. A refused compare-and-set logs `Compact recovery left durable continuity owner unchanged; session moved on` and returns; the compaction that already succeeded upstream is still returned to the client. The sticky rebind is unchanged (and still never a hard turn-state key).

Coverage:

- `tests/unit/test_proxy_utils.py::test_compact_previous_response_owner_unavailable_recovers_with_account_neutral_replay` now asserts the compare-and-set arguments come from the proving snapshot (`session_id`, `expected_owner_epoch`, `expected_account_id`, `expected_latest_response_id`).
- `tests/unit/test_proxy_utils.py::test_compact_recovery_keeps_compaction_when_durable_rebind_is_refused` — refused rebind still returns the compaction and still rebinds sticky.
- `tests/integration/test_proxy_compact.py::test_proxy_compact_recovery_keeps_a_concurrent_newer_turns_durable_continuity` — at the compact route, a newer response is registered on the proving durable session while the replacement account is serving the compaction; the durable row keeps the owner account, `latest_response_id=resp_cas_newer`, and that alias, while the client's sticky key still follows to the replacement account. Verified load-bearing: dropping the epoch/account/response predicates from the `WHERE` makes only this test fail.

OpenSpec: added the compare-and-set requirement paragraph and the scenario "A concurrent newer turn keeps its durable continuity" to `openspec/changes/fix-compact-previous-response-failover/specs/responses-api-compat/spec.md`, plus tasks item 13.

Gates: `ruff check` / `ruff format` / `ty check` clean, `openspec validate --strict` valid, `tests/unit` 4881 passed, `tests/integration` 1678 passed with the one pre-existing unrelated failure (`test_quota_planner_api.py::test_quota_planner_warm_now_keeps_bootstrap_for_metadata_less_primary_rows`, fails identically on clean `main`).

@codex review

Iweisc:
Addressed the P1 on the rebind's compare-and-set (`d991c853`).

**Fence rebinds against in-flight bridge turns** — confirmed, and thanks for the precise window. A turn that has been submitted upstream (`streaming.py:2300`) but has not returned yet changes none of the fields the compare-and-set compared: `account_id`, `owner_epoch`, and `latest_response_id` only move when the eventual response alias is registered. Stealing the row in that window clears the aliases and recorded anchor and advances the epoch, so the in-flight turn's own registration is fenced out and the response it already streamed to the client ends up with no durable continuity.

The field that *can* see that turn is ownership: `register_owned_alias` (`app/modules/proxy/durable_bridge_repository.py:717`) is itself fenced on `owner_instance_id == instance_id AND owner_epoch == …`, so a worker can only record continuity while it holds the row. The compare-and-set now also requires `owner_instance_id IS NULL` (`app/modules/proxy/durable_bridge_repository.py:406`), which refuses exactly while a worker could still register a response — no separate coordination channel or per-turn generation column needed, and no window where a submitted-but-unregistered turn looks idle. A held session keeps its account, anchor, and aliases; compact logs `… left durable continuity owner unchanged; session moved on or is still owned` and still returns the compaction. The sticky rebind is unaffected, so the client's own key still follows the recovered compaction.

Coverage:

- `tests/unit/test_durable_bridge_sessions.py::test_rebind_session_account_if_current_moves_only_an_unowned_matching_row` — held row refused, stale-snapshot row refused, unowned matching row moved with the epoch bumped and aliases/anchor cleared.
- `tests/integration/test_proxy_compact.py::test_proxy_compact_recovery_leaves_a_leased_durable_session_to_its_bridge_owner` — at the compact route, the proving durable session is still leased by a bridge worker; recovery compacts on the replacement account, the durable row keeps the owner account, `resp_lease_anchor`, its alias, and its lease, and the sticky key still moves. Verified load-bearing: dropping the ownership predicate makes only this test fail.
- `_register_durable_previous_response_anchor` now releases the lease after registering (what the bridge does when it closes the session that served the anchored turn), with `keep_lease=True` for the new test, so the recovery fixtures model a real post-turn row instead of a permanently held one.

OpenSpec: added the unowned-row requirement paragraph and the scenario "A durable session held by a bridge worker keeps its owner", plus tasks item 14.

Gates: `ruff check` / `ruff format` / `ty check` clean, `openspec validate --strict` valid, `tests/unit` 4882 passed, `tests/integration` 1679 passed with the same pre-existing unrelated failure (`test_quota_planner_api.py::test_quota_planner_warm_now_keeps_bootstrap_for_metadata_less_primary_rows`, fails identically on clean `main`).

@codex review

Soju06:
Appreciate the persistence on this — the selection-time failover intent is right and matches how ordinary turns already recover. Holding merge review until the codex review loop converges though: the current head still has multiple unresolved non-outdated P1 threads (full-history proof, alias binding, rebind CAS/fencing, legacy raw-session fail-closed), and the diff has grown to ~2.6k lines by absorbing durable-bridge proof/rebind logic. Two asks: (1) resolve or rebut the open P1s at the current head, and (2) if the coordinator/repository rebind work can stand alone, split it out — the compact-recovery core will land much faster at its original scope.

Soju06:
Status check on the two asks from the 07-26 review (resolve/rebut the open P1 threads at head d991c85; split the durable-bridge rebind work out): both still open. Since then #1394 landed (a66f7937) and reworked `app/modules/proxy/durable_bridge_repository.py` (+638) and `durable_bridge_coordinator.py` (+184) — exactly where this branch now conflicts, along with `tests/unit/test_proxy_utils.py`. Splitting the compact-recovery core from the rebind/CAS machinery is now also the cheapest path through the rebase; the core recovery gate itself merges far more cleanly on its own.

Soju06:
Thanks again for the fast iteration on 07-25/26 — the core diagnosis is correct and worth landing: a previous-response-pinned compact wedges when the owner account is quota-exhausted, while normal turns already escape via account-neutral fresh replay. Status as of this triage round, plus a concrete path to mergeable:

**Current blockers**
- 6 unresolved non-outdated P-threads (5x P1, 1x P2) at head `d991c853` — per repo gates these block merge until resolved or rebutted.
- `mergeable: CONFLICTING` — #1394 (a66f7937) reworked `app/modules/proxy/durable_bridge_repository.py` (+638) and `durable_bridge_coordinator.py` (+184), exactly where this branch's rebind/CAS/fencing machinery lives (plus `tests/unit/test_proxy_utils.py`). That machinery would need re-deriving against the new repository shape, not just a mechanical rebase.
- No branch activity since 07-26 despite the maintainer's asks on 07-26 and 08-04.

**Suggested rescope (moots most open P1s instead of fixing them)**
Every escalation since `3f2641b9` exists to make cross-account *continuity rebinding* safe — alias rebind, CAS, epoch fencing, in-flight turn fences, legacy raw-session rows (threads at `compact.py:1232`, `durable_bridge_repository.py:410`, `compact.py:666`). Soju06 already asked to split that out; I'd go further and drop it from this PR entirely: activate recovery only when the request carries no session/prompt-cache affinity and no durable-session row — i.e. the pin is exclusively `previous_response_id` over a self-contained history. With nothing to rebind, the three rebind-related P1s disappear by construction, and the diff returns to the compact.py-only scope the maintainer said "will land much faster".
- Full-history proof (threads at `compact.py:531` and `compact.py:603`): instead of new durable-metadata fingerprint plumbing, reuse the exact gates the normal-turn fresh-replay path already trusts (`responses_payload_is_account_neutral_fresh_replay` plus the retained-prior-output check the HTTP bridge uses). Where those gates cannot prove the history, fail closed on the existing `continuity_fail_closed` counter — today's behavior, no new invariants.
- P2 at `compact.py:1170`: gate recovery on *why* the owner was unselectable — recover only for quota/rate-limit-caused exclusion, keeping post-selection auth/transport failures owner-bound as your own spec text requires.

**If you don't have bandwidth**
Completely understandable given how much this grew. Say the word and we'll close this and carry the rescoped core recovery into a fresh PR on top of current main, with attribution. Otherwise, if there's no activity by the next triage round, we'll do that by default so the fix isn't lost.

Soju06:
Triage status for this round: thanks @Komzpa for the 08-14 rebase over #1394 — conflicts are cleared. However, three things keep this out of merge review:

1. **Head CI is red**: the unit-test job fails with an 'Event loop is closed' cascade (run 31836678299), which suggests the rebased durable-bridge helpers don't yet match #1394's async lifecycle. A red rebase can't clear the 'resolve at head' bar.
2. **The scope ask was not adopted**: the maintainer's 07-26/08-04/08-10 requests were to split out (or, per the 08-10 rescope, drop) the cross-account continuity-rebind/CAS/fencing machinery and land the compact.py-scoped core recovery. The branch still carries the full ~2.7k-line diff rewriting `durable_bridge_repository.py`/`durable_bridge_coordinator.py`. The previously blocking threads were largely outdated by the force-push rather than rebutted at the new head.
3. **Authorship**: @Iweisc has been inactive since 07-26; the 08-10 comment set close-and-rescope as this round's default in that case.

@Soju06 — owner call: (a) accept the third-party continuation, conditioned on green CI plus the rescope/split you asked for, or (b) proceed with the announced plan to close this and carry the rescoped core recovery (previous_response_id-only activation, existing account-neutral fresh-replay gates, quota-caused-exclusion gating) into a fresh PR with attribution to @Iweisc and @Komzpa.

Soju06:
Executing the plan announced in the 08-10 review: closing this PR and carrying the rescoped core forward in a fresh PR with attribution to @Iweisc (and @Komzpa for the continuation work). Rationale, for the record: the repeated ask was to split out or drop the cross-account continuity-rebind/CAS/fencing machinery and land only the compact.py-scoped core recovery — the current head still carries the full ~2.7k-line surface including rewrites of durable_bridge_repository.py/durable_bridge_coordinator.py on top of #1394's new shape, and its CI is red with event-loop lifecycle errors, so the rebase isn't sound either. The rescoped takeover will activate recovery only for previous_response_id-pinned self-contained histories, reuse the existing account-neutral fresh-replay gates, and gate on quota-caused owner exclusion. It will link back here and credit both of you in the commit trailers.

PR #1217, opened 2026-07-11
fix(accounts): sync free-to-paid plan upgrades

luawl:
## Summary

- trust a workspace-less Free ? recognized-paid transition from the per-account usage payload
- let the forced usage refresh behind **Force probe** persist Free ? Plus without requiring logout/relogin
- retain the existing guards against paid ? Free, unrecognized-plan rewrites, and conflicting workspace identity
- update the active usage-refresh OpenSpec change and convert the former rejection test into a Force-probe-path regression

## Root cause

The usage payload already reports the new plan and `_sync_identity_metadata()` can persist it. `_payload_mismatches_account_slot()` allowed paid ? paid changes but classified `stored_plan_type=free payload_plan_type=plus` as an identity mismatch, so Force probe discarded a legitimate upgrade.

## Scope

The behavior change is one condition in the existing plan-trust guard. The remaining runtime diff is explanatory comment reflow; all other changes are focused tests and OpenSpec synchronization.

## Validation

- `uv run pytest tests/unit/test_usage_updater.py -k "plan or workspace" -q` (10 passed)
- `uv run pytest tests/unit/test_accounts_service_probe.py -q` (11 passed)
- remaining usage-updater tests (81 passed)
- Ruff format/check
- `uv run ty check app/modules/usage/updater.py tests/unit/test_usage_updater.py`
- strict OpenSpec change validation
- strict validation of all 31 specs
- `git diff --check`

Local note: one unrelated Windows-only monthly-row freshness test fails identically on `upstream/main` because its naive local `datetime.now()` fixture is not fresh against UTC; it is not touched by this diff.

Fixes #1215

Issue #1340, opened 2026-07-15
Simplicity backlog: settings-surface reduction (164 → ~110 fields) & deferred follow-ups


Soju06:
# Simplicity backlog: settings-surface reduction & deferred follow-ups

Tracking issue for the simplicity effort (audit 2026-07-15). The effort's PR stack: #1336 (principles & gates), #1337 (docs site + README diet), #1338 (CI budgets), #1339 (dashboard progressive disclosure). This issue tracks everything the audit found that was deliberately **not** included in those PRs.

## Why

The settings surface is **164 env-settable fields** (`app/core/config/settings.py`), of which only ~8 are essential; zero are required for first boot. The audit ranked reductions by (fields removed × user-facing confusion) / risk. Full audit table in the collapsed section below. Net effect if all land: **164 → ~110 fields**.

Every item below is an operator-contract/compatibility change → **needs its own OpenSpec change folder** per merge gates. `extra="ignore"` means removals never crash existing deployments (stale env lines are dropped); deprecation policy: release note + one release of a startup WARN for removed `CODEX_LB_*` names.

## Settings-surface reduction (ranked, from the audit)

| # | Candidate | Fields | Action | Risk |
|---|-----------|--------|--------|------|
| 1 | ~~Fix `.env.example` default drift (5 values incl. `STREAM_IDLE_TIMEOUT_SECONDS=300` vs default 7200)~~ | 5 corrected | **shipped in PR B** (drifted lines deleted) | none |
| 2 | OAuth protocol constants → module constants (`oauth_client_id`, `oauth_originator`, `oauth_scope`, `oauth_redirect_uri`, `oauth_callback_port`, `auth_base_url`) | 6 | hardcode; keep `oauth_callback_host`/`oauth_timeout_seconds` as advanced | low |
| 3 | Auth guardian 8 → 1 (keep `auth_guardian_enabled`, hardcode interval/age/batch/concurrency/jitter/backoff) | 7 | hardcode | low |
| 4 | Debug log booleans 6 → 1 (`CODEX_LB_TRACE` CSV) | 5 | merge | low |
| 5 | DB pool tuning 6 → 2 (keep `database_pool_size` + one overflow; derive background pool) | 4 | derive/hardcode | low-med (Postgres HA operators may pin) |
| 6 | Soft-drain/probe thresholds → constants (keep the two `*_enabled` flags) | 6 | hardcode or dashboard Advanced group | med — check `deterministic-failover` capability first |
[...]

Issue #1215, opened 2026-07-11
bug(accounts): <one-line summary>


Lok3rn3t:
### Pre-flight checklist

- [x] I searched existing issues for similar reports.
- [x] I'm on the latest patch release of my minor version.

### codex-lb version

1.21.0-beta.2

### Affected account plan(s)

Mixed pool

### Symptom category

Other

### What does the dashboard / logs show?

```markdown
Here's the problem: when I buy a subscription—in my case, Plus my account information doesn't update, and I have to logout and back in, even `force probe` doesn't help, and neither does pausing and unpausing account's.
```

luawl:
I can confirm the current code has this gap. `Force probe` refreshes credentials only when they are already due (`ensure_fresh(..., force=False)`), sends the model probe, and then refreshes usage. The stored `plan_type` comes from authentication token claims, so a still-valid pre-upgrade token keeps the old plan; pause/unpause and usage refresh do not replace those claims. Logging in again works because it supplies a new identity token.

Could you confirm the exact old ? new plan shown and whether it updates after the next natural token refresh? If available, please include the nearby `account_token_refresh_*` log lines with IDs/tokens removed. A fix should refresh identity metadata deliberately without making every health probe unnecessarily rotate credentials.

luawl:
Correction after tracing the complete usage-refresh path: the usage payload *does* carry `plan_type`, and Force probe already tries to persist it. The concrete blocker is `_payload_mismatches_account_slot()`: for a workspace-less account it currently permits paid?paid changes (Plus?Pro) but deliberately rejects `free`?paid, so a Free?Plus upgrade is discarded with `Usage refresh payload identity mismatch; skipping account mutation`.

That matches this report more directly than token freshness. The requested log line/workspace information would still confirm the exact path, but this is now a bounded guard + regression fix rather than a need to force token rotation.

Lok3rn3t:
The plan changed from Free to Plus.
After logging out and signing in again, the Plus plan was displayed correctly on all accounts.
I’m unable to provide the account_token_refresh_* log entries.

Issue #1086, opened 2026-06-25
bug: Plus → Pro upgrade not reflected when workspace_id is null


ghgr:
### Pre-flight checklist

- [x] I searched existing issues and confirmed this is not a duplicate.
- [x] I am running a supported codex-lb version (see Releases).
- [x] I can reproduce the problem with the steps below.

### codex-lb version

1.20.1

### Deployment method

Docker (ghcr.io/Soju06/codex-lb)

### Client used against codex-lb

ОpenCode

### ChatGPT account plan(s) involved

Plus, Pro

### What happened?

After upgrading a ChatGPT account from Plus to Pro, codex-lb still shows the account as Plus.

The usage refresh payload contains the correct new plan, but the account update is skipped:

```text
Usage refresh payload identity mismatch; skipping account mutation
stored_workspace_id=None payload_workspace_id=None
stored_plan_type=plus payload_plan_type=pro
stored_seat_type=None payload_seat_type=None
```

This looks like the identity mismatch check is treating `plan_type` changes as a mismatch when there is no `workspace_id`, which prevents legitimate plan upgrades from being saved.

### What did you expect to happen?

Expected: account should update from `plus` to `pro`.

Actual: account remains `plus` until manually edited/reimported.

### Steps to reproduce

```markdown
1. Add/import a Plus account where `workspace_id` is null.
2. Upgrade that same ChatGPT account to Pro.
3. Trigger a usage refresh in codex-lb.
4. Observe the log:


Usage refresh payload identity mismatch; skipping account mutation
stored_workspace_id=None payload_workspace_id=None
stored_plan_type=plus payload_plan_type=pro
stored_seat_type=None payload_seat_type=None


5. Check the account in codex-lb. It still shows `plus`, even though the refresh payload reports `pro`.

Minimal config: Docker install with persistent codex-lb data volume. No custom workspace config.
```

### Relevant logs

```shell

```

### Configuration / environment

```yaml

```

PR #1491, opened 2026-07-25
fix(usage): confirm workspace-less paid to free plan downgrades

### LINKED ISSUE #1456: bug(usage): workspace-less Plus → Free entitlement changes are discarded
### Pre-flight checklist

### codex-lb version

1.21.0

### Deployment method

Docker Compose

Codex CLI

Free

### Model(s) involved

  Codex CLI / Codex backend API traffic. The defect itself occurs in codex-lb’s background usage-refresh job, so it is not client-specific.

### What happened?

 A workspace-less ChatGPT Plus account expired to Free, but codex-lb continued to store and display it as plus.

  The upstream usage-refresh payload correctly reported free once per minute, but codex-lb rejected every update as an identity mismatch:

  Usage refresh payload identity mismatch; skipping account mutation
  account_id=<redacted> stored_workspace_id=None payload_workspace_id=None
  stored_plan_type=plus payload_plan_type=free
  stored_seat_type=None payload_seat_type=None

  This left the dashboard/database with a stale paid-plan label and misleading quota/rate-limit state. Routing then treated the account inconsistently:

  Proxy preferred account unavailable ... error_code=no_plan_support_for_model
  error=No accounts with a plan supporting model 'gpt-5.6-terra'

 codex-lb should persist a verified recognized plan change from plus to free when the usage payload is fetched using that account’s token and there is no
  conflicting workspace ID.

  The dashboard, quota data, and routing should then reflect the Free entitlement consistently.

### Steps to reproduce

```markdown
Steps to reproduce

  1. Deploy codex-lb v1.21.0.
  2. Import a workspace-less ChatGPT Plus account (workspace_id=None).
  3. Let that account’s Plus subscription expire so its upstream usage payload reports plan_type=free.
  4. Wait for the background usage refresh, or trigger a force probe.
  5. Observe the repeated Usage refresh payload identity mismatch logs.
  6. Check the account: it remains stored as plus rather than being updated to free.
```

### Relevant logs

```shell
Usage refresh payload identity mismatch; skipping account mutation
  account_id=<redacted> stored_workspace_id=None payload_workspace_id=None
  stored_plan_type=plus payload_plan_type=free
  stored_seat_type=None payload_seat_type=None

  Proxy preferred account unavailable
  error_code=no_plan_support_for_model
  error=No accounts with a plan supporting model 'gpt-5.6-terra'
```

```yaml
services:
    codex-lb:
      image: ghcr.io/soju06/codex-lb:1.21.0
      ports:
        - "127.0.0.1:1455:1455"
        - "127.0.0.1:2455:2455"
      volumes:
        - codex-lb-data:/var/lib/codex-lb
```

### Additional context

  Related issue: #1086
  Related fix: PR #1217

  PR #1217 fixed Free→paid and paid→paid plan synchronization, but deliberately retained the workspace-less paid→Free rejection. Subscription expiration is
  a normal entitlement transition, so this leaves expired accounts with stale paid labels.

  If accepting one Free payload is considered unsafe, a conservative solution could require two consecutive fresh per-account-token free payloads before
  mutating the stored plan. Workspace-ID conflicts should remain rejected.

mrnaturalmonopoly-prog:
## Summary

A workspace-less account whose paid ChatGPT subscription expires is stored and displayed as paid forever, because usage refresh discards every `free` payload as an account-slot identity mismatch. This PR makes the paid -> `free` downgrade *confirmable* — it is persisted once two consecutive workspace-less refreshes agree — instead of either trusting one sample or rejecting it permanently.

## Type of change

- [ ] **Breaking change**

Linked issue: Fixes #1456

## OpenSpec

Change directory: `openspec/changes/confirm-workspaceless-plan-downgrade/`

This deliberately **modifies** an existing requirement rather than claiming to match it. The archived `sync-paid-plan-upgrade-without-workspace` change (PR #1217, issues #1086 / #1215) recorded the scenario *"Free downgrade without a workspace is rejected"* on purpose, because a single `free` payload is also the signature of a degraded or wrong-identity usage response. That scenario is replaced here with a confirmation-gated one, and the reasoning is recorded in the change proposal so the earlier decision is not silently reversed.

## Root cause

`_payload_mismatches_account_slot` (`app/modules/usage/updater.py`) whitelists a workspace-less `plan_type` change only when the payload reports a *recognized paid* plan. `free` -> `plus` and `plus` -> `pro` therefore pass, while `plus` -> `free` falls through to the mismatch branch on every refresh:

```text
Usage refresh payload identity mismatch; skipping account mutation
stored_workspace_id=None payload_workspace_id=None
stored_plan_type=plus payload_plan_type=free
```

The stale label is not cosmetic — plan-based routing keeps trusting it and then contradicts itself for the same account:

```text
Proxy preferred account unavailable error_code=no_plan_support_for_model
error=No accounts with a plan supporting model 'gpt-5.6-terra'
```

## Changes

- **Confirmation-gated downgrade** (`app/modules/usage/updater.py`): the first workspace-less `free` observation for an account is recorded without mutating the stored plan and without writing the usage sample; the second consecutive one persists the downgrade. This is the conservative approach suggested by the reporter in #1456.
- **Pending state is reset by any paid payload**, so a transient `free` blip never accumulates toward a downgrade.
- **Confirmation is scoped to `free` only.** An unrecognized plan value is never confirmable, so a degraded response reporting garbage still cannot rewrite a stored plan no matter how often it repeats.
- **The differing-`workspace_id` guard stays unconditional** — a payload reporting another workspace's slot is never trusted, regardless of repetition. Covered by an explicit regression test.
- **Evidence is durable and cluster-coherent** (`app/modules/usage/plan_downgrade_observations.py`): one row per account in the new `account_plan_downgrade_observations` table, advanced by a single atomic `INSERT .. ON CONFLICT DO UPDATE .. RETURNING` (mirroring `account_refresh_claims`), so replicas sharing a database observe and advance one sequence, and a restart never drops or short-circuits a pending confirmation.
- **Evidence is keyed to the credential lineage, not token bytes.** The stored `credential_fingerprint` is a non-reversible HMAC under a fixed, public, versioned salt over the account's stable seat identity (`chatgpt_account_id`, `chatgpt_user_id`, `email`, `codex_installation_id`). Routine token rotation — which rewrites every token ciphertext on each successful refresh — cannot reset a pending count, and since nothing is decrypted there is no decrypt-failure path at all.
- **Credential replacement discards pending evidence at the replacement seam.** Every in-place re-import or reauthentication flows through `AccountsRepository._apply_account_replacement`, which deletes the account's pending evidence in the same transaction that applies the fresh material (SAVEPOINT-guarded so a not-yet-migrated database degrades to a warning instead of failing the import). Delete + re-import is covered by `ondelete="CASCADE"`.
- **The paid-path clear is gated on a cheap existence check**: a primary-key read decides, and the writer lock + `DELETE` + `COMMIT` run only when a pending row exists, so healthy paid accounts stay read-only on the refresh hot path.
- **The new store, probe, repository, and migration round-trip tests run on PostgreSQL in CI** via `POSTGRES_PYTEST_TARGETS`, with explicit `String`/`DateTime` bind types on the raw observe statement.

## Schema change and migration

**This PR adds one Alembic revision**: `20260726_000000_add_account_plan_downgrade_observations`, re-parented onto `20260725_000000_add_http_bridge_pending_tool_calls` (main gained that revision after this branch forked, so the merged tree keeps a single head). It is DDL-only — a guarded, idempotent create (and drop on downgrade) of the per-account evidence table — with no data backfill. The row holds no secrets: a count, a plan value, timestamps, and the non-reversible seat-identity digest, stored outside the encrypted token columns.

## Simplicity

- [x] New feature defaults to **off** or works with **zero config** — behavior is automatic, no opt-in.
- New setting(s) and why each can't be a default: **none.** The confirmation threshold is the hardcoded constant `_FREE_PLAN_DOWNGRADE_CONFIRMATIONS = 2`, deliberately not a `CODEX_LB_*` setting: two observations is the minimum that distinguishes a real expiry from a single degraded sample, and operators gain nothing from tuning it (respects the settings-surface ratchet in #1340).
- [x] README sections / `.env.example` / dashboard nav within budget — the only README change is two rows in the budget-exempt all-contributors grid.

## Test plan

All numbers measured on a local merge of current main (CI-equivalent environment, `uv sync --dev`, Python 3.13):

```text
.venv/bin/python -m pytest tests/unit -q
# 5033 passed, 65 skipped

.venv/bin/python -m pytest tests/integration/test_accounts_api_probe.py \
    tests/integration/test_accounts_api.py \
    tests/integration/test_accounts_api_extended.py \
    tests/integration/test_live_usage_ingest.py \
    tests/integration/test_repositories.py \
    tests/integration/test_plan_downgrade_observation_store.py -q
# 123 passed

# Full POSTGRES_PYTEST_TARGETS allowlist against a local PostgreSQL 16.2
# (same major as the CI service):
# 100 passed (before the migration round-trip test joined the list)

.venv/bin/python -m pytest \
    "tests/integration/test_migrations.py::test_account_plan_downgrade_observations_migration_upgrade_and_downgrade" -q
# 1 passed on SQLite and, via CODEX_LB_TEST_DATABASE_URL, under PostgreSQL 16.2

# codex-lb-db upgrade head + check (SQLite and PostgreSQL):
# migration_policy=ok, schema_drift=none, single head on
# 20260726_000000_add_account_plan_downgrade_observations

.venv/bin/ruff check app/ tests/            # All checks passed!
.venv/bin/ruff format --check               # all files already formatted
.venv/bin/ty check                          # All checks passed!
.venv/bin/python scripts/check_proxy_architecture.py         # passed
.venv/bin/python .github/scripts/check_simplicity_budgets.py # all budgets OK
git diff --check                            # no whitespace errors
```

Each fix is mutation-verified: reverting the lineage digest to token material fails the rotation tests; removing the replacement-seam discard fails the two repository tests and the product-path re-import test; removing the existence gate fails the statement-capture test.

New coverage — unit:

- token rotation between two `free` observations still confirms the downgrade (the previously missing test);
- the fingerprint is unchanged by rotation and changed by seat-identity rebinding, exposes no identity or token material, and distinguishes missing from empty identity fields;
- a rebound seat identity restarts the confirmation count (defense in depth for the digest CASE);
- two-observation downgrade, paid-payload reset, unrecognized plans never confirming, per-account scoping, replica-sequence sharing — all preserved from the reviewed head, with the replica tests now stating exactly what the in-process harness proves.

New coverage — integration (all in the postgres allowlist):

- `tests/integration/test_migrations.py::test_account_plan_downgrade_observations_migration_upgrade_and_downgrade` — Alembic round-trip for the new revision: parent -> revision -> downgrade -> guarded re-upgrade over a pre-merge table -> head stays single-headed (Codex review P1);
- `tests/integration/test_plan_downgrade_observation_store.py`: the raw upsert (insert / increment / lineage restart / `RETURNING`), naive-UTC datetime binds round-tripped through the real driver, genuinely concurrent observations each advancing the count, the read-only clear gate (statement capture), the replacement discard's transaction atomicity, and the SAVEPOINT missing-schema degrade;
- `tests/integration/test_repositories.py`: `replace_reauthorized` and in-place `upsert_account_slot` discard pending evidence;
- `tests/integration/test_accounts_api_probe.py`: `test_reimport_clears_pending_downgrade_evidence` — the dashboard re-import flow (overwrite-on-import enabled) resets evidence and converges only on the new credential's own second observation, asserted through `/api/accounts`.

**Not validated locally:** `openspec validate --specs` (npm registry unreachable from this environment; CI's OpenSpec job is the authority) and any live upstream probe of a genuinely lapsed subscription — the failing path is covered hermetically through the same `fetch_usage` seam the existing upgrade tests use.

## Screenshots / output

No dashboard rendering change — this PR corrects the *data* the existing Accounts UI displays. The externally visible change is the `/api/accounts` payload, asserted directly in the integration tests above:

```text
# workspace-less account, stored plan_type=plus, upstream now reports free

first  POST /api/accounts/{id}/probe  ->  GET /api/accounts  ->  planType == "plus"
second POST /api/accounts/{id}/probe  ->  GET /api/accounts  ->  planType == "free"
```

Log excerpt from the confirmation path (previously a single repeating `identity mismatch` warning):

```text
Usage refresh observed a workspace-less downgrade to free; awaiting confirmation
  account_id=... stored_plan_type=plus observations=1 required=2

Usage refresh confirmed a workspace-less downgrade to free; persisting plan change
  account_id=... stored_plan_type=plus observations=2
```

## Checklist

- [x] Linked the related issue above (`Fixes #1456`).
- [x] Ran the relevant lint / typecheck / test / architecture / simplicity subsets locally (see Test plan).
- [ ] If touching specs: `openspec validate --specs` passes — **could not run locally** (npm registry unreachable); deferred to CI, see Test plan.


Soju06:
Good diagnosis and the two-observation approach matches what the issue asked for — the guardrails (free-only confirmation, paid resets, workspace-conflict check untouched) are right. Two edges to close before merge:

1. **Cross-replica observation coherence.** The pending map is process-local, so in a shared-DB multi-replica deployment the observation sequence diverges per replica: replica A records `free`, replica B observes a paid payload (which should reset), then A sees `free` again and confirms despite the intervening paid evidence — and conversely, probes split across replicas never confirm. Persisting the pending observation (e.g. a column or the existing settings/state storage) keyed to the account makes the sequence cluster-coherent; single-replica behavior stays identical.

2. **Pending state survives account replacement.** Account IDs are deterministic, so delete + re-import reuses the ID while the in-memory pending entry survives; the new credential's first degraded `free` payload then counts as observation two and downgrades immediately. Clearing pending state on account deletion/replacement (or keying it to a credential incarnation) preserves the safeguard.

CI is green and the OpenSpec delta reads well — with these two addressed this is ready for the merge gates. Fixes #1456 linkage confirmed.

Soju06:
The rework lands both of my review items conceptually — the durable `account_plan_downgrade_observations` table with a single atomic `INSERT .. ON CONFLICT DO UPDATE .. RETURNING` (mirroring `account_refresh_claims`) is the right shape for cross-replica coherence, and `ondelete="CASCADE"` plus the credential-fingerprint restart covers the delete/re-import and in-place reauth cases I raised. Thank you for taking it in that direction, and the self-audit in dce7f727 was appreciated.

But the deeper review of the new design surfaced issues I need addressed before this merges:

1. **Routine token rotation resets the confirmation count.** The fingerprint is an HMAC over the *decrypted refresh token* (`plan_downgrade_observations.py`, `credential_fingerprint`), but refresh tokens rotate on every token refresh (`auth_manager.py:346-348` — `_refresh_token_material_fingerprint` exists precisely because peers rotate material). Any rotation between the two required "free" observations restarts the count at 1 via the upsert's CASE — so an account whose token-refresh cadence interleaves with usage refresh can have the #1456 downgrade postponed indefinitely. The identity you want is "same account credential lineage," not "same literal refresh token"; and there's no test for rotation-between-observations (only the deliberate re-import case).
2. **The decrypt-failure fallback reintroduces the original bug.** On `TokenEncryptor` decrypt failure the code silently HMACs the Fernet *ciphertext* — which has a random IV, so every re-encryption produces a new fingerprint and the confirmation never converges, for exactly the rotated/lost-key rows the fallback exists to serve. Fail loudly or pin those rows to a stable per-account fallback identity instead.
3. **The raw SQL never runs on PostgreSQL in CI.** Production runs Postgres, but the postgres CI job executes the fixed `POSTGRES_PYTEST_TARGETS` allowlist in the Makefile, which contains none of the new store/probe tests — the upsert template, the `text()` datetime binds through asyncpg, and the missing-schema degrade path are exercised only on SQLite. Please add the new tests to that allowlist.
4. **Hot-path write amplification.** Every workspace-less refresh reporting a *paid* plan now awaits `_clear_workspace_less_free_plan_observations()` → `DELETE` + `COMMIT` inside `sqlite_writer_section` (a process-global writer lock) on every refresh cycle for every healthy account, though pending rows almost never exist. Gate it on a cheap existence check first.
5. **Test-evidence notes:** the two-replica unit tests share a single in-process `InMemoryPlanDowngradeObservationStore` via the autouse fixture in `tests/conftest.py`, so they demonstrate shared-object behavior rather than durable cross-process coherence — the DB-store integration test is the real evidence, worth saying so; and the barrier interleave test only covers the in-memory store, so the DB upsert's atomicity is asserted, not exercised.
6. **Please update the PR body.** It still says "No schema change and no migration: confirmation state is process-local," which stopped being true at 229b0fdd (new table + migration `20260726_000000`). The squash commit inherits the body, and I don't want a schema migration hidden from the release notes. While you're there, mention in the spec's security notes that `credential_fingerprint` is a stable HMAC(fixed public salt, decrypted refresh token) stored outside the encrypted columns.

On process: CI hasn't run on dce7f727 yet (fork approval gate) — I've approved the workflow runs. The migration itself is fine (guarded create-table only, single head on `20260724_..._rollups`). This is a `fix:` so it stays soak-exempt; with the six items above addressed and CI green on the final head, it merges.

mrunnaturalmonopoly-svg:
Quick note on authorship: @mrnaturalmonopoly-prog is me — I'll be completing the rest of this PR from this account. It has write access to the fork, so the work continues on this same branch and PR; the three commits just pushed (b419225, 40205ab, fb9a2bc) are the review response, with the item-by-item reply following in the next comment..

mrunnaturalmonopoly-svg:
All six items are addressed, pushed as new commits on top of the reviewed head (no force-push): `b419225` (migration re-parent), `40205ab` (the fix), `fb9a2bc` (OpenSpec docs), plus the base sync and the contributors commit noted at the end. Point by point:

**1 — Routine token rotation resets the confirmation count.** Confirmed against `rotate_tokens`: every successful refresh rewrites all three token ciphertexts, so the HMAC-over-decrypted-refresh-token digest restarted the upsert's CASE at 1 whenever rotation interleaved with usage refresh. I checked every existing column for a better anchor first — nothing both survives rotation and changes on in-place reauth — so the two jobs are now separate mechanisms (`40205ab`):

- The fingerprint digests **stable seat identity** (`chatgpt_account_id`, `chatgpt_user_id`, `email`, `codex_installation_id`) under a fixed public salt, bumped to `v2` so values from the old scheme can never read as an agreeing lineage. Rotation cannot move it by construction. The missing test exists at both levels: `test_token_rotation_between_observations_still_confirms_downgrade` (updater; the old replaced-credential test inverted — same token swap, now asserting the count **survives**) and `test_credential_fingerprint_is_unchanged_by_token_rotation` (store). The CASE restart stays as defense in depth for seat-identity rebinding (`test_rebound_seat_identity_restarts_free_downgrade_confirmation`).
- **Credential replacement resets evidence at the replacement seam itself**: every in-place re-import/reauth flows through `AccountsRepository._apply_account_replacement`, which discards the account's pending evidence in the same transaction that applies the fresh material (SAVEPOINT-guarded so an unmigrated database degrades to a warning instead of poisoning the caller's transaction on Postgres). Delete + re-import stays covered by `ondelete="CASCADE"`. Covered at the repository seam (`test_replace_reauthorized_discards_pending_downgrade_evidence`, `test_upsert_account_slot_discards_pending_downgrade_evidence_on_reimport`) and end to end through the dashboard flow (`test_reimport_clears_pending_downgrade_evidence`, with overwrite-on-import enabled; under the default import-without-overwrite a re-import creates a separate row and replaces no credential, so the original row's evidence rightly stands — the test says so).

**2 — Decrypt-failure fallback reintroduces the original bug.** Resolved *structurally* rather than by either of your literal suggestions, and I want that visible rather than slipped in: with no token material in the digest there is nothing to decrypt, so the ciphertext-HMAC fallback and `_shared_encryptor` are deleted outright — the failure mode ceases to exist instead of being handled. This is the "stable per-account fallback identity" option taken to its conclusion: the stable identity became the whole fingerprint. It also simplifies the security note item 6 asked for (below). If you'd rather keep a token-bound digest with a loud failure, say the word and I'll rework it.

**3 — The raw SQL never runs on PostgreSQL in CI.** `POSTGRES_PYTEST_TARGETS` now includes a dedicated `tests/integration/test_plan_downgrade_observation_store.py` (upsert insert/increment/restart through `ON CONFLICT` + `RETURNING`, naive-UTC datetime binds written and read back, genuinely concurrent observes, the gated clear, and the SAVEPOINT missing-schema degrade) plus the four probe tests and the two repository-seam tests. The observe statement now carries explicit `bindparams` types (`String`/`DateTime`), mirroring `refresh_claims`' `bindparam("ttl", Float)`. Measured, not asserted: I ran the **entire** allowlist — all pre-existing entries plus the new ones, 100 tests — against a local PostgreSQL 16.2 (same major as the CI service) on a merge of current main: 100 passed. `migration-check` and `migration-check-postgres` equivalents pass on the same merged tree (single head, no drift).

**4 — Hot-path write amplification.** `PlanDowngradeObservationStore.clear()` now reads first (primary-key SELECT, no writer section) and takes the writer lock + `DELETE` + `COMMIT` only when a row exists. Race semantics are unchanged and documented in the method: a row inserted concurrently after the read survives until the next paid observation, exactly as it would have survived an unconditional `DELETE` that committed before that insert. `test_clear_skips_the_write_path_when_no_evidence_exists` pins it by capturing cursor statements — the empty case issues SELECT only. (The capture listens at the `Engine` class level deliberately: the store runs on the lazily created background engine, so an instance-level listener goes blind once another test initializes it.)

**5 — Test evidence.** `_replica_module` and the two-replica tests now say plainly that the unit harness shares one in-process `InMemoryPlanDowngradeObservationStore` via the autouse fixture — they demonstrate the shared-sequence logic, not durable cross-process coherence — and point to the database-backed evidence: the new store file exercises the real upsert concurrently (each observation its own session/connection/transaction; genuinely racing on Postgres via item 3), and the probe tests remain the product-path proof. The barrier interleave test's docstring carries the same caveat.

**6 — PR body.** A corrected, ready-to-paste body is attached to this comment — it discloses the schema change and migration (including the re-parent in `b419225`: main gained `20260725_..._add_http_bridge_pending_tool_calls` after this branch forked, so the merge ref had two Alembic heads), drops the now-false "process-local, no migration" claims, and carries the measured test evidence. Neither of my accounts can edit the PR description (that needs the base repo or an editable-by-maintainers grant), so please paste the attached body over the current one — or take it as the squash message directly. The spec's security notes (`fb9a2bc`) now state exactly what `credential_fingerprint` is under the shipped design: a non-reversible HMAC under a fixed, public, versioned salt over stable seat-identity fields, stored outside the encrypted columns, derived without decryption — there is no longer any digest over secret material to caveat.

**Base sync + attribution.** Right after this comment the branch gets updated with main (merge, no force-push) and a follow-up commit adds both authoring accounts to `.all-contributorsrc` and the README grid, which is what the currently failing **Contributors attribution** job wants. The two contributor files were fully regenerated on main (#1518), so the additions only apply cleanly on the synced branch — that ordering is why they are not in the commits above. Grid cells are hand-built against main's regenerated layout (the npm CLI wasn't reachable from my environment) — happy to have them regenerated if the spacing is off.

**Verified locally** on a merge of current main: unit `5033 passed, 65 skipped`; the touched integration files (probe, accounts, accounts-extended, live-usage-ingest, repositories, new store file) `123 passed`; the full postgres allowlist `100 passed` on PostgreSQL 16.2; migration checks green on SQLite and PostgreSQL; `ruff check` / `ruff format --check` / `ty check` / proxy-architecture / simplicity budgets / `git diff --check` all clean. Each fix was mutation-verified: reverting the lineage digest fails the rotation tests, removing the replacement-seam discard fails all three reauth tests, and removing the gate fails the statement-capture test.

**Not verified locally:** `openspec validate` (npm registry unreachable here, same limitation as before — CI stays the authority on the spec delta) and any live upstream probe of a genuinely lapsed subscription; the frontend/e2e/bridge suites are untouched by this change and left to CI.

CI on the new head will need your fork-workflow approval again. Once it's green this should be ready for the merge gates..
[pr-body.md](https://github.com/user-attachments/files/30513461/pr-body.md)

Soju06:
Owner decision from today's triage: proceeding with this PR under the new account (treating the mid-PR handover mrnaturalmonopoly-prog → mrunnaturalmonopoly-svg as the same contributor). Two things before the merge gates can run:

1. **Fix the PR body** — it still claims "no migration", but the branch adds the `20260726_*` table migration. Please update the body to describe the migration (and confirm the alembic parent is current single-head after today's merges).
2. The six findings from the 07-27 review round (rotation-window fingerprint reset, Fernet-ciphertext HMAC fallback, postgres CI allowlist gap, unconditional DELETE+COMMIT in the hot loop, autouse-shared in-memory store in the replica test, and the migration-claim mismatch) — please confirm each is addressed on the current head or push the fixes; the review below will check against them.

Fork CI runs are approved.

@codex review

mrnaturalmonopoly-prog:
(Posting from the author account because the description edit needs it — same person as noted above; the work otherwise continues from @mrunnaturalmonopoly-svg.)

Done, point by point:

1. **PR body fixed** — the description now discloses the `20260726_000000_add_account_plan_downgrade_observations` migration (guarded, DDL-only create/drop, no backfill) and its re-parent onto `20260725_000000_add_http_bridge_pending_tool_calls`. Single-head confirmed **after today's merges**: main gained no new revisions since `b91d9bb8` (checked against `3fe0d6f2`), `codex-lb-db upgrade head && check` reports `migration_policy=ok, schema_drift=none` on both SQLite and PostgreSQL 16.2, and a merge of this head against today's main recomputes clean.

2. **The six 07-27 findings, each confirmed on the current head `4570163`** (details in the long reply above):
   - **rotation-window fingerprint reset** → the digest covers stable seat identity, not token material; rotation cannot move it — `test_token_rotation_between_observations_still_confirms_downgrade`, `test_credential_fingerprint_is_unchanged_by_token_rotation` (`40205ab`);
   - **Fernet-ciphertext HMAC fallback** → deleted outright — nothing decrypts anymore, so the failure mode no longer exists (`40205ab`);
   - **postgres CI allowlist gap** → store, probe, repository, and now the migration round-trip tests are in `POSTGRES_PYTEST_TARGETS`; the full allowlist passes against a local PostgreSQL 16.2 (`40205ab`, `4570163`);
   - **unconditional DELETE+COMMIT in the hot loop** → `clear()` reads first and takes the writer path only when a row exists — `test_clear_skips_the_write_path_when_no_evidence_exists` (`40205ab`);
   - **autouse-shared in-memory store in the replica tests** → annotated as exactly that, with the durable evidence in `tests/integration/test_plan_downgrade_observation_store.py` (`40205ab`);
   - **migration-claim mismatch in the body** → resolved by the description edit above.

3. **Codex review P1 addressed in `4570163`** — `test_account_plan_downgrade_observations_migration_upgrade_and_downgrade` round-trips the revision through Alembic itself (parent → revision → downgrade → guarded re-upgrade over a pre-merge table → upgrade to head on the single-head path), mirroring the rollups round-trip, and joins the postgres allowlist. Verified on SQLite and under PostgreSQL 16.2; a broken-`down_revision` mutation fails it.

One process note: the Actions runs on this PR still show **awaiting approval** on our side (the `c1229a4` run from 16:34 UTC yesterday and today's re-runs are all `action_required`) — the earlier approval may not have registered, and `4570163` will need a fresh one either way..

Soju06:
Body rewrite and the point-by-point confirmation of the six findings both check out against the current head — compliance accepted, and the 08-03 re-parent was correct at push time.

Unfortunately the head moved under you again: #1394 merged this morning and brought `20260730_000000_add_http_bridge_recovery_attempts` plus the no-op merge revision `20260803_000000_merge_http_bridge_recovery_and_capability_lineage_heads`, so main's Alembic head is now the 08-03 merge. With `down_revision = 20260731_000000_add_capability_lineage_markers` this branch re-creates two heads on merge and the startup migration check will fail. Please re-parent onto the 08-03 merge revision (and update the round-trip test's parent accordingly), and resolve the two remaining conflicts — `.all-contributorsrc` and `tests/integration/test_migrations.py`, both mechanical. After that this is at the merge gates.

Soju06 (approved):
Triage re-verification passed: the 08-04 head is a clean re-parent with the six-finding confirmation intact; substantive CI green. Approving for merge.

Issue #1530, opened 2026-07-30
bug: client-fault upstream 400 penalizes the serving account and 502s unrelated hard-sticky sessions (plus an unclassified 'tool search call' missing-tool-output variant)


yeongjun-cigro:
## Summary

A client-fault upstream 400 is charged against the upstream account that happened to serve it. Because that penalty feeds `error_count` -> exponential error backoff -> `in_error_backoff`, and because a hard Codex-session sticky request is narrowed to exactly one candidate account where the backoff rescue can never fire, **one tenant's poisoned conversation degrades shared account health and 502s unrelated tenants**.

The trigger we observed is the missing-tool-output 400, and it comes in a fourth wording that `_is_missing_tool_output_error` does not know about:

```json
{"type":"error","status":400,
 "error":{"type":"invalid_request_error",
          "message":"No tool output found for tool search call call_XXXX",
          "param":"input"}}
```

Two separable defects, described below. Verified against `origin/main` @ `b91d9bb8`.

---

## Defect A — the `tool search call` wording is unclassified

`_MISSING_TOOL_OUTPUT_MESSAGE_PREFIXES` (`app/modules/proxy/service.py:2105`) enumerates exactly three wordings:

```python
_MISSING_TOOL_OUTPUT_MESSAGE_PREFIXES = (
    "no tool output found for function call call_",
    "no tool output found for custom tool call call_",
    "no tool output found for apply patch call call_",
)
```

so `_is_missing_tool_output_error` (`service.py:2112`) returns `False` for the tool-search wording. Consequences at each consumer of that classifier:

- `app/modules/proxy/_service/http_bridge/upstream_events.py:670` and `app/modules/proxy/_service/websocket/mixin.py:605` use `is_missing_tool_output_event` as `prefer_previous_response_not_found` when matching an **anonymous** upstream error event (no `response.id`) to a pending request. On a multiplexed bridge/WS session the unclassified 400 falls through to the generic matcher (`_match_websocket_request_state_for_anonymous_event`, `websocket/helpers.py:1414`), so it can be attributed to a pending request that did not cause it.
[...]

PR #1531, opened 2026-07-30
fix(proxy): classify tool-search missing-tool-output rejections

yeongjun-cigro:
## Summary

`_MISSING_TOOL_OUTPUT_MESSAGE_PREFIXES` (`app/modules/proxy/service.py:2105`) enumerates only the `function call`, `custom tool call`, and `apply patch call` wordings, so `_is_missing_tool_output_error` returns `False` for upstream's fourth wording:

```json
{"type":"error","status":400,
 "error":{"type":"invalid_request_error",
          "message":"No tool output found for tool search call call_XXXX",
          "param":"input"}}
```

That unclassified 400 skips every consumer of the classifier: the raw upstream message and call id leak downstream instead of the `stream_incomplete` continuity rewrite (`_service/websocket/helpers.py:1041`/`:1303`, `_service/streaming/helpers.py:511`), `_record_continuity_fail_closed(reason="missing_tool_output")` is never recorded, and on a multiplexed HTTP bridge or WebSocket session the anonymous error event loses its `prefer_previous_response_not_found` matching (`_service/http_bridge/upstream_events.py:670`, `_service/websocket/mixin.py:605`) and can be attributed to a pending request that did not cause it.

Same enumeration gap as #1168. That fix made its sites 1 and 2 table-driven via `_PENDING_TOOL_CALL_OUTPUT_ITEM_TYPE_BY_CALL_TYPE` (`_service/support.py:60`); only site 3, the classifier's fixed wording list, is still hard-coded.

`tool_search_call` / `tool_search_output` are already modelled in-repo (`app/modules/proxy/replay_safety.py:20-22`, shapes fixed by `tests/unit/test_replay_safety.py:187-203`), so the item type is real — only the wording list was stale.

Refs #1168, #1530.

## Changes

- add `no tool output found for tool search call call_` to `_MISSING_TOOL_OUTPUT_MESSAGE_PREFIXES`, so the existing masking, continuity fail-closed recording, and anonymous-event matching engage for the tool-search wording exactly as they do for the other three;
- keep the hosted `web search call` wording unclassified (a `web_search_call` is executed upstream and has no `call_id`-addressed client output) and pin that with the spec and the existing negative test case;
- update `openspec/specs/responses-api-compat/spec.md` — "Missing-tool-output classification covers all tool call variants" — with the fourth wording, the hosted-web-search exclusion, and a bridge-masking scenario;
- add the OpenSpec change `classify-tool-search-missing-tool-output`;
- regression coverage: the classifier parametrisation gains the tool-search wording, and `tests/unit/test_proxy_http_bridge.py` gains `test_process_http_bridge_upstream_text_masks_missing_tool_search_output`, mirroring the existing custom-tool-call masking test (rewritten to `stream_incomplete`, `error_http_status_override == 502`, reconnect requested, call id absent from the downstream payload).

Deliberately **not** changed: `_PENDING_TOOL_CALL_OUTPUT_ITEM_TYPE_BY_CALL_TYPE` is left alone. `tool_search_output` carries a `tools` list rather than the `{"output": "<text>"}` field that `_synthetic_interrupted_function_call_output` (`_service/response_create.py:363`) emits for every table entry, so extending the synthesiser would mean asserting an upstream payload shape this repo cannot verify — and a wrong shape turns a maskable 400 into a different, unmaskable one. It would also silently pull `tool_search_call` into `_TTFT_OUTPUT_ITEM_TYPES` (`_service/support.py:67`). #1530 raises the shape question (synthesise `{"tools": []}` vs. drop the orphaned call, which `replay_safety.py` already proves upstream accepts) for maintainer decision.

## Validation

- `tests/unit` + `tests/test_request_logs_options_api.py`: 5013 passed, 65 skipped;
- affected focused suites (`test_proxy_utils.py`, `test_proxy_http_bridge.py`, `test_replay_safety.py`): 1412 passed;
- `uv run ruff check .` passed; `uv run ruff format --check .` passed (847 files);
- `uv run ty check` passed;
- `python scripts/check_proxy_architecture.py` passed;
- `openspec validate classify-tool-search-missing-tool-output --strict` valid; `openspec validate --specs` 48/48 passed;
- `git diff --check` passed.

Not run locally: integration/e2e/PostgreSQL/Helm/frontend jobs. Hosted CI is authoritative for those.

### DIFF
--- a/app/modules/proxy/service.py
+++ b/app/modules/proxy/service.py
@@ -2106,6 +2106,7 @@ _MISSING_TOOL_OUTPUT_MESSAGE_PREFIXES = (
     "no tool output found for function call call_",
     "no tool output found for custom tool call call_",
     "no tool output found for apply patch call call_",
+    "no tool output found for tool search call call_",
 )
 
 
--- a/openspec/changes/classify-tool-search-missing-tool-output/proposal.md
+++ b/openspec/changes/classify-tool-search-missing-tool-output/proposal.md
@@ -1,0 +1,63 @@
+# Classify the tool-search missing-tool-output rejection
+
+## Why
+
+Upstream rejects an anchored Responses request whose input contains a
+`tool_search_call` without its matching `tool_search_output` with:
+
+```json
+{"type":"error","status":400,
+ "error":{"type":"invalid_request_error",
+          "message":"No tool output found for tool search call call_xxx",
+          "param":"input"}}
+```
+
+`_MISSING_TOOL_OUTPUT_MESSAGE_PREFIXES` in `app/modules/proxy/service.py`
+enumerates only the `function call`, `custom tool call`, and `apply patch call`
+wordings, so `_is_missing_tool_output_error` returns `False` for the tool-search
+wording. The raw upstream 400 is then forwarded verbatim to the client, the
+continuity fail-closed counter is never recorded, and on a multiplexed HTTP
+bridge or WebSocket session the anonymous error event is no longer preferentially
+matched to the request that carries `previous_response_id`, so it can be
+attributed to an unrelated pending request on the same socket.
+
+This is the same enumeration gap #1168 fixed for the custom-tool and
+apply-patch wordings; `tool_search_call` / `tool_search_output` are already
+modelled elsewhere in the proxy (`app/modules/proxy/replay_safety.py`), so the
+item type is real, only the classifier list is stale.
+
+## What Changes
+
+- Add the `No tool output found for tool search call call_` prefix to the
+  missing-tool-output classifier so the existing masking, continuity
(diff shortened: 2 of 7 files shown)


Soju06:
Reviewed against current main. The prefix addition matches the upstream tool-search wording captured in #1530, the bridge regression asserts the stream_incomplete rewrite plus call-id masking end to end, and keeping `No tool output found for web search call` unclassified is correct — `web_search_call` output is server-owned. Spec delta and base-spec update line up with the implementation.

Remaining blocker is mechanical: the Contributors attribution check fails because `yeongjun-cigro` is missing from `.all-contributorsrc` (cigro-manager has since been added on main). Add the entry and CI Required clears; the substantive suites are already green.

Issue #1505, opened 2026-07-27
■ unexpected status 502 Bad Gateway: No available accounts, url: http://127.0.0.1:2455/backend-api/codex/responses


K4leri:
### Pre-flight checklist

- [x] I searched existing issues and confirmed this is not a duplicate.
- [x] I am running a supported codex-lb version (see Releases).
- [x] I can reproduce the problem with the steps below.

### codex-lb version

1.22.0

### Deployment method

uvx (codex-lb)

### Client used against codex-lb

Codex CLI

### ChatGPT account plan(s) involved

Plus

### Summary

When ChatGPT downgrades an account from **Plus** to **Free** (e.g., payment failure, manual downgrade), `_payload_mismatches_account_slot` in `app/modules/usage/updater.py:839` blocks the usage refresh for that account because `free` is excluded from the set of trusted "recognised paid plans". This leaves the account with a stale `plan_type=plus` in the DB, causing:

1. Codex-lb continues routing Plus-gated model requests to the account
2. ChatGPT rejects those requests (the account is actually Free)
3. Error count rises → cooldown → backoff → routing unavailable
4. If all active accounts hit this simultaneously: **502 No available accounts**

### Affected Version

`1.22.0` (also present in `1.20.2-beta.1`)

### Root Cause

`app/modules/usage/updater.py:858-866`:

```python
if payload_plan_type != stored_plan_type and not (
    stored_plan_type == "unknown"
    and normalized_payload_plan_type in recognized_paid_plans
    or (
        not account.workspace_id
        and stored_plan_type in recognized_paid_plans
        and payload_plan_type in recognized_paid_plans
    )
):
    return True  # mismatch → skip refresh
```

The second clause allows paid→paid transitions (Plus→Pro, etc.) but explicitly excludes any transition where the payload reports `plan_type=free`. The comment says:

> *"A workspace-less payload that instead introduces 'free' or an unrecognized plan stays untrusted — the signature of a degraded or wrong-identity usage response"*

However, there is a legitimate scenario: an account was genuinely downgraded to Free by ChatGPT (expired subscription, payment decline). The heuristic incorrectly treats every Plus→Free payload as a "wrong-identity" signal rather than a valid downgrade.

### Symptoms

1. **Log warning** repeated every usage refresh cycle (every ~8 days by default):
   ```
   Usage refresh payload identity mismatch; skipping account mutation
   account_id=... stored_plan_type=plus payload_plan_type=free
   ```

[...]

Issue #876, opened 2026-06-02
bug: codex-lb lists gpt-5.3-codex as available even though ChatGPT-backed Codex accounts cannot use it


cowwoc:
## Summary

`codex-lb` lists `gpt-5.3-codex` as an available model for Codex clients, but attempts to use it fail immediately for ChatGPT-backed Codex accounts.

## Actual behavior

When Codex is pointed at `codex-lb` and `gpt-5.3-codex` is selected, Codex returns:

```json
{"type":"error","status":400,"error":{"type":"invalid_request_error","message":"The 'gpt-5.3-codex' model is not supported when using Codex with a ChatGPT account."}}
```

At the same time, `codex-lb` still shows `gpt-5.3-codex` as available in its model listing.

## Expected behavior

If `gpt-5.3-codex` is no longer usable for ChatGPT-backed Codex accounts, `codex-lb` should not advertise it as available for those accounts.

More specifically, the model catalog exposed through `codex-lb` should be consistent with actual routability/usability.

## Comparison

Without `codex-lb`, Codex does not list `gpt-5.3-codex` as an available model.

With `codex-lb`, `gpt-5.3-codex` is shown as available, but using it fails with the 400 error above.

## Reproduction

1. Configure Codex to use `codex-lb` as the provider/base URL.
2. Request/list available models.
3. Observe that `gpt-5.3-codex` is shown as available.
4. Select or configure `gpt-5.3-codex` as the model.
5. Run a prompt using a ChatGPT-backed Codex account.
6. Observe the 400 response:

```json
{"type":"error","status":400,"error":{"type":"invalid_request_error","message":"The 'gpt-5.3-codex' model is not supported when using Codex with a ChatGPT account."}}
```

## Notes

This looks like a stale/bootstrap catalog problem or a model-registry filtering gap:

- the model is presented as selectable/available
- but the backend/account type can no longer actually use it

A fix could be either:

- stop listing `gpt-5.3-codex` for ChatGPT-backed Codex accounts, or
- mark it unavailable/hidden when the active account pool cannot route it successfully

cowwoc:
## Updated scope

Per https://x.com/thsottiaux/status/2059650685948551384, this should **not** be treated as a request to remove `gpt-5.3-codex` unconditionally.

`gpt-5.3-codex` still appears to remain available for **API** use. The problem is narrower:

[...]

PR #1532, opened 2026-07-30
fix(proxy): keep upstream request-shape rejections account neutral

yeongjun-cigro:
## Summary

`_handle_stream_error` (`app/modules/proxy/_service/streaming/helpers.py:785`) is the only call site of `LoadBalancer.record_error` in the app, and it skips the account-health penalty only when `_is_account_neutral_error_code(code)` matches — local overload, process-network, `proxy_unavailable`, `responses_compact_input_too_large` (`_service/support.py:1307`). An upstream `invalid_request_error` is therefore charged to whichever account served the request, including the missing-tool-output 400 that upstream raises when the **client's own** input array references a tool call with no matching output.

That 400 reproduces identically on every account and on every retry, and the penalty compounds through routing:

1. `app/core/balancer/logic.py:484` — `error_count >= 3` puts the account into `min(300, 30 * 2 ** (error_count - 3))` seconds of backoff and appends it to `in_error_backoff` instead of `available`.
2. `logic.py:519` — the rescue back out of backoff needs `len(in_error_backoff) > 1 or (in_error_backoff and hard_blocked_exists)`, and `hard_blocked_exists` (`:507`) excludes accounts already in `in_error_backoff`.
3. `app/modules/proxy/_load_balancer/sticky_selection.py:356-361` — a hard Codex-session sticky request with a resolved owner is narrowed to that single account. With one candidate, `in_error_backoff` can only have length 1 and `hard_blocked_exists` is `False` by construction, so step 2 is structurally unreachable: `sticky_selection.py:436` returns `hard_affinity_saturated` and `load_balancer.py:709/721` turns it into an immediate 502 — no failover, no wait.

Net effect: one client's poisoned conversation degrades shared account health and 502s unrelated, hard-pinned sessions. Field evidence over one 24h window on a shared deployment: ~25.7k upstream 400s from a single repeated tool-call id, matched 1:1 by `Recorded transient account error code=invalid_request_error`, and ~2.2k 502s of which 72% were `hard_affinity_saturated`, ~95% of those belonging to a tenant that never sent the offending payload. Only 11 of those 1,583 reached `request_logs` — selection fails before an account is assigned, so the failure is nearly invisible on the dashboard.

Full write-up: #1530.

The invariant this asserts:

> **An error that would reproduce identically on every account must never mutate one account's health.**

Refs #1530, #1505, #876, #1168.

## Changes

- add `_is_account_neutral_request_rejection` (`_service/streaming/helpers.py`) and gate `_handle_stream_error` on it before any health mutation, so a payload rejection records no transient error, rate-limit, quota, or permanent-failure penalty. The skip is logged (`Skipped account error penalty for account-neutral request rejection …`) so the decision stays observable alongside the existing `Recorded transient account error` line;
- keep the predicate deliberately narrow — `invalid_request_error` **and** HTTP 400 whenever a status is known **and** a classified missing-tool-output message. Membership is decided by the message, never by the code alone, so the account-scoped `invalid_request_error` from #876 (`The '<model>' model is not supported when using Codex with a ChatGPT account.`, matched by `_is_account_model_unsupported_error`) keeps backing its account off;
- split `_is_missing_tool_output_error` into `_is_missing_tool_output_message` + the existing `code`/`param` gate, because `UpstreamError` (`app/core/balancer/types.py`) carries `message` but no `param`. Behaviour of the existing classifier is unchanged;
- add the `account-routing` requirement "Upstream rejections of the request payload are account neutral" plus the OpenSpec change `keep-request-shape-rejections-account-neutral`;
- regression coverage in `tests/unit/test_proxy_utils.py`: a parametrised predicate table (missing-tool-output at 400 and at unknown status → neutral; the #876 model-entitlement message, a 500, a non-`invalid_request_error` code, and a `None` message → not neutral), plus two `_handle_stream_error` tests asserting that a missing-tool-output 400 awaits none of `record_error` / `mark_rate_limit` / `mark_quota_exceeded` / `mark_permanent_failure` while still classifying `non_retryable`, and that the model-entitlement 400 still awaits `record_error` exactly once.

`classify_upstream_failure`, `failover_decision`, and the status/body returned to the client are untouched. `invalid_request_error` already classifies `non_retryable` → `failover_decision` returns `"surface"`, so there was never failover amplification to remove; the only behaviour change is that a bogus health signal is no longer produced.

Deliberately **not** included, to keep the diff reviewable:

- `previous_response_not_found` is left penalizing. It is arguably anchor bookkeeping rather than account fault, but a stale anchor *is* account-scoped state, so that is a separate judgement call (raised in #1530).
- Letting hard-sticky selection rescue a single-element `in_error_backoff`. A hard-pinned session has no alternative account by construction, so the rescue condition is structurally unreachable for it — but that changes routing policy, whereas this PR only stops the bogus signal at the source. Proposed separately in #1530.

## Validation

- `tests/unit` + `tests/test_request_logs_options_api.py`: 5021 passed, 65 skipped;
- affected focused suites (`test_proxy_utils.py`, `test_proxy_http_bridge.py`, `test_proxy_api_websocket_auth.py`, `test_load_balancer_concurrency.py`, `test_proxy_load_balancer_refresh.py`): 1477 passed, 3 skipped;
- `uv run ruff check .` passed; `uv run ruff format --check .` passed (847 files);
- `uv run ty check` passed;
- `python scripts/check_proxy_architecture.py` passed;
- `openspec validate keep-request-shape-rejections-account-neutral --strict` valid; `openspec validate --specs` 48/48 passed;
- `git diff --check` passed.

Not run locally: integration/e2e/PostgreSQL/Helm/frontend jobs. Hosted CI is authoritative for those.

### DIFF
--- a/app/modules/proxy/_service/streaming/helpers.py
+++ b/app/modules/proxy/_service/streaming/helpers.py
@@ -782,6 +782,32 @@ async def _select_account_with_budget_for_stream(proxy: Any, deadline: float, **
     return await selector(deadline, **kwargs)
 
 
+def _is_account_neutral_request_rejection(
+    *,
+    code: str,
+    http_status: int | None,
+    message: str | None,
+) -> bool:
+    """Return whether upstream rejected the request payload, not the account.
+
+    A payload-shape rejection reproduces identically on every account, so it
+    must never mutate one account's health: otherwise a single client looping
+    on a self-inconsistent conversation drives its serving accounts into
+    ``error_count`` backoff and starves unrelated tenants.
+
+    Keep this set narrow. Not every upstream ``invalid_request_error`` is
+    account neutral -- the model-entitlement rejection matched by
+    ``_is_account_model_unsupported_error`` is genuinely account scoped -- so
+    membership is decided by the specific classified message, never by the
+    ``invalid_request_error`` code alone.
+    """
+    if code != "invalid_request_error":
+        return False
+    if http_status is not None and http_status != 400:
+        return False
+    return bool(_facade()._is_missing_tool_output_message(message))
+
+
 async def _handle_stream_error(
     proxy: Any,
     account: Account,
@@ -798,6 +824,18 @@ async def _handle_stream_error(
         phase="first_event",
     )
     if _facade()._is_account_neutral_error_code(code):
+        return classified
+    if _is_account_neutral_request_rejection(
+        code=code,
+        http_status=http_status,
+        message=error.get("message"),
+    ):
+        _facade().logger.info(
+            "Skipped account error penalty for account-neutral request rejection account_id=%s request_id=%s code=%s",
+            "<redacted>" if privacy_policy.redacts_sensitive_details else account.id,
+            get_request_id(),
+            code,
+        )
         return classified
     if classified["failure_class"] == "rate_limit":
         await proxy._load_balancer.mark_rate_limit(account, error)
--- a/app/modules/proxy/service.py
+++ b/app/modules/proxy/service.py
@@ -2109,16 +2109,22 @@ _MISSING_TOOL_OUTPUT_MESSAGE_PREFIXES = (
 )
 
 
+def _is_missing_tool_output_message(message: str | None) -> bool:
+    if message is None:
+        return False
+    normalized = " ".join(message.lower().split())
+    return normalized.startswith(_MISSING_TOOL_OUTPUT_MESSAGE_PREFIXES)
+
+
 def _is_missing_tool_output_error(
     *,
     code: str | None,
     param: str | None,
     message: str | None,
 ) -> bool:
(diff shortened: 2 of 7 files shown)


Soju06:
Reviewed against current main. Deciding neutrality by the classified message instead of `invalid_request_error` alone is the right scope — model-entitlement rejections stay account-scoped, and the guard mirrors the existing `_is_account_neutral_error_code` early-return in `_handle_stream_error` (app/modules/proxy/_service/streaming/helpers.py:825). Tests cover both the skip and the still-penalized account-scoped case.

Two notes:
- Merge order: the tool-search wording only becomes account-neutral once #1531's prefix lands, so #1531 should merge first.
- The only red check is Contributors attribution — add `yeongjun-cigro` to `.all-contributorsrc` and CI Required clears.

PR #1547, opened 2026-07-30
fix(proxy): retry silent bridge response.create upstreams

Komzpa:
## Summary

- retry an HTTP bridge request when the upstream Responses websocket accepts `response.create` but never emits `response.created`
- keep the existing fail-closed behavior when a request has already produced visible output or cannot be replayed safely
- add a regression test for the silent pre-created websocket case

## Test plan

- `uv run pytest tests/integration/test_http_responses_bridge.py -k 'missing_created_retry or precreated or response_create_gate or retired_gate'`


Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
Superseded: the core retry-before-claim idea landed via #1394 and #1563, and today's deployed #1625 covers the dead-owner recovery side. The remaining unique piece here (replaying client-owned anchors up to 20 times before response.created) is intentionally fenced off by the merged design, so closing this rather than rebasing it.

PR #1557, opened 2026-07-31
fix(proxy): replay full resend after bridge owner conflict

Komzpa:
## Summary

HTTP bridge full-resend recovery now treats `continuity_owner_conflict` the same way as an unavailable required owner when the existing account-neutral replay safety checks pass. This prevents stale Codex session/turn affinity from surfacing a one-turn 503 when the client has resent enough plaintext context to start a fresh bridge safely.

## Type of change

Linked issue: n/a

## OpenSpec

- [ ] This PR includes / updates an OpenSpec change
- [x] This PR touches a codex-faithful path (image pipeline, request/response shape, SSE framing, OAuth flow) and preserves upstream-equivalent behavior

Change directory: n/a

## Changes

- Allow account-neutral full-resend replay after `continuity_owner_conflict`, while keeping the existing safety gates: no forwarded request, no input-file owner, matching stored context prefix fingerprint, retained prior output, and account-neutral replay payload.
- Add an integration regression where a stale owner conflict is recovered by stripping stale Codex/session headers and replaying without `previous_response_id`.

## Simplicity

- [x] New feature defaults to **off** or works with **zero config**
- [x] No new required setup step (or maintainer approval via `simplicity-budget-approved` label)
- New setting(s) and why each can't be a default: n/a
- [x] README sections / `.env.example` / dashboard nav within budget (or `simplicity-budget-approved` label requested)

## Test plan

```
uv run pytest tests/integration/test_http_responses_bridge.py -k "replays_full_resend_after_owner_conflict"
uv run pytest tests/integration/test_http_responses_bridge.py -k "reports_unavailable_required_owner_when_other_account_exists or replays_full_resend_once_then_stays_on_new_owner or replays_full_resend_after_owner_conflict"
uv run ruff check app/modules/proxy/_service/http_bridge/streaming.py tests/integration/test_http_responses_bridge.py
uv run ruff format --check app/modules/proxy/_service/http_bridge/streaming.py tests/integration/test_http_responses_bridge.py
git diff --check
```

## Screenshots / output

Live symptom observed before the fix:

```
status=503 code="continuity_owner_conflict" message="Account-owned continuity sources conflict; retry the logical turn"
```

The regression proves this class now succeeds through an account-neutral full-resend replay instead of surfacing 503.

## Checklist

- [ ] Linked the related issue / discussion above.
- [x] Ran `uv run pre-commit run local-ci --hook-stage manual --all-files` or the relevant `make <target>` subset locally.
- [ ] If touching specs: `openspec validate --specs` passes and `/opsx:verify` is clean.

### DIFF
--- a/app/modules/proxy/_service/http_bridge/streaming.py
+++ b/app/modules/proxy/_service/http_bridge/streaming.py
@@ -1311,6 +1311,7 @@ class _HTTPBridgeStreamingMixin:
             if (
                 durable_model_transition_lookup is None
                 or not _http_bridge_owner_failure_allows_account_neutral_replay(exc)
+                or owner_unavailable_allows_account_neutral_replay(exc)
                 or request_state.previous_response_id is not None
                 or rewritten_file_account_id is not None
             ):
--- a/app/modules/proxy/_service/http_bridge/streaming.py
+++ b/app/modules/proxy/_service/http_bridge/streaming.py
@@ -68,7 +68,6 @@ from app.modules.proxy._service.http_bridge.helpers import (
     _effective_http_bridge_idle_ttl_seconds,
     _http_bridge_durable_lookup_allows_turn_state_takeover,
     _http_bridge_is_context_overflow_error,
-    _http_bridge_is_previous_response_owner_unavailable,
     _http_bridge_models_compatible,
     _http_bridge_owner_lookup_unavailable_error_envelope,
     _http_bridge_payload_looks_like_full_resend,
@@ -231,6 +230,11 @@ def _proxy_error_code_message(exc: ProxyResponseError) -> tuple[str | None, str 
     code = error.get("code")
     message = error.get("message")
     return (str(code) if code is not None else None, str(message) if message is not None else None)
+
+
+def _http_bridge_owner_failure_allows_account_neutral_replay(exc: ProxyResponseError) -> bool:
+    code, _message = _proxy_error_code_message(exc)
+    return code in {"previous_response_owner_unavailable", "continuity_owner_conflict"}
 
 
 def _http_bridge_account_capacity_wait_seconds(exc: ProxyResponseError) -> float | None:
@@ -1190,7 +1194,7 @@ class _HTTPBridgeStreamingMixin:
             nonlocal durable_full_resend_retains_prior_output
 
             if (
-                not _http_bridge_is_previous_response_owner_unavailable(exc)
+                not _http_bridge_owner_failure_allows_account_neutral_replay(exc)
                 or forwarded_request
                 or rewritten_file_account_id is not None
                 or durable_full_resend_anchor_count is None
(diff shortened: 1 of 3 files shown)


Komzpa:
@codex review

Soju06:
Two issues before this can land, beyond the rebase (#1562 rewrote the reconnect paths in streaming.py; #1394 merged on top):

1. Spec: `openspec/specs/sticky-session-operations/spec.md` ("Conflicting hard owners fail closed" at :130, plus :206/:298) requires `continuity_owner_conflict` to fail before upstream dispatch, and the active `openspec/changes/classify-required-http-bridge-owner-unavailable` spec says conflicting aliases MUST still fail with `continuity_owner_conflict`. This PR relaxes that, so "matches the existing spec" doesn't hold — it needs an OpenSpec delta amending those scenarios (e.g. "unless a verified account-neutral full resend permits fresh replay"). Conflict is also semantically different from owner-unavailable: both owners are alive, and fail-closed was deliberate to surface alias split-brain rather than paper over it. Worth an explicit maintainer call.

2. `switch_model_transition_to_account_neutral_fork` is weaker-gated than every analogous recovery path:
   - No `forwarded_request` guard. `owner_unavailable_allows_account_neutral_replay` refuses forwarded requests; this fork can fire on an owner-forwarded creation failure and start a local neutral lane on the wrong instance.
   - No payload-neutrality check. `switch_to_account_neutral_replay` requires `_http_bridge_payload_is_account_neutral_fresh_replay` (streaming.py:240 on main) and the cap-spill path requires `_request_allows_bare_session_cap_spillover` (affinity.py:294); the fork checks only `rewritten_file_account_id`, which misses opaque upstream-registered `input_file.file_id`s with no live local pin (forwarded verbatim per responses-api-compat), encrypted/account-scoped input items, and `conversation`. Repro: model-transition request on a hard session key whose input carries an unpinned upstream `file_id` living on the pinned account; owner conflict -> fork excludes that account -> replay on another account -> deterministic upstream file-not-found instead of a retryable 503.
   - Scope: `_http_bridge_owner_failure_allows_account_neutral_replay` includes `previous_response_owner_unavailable`, so the fork also fires for model transitions that fail full-resend fingerprint verification — wider than the conflict bug this PR describes, with no prefix verification at all.

The conflict->verified-full-resend lane itself keeps all the existing gates and looks sound for the observed one-turn 503; with the spec delta and the fork gates tightened (or the fork split into its own PR), this is close.

Soju06:
Heads-up before you invest in the rebase: #1599 is now open against the same durable full-resend / continuity_owner_conflict lane and touches the same three files as this branch (app/modules/proxy/_service/http_bridge/streaming.py, tests/integration/test_http_responses_bridge.py, tests/unit/test_proxy_http_bridge.py). It also carries an OpenSpec delta for the owner-conflict scenario (openspec/changes/reconcile-durable-full-resend-owner/), which is the spec amendment the 08-04 review noted was missing here. Suggest holding the rebase until #1599 settles, then reassessing what remains of this PR on top of it — likely the switch_model_transition_to_account_neutral_fork path, which per the earlier review needs a forwarded_request guard and a payload-neutrality check and is best split into its own PR against the post-#1599 baseline.

PR #714, opened 2026-05-19
chore(main): release 1.19.0

Soju06:
:robot: I have created a release *beep* *boop*
---

### Features

* **accounts:** add export action with audit and no-store safeguards ([#412](https://github.com/Soju06/codex-lb/issues/412)) ([c03e310](https://github.com/Soju06/codex-lb/commit/c03e31023bc0481696f1e85a0d373eaa086ef531))
* **accounts:** add operator-controlled account aliases ([#759](https://github.com/Soju06/codex-lb/issues/759)) ([781e259](https://github.com/Soju06/codex-lb/commit/781e2598b44356c90fc9e5e9c0780a87b59db202))
* **api-ui:** add account cost distribution for API ([#734](https://github.com/Soju06/codex-lb/issues/734)) ([d0a6737](https://github.com/Soju06/codex-lb/commit/d0a6737fb6016a627b23d12810e4a89d7b062ac3))
* **api-ui:** add account pool window usage remaining bar ([#785](https://github.com/Soju06/codex-lb/issues/785)) ([8eee9e2](https://github.com/Soju06/codex-lb/commit/8eee9e2c00e3739d1108f3d98cc4bf075d71900e))
* **api:** add codex /model support for allowed models ([#607](https://github.com/Soju06/codex-lb/issues/607)) ([15874aa](https://github.com/Soju06/codex-lb/commit/15874aa80795486c0bb0a33489a417c8b5613f51))
* **dashboard:** account burn projection card ([#752](https://github.com/Soju06/codex-lb/issues/752)) ([c48a20a](https://github.com/Soju06/codex-lb/commit/c48a20a566aa74213ec909bcc3e798bf835cbeef))
* **dashboard:** present hourly/weekly credits as raw remaining/total ([#612](https://github.com/Soju06/codex-lb/issues/612)) ([b6b2f8b](https://github.com/Soju06/codex-lb/commit/b6b2f8b0b64dbef8b66459e21b9406bdc24a8e94))
* **frontend:** add GitHub link to status bar ([#508](https://github.com/Soju06/codex-lb/issues/508)) ([8577edc](https://github.com/Soju06/codex-lb/commit/8577edc9a2eaeeabc736c2dcc6f3597e2da6e08f))
* **release:** add PR-driven beta release channel ([#732](https://github.com/Soju06/codex-lb/issues/732)) ([72b2962](https://github.com/Soju06/codex-lb/commit/72b2962b1a45f9b4796f4fc11f3f9db69cbceaf0))
[...]

PR #710, opened 2026-05-19
docs: FAQ entry on /wham/usage vs Settings UI divergence

ozpool:
## Summary

Adds \`docs/troubleshooting/usage-and-quota.md\` covering the recurring support pattern around OpenAI-side team reset events where:

- Codex Desktop *Settings → Account* UI shows the account as reset
- but \`/wham/usage\` still returns \`used_percent: 100\` for a short window afterwards
- codex-lb mirrors \`/wham/usage\` and so reports \`rate_limited\` / \`quota_exceeded\` until upstream catches up

Linked from a new \`## Troubleshooting\` section in the README.

The doc covers (matching #678's outline):
1. Which endpoint codex-lb reads (link to \`app/core/clients/usage.py\` + \`app/modules/usage/service.py\`)
2. How \`apply_usage_quota\` derives status from \`used_percent\` (\`>=100\` → \`rate_limited\` / \`quota_exceeded\`, auto-recovery on \`<100\`)
3. Why Settings UI and \`/wham/usage\` can legitimately disagree (separate OpenAI-side data sources, different update cadences)
4. What an operator can do today (wait for next request), and a forward-pointer to the Probe button planned in #677
5. A \`curl\` snippet to verify directly against \`/wham/usage\` that the disagreement is upstream rather than a codex-lb bug

Refs #676 / #677.

Fixes #678.

## Test plan
- [x] Doc renders cleanly in GitHub markdown (links + fenced code blocks)
- [x] All in-doc file references resolve against current \`main\`

PR #1560, opened 2026-07-31
Back off Codex review triggers on usage limits

Komzpa:
## Summary

- stop the Codex label sync from posting more `@codex review` comments when the same GitHub comment sender received a recent Codex usage-limit reply
- allow review requests again only when that same sender account has a newer normal Codex response within the 24-hour window
- make no-data runs probe once, reread the PR timeline, and suppress the remaining review requests if that probe hits the usage limit

## Tests

- `uv run pytest tests/unit/test_sync_codex_ok_labels.py -q`
- `uv run ruff check .github/scripts/sync_codex_ok_labels.py tests/unit/test_sync_codex_ok_labels.py`
- `uv run ruff format --check .github/scripts/sync_codex_ok_labels.py tests/unit/test_sync_codex_ok_labels.py`
- `make test-unit`

### DIFF
--- a/.github/scripts/sync_codex_ok_labels.py
+++ b/.github/scripts/sync_codex_ok_labels.py
@@ -10,7 +10,9 @@ import re
 import subprocess
 import sys
 import time
+from collections.abc import Callable
 from dataclasses import dataclass
+from datetime import UTC, datetime, timedelta
 from typing import Any
 from urllib.parse import quote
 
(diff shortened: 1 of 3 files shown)


Komzpa:
@codex review

Soju06:
The backoff logic itself looks sound (per-sender attribution, newer-normal-response unlatch, probe-and-reread), but splitting classification from apply introduced a log misattribution bug:

**Stale `number` in the apply loop** — `.github/scripts/sync_codex_ok_labels.py:1687` and the except handler at :1706 still format `f"{repo}#{number}"`, but the apply loop now iterates `for decision in decisions:` while `number` is the leftover from the classification loop (always the last classified PR). With `--all-open` over PRs 710 and 714, both status lines print `apply Soju06/codex-lb#714 ...`, and any exception during apply is attributed to #714 regardless of which PR failed. These lines are the script's audit trail, so this actively misleads. Should be `decision.number`. Note `test_main_stops_codex_review_triggers_after_probe_hits_usage_limit` passes only because it asserts on `#714`, the last PR — asserting `apply Soju06/codex-lb#710` appears in output would catch it.

Minor: the PR description says no-data runs "probe once", but the sleep+reread at :1678-1682 runs after every posted `@codex review`, not just the first — with N eligible PRs and no limit hit that's N x 10s plus N extra timeline fetches per run. Consider probing only until the first non-limit observation, or only when `latest_normal_response_at` is still None.

Soju06:
Superseded by #1631, which landed today as a first-party takeover (owner decision — the quota-burn this mitigates was live daily and the remaining review delta had no response since 07-31). Your commit was cherry-picked with authorship preserved; on top of it the review findings were addressed: per-decision log attribution, anchored quota-envelope matching, apply-time reclassification, installation-token-compatible sender resolution, reaction-based unlatch, and cross-repo backoff state. Thanks for the groundwork.

Issue #1534, opened 2026-07-30
HTTP-bridge reattach stalls with no response.created and needs a client restart to recover


lkraider:
**Version:** `ghcr.io/soju06/codex-lb:1.22.0`
**Client:** pi coding agent 0.82.1, transport `@earendil-works/pi-ai` 0.82.1, model `gpt-5.6-sol`, provider `codex-lb`, WebSocket endpoint `/backend-api/codex/responses`.

### Summary
When a session is resumed on the HTTP-bridge path, codex-lb reattaches by injecting a stored anchor and full-resending the conversation. If upstream never emits `response.created`, codex-lb waits its client-safe deadline (`_HTTP_BRIDGE_EVENTLESS_RESPONSE_CREATED_MAX_SECONDS = 240`), then ends the turn with "Upstream did not acknowledge response.create before the client-safe deadline". Retrying within the same client session rebuilds the identical reattach to the same account and stalls the same way, so the client cannot recover by retrying. Only starting a fresh client session recovers.

### Evidence (one session, two consecutive attempts)
```
event=fresh_reattach_anchor_injected  bridge_key=sha256:7dfa… account_id=…7cb39f74
event=durable_full_resend_anchor_injected  response_id=resp_0563… stored_items=425
event=create  request_stage=follow_up  durable_session_id=40b06ad6…
(~240s later) WARNING event=missing_response_created_timeout  pending=1
event=close ; event=retire_stale_pending
```
- Both attempts used the identical `bridge_key`, account `…7cb39f74`, `durable_session_id`, anchor `resp_0563…`, and 425 resent items. Each stalled ~241s. `reallocate_sticky=False`.
- The client recorded each as a terminal error and did not auto-recover.
- A fresh client session started under the same conditions came in as `kind=websocket request_stage=first_turn` with no injected anchor and proceeded normally. So a recoverable path exists; the running session could not reach it.
- This deadline/stall path is HTTP-bridge only; the WebSocket-upstream path has no equivalent deadline.

### Steps to reproduce
1. Run a long multi-turn session (codex session affinity on).
2. Let the upstream bridge go idle long enough to be retired (about 4 minutes).
3. Send the next turn. codex-lb does a fresh reattach with a full resend.
4. Upstream does not emit `response.created`; ~240s later the turn fails with the deadline error.
5. Retry in the same session: it reproduces identically. Restart the client (new session): it recovers.

### Expected behavior
[...]

PR #1574, opened 2026-08-03
Penalize eventless HTTP bridge accounts

Komzpa:
## Summary

Fixes an HTTP bridge recovery gap where a `response.create` send that never
receives `response.created` retires the stale bridge session but leaves the
selected account fully healthy.

That made later requests eligible to select the same eventless account again,
turning a single missing-created timeout into repeated
`upstream_request_timeout` failures. The request itself is still failed closed
without replaying or moving it to another account, but the selected account now
records the normal transient health failure so subsequent selection can avoid
the unhealthy route.

## Changes

- Record a transient account-health penalty when the HTTP bridge owner-side
  missing-`response.created` deadline expires.
- Keep the existing fail-closed behavior: no transparent replay and no
  submission of the timed-out request on another account.
- Update the active OpenSpec change so the spec matches the health-accounted
  recovery behavior.
- Extend HTTP bridge regression coverage to verify the penalty reaches
  `_handle_stream_error` for a real `Account`.

## Tests

- `uv run pytest -q tests/unit/test_proxy_http_bridge.py`
- `uv run ruff check app/modules/proxy/_service/http_bridge/upstream_events.py tests/unit/test_proxy_http_bridge.py`
- `git diff --check`
- `openspec validate recover-codex-desktop-idle-bridge --strict`

Full `openspec validate --specs` still reports existing normative-keyword
issues in `api-keys` and `conversations-api`, outside this change.

### DIFF
--- a/app/modules/proxy/_service/http_bridge/upstream_events.py
+++ b/app/modules/proxy/_service/http_bridge/upstream_events.py
@@ -661,7 +661,7 @@ class _HTTPBridgeUpstreamEventsMixin:
                                 session,
                                 error_code="upstream_request_timeout",
                                 error_message=receive_timeout.error_message,
-                                penalize_account=False,
+                                penalize_account=True,
                                 retire_detail=_HTTP_BRIDGE_MISSING_RESPONSE_CREATED_TIMEOUT_DETAIL,
                                 force_retire=True,
                             )
--- a/openspec/changes/recover-codex-desktop-idle-bridge/design.md
+++ b/openspec/changes/recover-codex-desktop-idle-bridge/design.md
@@ -53,7 +53,7 @@ Leading non-response telemetry such as `codex.rate_limits` does not change those
 
 When the deadline expires, reuse the reader-owned terminal failure and whole-session retirement path. Emit a stable `missing_response_created_timeout` detail, increment the existing stuck-retirement metric, settle every pending request exactly once, and close the bridge session.
 
-Do not transparently replay the timed-out request, submit it on another account, or mark the selected account unhealthy. Upstream acceptance is unknown, so duplicate submission and account movement are less safe than an explicit terminal failure. A later client request creates a fresh session through existing behavior.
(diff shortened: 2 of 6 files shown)


Soju06:
The gap you describe is real — an account that accepts response.create but never emits response.created stays fully selectable, which matches the #1534 stall pattern. But this needs a rework against current main and a maintainer call on the design:

- The missing-created path was restructured by the fresh-hard recovery work (#1394, merged 08-04): there are now two `penalize_account=False` sites (app/modules/proxy/_service/http_bridge/upstream_events.py:858 and :895), and `_retry_http_bridge_precreated_request` already gives an eligible fresh hard request one bounded retry that can switch accounts. The diff patches the old single call site and rewrites a stale spec revision (240s deadline; main is 60s and integrates `recover-fresh-hard-bridge-timeouts`).
- Main's spec for this change now states normatively: "Neither path may write an account-health failure solely because `response.created` was missing," and upstream_events.py carries a comment deliberately retaining the account-neutral classification. Reversing that is a design decision, not a fix — during upstream-wide silence (OpenAI-side flapping has produced exactly this signature fleet-wide), penalizing every account that hits missing-created drains the eligible pool, and interacts badly with the in-flight eligible-pool-exhaustion 429 work (#1541).
- This also overlaps the session-quarantine direction in #1405/#1402, which targets the same reselection gap without account-health writes.

If health accounting is wanted here, the rebased version should cover both call sites, rewrite the current spec text (not the pre-fresh-hard revision), and probably gate the penalty on repeated missing-created on the same account rather than a single occurrence. Also note tests/unit/test_proxy_http_bridge.py (`..._force_retires_with_admission_waiter`) now omits the explicit `penalize_account` kwarg, so it asserts the method default rather than the production call site.

Soju06:
Closing per maintainer decision: for silent/eventless upstreams we are going with session quarantine (the #1405 direction) rather than an account-health penalty — fleet-wide upstream silence must not drain the eligible pool, and the merged spec deliberately forbids health writes solely for missing response.created. The reselection gap you identified is real and will be addressed on that track.

PR #1484, opened 2026-07-24
fix(proxy): reconcile durable full resend owner

mhooooo:
## Summary

Stop a deterministic pre-dispatch retry loop when a durable HTTP bridge owner and a broad legacy `CODEX_SESSION` sticky row disagree.

A verified complete resend now starts the fresh bridge on the durable owner without consulting or forwarding the stale broad alias. The proof supports both retained prior assistant output and the response-bound pending-tool-call manifest introduced by #1486.

## Root cause

#1486 preserves verified unanchored full resends, but account selection still receives the broad session-header affinity. If the durable bridge row owns the task on account B while a legacy raw session row points at account A, selection returns retryable `continuity_owner_conflict`. Repeating the same request cannot change either persisted owner, so the client loops.

## Changes

- Restack the change directly on current `main`; the obsolete #1481 dependency commits are gone.
- Represent owner-reconciliation eligibility with an immutable request-local proof bound to the payload fingerprint, durable session, owner, latest response, stored count/fingerprint, and pending-tool-call manifest identity.
- Accept either retained prior assistant output or an exact `responses_input_suffix_matches_pending_tool_calls` settlement.
- For that proved fresh reattach only, remove the broad session-header selection source and downstream session/turn aliases.
- Keep the durable canonical bridge key and owner as hard constraints, preserve Codex session behavior for later incremental turns, and leave the legacy row unchanged.
- Preserve fail-closed behavior for incomplete resends and conflicting turn-state, previous-response, bridge, and file owners.

## Impact

- Fixes the stale broad-owner conflict loop for both ordinary full resends and the tool-loop resend shape covered by #1486.
- Adds no setting, dependency, endpoint, schema, migration, dashboard surface, or required setup step.
- No dashboard-visible change.

Refs #1393.

## Test plan

```text
python -m pytest tests/unit/test_proxy_http_bridge.py \
  tests/integration/test_http_responses_bridge.py -q
502 passed

make lint
passed

make typecheck
passed

npx --yes @fission-ai/openspec@1.6.0 validate \
  reconcile-durable-full-resend-owner --strict
valid

npx --yes @fission-ai/openspec@1.6.0 validate --specs --strict
48 passed, 0 failed

npx --yes @fission-ai/openspec@1.6.0 validate --changes --strict
71 passed, 0 failed

codex review --base origin/main
no actionable regression identified
```

## Checklist

- [x] Conventional Commit title.
- [x] OpenSpec change included and strictly validated.
[...]

PR #1599, opened 2026-08-04
fix(proxy): reconcile durable full resend owner

Soju06:
The fresh durable full-resend path can still fail before upstream dispatch when a broad legacy session-header sticky row points at a different account than the durable bridge owner: codex-lb returns `continuity_owner_conflict` as a retryable 503, the client repeats the identical request, and neither persisted owner ever changes — a deterministic retry loop even though the request already carries fingerprint-verified complete context.

This change represents complete durable full-resend eligibility as a sealed, immutable, request-local proof (`_VerifiedDurableFullResend`) created only by the count/fingerprint/retained-output or response-bound pending-tool-call verifier. For that proved fresh reattach only, the broad session-header alias is dropped from account selection and the upstream handshake, while the durable canonical key and durable owner account remain required. The broad sticky row is left intact for sibling traffic; specific turn-state/previous-response/file owner conflicts remain fail-closed. See `openspec/changes/reconcile-durable-full-resend-owner/` for the full design.

Supersedes #1484 — original implementation by @mhooooo, taken over per the 07-30 maintainer notice. The author's commit is preserved with original authorship; takeover fixes sit on top.

## What survived vs. what changed in the takeover

**Survived unchanged (author's commit, rebased onto current main):**
- The sealed `_VerifiedDurableFullResend` proof, including the round-4 restack requirement (`responses_input_suffix_matches_pending_tool_calls` settlement path) — the author had already addressed it before going silent.
- The owner-bound broad-session reconciliation (`fresh_reattach_broad_session_owner_ignored`), affinity-header stripping on session creation, the anchor-injection fallback for proof-failed full resends, the openspec change, and the author's unit + integration tests (including the real, unstubbed `test_backend_responses_verified_full_resend_ignores_stale_broad_owner_on_durable_account`).

**Dropped:**
- The three #1481-shape dependency commits are no longer in the series — they were superseded by what already landed on main (#1476/#1499 et al.), so the branch carries only the single implementation commit.

**Takeover fixes on top (reconciliation with #1486/#1394/#1558/#1527/#1531/#1532 and the codex P1):**
- Restacked onto main's `classify_durable_full_resend` structure and the post-#1394 recovery-attempt journal flow; the unit test's account-neutral classifier assertion is now conditional because main computes neutrality eagerly for the recovery journal on the preserved path. None of the reconciliation was subsumed by #1394 — main still leaves the broad session alias in selection for a proved full resend, so the conflict loop remained live.
- Updated the stale anchor-injection comment (injection can now also run for a full resend that failed proof verification).
- Addressed the open codex P1 on #1484 (the unit regression test pre-binds the session and stubs `_get_or_create_http_bridge_session`/`_submit_http_bridge_request`): added bridge-level coverage `test_backend_responses_verified_full_resend_fails_over_to_new_account_after_owner_loss` — a real turn completes on account A through the actual create/response-create gate, A is paused, and the verified session-header full resend fails over to account B with the complete history and no `previous_response_id`.
- Reworded the modified openspec requirement so the MUST clause sits on the first body line, per the current validator.

## Gates

- `architecture-check`: pass (no ratchet changes)
- `test-unit`: 5283 passed, 3 skipped
- `test-integration-bridge` (`test_http_responses_bridge.py` + `test_proxy_websocket_responses.py`): 220 passed
- `ruff check` / `ruff format --check`: clean
- `openspec validate reconcile-durable-full-resend-owner`: valid

### DIFF
--- a/app/modules/proxy/_service/http_bridge/streaming.py
+++ b/app/modules/proxy/_service/http_bridge/streaming.py
@@ -237,6 +237,155 @@ def _http_bridge_continuity_bound_without_safe_replay(request_state: _WebSocketR
     )
 
 
+class _VerifiedDurableFullResend:
+    """Immutable proof that one payload contains a durable turn's complete context."""
+
+    _durable_session_id: str
+    _full_input_fingerprint: str
+    _latest_response_id: str
+    _owner_account_id: str
+    _pending_tool_calls: tuple[tuple[str, str], ...] | None
+    _stored_input_fingerprint: str
+    _stored_input_item_count: int
+    __slots__ = (
+        "_durable_session_id",
+        "_full_input_fingerprint",
+        "_latest_response_id",
+        "_owner_account_id",
+        "_pending_tool_calls",
+        "_stored_input_fingerprint",
+        "_stored_input_item_count",
+    )
+    __construction_token = object()
+
+    def __init__(
+        self,
+        *,
+        _token: object,
+        durable_session_id: str,
+        owner_account_id: str,
+        latest_response_id: str,
+        stored_input_item_count: int,
+        stored_input_fingerprint: str,
+        full_input_fingerprint: str,
+        pending_tool_calls: tuple[tuple[str, str], ...] | None,
+    ) -> None:
+        if _token is not self.__construction_token:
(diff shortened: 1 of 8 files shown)


Komzpa:
@codex review

Issue #578, opened 2026-05-10
[design] Reconsider budget-safe routing gate vs health-tier overlap and per-window thresholds


Soju06:
## Background

[#421](https://github.com/Soju06/codex-lb/pull/421) introduced `_state_above_budget_threshold` and `_select_account_preferring_budget_safe` to avoid sending fresh requests to upstream accounts whose primary or secondary usage was already above a configured budget threshold. [#446](https://github.com/Soju06/codex-lb/issues/446) found a P1 edge case in the original `any(primary, secondary)` shape, and [#561](https://github.com/Soju06/codex-lb/pull/561) (merged in `3ed7834`) narrowed the hard gate to primary-only and added a primary-pressure-aware fallback.

That fix is a strict improvement over the previous behavior: the most obvious `usage_limit_reached` reproducer from #446 no longer fires. But the budget-safe gate still has open design questions that #561 deliberately did not address. This issue tracks them so we can revisit the design before the gate accumulates more callers / strategies that depend on its current shape.

### 1. `primary` and `secondary` are not the same kind of signal

(originally raised by @egusev [in the #446 discussion](https://github.com/Soju06/codex-lb/issues/446#issuecomment-XXXXXX) — link the actual comment if convenient.)

`primary` is the short-window quota (~5h) that throws `usage_limit_reached` when exceeded. `secondary` is a longer rolling window (~7d). Putting both behind the same single percentage threshold is misleading:

- `5%` remaining in the weekly window is a fundamentally different operational state from `5%` remaining in the 5h window.
- The 5h window mostly drives the immediate reject; the 7d window drives long-tail capacity planning.
- A single `sticky_reallocation_budget_threshold_pct` knob conflates the two and forces operators to pick the lowest reasonable threshold for both, which under-uses the 5h window when 7d is the binding constraint, and over-uses the 5h window when 7d is plenty.

#561 only closed the hard-gate side of this (drop secondary from the gate). The strategy-side sort_key in `usage_weighted` still uses `(secondary_used, primary_used, ...)`, so when the gate degrades to the full pool, the sort still prefers low-secondary accounts even when their primary is already 99 %. The narrowed gate makes that case rarer, but does not eliminate it.

Possible directions:

- separate `sticky_reallocation_primary_threshold_pct` and `sticky_reallocation_secondary_threshold_pct` knobs, both used as hard gates with the union semantics restored, but with sane defaults (primary aggressive, secondary loose).
[...]

PR #1603, opened 2026-08-05
fix(proxy): allow fresh burn-first usage drain

jaekwonhong:
﻿## Summary

- allow an unassigned fresh request to consume a `burn_first` account that is draining only because of quota usage
- require a separately selectable healthy fallback before admitting that exception
- preserve recovery-probe precedence, existing soft-sticky ownership, hard continuity, and explicit sequential/reset/single-account routing

## Root cause

Runtime health synchronization marks an account as `DRAINING` before routing-policy preference runs. A `burn_first` account already above the soft-drain threshold was therefore removed by health-tier narrowing and could never consume its remaining quota, even though another healthy account was available for failover.

## User and operator impact

Fresh traffic can now safely exhaust the remaining quota of a usage-only draining `burn_first` account. Accounts with current error-drain evidence, fully exhausted accounts, and requests with an existing owner remain excluded from the exception. No API, database, dashboard, or configuration contract changes are introduced.

## OpenSpec

- `openspec/changes/allow-fresh-burn-first-usage-drain/`
- Related to #578 as a narrow follow-up; this PR does not close that broader routing-design issue.

## Simplicity

- zero-config behavior: no new environment variables, settings, migrations, external accounts, or manual setup steps
- no README, `.env.example`, dashboard navigation, or dashboard-visible changes
- the shared selector option defaults off and is enabled only by the two fresh, unowned request paths covered by regression tests

## Validation

- `pytest tests/unit/test_load_balancer.py tests/unit/test_proxy_load_balancer_refresh.py tests/unit/test_select_with_stickiness.py -q` — 402 passed, 3 skipped
- `ruff check` on all changed Python files — passed
- `ruff format --check` on all changed Python files — passed
- `ty check` on all changed Python files — passed
- `openspec validate allow-fresh-burn-first-usage-drain --strict` — passed
- `openspec validate --specs --strict` — 49 passed, 0 failed
- full Windows-host `ty check` reports five unchanged POSIX-only diagnostics in `scripts/run_dashboard_browser_smoke.py` and `tests/unit/test_refresh_claims.py`; changed-file type checking is clean and CI will run the repository-wide Linux gate


Soju06:
Thanks — the underlying bug is real and your root-cause matches main: `_select_account_preferring_budget_safe` narrows via `_best_health_tier_states` before the burn-first preference, and `select_account` itself pools `healthy or probing or draining or available` (app/core/balancer/logic.py ~L630), so a `burn_first` account above the soft-drain thresholds is never selected fresh while any healthy account exists. The overall mechanism (default-off flag, canonical `evaluate_health_tier` decomposition, recovery-probe precedence, sticky/continuity exclusion) is well-constructed.

One blocking issue before this leaves draft:

1. **The no-fallback `elif` branch is an availability regression.** In `sticky_selection.py`, when `usage_draining_burn_first` is non-empty but no healthy fallback is selectable, you remove those accounts from `state_list` for the rest of the function. On main they are still served as last resort — `_best_health_tier_states` falls back to draining, and `select_account`'s own tier pooling does the same. A deployment whose only usable account is a `burn_first` at 86%+ serves traffic today but returns no account under this PR; `test_fresh_selection_requires_healthy_fallback_to_drain_burn_first` asserts `result.account is None`, codifying the regression. It also inverts burn-first policy in all-draining pools (the burn-first account becomes excluded instead of preferred). Your own spec scenario only says the account is not admitted "through the usage-drain exception" — it doesn't cover removing it from subsequent passes. Please drop the `elif` and fall through with the unchanged `state_list`; main's existing ordering already handles the no-fallback case.

Two smaller items:

2. The `fallback = select_account(healthy_fallbacks, ...)` probe runs before checking `if usage_draining_burn_first`, so every fresh selection in every deployment pays an extra full `select_account` pass under the runtime lock even with zero burn-first accounts. Guard the block on `usage_draining_burn_first` first.

3. Sequencing: this collides textually with #1541 (`sticky_selection.py`, `unbound_selection.py`, `tests/unit/test_load_balancer.py`) and #1536; expect a rebase after those land. Note that with the current `elif` in place, the new empty-pool outcomes would surface as #1541's pool-exhaustion 429s — another reason to remove it.

Whether fresh traffic should require a healthy fallback at all before draining to zero is really the #578 design question, so flagging that part for owner sign-off. The rest of the test coverage (probe precedence, error-drain exclusion, sticky/continuity invariance) looks solid.

Soju06:
Policy decision (maintainer, recorded in #578 as well): **option (a) — healthy-fallback-required.** A usage-draining `burn_first` account joins fresh selection only when a separately selectable healthy fallback exists, so a mid-turn 429 near the limit always has somewhere to fail over to. We're not lifting the health-tier narrowing unconditionally.

That confirms this draft's direction. Before marking it ready, two things from the review above must land:

1. **P1** — the no-fallback branch must degrade to main's current behavior (burn_first stays excluded from fresh selection but untouched otherwise), not strip those accounts from all selection passes; as written it's an availability regression when the pool has no healthy fallback.
2. **P3** — don't pay the `select_account(healthy_fallbacks, ...)` pass unconditionally on every fresh selection; gate it on `usage_draining_burn_first` being non-empty.

Also run the full CI gate once you mark it ready (draft runs skip the substantive jobs). The fresh-sticky-pins-to-98% behavior noted in the review is accepted as inherent to burn_first semantics given the fallback guarantee.

jaekwonhong:
Addressed both maintainer blockers in commit efa18c7c: the no-fallback path now preserves the baseline last-resort availability behavior, and the healthy-fallback probe runs only when a usage-draining burn-first candidate exists. The branch is also synchronized with current main at 410bfaea via merge commit 9b7039b1. Revalidation: 402 passed / 3 skipped; changed-file Ruff, format, and ty checks passed; the OpenSpec change and all 49 canonical specs pass strict validation.

jaekwonhong:
@codex review

PR #1605, opened 2026-08-05
fix(proxy): enforce fair share on idle bridge reacquire

jaekwonhong:
## What changed

- preserve API-key attribution when an idle HTTP bridge directly reacquires its stream lease
- reconstruct the enabled fair-share candidate pool with model/service-tier eligibility, API-key account scope, single-account routing, exclusions, and security authorization
- fail closed when the existing bridge owner is outside that pool
- evaluate membership, fair-share admission, and lease acquisition atomically under the runtime lock
- document and test the direct-reacquire contract in the existing OpenSpec change

## Why

PR #1536 applies fair-share admission during normal account selection, but an idle HTTP bridge keeps its upstream session and directly reacquires a stream lease on the next turn. That path could bypass the selection-only gate and omit API-key attribution, allowing a heavy key to retain capacity outside the intended fair-share accounting.

## Impact

The feature remains default-off. With the threshold disabled, direct reacquisition only records API-key attribution and performs no candidate-pool load. Reattach remains exempt. Enabled requests now receive the existing stable 429 api_key_stream_fair_share envelope when the key is over share; normal hard-cap behavior and session ownership are preserved.

## Validation

- targeted proxy/load-balancer/HTTP bridge suite: 409 passed
- Ruff check and format check: passed
- openspec validate add-api-key-stream-fair-share --strict: passed
- openspec validate --specs --strict: 49 passed, 0 failed
- git diff --check: passed

Local validation used Python 3.14.4; project CI uses Python 3.13.

Stacked on #1536. This PR should merge into feat/api-key-stream-fair-share before the parent PR lands.


jaekwonhong:
@codex review

jaekwonhong:
Closing this stacked PR because the repository owner has directed all further publication to XZ-Organization/Codex_Model-provider-selector_codex-lb only.

PR #1606, opened 2026-08-05
fix(compact): preserve tool search pairs

Komzpa:
## Summary

Preserve Codex `tool_search_call` / `tool_search_output` pairs during compact input trimming so oversized sessions do not forward an orphan `tool_search_output` upstream. This fixes a live compact failure where OpenAI rejected session `019fcc77-8e08-7f02-b5d2-3388c5122004` with `No tool call found for tool search output with call_id call_d2chifMNZfXqfFWin5jz0dQf.`

## Type of change

Linked issue: none

## OpenSpec

- [x] This PR touches a codex-faithful path (image pipeline, request/response shape, SSE framing, OAuth flow) and preserves upstream-equivalent behavior

Change directory: n/a

## Changes

- Adds `tool_search_call` and `tool_search_output` to the compact tool pair reconciliation sets.
- Adds a regression test using the live-shaped `tool_search_call` / `tool_search_output` history item pair.

## Test plan

```sh
.venv/bin/python -m pytest tests/unit/test_openai_requests.py -k 'tool_search_outputs or compact_trimming_keeps_selected_tool_outputs_with_matching_calls or compact_trimming_drops_selected_tool_outputs_without_matching_calls or compact_trimming_reconciles_duplicate_tool_call_ids_by_occurrence' -q\n# 4 passed, 146 deselected\n\n.venv/bin/python -m pytest tests/integration/test_proxy_compact.py -k 'compact_strips_tool_fields or side_effect_pair' -q\n# 2 passed, 25 deselected\n\ngit diff --check\n```\n\n## Screenshots / output\n\nLive replay from `/home/kom/.codex/sessions/2026/08/04/rollout-2026-08-04T15-10-15-019fcc77-8e08-7f02-b5d2-3388c5122004.jsonl` through `ResponsesCompactRequest.to_payload()` now keeps both `tool_search_call` and `tool_search_output` for `call_d2chifMNZfXqfFWin5jz0dQf`; `has_orphan_tool_search_output` is `False`.\n\n## Checklist\n\n- [x] Title is in Conventional Commits format (`<type>(<scope>)?: <subject>`).\n- [ ] Linked the related issue / discussion above.\n- [x] Added or updated tests covering the change.\n- [ ] Ran `uv run pre-commit run local-ci --hook-stage manual --all-files` or the relevant `make <target>` subset locally.\n- [ ] If touching specs: `openspec validate --specs` passes and `/opsx:verify` is clean.\n- [x] Simplicity gates reviewed: the five simplicity rules (PRINCIPLES.md P1-P5).\n- [x] CHANGELOG is **not** edited by hand (release-please handles it).\n

### DIFF
--- a/app/core/openai/requests.py
+++ b/app/core/openai/requests.py
@@ -48,6 +48,7 @@ _COMPACT_TOOL_CALL_TYPE_BY_OUTPUT_TYPE: dict[str, str] = {
     "function_call_output": "function_call",
     "custom_tool_call_output": "custom_tool_call",
     "apply_patch_call_output": "apply_patch_call",
+    "tool_search_output": "tool_search_call",
 }
 _COMPACT_TOOL_CALL_ITEM_TYPES = frozenset(_COMPACT_TOOL_CALL_TYPE_BY_OUTPUT_TYPE.values())
 _COMPACT_TOOL_CALL_OUTPUT_ITEM_TYPES = frozenset(_COMPACT_TOOL_CALL_TYPE_BY_OUTPUT_TYPE)
--- a/app/modules/proxy/_service/http_bridge/helpers.py
+++ b/app/modules/proxy/_service/http_bridge/helpers.py
@@ -530,7 +530,7 @@ def _trim_http_bridge_previous_response_input_items(input_items: list[JsonValue]
             index
             for index, item in enumerate(input_items)
             if _http_bridge_input_item_type(item)
-            in {"function_call_output", "custom_tool_call_output", "apply_patch_call_output"}
+            in {"function_call_output", "custom_tool_call_output", "apply_patch_call_output", "tool_search_output"}
         ),
         None,
     )
@@ -544,7 +544,7 @@ def _trim_http_bridge_previous_response_input_items(input_items: list[JsonValue]
 
 def _is_http_bridge_previous_response_output_item(item: JsonValue) -> bool:
     item_type = _http_bridge_input_item_type(item)
-    if item_type in {"reasoning", "function_call", "custom_tool_call", "apply_patch_call"}:
+    if item_type in {"reasoning", "function_call", "custom_tool_call", "apply_patch_call", "tool_search_call"}:
         return _has_http_bridge_response_output_marker(item)
     if item_type != "message" or not isinstance(item, dict):
         return False
--- a/app/modules/proxy/_service/http_bridge/request_submit.py
+++ b/app/modules/proxy/_service/http_bridge/request_submit.py
@@ -3020,12 +3020,13 @@ class _HTTPBridgeRequestSubmitMixin:
                 # Account-scoped uploaded files cannot be replayed on a
                 # different owner. Keep the preferred account mandatory for
                 # both silent recovery and clean-close recovery.
-                require_preferred_reconnect = account_neutral_recovery or request_state.file_required_preferred_account
+                require_preferred_reconnect = request_state.file_required_preferred_account
                 request_text = _prepare_websocket_request_state_for_visible_output_replay(request_state)
                 if request_text is None:
                     return False
                 if account_neutral_recovery:
-                    request_state.preferred_account_id = session.account.id
+                    request_state.preferred_account_id = None
+                    request_state.excluded_account_ids.add(session.account.id)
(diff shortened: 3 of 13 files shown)


Soju06:
Verified the core fix against main and it is correct: `_COMPACT_TOOL_CALL_ITEM_TYPES` / `_COMPACT_TOOL_CALL_OUTPUT_ITEM_TYPES` (app/core/openai/requests.py:45-47) omit the tool-search pair, so `_compact_reconciled_tool_call_indices` treats a `tool_search_output` in the retained tail as ordinary context and forwards it without its call. The orphan-output direction (`No tool call found for tool search output ...`) is also not covered by the missing-tool-output classifier (it only matches `No tool output found for ...`), so the upstream 400 surfaces unmasked — the regression test models this correctly and fails without the fix.

Two items before this can merge:

1. Scope disclosure: commit ca7d010 changes the compact trim marker role from `user` to `developer` (app/core/openai/requests.py:1472), but the PR body says "no behavior change beyond the bug" and the Changes section does not mention it. Please drop that commit or split it into its own PR with rationale. It is not a pure cosmetic change: a `{"type": "message", "role": "developer"}` item matches `_compact_item_is_state_anchor` (requests.py:1370) and the `_normalize_responses_input_instructions` hoist shape, so a marker that re-enters validation or trimming behaves differently from the user-role marker (instructions hoist / mandatory trim anchor). If keeping synthetic text out of user turns is the goal, that deserves its own review and test.

2. CI: integration-bridge failed on `test_v1_responses_http_bridge_reconnects_after_clean_upstream_close` (409 vs 200), which is not the known flaky WS masking test. I don't see a causal path from this diff, but it needs a green rerun.

Heads-up: your open #1235 rewrites these same frozensets into `_COMPACT_TOOL_CALL_TYPE_BY_OUTPUT_TYPE`; whichever lands second needs a rebase that keeps `tool_search_call`/`tool_search_output` in the mapping.

Komzpa:
@codex review

Soju06:
Thanks — the core fix here is real and worth landing: pairing `tool_search_call` / `tool_search_output` during compact trimming addresses a live upstream rejection (`No tool call found for tool search output...`), and the regression test reproduces the shape faithfully. However, the Codex review on `433d75e5` raised four findings that all check out on inspection, and they need to be addressed before merge:

1. **Drop the api-keys commit entirely** (`fix(api-keys): retain explicit last-used compatibility touch`). The new `update_last_used()` in `app/modules/api_keys/repository.py:383` has zero call sites in this PR or on main, no test, and commits by default — it is dead public API unrelated to compact trimming. It also writes `last_used_at` unconditionally, which would violate the greatest-wins monotonic contract from `openspec/changes/coalesce-api-key-last-used-writes/specs/api-keys/spec.md` if it were ever called. If this belongs to another PR in your series, please move it there (guarded with the coalescer predicate); here it just couples two independently revertible concerns.

2. **Use a dedicated pairing set instead of widening `_TOOL_CALL_ITEM_TYPES`** (`app/core/openai/requests.py:45`). That set is shared with `strip_replayed_tool_call_namespaces_from_payload()` (line ~891), so this change silently makes namespace stripping mutate `tool_search_call` items on all request paths — beyond the scope allowed by `openspec/changes/fix-replayed-namespaced-function-call/context.md`. Something like `_COMPACT_TOOL_CALL_PAIR_ITEM_TYPES = _TOOL_CALL_ITEM_TYPES | {"tool_search_call"}` used only in the compact reconciliation (line ~1330) keeps the fix scoped.

3. **Add coverage at the failing surface.** The bug manifested at `/backend-api/codex/responses/compact`; per AGENTS.md the regression should go through that route (e.g., in `tests/integration/test_proxy_compact.py` alongside the existing side-effect pair tests), not only through `ResponsesCompactRequest.to_payload()`. Keeping the unit test is fine, but the route-level assertion is what proves routing and final upstream preparation preserve the pair.

Once those are in (and the Codex threads resolved), this looks mergeable — CI is already green.

Komzpa:
@codex review

Komzpa:
Folded into #1720: tool-search pair preservation joins the compact payload sanitization carrier; commits replayed onto a clean base to shed stale branch ancestry.

PR #1608, opened 2026-08-05
fix(db): recognize deployed migration head

Komzpa:
## Summary

- add the Alembic revisions needed for a current `main` build to recognize the deployed database migration head
- keep migration policy checks tolerant of older SQLite files that still carry legacy bridge columns or foreign-key wiring
- unblock helper-built live stacks that combine this PR with Soju06/codex-lb#1606 from `origin/main`

## Validation

- `uv run --frozen python -c "from app.db.migrate import check_migration_policy; check_migration_policy(); print('policy_ok')"`
- `uv run --frozen python -m app.db.migrate --db-url sqlite:///migration-smoke.db upgrade`
- `uv run --frozen python -m app.db.migrate --db-url sqlite:///migration-smoke.db current`
- `uv run --frozen python -m app.db.migrate --db-url sqlite:///migration-smoke.db check`
- `git diff --check`

### DIFF
--- a/app/db/alembic/versions/20260722_000000_add_security_lineage_persistence.py
+++ b/app/db/alembic/versions/20260722_000000_add_security_lineage_persistence.py
@@ -1,0 +1,270 @@
+"""Reconcile durable security-lineage persistence without a second head.
+
+Revision ID: 20260722_000000_add_security_lineage_persistence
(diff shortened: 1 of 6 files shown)


Komzpa:
Closing: this was opened for a local deployment-state repair and should not be an upstream PR.

PR #1638, opened 2026-08-06
fix(proxy): never inject a cross-account previous_response_id anchor (takeover of #1274)

Komzpa:
Takeover of #1274 by @aididhaiqal — the original three commits are carried with authorship intact; two additions on top. The original diagnosis was exactly right and turned out to be the root cause of the bridge-wedge family we have been mitigating from the other end all day (#1580 detection, #1630 quarantine): an injected `previous_response_id` belonging to a different account is accepted by upstream, which then never emits `response.created`, and the client waits forever.

**What this adds over #1274 as-is:**
- Rebased onto current main (one mechanical conflict in `mixin.py`; main's newer try/abort structure kept verbatim).
- A second unguarded post-binding injection site closed: the owner-forward recovery rebind (`allow_previous_response_recovery_rebind` / `allow_bootstrap_owner_rebind`) could legally land on a non-owner account and then injected the durable anchor — same wedge, different door. Now gated the same way.
- Real-bridge integration coverage answering the review's P1: the new test drives real session creation, framing, response_create gate and created-wait against a fake upstream that models the account scope of `previous_response_id` (accepts the alien anchor, answers with silence). Removing the guard reproduces the production wedge (5s POST timeout, gate held); keeping it settles via full-history resend, and a follow-up turn proves same-account continuity still anchors.
- Both declines log `cross_account_anchor_declined` with site + owning account — a concrete key for the #1580/#1630 detection path.
- The pre-binding `fresh_reattach_anchor_injected` site is documented as safe-by-construction (injection there makes the durable owner a required account) rather than double-guarded.

**Verification:** bridge suites 604 passed; full unit suite 5681 passed / 3 skipped; ruff + format clean; `openspec validate fix-cross-account-previous-response-anchor --strict` green.

Happy to fold this back into #1274's branch instead if @aididhaiqal prefers to land it under the original PR.

### DIFF
--- a/app/modules/api_keys/repository.py
+++ b/app/modules/api_keys/repository.py
@@ -380,6 +380,12 @@ class ApiKeysRepository:
     async def commit(self) -> None:
         await self._session.commit()
 
+    async def update_last_used(self, key_id: str, *, commit: bool = True) -> None:
+        """Compatibility touch for maintenance and durability checks."""
+        await self._session.execute(update(ApiKey).where(ApiKey.id == key_id).values(last_used_at=utcnow()))
+        if commit:
+            await self._session.commit()
+
     async def rollback(self) -> None:
         await self._session.rollback()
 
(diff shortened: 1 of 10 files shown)


Komzpa:
@codex review

PR #1664, opened 2026-08-08
feat(proxy): add OpenAI-compatible fallback for exhausted subscriptions

oformaniuk:
## Summary

Add an opt-in OpenAI-compatible overflow path for Responses traffic when every eligible ChatGPT subscription account is exhausted by upstream usage quota.

This reuses the existing dashboard-managed **Model Sources** capability rather than introducing fallback-specific environment variables or secret storage. Operators can designate one enabled Responses-capable source as the subscription fallback and optionally configure a model override; the existing Model Source API key remains encrypted with the normal credential path.

## Behavior

- fallback is attempted **only** after terminal aggregate `usage_limit_reached` from normal subscription account selection
- local capacity/fair-share/admission failures, arbitrary upstream 429s, auth/model errors, and other failures do not activate fallback
- `/backend-api/codex/responses` and `/v1/responses` are covered for streaming/non-streaming HTTP flows
- requested model is preserved unless the designated source has a configured fallback-model override
- API-key Model Source assignment scope is respected
- fallback provider errors are terminal and do not loop back into subscription selection

## Continuity / safety

Fallback is fail-closed for account-owned state. It reuses the existing Responses replay-safety machinery:

- fresh account-neutral requests can cross providers
- file-pinned/provider-owned hosted state is rejected
- requests using `previous_response_id` cross only when existing continuity evidence can project them into a self-contained fresh replay; the ChatGPT-owned anchor is stripped before forwarding

## Accounting

The existing downstream API-key reservation is transferred to the Model Source forwarding path on aggregate quota exhaustion. The fallback path reuses that reservation rather than running request admission a second time, preserving exactly-once settlement/release behavior.

## Configuration / UI

- add `is_subscription_fallback` and optional `fallback_model` to Model Sources
- enforce at most one designated fallback with both service transaction behavior and a partial unique DB index
- validate that a fallback is enabled, Responses-capable, and that any model override names an enabled model on the source
- expose the controls in the existing Model Source create/edit UI with localized copy
- stored API keys are not returned to the dashboard

## OpenSpec

Includes `openspec/changes/openai-api-fallback/` with proposal, context, completed task checklist, and a normative `responses-api-compat` delta.

## Validation

Fork-side verification passed:

- fallback unit tests: 11 passed
- fallback Responses route integration tests: 6 passed
- existing Model Source Responses routing tests: 2 passed
- designated-fallback replacement transaction regression: passed
- dashboard fallback-control regression tests: passed
- proxy architecture checks
- Ruff lint + format checks
- Python `ty` typecheck
- frontend typecheck and focused Model Source/i18n tests
- strict OpenSpec validation

Upstream `CI` and `Simplicity budgets` are currently reported by GitHub as `action_required` with no jobs created. This is the fork-workflow approval gate; a maintainer needs to approve the Actions run before upstream CI can execute.

## Scope

This change intentionally targets Responses HTTP routing. It does not add overflow routing for realtime/websocket, file, compact, image, or standalone Chat Completions paths.

## UI evidence

Screenshot 2026-08-09 at 4 34 21 PM

Screenshot 2026-08-09 at 4 34 05 PM


Soju06:
Thanks for this contribution — the design itself is well aligned with this repo's conventions (opt-in via the existing Model Sources capability, default off, no new environment variables, encrypted credentials reused, and an `openspec/changes/openai-api-fallback/` change included). The replay-safety and reservation-transfer reasoning in the spec delta is also solid. That said, there are four concrete blockers before this can merge:

**1. The migration must be replay-safe (this is what's breaking CI).**
All of the red CI legs (`Tests (pytest, unit)`, `integration-core-3`, `PostgreSQL`) fail with `duplicate column name: is_subscription_fallback` in the migration-contract tests (`tests/unit/test_db_migrate.py::test_run_upgrade_auto_remaps_legacy_revision_ids`, `tests/integration/test_migrations.py::test_run_startup_migrations_*`, `test_postgresql_migration_contract_policy_and_drift_match`, ...). These tests replay the migration chain against a schema that is already at head (legacy revision-id remap path), so every migration in this repo must guard against re-application. Your migration (`app/db/alembic/versions/20260807_163500_add_model_source_subscription_fallback.py`) uses bare `op.add_column` / `op.create_index`. Please follow the established pattern — see `app/db/alembic/versions/20260804_230000_add_request_log_connection_request_kind.py`, which inspects existing columns via `sa.inspect(op.get_bind())` and returns early when the column/index already exists. Apply the same guard to both columns and the partial unique index (and the symmetric guards in `downgrade()`).

**2. Rebase onto current main and re-chain the migration head.**
`main` has since gained `20260808_000000_tune_usage_history_autovacuum`, which has the same `down_revision` (`20260806_020000_add_usage_history_bulk_covering_indexes`) as your migration. Merging as-is would create two Alembic heads, which this repo forbids (single linear head — see AGENTS.md). After rebasing, please set your migration's `down_revision` to `20260808_000000_tune_usage_history_autovacuum`.

**3. Please address the open Codex review thread (reservation leak on fallback preparation failure).**
`_prepare_subscription_fallback(...)` is awaited at `app/modules/proxy/api.py:5081` (streaming) and `:5242` (non-streaming) after `_enforce_request_limits` has created the API-key reservation, but outside any cleanup block. If the fallback lookup or replay projection raises (e.g. a transient DB error), the request fails while the reservation stays held until stale-reservation cleanup — violating the exactly-once settlement/release invariant your own spec delta states. Simplest fix: wrap the call so a preparation failure degrades to "no fallback" (`fallback = None`) instead of propagating, or explicitly release the reservation before re-raising. Either way, please add a test for the preparation-failure path and resolve the review thread.

**4. Contributors attribution check.**
The `Contributors attribution` CI job fails because `oformaniuk` isn't listed yet — please add yourself via all-contributors as the job output instructs.

One note for the maintainer side: this PR carries the `db migration` label, so merge ordering will need coordination regardless — but items 1–4 are all author-actionable. Happy to re-review once CI is green and the Codex thread is resolved.

Soju06:
Thanks — the design holds up on the ownership/settlement invariants (terminal usage_limit_reached only, API-key source scope respected, account-neutral replay gating, reservation transfer instead of re-admission). Three things block this on the current head:

1. **Codex P1 (unresolved)** — if `_prepare_subscription_fallback` raises after `_enforce_request_limits`, the API-key reservation leaks (both `_stream_responses` and `_collect_responses`). Please treat preparation failure as no-fallback or explicitly release the reservation before propagating.
2. **Alembic re-parent** — `20260807_163500_add_model_source_subscription_fallback` revises `20260806_020000_...`, but main's `20260808_000000_tune_usage_history_autovacuum` already sits on that parent, so the graph forks and the legacy-remap migration tests fail with `duplicate column name: is_subscription_fallback` (unit, integration-core-3, PostgreSQL). Re-parent onto the current head `20260811_000000_add_hourly_rollup_cancelled_count`.
3. **Branch hygiene** — update-branch to clear the Beta release guard false positive, and add yourself to `.all-contributorsrc` for the Contributors attribution check.

Minor: consider gating `_prepare_subscription_fallback` behind a cheap configured-fallback check so every Responses request doesn't pay the replay-projection + DB lookup, and replace the `except AttributeError` probe with an explicit capability check. Ping for a fresh `@codex review` once pushed.

Soju06:
Correction to the re-parent target: #1657 just merged, so main's Alembic head is now `20260812_000000_merge_recovery_dispatch_and_hourly_cancelled_heads` — please re-parent `down_revision` onto that (not `20260811_000000_add_hourly_rollup_cancelled_count` as I said above).

Soju06:
Re-parent target update (second correction, sorry for the churn): #1618 just merged and its migration sits on the previous head, so main's Alembic head is now `20260806_000000_add_anonymous_telemetry` (note: the revision ID is dated earlier than its parent `20260812_000000_merge_recovery_dispatch_and_hourly_cancelled_heads` — go by the graph, not the filename date). Please point `down_revision` there, or add a merge revision, so the single-head CI check passes after update-branch.

Soju06:
Thanks for the update — I verified the reservation-leak P1 is now closed (both `_prepare_subscription_fallback` call sites release the reservation before re-raising) and the attribution check passes. Four things still block this:

1. **CI is red with two real defects in the new code.** `ty` fails at `app/modules/proxy/api.py:4238` — `_yield_response_as_public_stream` yields `str | memoryview[int]` but is annotated `AsyncIterator[str]` (decode the memoryview branch too). And the unit suite hangs: `tests/unit/test_proxy_errors.py::test_indefinite_recovery_stops_after_retry_output_then_transport_error` times out inside `_stream_response_error_events` (`api.py:6672`, the `while True:` loop) — the delayed-fallback error-handling change appears to have broken loop termination on the retry-output → transport-error sequence. PostgreSQL and integration-core-2 legs fail on the same head.
2. **Merge conflict.** The branch now conflicts with main in `app/modules/proxy/api.py` — please rebase/update-branch.
3. **Migration re-parent is stale again (third time, sorry).** Main's Alembic head has moved past `20260806_000000_add_anonymous_telemetry`: #1521 added `20260813_000000_add_file_account_pins` and #1680 added `20260812_120000_add_sticky_abandonment_scope`, which is the current head. Please point `down_revision` at `20260812_120000_add_sticky_abandonment_scope` after rebasing (as before: go by the graph, not filename dates).
4. **Branch hygiene.** The rewritten history carries ~50 scratch commits (`chore: capture fallback build failure`, `chore: probe fork actions`, ...). We'll squash-merge regardless, but please make sure no scratch artifacts remain in the tree.

The two minor suggestions from the last round (cheap configured-fallback pre-check before the replay projection, and replacing the `except AttributeError` probe with an explicit capability check) are still open — fine to fold into the same push. Ping for a fresh `@codex review` once CI is green on a rebased head.

Soju06:
Friendly ping — this is still blocked on the same items from the 08-16 round, since no new commits have landed on the branch (head is still af1a293 from 08-15):

1. **CI defects**: `ty` fails at `app/modules/proxy/api.py:4238` (`_yield_response_as_public_stream` yields `str | memoryview[int]` against an `AsyncIterator[str]` annotation — decode the memoryview branch), and the unit suite hangs in `tests/unit/test_proxy_errors.py::test_indefinite_recovery_stops_after_retry_output_then_transport_error` (loop termination in `_stream_response_error_events`, `api.py:6672`).
2. **Two open Codex P1 threads (08-16)**: (a) reservation transfer is disabled once the startup-probe context manager resets, so a late `usage_limit_reached` on a delivered stream releases the reservation mid-fallback; (b) when quota exhaustion arrives after `response.created` was already yielded, the fallback stream gets spliced into an already-started response — clients would see two response lifecycles in one SSE stream. Both need fixes + tests, and please resolve the threads.
3. **Rebase + migration re-parent (target moved again, sorry)**: main's Alembic head is now `20260816_000000_add_account_pending_deletion` — point `down_revision` there after rebasing. Note main has also merged significant rewrites of the Responses streaming path (SSE verbatim relay, pure ASGI), so the `api.py` conflict resolution will need the fallback splice/relay logic re-checked against the new relay code, not just a textual merge.
4. The minor suggestions (cheap configured-fallback pre-check before the replay projection; explicit capability check instead of `except AttributeError`) are still open — fine to fold in.

The feature direction remains good and the reservation-leak fix from the earlier round still holds. Happy to re-review as soon as a rebased head with green CI is up.

Soju06:
Thanks @oformaniuk — the product intent here is wanted, and the owner has decided to pursue it design-first rather than rebase this branch.

Why not a rebase: since this branch forked (2026-08-14) `app/modules/proxy/api.py` gained 2,500+ lines and `retry.py` 650+ (SSE parse-once verbatim relay, pure-ASGI middleware, retry-loop changes), the two confirmed P1s (reservation transfer scoped to the startup probe; `response.created` treated as non-substantive splicing a second lifecycle and breaking the indefinite-recovery fence) are structural to where the fallback hooks in, and the migration is parented on a superseded alembic head. Fixing those means moving the decision out of the retry loop entirely.

The design that supersedes this PR is tracked in #2123 (decide once at route admission via a read-only selector probe, dispatch through the hardened direct source route, pin threads that received source content, explicit fail-closed rules). It carries forward your product decisions (opt-in, default off, encrypted credentials, no env var / nav item), the trigger-narrowness wording, the key-scope enforcement, the openspec structure and the route-test matrix, with credit in the issue.

Closing this PR in favour of #2123. If you want to pick up any of the work packages there, comment on the issue.

PR #1681, opened 2026-08-10
docs(client-setup): add zsh shell wrapper for bi-directional session auto-toggling and retagging

guitarbeat:
### Description
When users frequently switch between direct OpenAI and `codex-lb` (e.g. running `uvx codex-lb` on demand), Codex CLI and the Desktop App filter session history based on the active `model_provider`. Switching between the two without re-tagging causes past chat sessions to temporarily disappear.

This PR updates `docs/client-setup.md` to document an automated shell wrapper function for `~/.zshrc` / `~/.bashrc`.

### Changes Included
- Added an automatic shell wrapper section under **Codex CLI / IDE Extension** setup.
- Intercepts `uvx codex-lb` to dynamically update `~/.codex/config.toml` (`model_provider = "codex-lb"`), `~/.codex/state_5.sqlite` (`UPDATE threads`), and `.jsonl` session files.
- Automatically restores `config.toml` and re-tags sessions back to `openai` on server exit (`EXIT`, `INT`, `TERM`) with idempotent single-execution cleanup.


Soju06:
Thanks — the pain point is real (provider toggling hides session history in `codex resume`), but this snippet can't go into the docs as written:

1. **It contradicts the section right above it.** docs/client-setup.md explicitly says to use the built-in `codex-lb codex-sessions retag` command *instead of editing Codex files by hand* — the CLI does dry-runs, backups, and rollback (see the runtime-portability context doc). This wrapper hand-edits `config.toml`, `state_5.sqlite`, and every `.jsonl` with raw sed/sqlite and `2>/dev/null` swallowing failures, which risks silent half-retagged or corrupted session stores (the sed pattern can also match `"model_provider":"openai"` inside message *content*).
2. **It's broken on Linux/WSL.** `sed -i ''` is BSD/macOS syntax; GNU sed treats `''` as the script and fails. Most of our documented setups are Docker/WSL.
3. Smaller issues: `state_5.sqlite` is a version-pinned filename that will drift with Codex releases; sqlite writes race a running Desktop App; `trap ... EXIT` inside a function behaves differently in bash vs zsh; and the `"$*" == *codex-lb*` match over-triggers.

If you'd like to pursue this, I'd welcome a reworked version where the wrapper simply calls `codex-lb codex-sessions retag --from ... --to ... --yes` on start/stop and only touches `model_provider` in `config.toml` — no direct sqlite/jsonl edits — with portable sed (or a small `python -c`) and a note about closing Codex first. Alternatively, open a feature request for a first-class `codex-lb codex-sessions toggle` subcommand and we can spec it properly via OpenSpec. Either way the Contributors attribution check needs you added to `.all-contributorsrc` before merge.

Komzpa:
@codex review

Soju06:
Thanks for the rework — this now matches the requested shape: config-only `model_provider` toggling via portable Python (no sed/sqlite/jsonl edits), retagging deferred to the built-in `codex-lb codex-sessions retag` with a close-Codex-first note, a dedicated `codex_lb_uvx` wrapper instead of overriding `uvx`, saved exit status, and provider restore only when this invocation changed it. All of the earlier review findings are resolved. Three new issues in the rewritten snippet, though:

1. **Broken under zsh (the headline shell).** `status` is a read-only special parameter in zsh, so `local status` / `status=$?` abort with `read-only variable: status` — reproducible with `zsh -c 'f() { local status; }; f'`. Please rename it (e.g. `rc`).
2. **Hard failure when `model_provider` is not explicitly set.** `_codex_set_provider` exits with "expected one top-level model_provider entry" when `~/.codex/config.toml` omits the key (fresh installs use the implicit `openai` default — `_codex_current_provider` already handles that case). Consider appending the line when `count == 0` instead of failing.
3. **The new test is never run.** `docs/tests/test_client_setup_docs.py` sits outside `testpaths = ["tests"]` in pyproject.toml, so pytest/CI never collects it. Move it under `tests/` (e.g. `tests/unit/test_client_setup_docs.py`) so the guard actually guards.

Minor: a one-line note that the snippet needs Python ≥ 3.11 (`tomllib`) would help. Happy to re-review after these.

Soju06:
Fresh full review of a2d55cd7 against current main. The rework is the right shape now — config-only toggling, portable `python3` + `tomllib`/`re`, no state-db or `.jsonl` edits, retagging deferred to `codex-lb codex-sessions retag` — but the head predates Soju06's latest review and all three of those findings still reproduce, plus the branch now conflicts with main:

1. **zsh abort confirmed (blocking).** `status` is a read-only special parameter in zsh; I reproduced the exact failure mode: the function dies at `status=$?` immediately after `command uvx codex-lb` returns, so the restore branch never runs and `~/.codex/config.toml` is left pointing at `codex-lb` — the stuck state this wrapper exists to prevent, in the shell the section is titled for. Rename to `rc` (fine in both zsh and bash).

2. **Hard fail when `model_provider` is absent (blocking).** `_codex_set_provider` raises `SystemExit` on `count != 1`, but the wrapper's target audience (direct-OpenAI users) commonly has no explicit top-level `model_provider` — `_codex_current_provider` already treats that as `openai`. Append the line when `count == 0` instead. Two more edges of the same regex worth fixing while you're in there: an inline comment after the value (`model_provider = "openai" # note`) defeats the `$` anchor and hard-fails too, and if the key only exists inside a `[profiles.*]` table, the `(?m)^` anchor will rewrite the profile's key rather than failing cleanly.

3. **The guard test never runs (blocking).** `pyproject.toml` has `testpaths = ["tests"]`, so `docs/tests/test_client_setup_docs.py` is never collected by pytest or CI — the green "Tests (pytest, unit)" check doesn't include it. Move it to `tests/unit/test_client_setup_docs.py`. (As a bonus, a `.py` file under `docs/` also gets copied into the built mkdocs site.)

4. **Merge conflict.** `.all-contributorsrc` and `README.md` now conflict with main (new contributor rows landed since your base). Please rebase; `docs/client-setup.md` itself merges cleanly.

Minor: a one-liner noting the snippet needs Python ≥ 3.11 (`tomllib`) would help WSL/older-distro users.

Happy to re-review once these land — the approach itself is accepted.

Soju06:
Second follow-up: two review rounds with a reproducing blocking bug in the shipped snippet and no author engagement yet. Docs-only so no rush, but if the 08-16 findings stay unanswered by the next triage round this will be closed as unresponsive — comment if you want it kept open.

Soju06:
Closing as unresponsive, per the 08-19 notice. The reworked approach was accepted (config-only toggling, retag deferred to the built-in CLI), but the three blocking findings from the 08-16 review still reproduce on the current head (a2d55cd7): the zsh `local status` read-only abort that leaves `config.toml` stuck on `codex-lb`, the hard `SystemExit` when `model_provider` is absent from a fresh config, and the guard test under `docs/tests/` that pytest never collects (`testpaths = ["tests"]`). The branch has also picked up merge conflicts in `.all-contributorsrc` / `README.md`.

The pain point is real and the direction is welcome. If you'd like to continue: fix the three items (rename `status` to `rc`, append the key when `count == 0`, move the test to `tests/unit/`), rebase, and comment here — we'll reopen. Alternatively, open a feature request for a first-class `codex-lb codex-sessions toggle` subcommand and we can spec it via OpenSpec. Thanks for the work so far.

PR #903, opened 2026-06-03
fix(proxy): bypass HTTP bridge for input images

Soju06:
## Summary

- Bypass the HTTP responses bridge for Responses requests that contain `input_image` parts and force those requests onto the upstream HTTP stream path
- Detect top-level images, nested message-content images, and tool-output image parts
- Reject unsupported `input_image.file_id` / `sediment://` references even when they appear inside tool-output image parts
- Preserve raw-path external image URL inlining for top-level `input_image` items
- Add regression coverage for image-shape detection, bridge bypass, preserving later text-only bridge behavior, and top-level external image URL inlining
- Add OpenSpec coverage for the image bridge bypass behavior

## Why

During live beta validation, an invalid inline image failed quickly on the raw HTTP path but could occupy a HTTP bridge pending slot until local bridge timeouts. Image-bearing requests should favor upstream HTTP image validation and error semantics over bridge prompt-cache/session affinity.

## Test Plan

- `uv run ruff check app/core/openai/requests.py app/core/clients/proxy.py app/modules/proxy/service.py tests/unit/test_openai_requests.py tests/unit/test_input_image_proxy.py tests/unit/test_proxy_utils.py`
- `uv run pytest tests/unit/test_openai_requests.py::test_extract_input_image_file_references_collects_multi_message_paths -q`
- `uv run pytest tests/unit/test_proxy_utils.py::test_responses_request_contains_input_image_detects_supported_shapes -q`
- `uv run pytest tests/unit/test_proxy_utils.py::test_raise_for_unsupported_input_image_references_rejects_tool_output_file_id -q`
- `uv run pytest tests/unit/test_proxy_utils.py::test_stream_http_bridge_or_retry_bypasses_bridge_for_input_image -q`
- `uv run pytest tests/unit/test_openai_requests.py tests/unit/test_input_image_proxy.py tests/unit/test_proxy_utils.py -q`
- `uv run ty check app/core/openai/requests.py app/core/clients/proxy.py app/modules/proxy/service.py tests/unit/test_openai_requests.py tests/unit/test_input_image_proxy.py tests/unit/test_proxy_utils.py`
- `uv run openspec validate bypass-http-bridge-input-images --strict`

## Live Validation

Validated against the beta candidate environment before splitting this into a clean main-based PR:

- invalid 1x1 inline PNG returns `400 invalid_value` in ~0.5s instead of timing out
- valid 64x64 inline PNG completes successfully with output `Red`
- text-only requests continue to use the HTTP responses bridge

PR #1690, opened 2026-08-10
fix(proxy): allow validated inline images on the HTTP responses bridge

Heyaaz:
## Summary

Since #903, every Responses request carrying an `input_image` part bypasses the HTTP responses bridge and is forced onto the raw HTTP stream path. That means one historical inline image (e.g. an agent screenshot read mid-task) drops a long coding session onto HTTP for every later turn, losing the bridge's prompt-cache reuse and `previous_response_id` session continuity — even though the image is a perfectly valid inline `data:` URL that the upstream WebSocket accepts.

This PR makes the bypass conditional: the HTTP responses bridge may serve image-bearing requests when **all** of:

- the serialized payload fits the WebSocket transport payload budget (`_ws_transport_payload_budget_bytes`);
- every `input_image` part is an inline `data:image/<type>;base64,...` URL;
- the base64 decodes and the bytes match the declared media type with parseable dimensions (PNG IHDR / GIF header / JPEG SOF / WebP VP8X/VP8/VP8L);
- each dimension is ≥ 64px and total pixels ≤ 100M (degenerate-image guard).

Any failed check preserves the #903 behavior exactly: raw HTTP stream path with `upstream_stream_transport_override="http"`, so invalid images still fail fast with upstream validation errors instead of occupying bridge pending slots. `image_generation` tool requests keep the unconditional bypass. `_raise_for_unsupported_input_image_references` still runs before this decision.

## Why

Live validation on codex-lb 1.22.0: image-heavy sessions over HTTP cached ~20–43% of input tokens while equivalent text-only WebSocket traffic cached ~93%; the same inline-image request over the WebSocket bridge cached 89% on the second call. OpenAI's Responses WebSocket mode accepts inline `data:image/...` inputs, and the bridge already carries the external-URL inlining logic (`_inline_http_bridge_image_urls`).

## Test Plan

- `uv run ruff check app/modules/proxy/_service/http_bridge/streaming.py app/modules/proxy/service.py tests/unit/test_proxy_utils.py`
- `uv run ty check app/modules/proxy/_service/http_bridge/streaming.py app/modules/proxy/service.py tests/unit/test_proxy_utils.py`
- `uv run pytest tests/unit/test_proxy_utils.py -q` (993 passed)
- New unit coverage: validation helper accepts real inline PNG (nested + tool output), rejects tiny/invalid/external images; bridge routing uses the bridge for a valid 64×64 PNG and bypasses to HTTP for a 1×1 PNG. Existing #903 regression tests pass unchanged.

## Live Validation

- Valid inline PNG → `upstream_transport=websocket`; second identical request `cached_tokens` 7,936/8,916 (~89%).
- Invalid base64 image → HTTP 400 `invalid_value` in ~0.45 s.
- 1×1 PNG → HTTP transport (dimension guard).

### DIFF
--- a/app/modules/proxy/_service/http_bridge/streaming.py
+++ b/app/modules/proxy/_service/http_bridge/streaming.py
@@ -1,6 +1,8 @@
 from __future__ import annotations
 
 import asyncio
+import base64
+import binascii
 import dataclasses
 import json
 import logging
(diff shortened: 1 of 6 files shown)


Heyaaz:
Closing per requester instruction — local patch only.

PR #1706, opened 2026-08-12
fix(proxy): terminate suppressed duplicate tool-call replays

Komzpa:
Fixes duplicate same-response tool-call replay handling without leaving clients nonterminal. Suppressed duplicates now receive `response.failed` with `duplicate_tool_call_replay_suppressed` across direct SSE, HTTP bridge, and WebSocket; persistence, settlement, reconnect, and account health remain coherent. The unrelated `shell_command` dedupe expansion was removed from this PR as requested in review.\n\nValidation: 25 focused tests, ruff, ty, strict OpenSpec validation, and scope gate.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved replay validation to prevent requests from being sent through an incorrect account.
  * Duplicate tool-call replays now return a clear `response.failed` result with a dedicated error code.
  * Preserved request completion, operation persistence, and reconnect behavior across supported transport types.
  * Prevented suppressed duplicate replays from incorrectly affecting account health.
* **Tests**
  * Added regression coverage for duplicate tool-call replay handling over HTTP bridge, WebSocket, and SSE.

### DIFF
--- a/app/modules/proxy/_service/streaming/mixin.py
+++ b/app/modules/proxy/_service/streaming/mixin.py
@@ -921,11 +921,11 @@ class _StreamingMixin(_StreamingRetryMixin):
                         event_type,
                     ) = _facade()._build_rewritten_stream_response_failed_event(
                         response_id=response_id,
-                        error_code="stream_incomplete",
+                        error_code=_facade()._SUPPRESSED_DUPLICATE_TOOL_CALL_ERROR_CODE,
                         error_message=_facade()._SUPPRESSED_DUPLICATE_TOOL_CALL_MESSAGE,
                     )
                     status = "error"
-                    error_code = "stream_incomplete"
+                    error_code = _facade()._SUPPRESSED_DUPLICATE_TOOL_CALL_ERROR_CODE
                     error_message = _facade()._SUPPRESSED_DUPLICATE_TOOL_CALL_MESSAGE
                     settlement.record_success = False
                     settlement.account_health_error = False
--- a/app/modules/proxy/_service/websocket/helpers.py
+++ b/app/modules/proxy/_service/websocket/helpers.py
@@ -1386,7 +1386,7 @@ def _rewrite_websocket_suppressed_duplicate_tool_call_completion_event(
     request_state: _WebSocketRequestState,
 ) -> tuple[OpenAIEvent | None, dict[str, JsonValue] | None, str | None, str]:
     rewritten_event_payload = response_failed_event(
-        "stream_incomplete",
+        _facade()._SUPPRESSED_DUPLICATE_TOOL_CALL_ERROR_CODE,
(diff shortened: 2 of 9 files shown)


Komzpa:
@codex review

Soju06:
Thanks — the `shell_command` dedupe coverage (tool_call_safety.py, tool_call_dedupe.py + the unit test) and the `_ring_membership` monkeypatch hygiene fix both look good and would merge cleanly on their own.

I have concerns about the terminal-suppression half, though. Replacing the synthetic `response.failed("stream_incomplete")` with full downstream suppression changes more than the visible "echo":

1. **WS transport clients now hang** (`app/modules/proxy/_service/websocket/mixin.py:4915`). The request_state has already been popped by `_pop_terminal_websocket_request_state` and is finalized right after, and `reconnect_requested` only retires the *upstream* socket. With `suppress_downstream_event=True` and no rewrite, the downstream websocket client never receives any terminal event for its `response.create` — it waits until its own timeout instead of getting an immediately retryable failure. Something terminal still needs to reach the client on this path.

2. **Durable operations never reach a terminal state** (`app/modules/proxy/_service/http_bridge/upstream_events.py:2640`). Gating `_persist_http_bridge_operation_event(..., terminal=True, terminal_state=...)` behind the new flag means the spooled operation stays non-terminal forever. On main the rewritten event persisted `state="failed"`, so a durable-bridge reconnect/poll could settle; now a resuming client replays only non-terminal events and waits indefinitely.

3. **Accounting is no longer fenced**, contrary to the PR summary. Because `event`/`payload`/`event_type` stay `response.completed`, `_finalize_websocket_request_state` takes the completed branch: `status="success"`, `record_success=True`, success request log — for a turn whose client-visible outcome is a truncated stream. `error_http_status_override=502` is set but nothing consumes it anymore (the `http_bridge/streaming.py:4293` propagation only triggers on a `response.failed` block, which no longer arrives).

4. **Transport divergence**: `app/modules/proxy/_service/streaming/mixin.py:910` (direct SSE path) still emits the synthetic failed terminal for the same scenario, so the three transports now behave differently. (The ruff command in the PR body lists `streaming/mixin.py`, but the diff doesn't touch it — was a hunk dropped?)

If the goal is to stop showing an error block for an already-suppressed duplicate, could we keep a client-visible terminal but make it benign — e.g. forward the (response-id-rewritten) terminal, or keep the failed terminal only where the client hasn't already settled the turn — while still persisting a terminal operation state and keeping the request log marked as a fenced non-success? Happy to re-review; splitting the `shell_command` dedupe addition into its own PR would also let that part land immediately.

Soju06:
Status check: this is blocked on a design conversation, not a rebase — the benign-terminal alternative and the suggestion to split the dedupe into its own immediately-landable PR are both waiting on your take. The dedupe split could merge quickly if you want to peel it off.

PR #1720, opened 2026-08-13
fix(compact): harden replay recovery

Komzpa:
## Summary

- sanitize plaintext compaction replays while preserving opaque provider-encrypted state
- preserve valid tool-search and compact replay pairs without allowing owner-bound state to move accounts
- recover post-compaction bridge replays with bounded, fail-closed continuity handling

## Scope

Absorbs #1744. Its active-skill recovery and bridge replay safeguards are now in this branch, rebased on current `main`; #1744 will close once this current head passes CI.

## Validation

- `uv run ruff check` on changed Python paths
- `uv run ty check` on changed runtime paths
- `openspec validate preserve-opaque-compaction-replays --strict`
- `openspec validate recover-post-compact-bridge-replays --strict`
- compact/replay unit and integration suites, including bridge owner-affinity regressions

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved conversation recovery after compaction, preserving encrypted context and usable summaries.
  * Prevented duplicate or incomplete tool-search history during continued requests and retries.
  * Improved reconnect behavior while maintaining required session continuity and account restrictions.
  * Safer handling of replay boundaries reduces failures and prevents invalid historical data from being resent.
  * Corrected service-tier normalization and fallback behavior.

* **Documentation**
  * Added compatibility requirements and recovery guidance for compaction and tool-search replay scenarios.

### DIFF
--- a/app/core/openai/requests.py
+++ b/app/core/openai/requests.py
@@ -944,18 +944,20 @@ def _plaintext_compaction_replay_replacement(item: JsonValue) -> JsonValue | Non
         return None
     assert is_json_mapping(item)
     encrypted_content = item.get("encrypted_content")
-    if isinstance(encrypted_content, str) and _looks_like_provider_encrypted_content(encrypted_content):
+    if isinstance(encrypted_content, str):
         return None
 
     text = _compaction_plaintext_summary(item)
     if text is None:
-        return {"role": "assistant", "content": _COMPACT_STATE_TEXT_PREFIX + "[unverified compact state omitted]"}
-    return {"role": "assistant", "content": _COMPACT_STATE_TEXT_PREFIX + text}
+        return _assistant_compact_state_item("[unverified compact state omitted]")
+    return _assistant_compact_state_item(text)
 
 
-def _looks_like_provider_encrypted_content(value: str) -> bool:
-    stripped = value.strip()
-    return stripped.startswith("gAAAA") and len(stripped) >= 80
+def _assistant_compact_state_item(text: str) -> JsonValue:
+    return {
+        "role": "assistant",
+        "content": [{"type": "output_text", "text": _COMPACT_STATE_TEXT_PREFIX + text}],
+    }
 
 
 def _compaction_plaintext_summary(item: Mapping[str, JsonValue]) -> str | None:
--- a/app/core/openai/requests.py
+++ b/app/core/openai/requests.py
@@ -48,6 +48,7 @@ _COMPACT_TOOL_CALL_TYPE_BY_OUTPUT_TYPE: dict[str, str] = {
     "function_call_output": "function_call",
     "custom_tool_call_output": "custom_tool_call",
     "apply_patch_call_output": "apply_patch_call",
+    "tool_search_output": "tool_search_call",
 }
 _COMPACT_TOOL_CALL_ITEM_TYPES = frozenset(_COMPACT_TOOL_CALL_TYPE_BY_OUTPUT_TYPE.values())
 _COMPACT_TOOL_CALL_OUTPUT_ITEM_TYPES = frozenset(_COMPACT_TOOL_CALL_TYPE_BY_OUTPUT_TYPE)
--- a/app/core/openai/requests.py
+++ b/app/core/openai/requests.py
@@ -869,6 +869,7 @@ def strip_replayed_tool_call_namespaces_from_payload(payload: MutableJsonObject)
 
 
 _POISONED_LOCAL_COMPACT_FALLBACK_TEXT = "Local compact fallback preserved the latest encrypted reasoning state."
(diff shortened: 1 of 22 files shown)


Komzpa:
@codex review

Komzpa:
@codex review

Soju06:
Thanks — the compaction-replay core of this PR looks right now: preserving any string `encrypted_content` verbatim (no ciphertext-shape heuristics) and rewriting only plaintext summary/text/content shapes into canonical `output_text` assistant items resolves both live failure modes, and the iteration through the review rounds (opaque envelopes, canonical content, owner-bound fail-closed, server-executed tool-search rejection) reads well. A few things block merge as-is:

1. **CI is red because of an orphaned test.** `tests/integration/test_http_responses_bridge.py:6992` (`test_retry_account_neutral_precreated_request_switches_from_silent_account`, added in c5b5f9da47) asserts that the silent-account retry clears `preferred_account_id` and excludes the lane's account — but db7ce90fb4 reverted exactly that behavior in `request_submit.py` (correctly, per the recovery-lane review comment). The test now fails deterministically (`excluded_account_ids == set()`). Please delete it, or rewrite it to assert the current invariant (lane keeps its account, `require_preferred_account=True`), which would actually lock in the fix from that round.

2. **Empty-string `encrypted_content` edge.** `_plaintext_compaction_replay_replacement` (`app/core/openai/requests.py:943`) preserves any `str` `encrypted_content`, including `""`. An empty encrypted payload will still be replayed upstream and hit the invalid-encrypted-content failure this PR targets. Suggest `isinstance(encrypted_content, str) and encrypted_content`, matching the non-empty check the old poison-skip path used.

3. **Dead code in replay_safety.** `_compaction_item_is_self_contained` (`replay_safety.py:671`) is constant `False`, which makes the new `compaction` allowlist entries and the `continue` at line 1055 unreachable. The fail-closed outcome is what we want, but stating it directly (reject `item_type == "compaction"`) would be clearer than a predicate that implies some shapes could qualify.

4. **Scope and description drift.** The PR body still claims the diff touches only `requests.py` + its unit tests and cites test names that no longer exist, but the branch now also carries a full tool_search replay feature (previous-response trimming on both bridges, account-neutral classification) and edits an *archived* openspec change to document it. Could you update the body to match, and move the tool-search spec delta out of `openspec/changes/archive/2026-08-13-...` into a live change folder (as you already did for the compaction contract)? Splitting the tool-search work into its own PR would also make both easier to land, but at minimum the description and archive edit need fixing.

Happy to re-review once CI is green.

Soju06:
Rebase nudge: not superseded by #1809 — that changed the compact upstream transport/trigger path while this sanitizes plaintext compaction items on client replay. But #1809 and the http_bridge restructure moved the surfaces underneath you, so the rebase needs a semantic pass, and note #1815 (malformed compact item ids) is about to land on the same test file + spec surface — rebase after it merges to avoid doing it twice.

Soju06:
Thanks for the rebase onto beta.2 — CI is fully green now, the orphaned silent-account retry test is gone, and the core contract (preserve any string `encrypted_content` verbatim, rewrite only plaintext summary/text/content shapes into canonical `output_text` assistant items) is exactly right against current main, including post-#1809/#1815 surfaces. Three items from my 08-16 review are still open at d270f66 and remain the merge blockers:

1. **Empty-string `encrypted_content`.** `_plaintext_compaction_replay_replacement` still preserves `encrypted_content: \"\"` (`if isinstance(encrypted_content, str): return None`), so an empty encrypted payload is replayed upstream verbatim and reproduces the invalid-encrypted-content failure this PR targets. Please use `isinstance(encrypted_content, str) and encrypted_content`, matching the non-empty check the old poison-skip path used, and add a regression test for the empty-string shape.

2. **Dead code in replay_safety.** `_compaction_item_is_self_contained` is still constant `False`, which keeps the new `compaction` allowlist entries and the `continue` in `_contains_account_scoped_input_state` unreachable. The fail-closed behavior is correct — just state it directly (reject `item_type == \"compaction\"`) instead of a predicate that implies some shapes could qualify.

3. **Scope and description drift.** The body still claims the diff is only `requests.py` + its unit tests and cites test names that no longer exist, while the branch carries the full tool_search replay feature and edits the *archived* `2026-08-13-classify-tool-search-missing-tool-output` change. Archived changes are immutable history — please move that delta into a live change folder (as you already did with `preserve-opaque-compaction-replays`) and update the body to match the real diff. Splitting the tool-search work into its own PR is still my preference; it is independently mergeable and would let the compaction fix land immediately once items 1–2 are addressed.

Happy to do a final pass as soon as these land — the hard part of this PR is done.

Komzpa:
Superseded by #1849, the current-main maintainer carrier that absorbs the still-needed replay-safe compaction behavior from #1720 and #1744.

PR #1730, opened 2026-08-13
fix(proxy): demote quarantined bridge reattach keys

Komzpa:
## Summary
- route quarantined full-resend HTTP bridge reattaches onto a fresh account-neutral replay key
- clear stale session affinity inputs before session creation so poisoned hard keys cannot rebuild the same wedged bridge
- tighten the quarantine regression test to assert the fresh key and stripped headers

## Validation
- `uv run --frozen pytest -q tests/unit/test_proxy_http_bridge.py::test_stream_http_bridge_quarantined_full_resend_stays_unanchored_when_reattach_gate_already_false`
- `uv run --frozen pytest -q tests/unit/test_proxy_http_bridge.py -q`

## Notes
This is the focused public carrier for the post-compaction live hang class where a quarantined hard session key could suppress the durable anchor but still create/reuse the same poisoned bridge key.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recovery of interrupted HTTP bridge sessions while preventing stale recovery events from clearing newer quarantine states.
  * Added safer replay handling for quarantined requests, including account-neutral recovery when durable request validation succeeds.
  * Preserved session continuity, ownership, and response metadata during quarantine clearing and reconnection.
* **Tests**
  * Expanded coverage for restart recovery, session reclamation, replay safeguards, and generation-based quarantine behavior.

### DIFF
--- a/app/modules/proxy/_service/http_bridge/quarantine.py
+++ b/app/modules/proxy/_service/http_bridge/quarantine.py
@@ -43,6 +43,7 @@ class _HTTPBridgeQuarantineEntry:
     consecutive_eventless_timeouts: int = 0
     last_touched_monotonic: float = 0.0
     reason: str | None = None
+    generation: int = 0
 
 
 def _http_bridge_quarantine_registry(
@@ -100,6 +101,16 @@ def _http_bridge_session_key_quarantined(service: Any, key: _HTTPBridgeSessionKe
     return entry is not None and entry.quarantined_until > now
 
 
+def _http_bridge_session_key_quarantine_generation(service: Any, key: _HTTPBridgeSessionKey) -> int | None:
+    registry = _http_bridge_quarantine_registry(service)
+    now = time.monotonic()
+    _prune_http_bridge_quarantine_registry(registry, now)
+    entry = registry.get(key)
+    if entry is None or entry.quarantined_until <= now:
+        return None
+    return entry.generation
+
+
 def _quarantine_http_bridge_session(service: Any, session: _HTTPBridgeSession, *, reason: str) -> None:
     """Quarantine a bridge session that has proven silent/wedged.
 
@@ -113,6 +124,7 @@ def _quarantine_http_bridge_session(service: Any, session: _HTTPBridgeSession, *
     entry.quarantined_until = max(entry.quarantined_until, now + _HTTP_BRIDGE_QUARANTINE_TTL_SECONDS)
     entry.last_touched_monotonic = now
     entry.reason = reason
+    entry.generation += 1
     _prune_http_bridge_quarantine_registry(registry, now)
     session.quarantined = True
     if already_quarantined:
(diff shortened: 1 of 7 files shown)


Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Soju06:
Thanks — the quarantine-demotion core of this PR is in good shape after the review iterations: the generation fence, the projected account-neutral dispatch, the deterministic replay nonce, and the claim→rebind(clear_continuity)→renew advance of the original durable row all trace correctly, and the test coverage for them is thorough. Two blocking items remain in the bundled slimming/namespace scope, plus a scoping request:

1. **Orphaned archived-spec rewrite.** `openspec/changes/archive/2026-08-13-fix-replayed-namespaced-function-call/specs/responses-api-compat/spec.md` is rewritten to say replayed tool-call namespaces "MUST be preserved" on live Responses egress, but the final head no longer contains the serializer change that would implement it: `app/core/openai/requests.py::to_payload()` still strips namespaces by default, the live spec (`openspec/specs/responses-api-compat/spec.md:4595`) still says "MUST omit", and this PR's own integration test (`test_v1_responses_http_bridge_preserves_agent_wait_output_through_wire_normalization`) asserts the upstream `function_call` item has no `namespace` key. Please revert the archive edit (it's the historical record of what actually landed); if the requirement is genuinely being flipped, that should be its own openspec change with the serializer work.

2. **Core WebSocket slimmer not covered.** The new live-spec requirement ("Historical response.create slimming MUST preserve outputs for namespaced agent-control calls") is only implemented in the service-facade slimmer. The duplicate `_slim_historical_response_input_item` in `app/core/clients/proxy.py` (~L2244), used when the HTTP session bridge is disabled, still replaces `collaboration`/`multi_agent_v1` outputs with the omission notice — the last completed Codex round flagged this and the final push didn't address it (the follow-up review errored out). Either extend the protection there too or scope the spec text to the bridge path.

3. **Scope.** The title/body describe a focused quarantined-reattach fix, but the PR also carries the agent-control slimming protection (+ two spec files), a `_matching_websocket_request_states_for_previous_response_error` heuristic change, and test-flake fixes. Could you split the slimming/spec work into its own PR? The quarantine fix looks close to merge-ready on its own, and separating them keeps the spec question from blocking it.

Minor, non-blocking: `_AGENT_CONTROL_TOOL_NAMES` matches on bare name with no namespace, so a user tool that happens to be called `send_input` gets its large outputs permanently exempted from slimming and can newly fail size enforcement — consider requiring the namespace for the name-only legacy case or documenting the tradeoff. Also, in the reroute branch the `fresh_reattach_anchor_skipped_quarantined` event now logs the freshly minted recovery key because `bridge_session_key` is reassigned before the log call; logging the original quarantined key (with the new key in `detail=`) would keep the event useful for tracing.

Soju06:
Rebase nudge: per the 08-16 assessment this is close to mergeable — two scoped blockers plus the post-0816 conflict. Not superseded by anything on main; a rebase + those two fixes gets it over the line.

Soju06:
Thanks for the rebase onto beta.2 — CI is fully green and the quarantine-demotion core still traces correctly. However, checking head f38b650, both blocking items from the 08-16 assessment came through the rebase unchanged:

1. The archived-spec rewrite (`openspec/changes/archive/2026-08-13-fix-replayed-namespaced-function-call/specs/responses-api-compat/spec.md`) is still present. On current main, `app/core/openai/requests.py::to_payload()` still strips replayed tool-call namespaces by default, and this PR's own integration test (`test_v1_responses_http_bridge_preserves_agent_wait_output_through_wire_normalization`) asserts the upstream `function_call` item has no `namespace` key — so the rewritten archive describes behavior that never shipped. Please revert the archive edit; if the requirement is genuinely being flipped, that needs its own openspec change with the serializer work.

2. The new live-spec MUST ("Historical response.create slimming MUST preserve outputs for namespaced agent-control calls") is still only implemented in the service-facade slimmer. The duplicate `_slim_historical_response_input_item` in `app/core/clients/proxy.py` (~L2740), used when the HTTP session bridge is disabled, still replaces `collaboration`/`multi_agent_v1` outputs with the omission notice. Either extend the protection there or scope the spec text to the bridge path.

And the scoping ask stands: the quarantined-reattach fix itself is essentially merge-ready — splitting the slimming/spec work (plus the `_matching_websocket_request_states_for_previous_response_error` heuristic and flake fixes) into a follow-up PR would let us land the quarantine fix immediately. If splitting is inconvenient, say the word and we can carve out the quarantine core on our side with attribution preserved.

(Also noting the `db migration` label appears stale — this head touches no `migrations/` files — so I'll drop it.)

PR #1732, opened 2026-08-13
fix(proxy): keep hard continuity owners ahead of affinity

Komzpa:
## Summary
- allow compact turn-state live/durable aliases that agree on the same account to resolve the compact owner instead of failing with continuity_owner_conflict
- drop session/header affinity once compact or direct HTTP stream has a hard continuity owner, so image-capable live requests that bypass the bridge cannot reintroduce stale locality
- normalize live usage snapshots to persisted account IDs before writing usage_history, avoiding FK failures when an upstream chatgpt-account-id appears on the serving path
- update compact, direct-stream, and live-usage regressions

## Validation
- `uv run pytest tests/unit/test_proxy_utils.py::test_plain_stream_resolves_http_bridge_turn_state_owner tests/unit/test_live_usage_ingest.py::test_ingestor_resolves_upstream_account_id_before_history_write tests/unit/test_live_usage_ingest.py::test_stream_responses_tap_publishes_rate_limit_events`
- `uv run pytest tests/unit/test_proxy_utils.py::test_compact_turn_state_owner_never_falls_back_to_unscoped_sticky_sessions tests/unit/test_proxy_utils.py::test_compact_turn_state_owner_is_a_strict_selection_constraint tests/unit/test_proxy_utils.py::test_plain_stream_resolves_http_bridge_turn_state_owner tests/unit/test_proxy_utils.py::test_compact_file_pin_overrides_session_and_prompt_cache_locality tests/unit/test_live_usage_ingest.py::test_ingestor_resolves_upstream_account_id_before_history_write tests/unit/test_live_usage_ingest.py::test_stream_responses_tap_publishes_rate_limit_events`
- `git diff --check`

## Live note
Initial live recovery at `codex-lb:live-stack-790c93e` / `790c93ef2f0b0d94aac1fb72b0f249304271fb6b` removed the compact owner-alias failure, but a later live image-capable stream still hit `continuity_owner_conflict` after bypassing the HTTP bridge and live usage ingest logged `sqlite3.IntegrityError: FOREIGN KEY constraint failed`. This PR now includes those follow-up fixes; live redeploy/readback will be tracked separately from PR CI.

## Summary by CodeRabbit

- **New Features**
  - Improved live usage attribution by resolving account identities more accurately and consolidating duplicate activity.
  - Added safeguards to prevent ambiguous, conflicting, or outdated account matches from being recorded.

- **Bug Fixes**
  - Improved handling of session continuity conflicts with clearer retry behavior and more accurate routing.
  - Preserved compatibility for legacy session ownership scenarios while supporting durable turn-state ownership.

- **Reliability**
  - Added database indexing to improve account identity lookup performance.
  - Expanded coverage for account resolution, usage ingestion, streaming, and continuity conflict scenarios.

### DIFF
--- a/app/db/alembic/versions/20260814_000000_add_accounts_chatgpt_identity_index.py
+++ b/app/db/alembic/versions/20260814_000000_add_accounts_chatgpt_identity_index.py
@@ -1,0 +1,32 @@
+"""add account ChatGPT identity lookup index
+
+Revision ID: 20260814_000000_add_accounts_chatgpt_identity_index
+Revises: 20260816_000000_add_model_source_embeddings
+Create Date: 2026-08-14
+"""
+
+from __future__ import annotations
+
+import sqlalchemy as sa
+from alembic import op
+
+revision = "20260814_000000_add_accounts_chatgpt_identity_index"
+down_revision = "20260816_000000_add_model_source_embeddings"
+branch_labels = None
+depends_on = None
+
+_INDEX = "idx_accounts_chatgpt_account_id"
+_TABLE = "accounts"
+
+
+def upgrade() -> None:
+    bind = op.get_bind()
+    if bind.dialect.name == "postgresql":
+        with op.get_context().autocommit_block():
+            op.execute(sa.text(f"CREATE INDEX CONCURRENTLY IF NOT EXISTS {_INDEX} ON {_TABLE} (chatgpt_account_id)"))
+    else:
+        op.execute(sa.text(f"CREATE INDEX IF NOT EXISTS {_INDEX} ON {_TABLE} (chatgpt_account_id)"))
+
+
+def downgrade() -> None:
+    op.drop_index(_INDEX, table_name=_TABLE, if_exists=True)
--- a/app/db/models.py
+++ b/app/db/models.py
@@ -2160,6 +2160,7 @@ Index(
     postgresql_where=text("delete_requested_at IS NOT NULL"),
     sqlite_where=text("delete_requested_at IS NOT NULL"),
 )
+Index("idx_accounts_chatgpt_account_id", Account.chatgpt_account_id)
 Index("idx_api_keys_name", ApiKey.name)
 Index("idx_logs_account_time", RequestLog.account_id, RequestLog.requested_at)
 Index("idx_logs_model_source_time", RequestLog.model_source_id, RequestLog.requested_at)
--- a/app/modules/proxy/_service/compact.py
+++ b/app/modules/proxy/_service/compact.py
@@ -731,12 +731,11 @@ class _CompactMixin:
                 if owner_account_id == resolved_owner and session_identity is not None
             }
             if len(session_identities) > 1:
-                sources = ", ".join(source for source, _account_id, _session_id in owner_refs)
                 raise ProxyResponseError(
                     502,
                     openai_error(
                         "continuity_owner_conflict",
-                        f"Account-owned continuity sources conflict ({sources}); retry the logical turn.",
+                        "Turn-state owner sessions conflict; retry the logical turn.",
                         error_type="server_error",
                     ),
                 )
--- a/app/modules/proxy/_service/streaming/retry.py
+++ b/app/modules/proxy/_service/streaming/retry.py
@@ -1409,16 +1409,27 @@ class _StreamingRetryMixin:
                         )
                         return
                     if require_preferred_account and preferred_account_id is not None:
+                        error_code = "previous_response_owner_unavailable"
                         message = "Previous response owner account is unavailable; retry later."
+                        reason = "owner_account_unavailable"
+                        upstream_error_code = "no_accounts"
+                        if selection.error_code == "continuity_owner_conflict":
+                            error_code = "continuity_owner_conflict"
+                            message = (
+                                selection.error_message
+                                or "Account-owned continuity sources conflict; retry the logical turn"
+                            )
+                            reason = "owner_conflict"
+                            upstream_error_code = selection.error_code
                         _record_continuity_fail_closed(
                             surface="http_stream",
-                            reason="owner_account_unavailable",
+                            reason=reason,
                             previous_response_id=payload.previous_response_id,
                             session_id=headers.get("x-codex-turn-state") or headers.get("session_id"),
-                            upstream_error_code="no_accounts",
+                            upstream_error_code=upstream_error_code,
                         )
                         event = response_failed_event(
-                            "previous_response_owner_unavailable",
+                            error_code,
                             message,
                             response_id=request_id,
                         )
@@ -1430,7 +1441,7 @@ class _StreamingRetryMixin:
                             model=payload.model,
                             latency_ms=int((time.monotonic() - start) * 1000),
                             status="error",
-                            error_code="previous_response_owner_unavailable",
+                            error_code=error_code,
                             error_message=message,
                             reasoning_effort=payload.reasoning.effort if payload.reasoning else None,
                             transport=request_transport,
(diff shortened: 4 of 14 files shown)


Komzpa:
@codex review

Komzpa:
@codex review

Soju06:
Thanks — the surviving core of this PR (live-usage identity normalization + the continuity_owner_conflict error-code surfacing) fixes a real prod failure and the integration coverage is thorough. A few things block merge in the current state:

1. **CI is red from the PR's own test.** `tests/unit/test_proxy_utils.py::test_plain_stream_resolves_http_bridge_turn_state_owner` still asserts `affinity.kind is None` / `affinity.key is None`, but the head commit reverted the affinity-drop implementation (rightly, per the earlier review feedback about legacy raw-owner conflict detection and conversation ambiguity checks). The failure is deterministic: the policy arrives as `key='<turn_state>', kind=CODEX_SESSION, codex_session_source='turn_state'`. Please revert those two assertions (the rest of the test — owner preferred, no fallback — still holds).

2. **Migration head race.** `20260814_000000_add_accounts_chatgpt_identity_index` parents `20260806_000000_add_anonymous_telemetry`, but main's head has since moved to `20260812_120000_add_sticky_abandonment_scope` (via `20260813_000000_add_file_account_pins`). Merging as-is produces two alembic heads. On rebase, re-parent the new revision onto the current head.

3. **Merge conflicts / stale head pin.** The branch is conflicting with main; in particular `tests/unit/test_telemetry_migration.py` on main now derives the expected head via `inspect_migration_state(db_url).head_revision` — please adopt that instead of pinning `head_revision = "20260814_…"`, which will go stale again.

4. **Leftover scaffolding from the revert.**
   - `app/modules/proxy/_service/compact.py`: `selection_affinity = affinity` plus the two call-site renames are now a pure no-op — please revert to `affinity`.
   - `app/modules/usage/live_ingest.py`: `_resolve_account_id` writes `_resolution_cache` but no longer reads it, and `_resolve_account_id_by_id`'s cache hit still does a DB round-trip via `_account_id_exists`, so the cache no longer saves any queries while `_resolution_cache` (one `id:<raw>` entry per unique suffixed raw ID) and `_resolution_aliases` grow without bound. Either restore a bounded/TTL read path that still satisfies the moved-identity revalidation requirement, or drop the dead cache writes and add eviction for the alias map.
   - `_resolve_account_id_by_account_id` is a test-only alias of `_resolve_account_id_by_id` — the tests can call the real method.

5. **Title/body drift.** The description still promises "drop session/header affinity once … hard continuity owner", which the code intentionally no longer does. Please rewrite the summary around what remains: identity-proofed live-usage attribution (+ index/migration) and surfacing `continuity_owner_conflict` on the http_stream fail-closed paths — otherwise the merge history will misdescribe the behavior.

The live-ingest resolution logic itself traced correctly for me (exact-id proof, cross-check against a supplied ChatGPT identity, conflicting evidence dropped, alias coalescing bypassed when a ChatGPT identity is supplied), and the `CREATE INDEX CONCURRENTLY` / `IF NOT EXISTS` migration shape is right for both dialects. With the rebase, the test revert, and the cleanup above, this should be in good shape.

Soju06:
Rebase nudge: the prod-fix core here is acknowledged; still blocked on the failing test fix + rebase over the 0816 merges (blocker list from 08-16 unchanged).

Soju06:
Thanks for the 08-19 rebase — blockers 1-3 from the 08-16 review are resolved (test assertions now match the real policy, the migration is re-parented onto the current head with a single linear head, and the head-pin in test_db_migrate is replaced with a reachability check, which is a nice improvement). CI is fully green and the branch is clean against main, so the needs-rebase label is being dropped.

However, main moved under the live-usage half of this PR while it was open. #1773 (settle live snapshots after account consolidation) and #1783 landed `UsageRepository.settle_live_account_snapshot`, which already resolves the snapshot owner inside the write transaction (exact local ID first, then unique chatgpt_account_id, under an advisory lock on PostgreSQL) and drops unresolved/ambiguous snapshots before any insert — i.e. the FOREIGN KEY failure this PR targets is already fixed on main. That leaves three problems with the ingestor-level `_resolve_persisted_account_id` layer:

1. It duplicates settle's resolution with 1-3 extra unlocked queries per snapshot in separate sessions, then still passes the raw IDs to settle, which re-resolves authoritatively.
2. Its conflict semantics contradict the pending spec: when an exact local account ID exists but the supplied chatgpt identity resolves to a different account, this PR drops the snapshot, while `openspec/changes/settle-live-usage-after-account-consolidation` ("A valid local owner takes precedence") requires persisting under the captured local owner without substitution. Because your pre-filter runs outside the advisory lock, a consolidation racing the queue can make it drop snapshots settle would have persisted correctly.
3. The 08-16 blocker 4 items are still present at this head: the `chatgpt:*` `_resolution_cache` writes are never read, the `id:*` cache hit still does a DB round-trip via `_account_id_exists` (so the cache saves zero queries), `_resolution_cache`/`_resolution_aliases` grow without bound, and `_resolve_account_id_by_account_id` is a test-only alias.

Suggest re-scoping the PR to the parts that are still net-new against today's main, all of which look good: the `continuity_owner_conflict` surfacing on the http_stream fail-closed paths in retry.py (verified consistent with the load balancer's error code), the `idx_accounts_chatgpt_account_id` migration/index (it benefits settle's per-snapshot chatgpt lookups), the compact/route-level regression tests, and the test_db_migrate head-reachability change. Drop the live_ingest.py resolution layer and rework the openspec delta so it doesn't collide with the two pending live-usage-ingestion deltas — or, if you believe conflicting-evidence-drop is the right semantic, make that case against the consolidation spec explicitly so we can decide it as a spec change rather than ship two contradictory requirements. Please also rewrite the PR body: bullets 1-2 describe behavior that is either already on main or was reverted from this branch.

Komzpa:
@codex review

Soju06:
Following up on the 08-20 review — the current head (2ad8199fc) still doesn't reflect the re-scope request, and main has moved again since your last push. Concrete blockers:

1. **Migration head race recurred.** `app/db/alembic/versions/20260814_000000_add_accounts_chatgpt_identity_index.py` revises `20260816_000000_add_model_source_embeddings`, but main's `20260821_000000_add_retry_circuit_admission_generation` (merged 08-21) now revises that same parent. Merging as-is produces two alembic heads — same failure mode as the 08-16 blocker 2. Re-parent onto `20260821_000000_add_retry_circuit_admission_generation`.

2. **The live-usage resolution layer is still present.** Main's `app/modules/usage/live_ingest.py:243` already resolves snapshot owners inside `UsageRepository.settle_live_account_snapshot` (repository.py:762, under an advisory lock on PostgreSQL), so the FK failure this PR targets is fixed on main. Your `_resolve_persisted_account_id` preflight (live_ingest.py, added in this PR) runs 1-3 unlocked queries in separate sessions before settle re-resolves authoritatively — the 08-20 codex P1 shows a consolidation racing the queue can make the preflight drop (or mis-accept) a snapshot settle would handle correctly, and its conflict-drop semantics contradict the pending `settle-live-usage-after-account-consolidation` spec ("a valid local owner takes precedence"). Please drop the layer plus the `normalize-live-usage-account-identity` openspec delta, or explicitly propose the conflicting-evidence-drop semantic as a spec change against the consolidation spec so it can be decided as one.

3. **Unresolved codex P1s at this head (08-20 21:27):**
   - `app/modules/proxy/_service/streaming/retry.py:1420` — the SSE error envelope changes from `previous_response_owner_unavailable` to `continuity_owner_conflict`, but the added openspec artifact covers only `live-usage-ingestion` and claims response shapes are unchanged (proposal.md:25-35). Add the sticky-session/proxy delta for the error-contract change.
   - `app/db/alembic/versions/20260814_000000_add_accounts_chatgpt_identity_index.py:26` — an interrupted `CREATE INDEX CONCURRENTLY` leaves `indisvalid=false`; `IF NOT EXISTS` then skips the rebuild and stamps the revision with no usable index. Other concurrent-index migrations in this repo detect and drop the invalid index before rebuilding — apply the same recovery.

4. **P2s worth taking while you're in there:** bound the chatgpt-identity lookup to 2 rows (live_ingest.py:360) and don't cache negative exact-ID results for 300s (live_ingest.py:337) — a recreated account ID gets silently dropped for the TTL.

The parts that remain net-new and good against today's main are the `continuity_owner_conflict` surfacing on the http_stream fail-closed paths in retry.py, the `idx_accounts_chatgpt_account_id` index (it benefits settle's per-snapshot lookups), the compact/route regression tests, and the head-reachability test change. Re-scoping to those plus the re-parented migration should make this mergeable quickly. Please also rewrite the PR body — bullets 1-3 describe behavior that is either already on main or was reverted from this branch.

Soju06:
Hi @Komzpa — thanks again for this fix and for the earlier rebases. Since the re-scope requested in the 08-20 / 08-27 reviews has been pending for over a week and main has moved again underneath the branch (new alembic head), I've opened #1941 as an attribution-preserving takeover so the prod-fix core can land: your original commit is kept intact (with you as author) on that branch, with the maintainer re-scope applied as separate follow-up commits on top (live-ingest preflight layer dropped in favor of main's settle_live_account_snapshot, migration re-parented onto the current head with invalid-index recovery, and the proxy error-contract openspec delta added).

If you'd prefer to carry this forward yourself, happy to close #1941 in favor of an updated push here. Otherwise this PR will be closed as superseded once #1941 merges.

Soju06:
Superseded by #1941, which has now been merged — your original commit is preserved with authorship intact, with the maintainer-requested rescope applied on top (live-ingest preflight layer dropped in favor of the settle-side authority from #1773/#1783, migration re-parented onto the current alembic head with invalid-index recovery, and the sticky-session error-contract OpenSpec delta added). Thank you for the original fix!

PR #1745, opened 2026-08-14
fix(db): merge quota warmup migration head

Komzpa:
## Summary
- add the no-op Alembic merge revision for the quota warmup claim-expiry branch
- converge `20260806_030000_add_quota_warmup_claim_expiry` with current main's `20260806_000000_add_anonymous_telemetry` head

## Notes
This PR is intended to be stacked after Soju06/codex-lb#1646. The migration has no schema operations; it only records the Alembic graph convergence so deploy preflight and startup see a single canonical head.

## Validation
- `python -m py_compile app/db/alembic/versions/20260806_140000_merge_quota_warmup_claim_expiry_head.py`
- `git diff --check`


Komzpa:
Folded into #1646 (quota-warmup migration head merge travels with the feature family that needs it); history preserved via merge.

PR #1759, opened 2026-08-14
fix(ci): fall back to the workflow token when label-sync writes hit 403

Komzpa:
## Summary

- retry label/comment/approval writes once with `GH_FALLBACK_TOKEN` when the primary token fails with `gh: Resource not accessible by integration (HTTP 403)`
- track attempted vs succeeded writes and exit non-zero on `--apply` runs where every attempted write failed with a permission denial, even under `--tolerate-write-permission-errors` — total write failure is a configuration bug, not tolerable noise
- add focused unit tests for the fallback retry and the canary exit

## Why

Scheduled `sync-after-ci` runs (e.g. run 31821057434) computed correct label transitions for every open PR but lost 100% of their writes: the minted label-sync App token lacks Issues write permission on this repo, and the tolerate flag turned that into green runs. Stale `needs rebase` labels sat on mergeable PRs for days while truly conflicting PRs stayed unlabeled. The workflow already exports `GH_FALLBACK_TOKEN` (the workflow token) and grants `issues: write`, so the fallback path can write labels today. Granting the GitHub App Issues write permission is the complementary infra fix on the repo side.

## Validation

- `uv run pytest -q tests/unit/test_sync_codex_ok_labels.py` — 79 passed
- `python3 -m py_compile .github/scripts/sync_codex_ok_labels.py`
- workflow file unchanged

### DIFF
--- a/.github/scripts/sync_codex_ok_labels.py
+++ b/.github/scripts/sync_codex_ok_labels.py
@@ -192,6 +192,31 @@ _GH_RETRY_BASE_DELAY_SECONDS = 2.0
 _fallback_token_active = False
 
 
+@dataclass
+class WriteAttemptTracker:
+    attempted: int = 0
+    succeeded: int = 0
+    permission_denied_failures: int = 0
+
+    def record_success(self) -> None:
+        self.attempted += 1
+        self.succeeded += 1
+
+    def record_permission_denied_failure(self) -> None:
+        self.attempted += 1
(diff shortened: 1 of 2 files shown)


Komzpa:
Folded into #1554 (label-sync write fallback joins the workflow-approval guard; replayed via cherry-pick to keep the survivor's history clean).

PR #1762, opened 2026-08-14
fix(db): merge identity-index and quota-warmup migration heads

Komzpa:
No-op Alembic merge revision converging the #1732 accounts-identity-index head with the #1646 quota-warmup claim-expiry head, so deployments carrying both see one canonical migration head. Same pattern as the earlier #1745 merge-head PR. Standalone CI is expected red until both parent revisions are on main; the revision is inert (empty upgrade/downgrade).


Komzpa:
Withdrawn: overlay-local Alembic head convergence is deployment mechanics, not upstream material. The local deploy helper will generate the no-op merge revision inside the assembled stack instead.

PR #1765, opened 2026-08-15
fix(proxy): do not rewrite thread locality for a file-pin owner

mastertyko:
## Summary

A live `input_file.file_id` pin is hard ownership. After thread-scoped affinity, current Codex locality is the `thread_header` PROMPT_CACHE row, but preferred-owner selection still only bypassed `session_header`. A file-pinned Responses turn therefore rewrote the thread mapping to the upload account, so later unpinned turns followed the file owner.

## Type of change

Linked issue: None — composition leftover after #1521 and #1703.

## OpenSpec

Change directory: `openspec/changes/file-pin-does-not-rewrite-thread-locality/`

## Changes

- Bypass the writable `thread_header` sticky key in `preferred_owner_sticky_inputs`, matching the existing process-session bypass.
- Keep consulting the raw process-session row for hard-owner conflicts.
- Keep process-session seed insert-if-absent. Do not write or rebind the thread row on the required-owner path.

## Test plan

```
uv run pytest \
  tests/unit/test_proxy_utils.py::test_select_account_with_budget_keeps_thread_seed_for_first_exact_owner \
  tests/unit/test_load_balancer_concurrency.py::test_required_file_owner_does_not_rewrite_existing_thread_row \
  tests/unit/test_load_balancer_concurrency.py::test_legacy_raw_owner_conflict_blocks_resolved_preferred_owner -q
# 3 passed (plus adjacent thread-seed tests)

openspec validate file-pin-does-not-rewrite-thread-locality --strict
# valid
```

Intentionally not run locally: full `pre-commit` / `local-ci` and dashboard suites. Those stay on required GitHub CI.

## Related work

- #1521 persisted file pins and bypassed process-session locality only.
- #1703 made `thread_header` the current Codex soft mapping.
- #1761 keeps the file-pin owner on soft 1011 reconnect; it does not stop thread rewrite.
- Independent of #1764 (goal-restart + thread-id).

## Checklist

- [x] Focused unit checks and scoped OpenSpec validation passed.
- [x] CHANGELOG is **not** edited by hand.

## Summary by CodeRabbit

* **Bug Fixes**
  * File-pinned requests now consistently route to the pinned account without rewriting existing thread-local routing.
  * Process-session preferences are initialized after successful account selection, with leases safely released if persistence fails.
  * Existing conflict detection and sibling-thread behavior remain preserved.

* **Tests**
  * Added regression coverage for file-pin routing, thread locality, process-session seeding, and backend Responses flows.

* **Documentation**
  * Documented the updated sticky-session and file-pinning behavior.


mastertyko:
@codex review

mastertyko:
@codex review

mastertyko:
@codex review

Soju06:
Traced this end-to-end and the fix looks right. At main, `preferred_owner_sticky_inputs` only bypassed `session_header`, so a file-pinned turn on an existing thread reached the sticky selection path with the writable thread key; since the PROMPT_CACHE row is soft, no conflict fired and the persist step rewrote thread B→A. Nulling the key for `thread_header` while keeping `legacy_sticky_key` preserves the raw-owner fail-closed check (it runs in `select_account` before the path split), and the new insert-if-absent seed block in `load_balancer.py` correctly restores the first-ever process preference on the unbound owner path — it can only fire in this exact-owner thread bypass, since only the thread affinity policy sets a seed key. The route-level integration test covering the pinned turn, the later unpinned turn, and the sibling-thread seed is exactly the coverage this needed.

Two small pre-merge suggestions, both from the open CodeRabbit comments:

1. `openspec/changes/file-pin-does-not-rewrite-thread-locality/specs/sticky-session-operations/spec.md:5` — "The process-session compatibility row MAY still be consulted" understates what the code guarantees: `select_account` always performs the raw lookup when `legacy_sticky_key` is set, and the fail-closed scenario below depends on it. Suggest MUST.
2. `tests/integration/test_proxy_files.py` (~L938) — the post-pin assertions would still pass if an upsert refreshed the row's timestamp with the same owner, which would silently extend the bounded thread mapping. Capturing `updated_at` before the pinned request and asserting it unchanged would pin the no-write contract.

Neither changes runtime behavior; happy to see this merge with those addressed.

PR #912, opened 2026-06-04
feat(frontend): add upstream proxy controls

Soju06:
## Summary
- add dashboard schemas/API/hooks for upstream proxy admin endpoints
- add Settings UI for routing enablement, default pool, endpoint/pool/member management
- add Accounts UI for per-account proxy pool binding and disabling

## OpenSpec
- openspec/changes/add-upstream-proxy-dashboard-controls/

## Validation
- uv run openspec validate add-upstream-proxy-dashboard-controls --strict
- uv run openspec validate --specs
- cd frontend && bun run lint
- cd frontend && bun run typecheck
- cd frontend && bun run build
- cd frontend && bun run test src/features/settings/components/routing-settings.test.tsx src/features/settings/components/upstream-proxy-settings.test.tsx src/features/accounts/components/account-proxy-binding.test.tsx src/features/settings/schemas.test.ts
- cd frontend && bun run test src/features/accounts/components/account-list.test.tsx src/features/accounts/components/account-list-item.test.tsx src/features/accounts/components/account-proxy-binding.test.tsx src/features/settings/components/upstream-proxy-settings.test.tsx src/features/settings/schemas.test.ts

PR #878, opened 2026-06-02
feat(proxy): add account-bound Codex upstream routing

Soju06:
## Summary

Recreates the #869 account-bound Codex upstream proxy routing work on top of current `main` as a clean, focused PR.

This intentionally supersedes the closed stacked PR #869 after its base branch was deleted when #868 merged. It does **not** use the broader #873 repair/consolidation branch, because #873 also mixes in #804 account OAuth proxy work and merge/consolidation commits.

## Scope

- Adds account-bound upstream proxy routing for Codex/ChatGPT upstream HTTP and websocket traffic.
- Uses the curl_cffi async Codex upstream client with the built-in Codex CLI TLS fingerprint.
- Keeps route-required account-bound traffic fail-closed: no default pool, environment proxy, or direct fallback for bound routes.
- Preserves route metadata in request logs/audit paths and websocket handshakes.
- Adds settings validation for proxy pools/endpoints and route-related regression coverage.
- Includes OpenSpec change: `openspec/changes/add-codex-proxy-pool-egress/`.

## What this excludes

- Excludes #804 `feature/account-egress-isolation` / per-account OAuth proxy support.
- Excludes #873 repair/consolidation merge commits.

## Validation

```bash
uv run openspec validate add-codex-proxy-pool-egress --strict
uv run openspec validate --specs
git diff --name-only origin/main..HEAD -- '*.py' | xargs uv run ruff check
git diff --name-only origin/main..HEAD -- '*.py' | xargs uv run ruff format --check
git diff --name-only origin/main..HEAD -- '*.py' | xargs uv run ty check
uv run pytest \
  tests/unit/test_auth_dependencies_upstream_proxy.py \
  tests/unit/test_auth_manager.py \
  tests/unit/test_codex_client.py \
  tests/unit/test_codex_upstream_paths.py \
  tests/unit/test_files_client.py \
  tests/unit/test_limit_warmup.py \
  tests/unit/test_model_fetcher.py \
  tests/unit/test_model_refresh_scheduler.py \
  tests/unit/test_proxy_utils.py \
  tests/unit/test_proxy_websocket_client.py \
  tests/unit/test_request_log_upstream_proxy_metadata.py \
  tests/unit/test_upstream_proxy_resolver.py \
  tests/unit/test_usage_client.py \
  tests/unit/test_usage_updater.py \
  tests/integration/test_oauth_flow.py \
  tests/integration/test_proxy_files.py \
  tests/integration/test_settings_api.py \
  tests/integration/test_migrations.py \
[...]

PR #1767, opened 2026-08-15
fix(dashboard): surface upstream route metadata

mastertyko:
## Summary

Expose the five existing credential-safe upstream route diagnostics through request-log API responses and the dashboard details dialog. Populated values are shown only in request details; absent values stay omitted and proxy credentials remain excluded.

Related: #878
Related: #912

## Type of change

- [x] `fix:` — bug fix
- [ ] Breaking change

## Changes

- map route mode, pool ID, endpoint ID, fallback use, and fail-closed reason into `RequestLogEntry`
- preserve all five fields through the frontend Zod boundary
- render populated routing metadata with English, Korean, and Simplified Chinese labels
- constrain the dialog body to an internal scroll region so lower metadata stays reachable
- add API, schema, component, locale, and OpenSpec regressions

## Verification

- focused request-log API regression — 1 passed
- frontend focused suite — 3 files, 69 tests passed
- Ruff and focused ESLint — passed
- frontend native TypeScript typecheck — passed
- frontend production build — passed
- `openspec validate surface-upstream-route-metadata --strict` — passed
- local CI parity passed all test/type/build/e2e gates; Docker rerun after installing the missing local Trivy tool completed with 0 critical findings
- real Chromium QA against isolated seeded servers — API 200, desktop/mobile/CJK verified
- two independent visual reviews — PASS
- five independent commit reviews — PASS, no blocking findings
- `git diff --check` — passed

## OpenSpec

- `openspec/changes/surface-upstream-route-metadata/`

### Before

The same successful request has no route diagnostics in the baseline details dialog.

Before: request details without route metadata

### After

Successful routing exposes mode, pool, endpoint, and same-pool fallback.

After: successful request route metadata

Fail-closed requests expose the persisted reason without credentials.

After: fail-closed route metadata

Mobile lower content remains reachable through the bounded dialog scroller.

After: mobile scroll reaches lower content

Simplified Chinese route labels render without clipping at 390x844.

After: Simplified Chinese mobile route metadata

## Simplicity

No new configuration, dependency, navigation item, table column, filter, or database migration.

## Summary by CodeRabbit

- **New Features**
  - Request details now show upstream routing information, including route mode, proxy pool, endpoint, fallback status, and fail-closed reasons.
  - Routing information is displayed only when available and is localized in English, Korean, and Simplified Chinese.
  - Request-log API responses now include these diagnostic details.

- **Bug Fixes**
  - Improved request-details layout with a scrollable dialog for easier viewing of routing metadata.


mastertyko:
@codex review

Soju06:
Traced this end to end and it looks solid: the five fields already exist on the model and `add_log`, the read-model widening is purely additive, the Zod fields are nullable+optional so mixed-version deployments are safe, and the `!= null` guard means `fallback_used: false` still renders "Not used". Two non-blocking notes:

1. `app/modules/request_logs/mappers.py:61-65` — the route fields are exposed unconditionally, i.e. also to `GUEST` dashboard sessions, while `useragent`/`client_ip`/`conversation_id` are gated behind `include_sensitive_metadata`. Pool/endpoint IDs are operator-chosen infrastructure identifiers rather than credentials, so this seems acceptable (same class as the already-ungated `account_id`), but if you consider proxy topology admin-only it would be a one-line gate here.

2. The two CodeRabbit nitpicks (a negative assertion that credential-bearing fields never serialize, and a UI test case for `upstreamProxyFallbackUsed: false` rendering "Not used") are both cheap and worth taking, either here or as a follow-up.

Neither blocks merge from my side.

PR #1776, opened 2026-08-16
feat(model-sources): embeddings source capability with request-log accounting

### LINKED ISSUE #261: Support embeddings with codex-lb
Hello,

Is it possible to generate embeddings using codex-lb?

I tried sending a POST request to /v1/embeddings, but the API returned 405 Method Not Allowed.

Could you confirm whether embeddings are supported for codex-lb? If not, is there a recommended model or endpoint for this use case?

Thanks.

Image

Komzpa:
Closes Soju06/codex-lb#261.

## What

`POST /v1/embeddings` routed to OpenAI-compatible model sources, with the same API-key enforcement and `request_logs` accounting as chat completions. Before this, `/v1/embeddings` returned `405 Method Not Allowed`, so local embedding traffic (e.g. an ollama source) could not go through codex-lb at all — and never appeared in the ledger.

The change mirrors the existing audio-transcriptions capability shape end to end:

- `model_sources.supports_embeddings` boolean (additive migration, defaults false) with the dashboard capability toggle and list badge
- repository finder `find_embeddings_source_for_model` (same filters as chat/audio: enabled source, enabled model, kind `openai_compatible`, API-key source assignment scope)
- `forward_embeddings` posts the client payload to the source's `/embeddings` and passes the response through; usage is read from `prompt_tokens`/`total_tokens` (embeddings responses carry no completion tokens)
- the route validates `model` + `input` and passes every other OpenAI param (`encoding_format`, `dimensions`, `user`, …) through verbatim
- ledger: success and error rows are written exactly like source-routed chat (`source="model_source"`, token counts, cost from the source model's input rate, upstream status)

## Behavior details

- Unknown model → OpenAI-shaped `model_not_found` (404). Embeddings have no subscription-backed fallback, so only configured sources can serve them.
- Upstream source errors pass through with the source's status and are logged as error rows.
- A usage-less upstream response fails closed (502 `usage_unavailable`) for API keys with usage limits, matching the transcription route's contract.

## Tests

`tests/integration/test_model_source_routing.py`: four new tests — happy path (payload passthrough, auth header, ledger row with token counts), unknown model 404, upstream error pass-through + error log row, and the limited-key fail-closed path. All 55 tests in the touched backend files pass locally (`pytest tests/integration/test_model_source_routing.py tests/unit/test_model_sources_service.py tests/unit/test_model_sources_catalog.py`). Frontend changes mirror the audio-capability commit (schema default, form toggle, badge, mocks, fixtures); frontend tests were not run locally and rely on CI.

## Summary by CodeRabbit

* **New Features**
  * Added an OpenAI-compatible `POST /v1/embeddings` endpoint.
  * Added embeddings capability settings for model sources, including create, edit, and display support.
  * Model sources can now be selected automatically for embeddings requests.
  * Added support for forwarding usage details and preserving request fields.
* **Bug Fixes**
  * Improved handling of unavailable sources, invalid responses, upstream errors, and usage-limit scenarios.
* **Localization**
  * Added embeddings capability labels in English, Korean, and Simplified Chinese.
* **Tests**
  * Added coverage for routing, payload forwarding, errors, and restricted usage behavior.


Soju06:
Thanks — this is a clean implementation that mirrors the audio-transcriptions capability shape faithfully, and the four integration tests are exactly the coverage we want. A few concrete items before this can land:

1. **CI red (the real one):** `tests/integration/test_daybreak_capability_routes.py::test_registered_proxy_route_inventory_has_one_explicit_capability_policy` requires every proxy route to be classified in a capability-policy group. Add `("HTTP", "POST", "/v1/embeddings")` to `_FAIL_CLOSED_HTTP_ROUTES` — your handler already calls `_required_capability_http_transport_denial` first, so fail-closed is the right group. This test predates your branch point, so it will fail regardless of rebase.

2. **Pre-body capability denial parity:** `/v1/embeddings` is a JSON-body route validated via pydantic `Body(...)`, like `/v1/chat/completions` and `/v1/images/generations`. Those routes are listed in `_JSON_BODY_DENY_PATHS` in `app/core/middleware/required_capability_http.py:26` so a capability-marked POST is denied before the body is parsed (a malformed body must yield the transport denial, not a 422). Please add `/v1/embeddings` there and to the `test_daybreak_json_routes_reject_capability_before_body_validation` param list.

3. **Alembic head conflict:** main has since gained `20260816_000000_add_account_pending_deletion`, which chains off the same `20260812_120000_add_sticky_abandonment_scope` parent as your migration — merging as-is produces two heads. Please re-chain your `down_revision` to `20260816_000000_add_account_pending_deletion` (and consider bumping your revision timestamp so we don't have two `20260816_000000_*` files).

4. **Rebase:** there's a textual conflict in `tests/integration/test_model_source_routing.py` against current main; everything else merges cleanly, and I verified all the proxy/reservation helpers you call are unchanged after the recent perf work on main.

5. **OpenSpec:** per AGENTS.md, behavior/contract/schema changes need an entry under `openspec/changes/<change>/` (proposal -> tasks) covering the new capability flag, the `/v1/embeddings` contract (404 fallbackless behavior, usage_unavailable fail-closed rule), and accounting. Heads-up that the open change `route-model-sources-off-websocket` plans to extract source-selection helpers into `app/modules/model_sources/selection.py`, so keep `_select_embeddings_model_source` shaped like its siblings.

Minor: `supportsEmbeddings` is missing from the update-payload mock schema in `frontend/src/test/mocks/handlers.ts`. The CodeRabbit locale comment is a false positive — ko/zh-CN intentionally keep capability labels in English, so your strings are consistent as-is.

PR #1817, opened 2026-08-19
fix(proxy): recognize the terse upstream previous_response_id rejection

yeongjun-cigro:
## Why

Detection of an unresolvable `previous_response_id` is string-shaped. `is_previous_response_not_found_error` matches only:

- `code=previous_response_not_found`, or
- `code=invalid_request_error` **with** `param=previous_response_id` **and** a message containing "previous response … not found"

The Codex backend now rejects an unresolvable anchor like this instead:

```json
{"error":{"message":"Invalid `previous_response_id`.","type":"invalid_request_error","code":"invalid_request_error"}}
```

No `param`, no "not found". The predicate returns `False`, and every recovery site gated on it is skipped — the WebSocket full-resend retry, the HTTP bridge anchor rewrite, and the public-error masking that is supposed to turn this into a retryable `stream_incomplete`. The raw upstream 400 reaches the client, and the user's turn dies with `Invalid previous_response_id.` on a thread they then have to abandon.

## Observed impact

On a production deployment (~8k successful requests/hour), the changeover is unmistakable in the request log:

| hour (UTC) | recoveries (`codex_previous_response_stale`) | raw 400s reaching clients |
|---|---|---|
| 08-18 04:00 → 16:00 | 1–8 per hour | 0 |
| 08-18 22:00 | 0 | 1 |
| 08-19 04:00 | 0 | 13 |
| 08-19 06:00 | 0 | 17 |

24-hour totals: **71 raw 400s across 10 API keys**, on both transports (56 HTTP bridge, 15 WebSocket). Recoveries: zero since the wording changed. Before it changed, users never saw this error at all — the retry absorbed it.

## What changes

- `is_previous_response_not_found_message` also matches a message that names `previous_response_id` alongside "invalid".
- A missing `param` becomes inconclusive rather than disqualifying — only a *different* `param` rules the anchor out. `code` must still be `invalid_request_error` (or the canonical `previous_response_not_found`).

What recovery *does* is unchanged. Retrying with full input and no anchor is already the right response to any upstream rejection of the anchor itself, including a malformed one, so widening detection cannot produce a wrong recovery — at worst one extra full-context retry on a 400 that mentions the anchor.

## Spec

`openspec/changes/recognize-terse-previous-response-rejection/` modifies the `responses-api-compat` requirement "Public Responses errors mask previous-response misses" to state that the miss is recognized from wording as well as canonical code, and that a different `param` still disqualifies.

## Tests

`tests/unit/test_openai_errors.py::test_is_previous_response_not_found_error_terse_upstream_shape` covers the terse shape, the two pre-existing shapes, a different `param`, an unrelated `invalid_request_error` message, and a non-400 code that merely mentions the anchor. `ruff check`, `ruff format --check`, and `ty check` clean; unit suite 4,299 passed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recovery when a previous response cannot be found or is rejected by the service.
  * Recognizes both detailed and terse error messages, including cases where parameter details are omitted.
  * Prevents unrelated parameter errors from being incorrectly treated as recoverable.
  * Continues masking missing response identifiers from public error messages.

* **Tests**
  * Added coverage for valid recovery scenarios and unrelated or mismatched errors.

### DIFF
--- a/app/core/errors.py
+++ b/app/core/errors.py
@@ -76,7 +76,14 @@ def is_previous_response_not_found_message(message: str | None) -> bool:
     if message is None:
         return False
     normalized = " ".join(message.lower().split())
-    return "previous response" in normalized and "not found" in normalized
+    if "previous response" in normalized and "not found" in normalized:
+        return True
+    # The Codex backend also rejects an unresolvable anchor with a terse
+    # ``Invalid `previous_response_id`.`` that carries neither "not found" nor an
+    # error ``param``. Recognize that shape too: the recovery (retry with full
+    # input and no anchor) is the correct response to any upstream rejection of
+    # the anchor itself, including a malformed one.
+    return "previous_response_id" in normalized and ("invalid" in normalized or "not found" in normalized)
 
 
 def previous_response_id_from_not_found_message(message: str | None) -> str | None:
@@ -102,7 +109,11 @@ def is_previous_response_not_found_error(
 ) -> bool:
     if code == PREVIOUS_RESPONSE_NOT_FOUND_CODE:
         return True
-    if code != "invalid_request_error" or param != "previous_response_id":
+    if code != "invalid_request_error":
+        return False
+    # ``param`` is absent on the terse rejection shape; only a *different* param
+    # rules the anchor out as the cause.
+    if param is not None and param != "previous_response_id":
         return False
     return is_previous_response_not_found_message(message)
 
--- a/openspec/changes/recognize-terse-previous-response-rejection/proposal.md
+++ b/openspec/changes/recognize-terse-previous-response-rejection/proposal.md
@@ -1,0 +1,19 @@
+## Why
+
+Upstream detection of an unresolvable `previous_response_id` is string-shaped: `is_previous_response_not_found_error` requires `code=previous_response_not_found`, or `code=invalid_request_error` **with** `param=previous_response_id` **and** a message containing "previous response … not found".
+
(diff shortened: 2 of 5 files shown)


rknightion:
I tested this branch at `a983b8ac` against the exact nested production frame with no `code` or `param`:

```json
{"type":"error","status":400,"error":{"type":"invalid_request_error","message":"Invalid `previous_response_id`."}}
```

The classifier change is enough for the self-contained full-resend path, and that route test passes. The public `/v1/responses` WebSocket masking path still fails though: `_maybe_rewrite_websocket_previous_response_not_found_event` rechecks the event using only the missing `error.code`, so the client receives raw `type: error` rather than sanitized `response.failed`.

I opened #1818 with that extra normalization and route-level regressions for Codex-native delta handling, full resend, and public masking. It also keeps the new message match limited to the observed normalized wording rather than any invalid message mentioning `previous_response_id`.

Soju06:
Closing as superseded by #1818, which just merged — thank you for this PR: it independently identified the same production regression with the strongest evidence in either PR (71 raw 400s/24h on the HTTP transport with recoveries dropping to zero), and that evidence is what confirmed the urgency. The two PRs hard-conflicted in `app/core/errors.py` and `tests/unit/test_openai_errors.py`, and we verified empirically that #1818 additionally fixes the WS rewrite-boundary normalization (`_maybe_rewrite_websocket_previous_response_not_found_event` reading raw `error.code`) that this branch didn't cover — the /v1/responses masking replay test for the observed code-less frame fails at this head, so the masking claim didn't hold on that path. Nothing salvageable remains after #1818: rebasing this branch would only loosen the deliberately exact predicate. If your production data shows backtick-less or reworded variants of the terse rejection that #1818's exact-match predicate misses, please open a focused follow-up widening `_is_invalid_previous_response_id_message` with those frames as fixtures.

PR #1840, opened 2026-08-20
fix(proxy): bound HTTP bridge live event queue

mastertyko:
## Summary

Bounds each admitted HTTP Responses bridge's live upstream-event queue so a paused SSE consumer backpressures the shared relay instead of allowing unbounded per-request memory growth. Downstream cancellation revokes blocked queue producers while preserving terminal persistence, settlement, logging, replay, and cleanup ownership.

## Type of change

- [ ] **Breaking change**

Linked issue: None found. Related continuity work: #232 (merged; contextual, not overlapping).

## OpenSpec

- [x] This PR touches a codex-faithful path and preserves upstream-equivalent SSE framing and ordering

Change directory: `openspec/changes/bound-http-bridge-event-queue/`

## Changes

- Use a two-event finite live queue: enough for terminal event plus end marker before consumption starts, while the next unread event applies backpressure.
- Revoke blocked queue puts when the downstream detaches, including cancellation before the consumer loop begins, without changing shared-reader drain or settlement ownership.
- Keep completed durable replay finite with transcript-sized startup capacity.
- Add deterministic paused, resumed, paced, detached, startup-cancelled, and durable-replay regressions.

## Simplicity

- [x] Works with zero config and adds no setting
- New setting(s): None
- [x] README, environment example, and dashboard navigation are unchanged

## Test plan

```text
10 focused queue and bridge lifecycle tests: passed
Responses integration suite: 65 passed
Responses API contract suite: 65 passed
Full bridge unit module: 637 passed; 1 unrelated existing fixture failure because file_account_pins was absent
Ruff check and format check on changed Python files: passed
ty check on changed Python files: passed
OpenSpec strict validation for bound-http-bridge-event-queue: passed
Actual-path async surface proof: bounded at maxsize=2, paused producer blocked, resumed terminal/EOS delivery remained ordered, and detach released the producer without queue-task leaks
```

Global strict spec validation was also attempted; it reports eight existing repository spec failures outside this change. The change-specific strict validation is clean. The all-files local CI hook was intentionally not run; Sensitive-path focused tests, lint, formatting, type checks, diagnostics, OpenSpec validation, and committed-diff review were run instead.

## Output

The regression drives request preparation through upstream event relay. With consumption paused, the queue remains at two events and the producer blocks; after consumption resumes, all events arrive in order through the terminal event and end marker. Cancelling before the consumer loop revokes the full queue and terminates producer-owned put/revoke tasks.

## Checklist

- [x] Related work was checked and documented above.
- [ ] Ran the all-files local CI hook (focused Sensitive-path verification used instead; details above).
- [x] Change-specific strict OpenSpec validation and verification are clean.
- [x] Simplicity gates reviewed; no setting, default, README, environment, or navigation change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved HTTP bridge event delivery for slow or paused consumers by applying bounded buffering and backpressure.
  * Prevented event producers from remaining blocked when a downstream request disconnects or is cancelled.
  * Preserved event ordering, terminal responses, durable replay, and cleanup during disconnects and retries.

* **Tests**
  * Added coverage for backpressure, cancellation, queue revocation, cleanup, and durable event replay.

### DIFF
--- a/app/modules/proxy/_service/http_bridge/request_submit.py
+++ b/app/modules/proxy/_service/http_bridge/request_submit.py
@@ -221,6 +221,9 @@ logger = logging.getLogger("app.modules.proxy.service")
 
 _HTTP_BRIDGE_CLEAN_CLOSE_RETRY_MAX_COUNT = 1
 _HTTP_BRIDGE_CLEAN_CLOSE_RETRY_JITTER_MAX_SECONDS = 2.0
+# A local startup failure must enqueue its terminal frame and end marker before
+# the downstream consumer loop starts; the third unread live event backpressures.
+_HTTP_BRIDGE_LIVE_EVENT_QUEUE_MAX_SIZE = 2
 
 _REQUEST_TRANSPORT_HTTP = "http"
 _WEBSOCKET_AUTH_INVALIDATED_FAILURE_CODE = "account_auth_invalidated"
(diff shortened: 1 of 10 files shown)


Komzpa:
@codex review

Soju06:
Closing in favor of #1903, which embeds this PR's exact backpressure/revoke mechanism and closes the gaps around it — it also bounds the prewarm queue at the second constructor site, adds an out-of-band terminal+EOS delivery path (with maxsize=2 here, `_settle_aborted_http_bridge_terminal_states`'s `put_nowait(None) except QueueFull: pass` becomes reachable and silently drops EOS), revokes pre-consumer queues on liveness-failure/submit-interruption paths, and adds a process-wide byte budget so the bound is in bytes rather than event count. The two PRs also create the same `openspec/changes/bound-http-bridge-event-queue/` directory, so they're mutually exclusive.

Independent of the comparison, this version has a session-wide wedge under the shared-reader topology: the per-session upstream reader awaits `_process_http_bridge_upstream_text` inline, whose delta relay and terminal delivery block in the no-timeout revocable `put`; revocation only fires from the stalled request's own consumer detach, but a connected-not-reading SSE client parks its consumer in `yield` (uvicorn has no write timeout), so one slow client wedges every request multiplexed on the session until its TCP dies — on main the same input costs only memory growth. The tests here cover single-request backpressure only; a test pausing one consumer while a second request shares the session would surface it.

Thanks for pushing this problem forward — the bounding direction was right, and #1903 builds on the same mechanism.

PR #1856, opened 2026-08-20
fix(proxy): refuse disabled model sources and spare account health from model-scoped rejections

Komzpa:
## Summary

Two linked defects on the same path, found while diagnosing a live incident on 2026-08-20. A model whose only OpenAI-compatible model source had been switched off was routed to ChatGPT subscription accounts instead. Those accounts rejected it with HTTP 400, each rejection was recorded as a transient account error, and accounts that were serving unrelated traffic went into error backoff.

They ship together because either commit alone leaves half the loop in place: the routing fix alone still leaves a rejection that any other entitlement mismatch can produce charging account health, and the health fix alone still spends one subscription account selection per attempt, per account, on a request no account can serve.

## Type of change

Linked issue: none

### 1. A disabled model source is indistinguishable from an unknown model

`find_chat_source_for_model` and `find_responses_source_for_model` (`app/modules/model_sources/repository.py`) filtered on `ModelSource.is_enabled.is_(True)` and `ModelSourceModel.is_enabled.is_(True)` inside the lookup, so a disabled source and a model no source claims both return `None`. `app/modules/proxy/api.py` then dispatches to subscription account selection, and upstream answers:

```
HTTP 400
The '<model>' model is not supported when using Codex with a ChatGPT account.
```

The caller learns nothing about the source being off, and every attempt costs one subscription account selection for each account tried.

### 2. That rejection was charged to account health

`_handle_stream_error` (`app/modules/proxy/_service/streaming/helpers.py`) is the single funnel every transport uses for stream-error account health. Upstream delivers this 400 on the streaming path with neither `code` nor `type`, so `_normalize_error_code` produces the `upstream_error` fallback. `classify_upstream_failure` (`app/modules/proxy/helpers.py`) classifies `upstream_error` as `retryable_transient`, so `_handle_stream_error` reaches `record_error(account)`, which increments the runtime `error_count` and, past `ERROR_BACKOFF_THRESHOLD`, applies error backoff up to 300 s (`app/core/balancer/logic.py`).

The existing matcher `_is_account_model_unsupported_error` does not cover this path: it requires `code == "invalid_request_error"` and the requested model string, and on the streaming path the code has already been normalized away.

Measured on the live instance before the source was re-enabled at 17:03:35Z:

- 80 rejections produced 240 `record_error` calls — 80 against each of three Pro accounts, because one client request fans out to all three.
- 124 `No available accounts` responses in the same 15 minutes. Sticky selection passes `allow_backoff_fallback=False`, so sessions hard-pinned to those accounts failed with `continuity_owner_unavailable` / `No available accounts` while the accounts themselves were at 1–7% quota used.
- 4,801 of these 400s in 24 hours. All of it stopped when the source was re-enabled.

`error_count` is runtime-only and never persisted (`app/modules/proxy/load_balancer.py`), so the `accounts` table shows no trace of this afterwards.

## Changes

Two commits with disjoint file sets.

**`fix(proxy): stop model-entitlement rejections from penalizing account health`**

- `is_model_scoped_upstream_rejection(message)` in `app/modules/proxy/helpers.py` matches the rejection from the normalized message alone, with no dependency on the error code and without the caller having to know the requested model.
- `_handle_stream_error` returns the classified failure before any account-health mutation when the upstream status is 400 and the message matches, and logs the skip in the same shape as the existing account-neutral skip.
- Failure classification, the failover decision, and the client-visible status and body are unchanged, so an account with a different entitlement is still attempted.

**`fix(proxy): refuse models whose source is disabled instead of routing them to a subscription`**

- `_enablement_filter(only_disabled)` in the model-sources repository, with `only_disabled` threaded through `selection.py`, `select_responses_model_source` and `_select_chat_model_source`. It inverts the enabled-state predicate and leaves every other rule alone — candidate order, API-key model allowlist, source assignment scope, subscription-registry precedence, route shape, streaming — so a hit is exactly the source the request would have used had it not been switched off.
- `_disabled_model_source_denial()` on `/v1/chat/completions`, `/v1/responses` and `/backend-api/codex/responses`, consulted only after the ordinary lookup misses, returning 503 `model_source_disabled`:

  ```json
  {
    "error": {
      "type": "upstream_error",
      "code": "model_source_disabled",
      "message": "The model '<model>' is served by an OpenAI-compatible model source that is disabled. Enable the source and its model in codex-lb, or request a different model."
    }
  }
  ```

  The envelope names the model and the condition but not the source, since the source name is operator-facing configuration rather than a client-visible identifier.
- On the chat route the check runs before `_enforce_request_limits`, so a refusal strands no usage reservation.
- File-pinned Responses requests and compaction triggers still skip the check.
- Every other miss keeps its existing path: a model no source claims, a source scoped away from the API key, a route-shape mismatch, and a subscription slug that an unscoped key never source-routes.

## Status code: 503 vs a terminal 4xx — maintainer decision welcome

The refusal returns 503 `model_source_disabled` rather than a terminal 4xx. The reasoning is that this is a source-availability condition and belongs to the existing `model_source_*` family, and that a "model does not exist" shape would send an operator to the model registry rather than to the source they switched off.

The trade-off is that 503 is retryable-shaped, so a client retry loop will keep retrying — although it no longer spends a subscription account selection per attempt, which is the cost this PR is removing. If a terminal 4xx is preferred, it is a status change in `_disabled_model_source_denial` plus the matching line in the spec delta.

## OpenSpec

Two change directories, targeting different capabilities and each coherent on its own:

- `openspec/changes/spare-account-health-from-model-rejections/` — MODIFIED `account-routing`. Required because `openspec/specs/account-routing/spec.md` currently mandates deciding payload-rejection membership from the `invalid_request_error` code, which is what missed this rejection on the streaming path.
- `openspec/changes/refuse-disabled-model-source-routing/` — `responses-api-compat`.

```
openspec validate spare-account-health-from-model-rejections --strict --no-interactive   # valid
openspec validate refuse-disabled-model-source-routing --strict --no-interactive         # valid
```

`openspec validate --specs` reports 49 passed / 8 failed on this branch and the identical 49 passed / 8 failed on clean `main` at `b6c217fa`, so those 8 are pre-existing and untouched here.

## Simplicity

No new setting, no new required setup step, no README section, no dashboard nav item, and no changed default. Both commits narrow existing behavior on paths that already exist.

## Test plan

All numbers below are from the current head of this branch.

```
uv run pytest tests/unit -q -p no:randomly
# 6432 passed, 3 skipped, 12 warnings in 136.00s

uv run pytest tests/integration/test_model_source_routing.py \
             tests/unit/test_proxy_websocket_model_source_guard.py -q -p no:randomly
# 109 passed in 21.54s

uv run ruff check                                   # All checks passed!
uv run ruff format --check <the three changed files># 3 files already formatted
uv run ty check                                     # All checks passed!
```

Both halves were also run with their `app/` changes reverted and their tests kept, to show the tests fail without the fix.

Account-health half, with `app/modules/proxy/helpers.py` and `app/modules/proxy/_service/streaming/helpers.py` reverted:

```
uv run pytest tests/unit/test_proxy_utils.py -q -p no:randomly \
  -k "model_scoped or genuine_upstream_error_still_penalizes or non_400_model_rejection or rate_limit_still_marks_rate_limit"
# 9 failed, 4 passed, 1117 deselected
```

The four that still pass are the negative controls: classification and failover are unchanged, a genuine `upstream_error` still penalizes the account, a non-400 carrying the same message still penalizes it, and a rate limit still marks the account rate-limited.

Routing half, with `app/modules/model_sources/repository.py`, `app/modules/model_sources/selection.py` and `app/modules/proxy/api.py` reverted:

```
uv run pytest tests/integration/test_model_source_routing.py -q -p no:randomly \
  -k "disabled_source or unknown_model_still_falls_through"
# 4 failed, 2 passed, 90 deselected
```

The two that still pass are the negative controls: a model no source claims still falls through to subscription routing, and a subscription model slug shadowed by a disabled source still resolves by registry precedence.

## Related

Adjacent open PRs, and why this is separate from them:

- Soju06/codex-lb#1853 also moves account-health writes on the streaming path, but in `_service/streaming/retry.py`, and it concerns *when* a penalty is applied relative to reservation settlement on keyed streams. This PR concerns *whether* a model-scoped rejection is an account-health event at all, in `_service/streaming/helpers.py`. The file sets do not overlap.
- Soju06/codex-lb#1664 also edits `app/modules/model_sources/repository.py`, adding a subscription-exhausted fallback *into* a model source. It adds new queries and keeps the `is_enabled.is_(True)` predicate in them; it does not change the enabled-state filter on the existing per-model lookups. The two changes are compatible in behavior. Whichever lands second will need a one-line import merge in `repository.py` (`update` on one side, `ColumnElement, and_, or_` on the other).

No other open PR touches `_handle_stream_error`, account-health recording, or model-source lookup.

## Checklist

- [x] Ran the local subset above.
- [x] If touching specs: `openspec validate --strict` passes for both change directories; `--specs` is unchanged from `main`.
- [x] Simplicity gates reviewed: the five simplicity rules (PRINCIPLES.md P1–P5).

## Summary by CodeRabbit

* **Bug Fixes**
  * Requests targeting disabled model sources now return a structured HTTP 503 error instead of being routed elsewhere.
  * Disabled models are correctly detected across Chat and Responses endpoints.
  * Model-entitlement rejections no longer negatively affect account health, while failover behavior is preserved.
  * Unknown models and subscription-owned models retain their existing routing behavior.

* **Tests**
  * Added integration and unit coverage for disabled-source routing, rejection handling, failover, and account-health behavior.

* **Documentation**
  * Added specifications describing disabled-source refusal and health-neutral model rejections.

### DIFF
--- a/app/modules/model_sources/repository.py
+++ b/app/modules/model_sources/repository.py
@@ -1,10 +1,25 @@
 from __future__ import annotations
 
-from sqlalchemy import delete, select
+from sqlalchemy import ColumnElement, and_, delete, or_, select
 from sqlalchemy.ext.asyncio import AsyncSession
 from sqlalchemy.orm import selectinload
 
 from app.db.models import ModelSource, ModelSourceModel
+
+
+def _enablement_filter(only_disabled: bool) -> ColumnElement[bool]:
+    """Enabled-state predicate for a per-capability source lookup.
+
+    ``only_disabled`` selects the exact complement of the routable set: rows
+    that match the model and the route shape but that an operator switched off,
+    at the source or at the individual model. Routing needs that complement to
+    tell "no source serves this model" apart from "the source that serves this
+    model is switched off" -- the latter must not fall through to a
+    subscription account, which rejects the model outright.
+    """
+    if only_disabled:
+        return or_(ModelSource.is_enabled.is_(False), ModelSourceModel.is_enabled.is_(False))
+    return and_(ModelSource.is_enabled.is_(True), ModelSourceModel.is_enabled.is_(True))
 
 
 class ModelSourcesRepository:
@@ -38,16 +53,16 @@ class ModelSourcesRepository:
         *,
         allowed_source_ids: set[str] | None = None,
         require_streaming: bool = False,
+        only_disabled: bool = False,
     ) -> ModelSource | None:
         stmt = (
             select(ModelSource)
             .options(selectinload(ModelSource.models))
             .join(ModelSourceModel, ModelSourceModel.source_id == ModelSource.id)
             .where(ModelSource.kind == "openai_compatible")
-            .where(ModelSource.is_enabled.is_(True))
             .where(ModelSource.supports_chat_completions.is_(True))
             .where(ModelSourceModel.model == model)
-            .where(ModelSourceModel.is_enabled.is_(True))
+            .where(_enablement_filter(only_disabled))
             .order_by(ModelSource.name, ModelSource.id)
             .limit(1)
         )
@@ -66,16 +81,16 @@ class ModelSourcesRepository:
         *,
         allowed_source_ids: set[str] | None = None,
         require_streaming: bool = False,
+        only_disabled: bool = False,
     ) -> ModelSource | None:
         stmt = (
             select(ModelSource)
             .options(selectinload(ModelSource.models))
             .join(ModelSourceModel, ModelSourceModel.source_id == ModelSource.id)
             .where(ModelSource.kind == "openai_compatible")
-            .where(ModelSource.is_enabled.is_(True))
             .where(ModelSource.supports_responses.is_(True))
             .where(ModelSourceModel.model == model)
-            .where(ModelSourceModel.is_enabled.is_(True))
+            .where(_enablement_filter(only_disabled))
             .order_by(ModelSource.name, ModelSource.id)
             .limit(1)
         )
--- a/app/modules/model_sources/selection.py
+++ b/app/modules/model_sources/selection.py
@@ -34,8 +34,19 @@ async def select_responses_model_source(
     *,
     raw_model: str | None = None,
     require_streaming: bool = False,
+    only_disabled: bool = False,
 ) -> tuple[ModelSource, str] | None:
-    """Resolve ``model`` to a Responses-capable model source, if any."""
+    """Resolve ``model`` to a Responses-capable model source, if any.
+
+    ``only_disabled`` inverts the enabled-state filter and leaves every other
+    rule -- candidate order, API key model allowlist, source assignment scope,
+    subscription-registry precedence, route shape, streaming -- untouched, so
+    the answer is exactly "the source this request would have used, had the
+    operator not switched it off". Callers use it after the ordinary lookup
+    misses, to refuse the request instead of handing a source-owned model to a
+    subscription account that cannot serve it. Sharing one function keeps the
+    two lookups from drifting apart the way the transports once did.
+    """
     assigned_source_ids = allowed_source_ids_for_api_key(api_key)
     exact_allowed_models = set(api_key.allowed_models) if api_key and api_key.allowed_models else None
     candidates = [candidate for candidate in (raw_model, model) if candidate]
@@ -55,6 +66,7 @@ async def select_responses_model_source(
                 candidate,
                 allowed_source_ids=assigned_source_ids,
                 require_streaming=require_streaming,
+                only_disabled=only_disabled,
             )
             if source is not None:
                 break
--- a/app/modules/proxy/_service/streaming/helpers.py
+++ b/app/modules/proxy/_service/streaming/helpers.py
@@ -387,6 +387,7 @@ from app.modules.proxy.durable_bridge_coordinator import (
 from app.modules.proxy.helpers import (
     _normalize_error_code,
     classify_upstream_failure,
+    is_model_scoped_upstream_rejection,
     is_upstream_model_capacity_error,
 )
 from app.modules.proxy.http_bridge_forwarding import (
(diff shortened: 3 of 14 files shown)


Komzpa:
Rebased this branch onto current \`main\` after the five-squash-merge revert (#1858, now \`d4b00fd0\`). Old head \`5e14610c9\` -> new head \`ad0b724e2\`; merge-base moved from \`b6c217fad\` to current \`main\`, so the diff again shows only this PR's own change (model-source routing + account-health handling), with none of the reverted HTTP-bridge content.

Typecheck, full unit suite, and the touched integration module all pass on the rebased branch.

Soju06:
Thanks — this is an unusually well-evidenced PR (live incident numbers, reverted-fix test verification, tight scoping of the enablement-filter inversion), and both halves read correct for the HTTP routes. CI is fully green. Three things stand between this and merge:

1. **WebSocket transport still routes disabled-source models to subscription accounts** (the codex P2 from Aug 20, still open on the current head). `app/modules/proxy/_service/websocket/mixin.py` is unchanged: both the connect-time guard (~line 3412) and the socket-reuse guard (~line 1762) call `responses_model_is_source_owned()`, which uses the default `only_disabled=False` lookup. So a model owned only by a disabled source over WS returns `False` from the guard, the turn dispatches to a subscription account, and upstream answers the same `The '<model>' model is not supported…` 400 — on exactly the transport your own docstring notes delivered the incident's rejections. The health half now keeps that 400 out of `error_count` (good — the worst symptom is capped), but each WS attempt still burns a subscription selection per account and the client sees the opaque 400 instead of `model_source_disabled`. Suggested shape: have the guard also recognize disabled-source ownership (e.g. a second `only_disabled=True` probe or an ownership mode that ignores enablement), so the WS turn bounces to HTTP where your new 503 denial fires — plus a WS regression test for both guard sites.

2. **Routed regression for health neutrality** (codex P1): the new tests exercise `_handle_stream_error` directly. An externally routed test that injects the exact upstream rejection and asserts the selected accounts' `error_count`/backoff are unchanged after failover would protect against a transport path that loses the message or writes health elsewhere.

3. Two small CodeRabbit test suggestions are still open: asserting `failure_class == "retryable_transient"` in the parameterized failover test (tests/unit/test_proxy_utils.py ~404-410), and asserting the `gpt-5.6-sol` registry precondition in the disabled-shadow-source integration test (tests/integration/test_model_source_routing.py ~647-665). Both are one-liners.

On 503 vs terminal 4xx for `model_source_disabled`: your reasoning (source-availability family, operator-actionable) is sound as a default; flagging it for the maintainer to confirm at merge, but I don't consider it a blocker.

One optional note: on the hot path, every subscription-model request now takes a second `get_background_session()` context in `_disabled_model_source_denial` (the registry-model candidates `continue` past the DB query, but the session is still opened). A cheap short-circuit — skip the denial probe when the model resolves in the subscription registry for an unscoped key, mirroring the precedence rule that already guarantees a miss — would keep this off the per-request cost after the recent hot-path CPU work.

Soju06:
Thank you @Komzpa for identifying and fixing both halves of this problem — the disabled-source fallthrough to subscription accounts and the account-health penalty from the resulting model-entitlement 400. Both fixes were sound and the owner approved the direction.

Since there has been no response since 2026-08-20 across two triage rounds, and the itemized blockers from the 2026-08-26 review (WebSocket guard gap, routed health-neutrality regression test, two one-line asserts) remained unaddressed while main kept advancing (the dispatch path this touches was reshaped by the continuity-aware selection work), we have taken this over in #1978 to get the fix landed.

Your two commits are preserved verbatim on the takeover branch with your author fields intact — attribution is fully retained; the takeover commits sit on top and only add the rebase adaptation, the blocker fixes, and tests.

Closing this PR in favor of #1978.

PR #1857, opened 2026-08-20
fix(proxy): stop hard bridge keys wedging on a leaked half-open probe and a rejected continuity anchor

Komzpa:
Two independent ways a hard HTTP bridge key stops making progress on its own,
both observed on the same deployment, both ending in a client-visible failure
with no upstream fault: a retry-circuit half-open lease that is taken and never
returned, and a continuity anchor that upstream has rejected but the proxy keeps
replaying. They share the file and the recovery machinery, so they are fixed
together.

## First defect: a leaked half-open probe

A hard bridge key that carries any durable retry-circuit row can be refused for
up to the full half-open lease (600s) while the client is told `retry_after=1`.
The client reconnects roughly once a second, is suppressed each time, and the
key makes no progress until the lease expires on its own.

### Live evidence

Measured over a 3 hour window on a running deployment:

- **103** client-visible `submit_retry_circuit_suppressed`
- **0** `http_bridge_retry_circuit event=suppressed`

The cooldown branch is the only branch that emits `event=suppressed`, and it
never fired. All 103 suppressions came from the half-open branch, which reports
`cooldown=0`, so the client receives `retry_after=1` while the key stays locked
for the remainder of the lease.

Circuit transitions in the same window: 38 `half_open`, 5 `opened`, 0 `reset`.
All 5 `opened` records read `failures=2 cooldown_seconds=60.0`; the consecutive
failure counter never reached 3 during the container's lifetime, so the observed
lockout was not produced by a key crossing the open threshold.

### Root causes

**1. A cooldown that had already elapsed loads as a live one.**
`_load_http_bridge_retry_circuit` computed `persisted_cooldown_until =
now_monotonic + cooldown_remaining` unconditionally. When the durable cooldown
had already elapsed, `cooldown_remaining` is `0.0` and the deadline becomes
`now_monotonic`: non-zero and in the past at the same time. That is exactly the
state `_http_bridge_precreated_retry_allowed` reads as "a cooldown just ended",
so it takes the exclusive half-open probe lease on a key that was not cooling
down. Below-threshold rows reach this on every load, because
`persist_retry_circuit` writes `now_wall` as `cooldown_until_epoch` for a
failure count under the open threshold.

The deadline now loads as `0.0` when nothing remains, which is the value the
rest of the circuit already uses for "not cooling down".

**2. A probe lease is only ever released by settling the circuit.**
A half-open lease admits one probe and suppresses every other request on the key
until that probe records a failure or clears the circuit on success. A probe
that ends in this proxy's own continuity-ownership loss does neither, so the
lease is held for its full duration while the upstream is healthy.

New `_release_http_bridge_retry_circuit_half_open` hands the lease back without
touching `consecutive_failures` or any live cooldown: the next request simply
becomes the probe. It is called from
`_reset_http_bridge_session_after_local_terminal_error` and from the
`CONTINUITY_OWNER_UNAVAILABLE` reattach path. The six server-continuity details
(`continuity_owner_unavailable`, `previous_response_owner_unavailable`,
`previous_response_not_found`, `bridge_previous_response_not_found`,
`bridge_owner_unreachable`, `bridge_instance_mismatch`) release the lease and no
longer charge the circuit, since they describe this proxy losing its anchor
rather than an upstream transport failure.

The half-open suppression branch had no log line of its own. It now logs
`event=half_open_suppressed` with the remaining lease, so a key held behind a
lease is distinguishable in the logs from one that was admitted. Returning a
lease logs `event=half_open_released` and increments the existing
`http_bridge_retry_circuit_total` counter under a new `half_open_released`
outcome.

**3. The reset that escapes a stale anchor produces the failure that re-arms the cooldown.**
`_reset_http_bridge_session_after_local_terminal_error` closes the bridge
session, which tears down a websocket that sibling turns may still be pending
on. Those readers wake on the teardown and report the ordinary transport class
`stream_incomplete`, which the circuit charges. The sequence closes on itself
with no upstream input:

```
cooldown -> reattach cannot reuse the durable session
        -> anchor replayed against an account that never held it
        -> previous_response_not_found
        -> local reset -> teardown reported as stream_incomplete
        -> cooldown
```

When the reset is performed for our own continuity loss, it now marks the
physical sends it is already settling with the existing
`_HTTPBridgeResponseCreateAttempt.disarmed` flag, the same marker a failed send
sets. The reader's attempt selection then classifies them as `settled` instead
of `eligible`, so the teardown is not charged. A teardown without that marker
keeps the send eligible and charges the circuit exactly as before.

### Proof of oracle

Eight tests are added. Run against this branch's `app/` tree they all pass. Run
unchanged against `main`'s `app/` tree (`52092bc9`), four of them fail:

```
FAILED test_http_bridge_retry_circuit_ignores_server_continuity_ownership_loss
FAILED test_http_bridge_local_terminal_error_reset_returns_half_open_probe
FAILED test_http_bridge_retry_circuit_elapsed_durable_cooldown_does_not_burn_half_open_probe
FAILED test_http_bridge_stale_anchor_reset_does_not_arm_its_own_cooldown
4 failed, 4 passed, 654 deselected in 2.06s
```

### Negative controls

The other four pass in both trees by design. They fail if this change
over-reaches:

- `test_http_bridge_upstream_teardown_without_continuity_loss_still_arms_cooldown`
  — a teardown with no continuity-loss marker still arms the cooldown.
- `test_http_bridge_retry_circuit_still_trips_on_genuine_upstream_stream_incomplete`
  — genuine upstream `stream_incomplete` still trips the circuit.
- `test_http_bridge_retry_circuit_suppressed_attempt_does_not_extend_cooldown`
  — a suppressed attempt does not extend the cooldown.
- `test_http_bridge_live_durable_owner_is_not_reclaimable`
  — ownership, reclaim, and takeover behavior are unchanged.

## Second defect: a continuity anchor upstream has rejected is never dropped

The same hard bridge key has a second way to stop making progress, with no
cooldown involved. `previous_response_not_found` is upstream's definitive answer
about one response id. The bridge treats it as terminal for the request, but it
never invalidates its own copies of that id, so the next turn on the session
sends the same rejected anchor again.

Two carriers hold it, and both re-inject it:

- `http_bridge_sessions.latest_response_id`, read by the fresh-reattach path
  (`fresh_reattach_anchor_injected`), which adds `previous_response_id` to a
  payload that arrived without one. It survives process restarts.
- the in-memory `session.last_completed_response_id`, read by the session-level
  anchor path (`session_anchor_injected`).

`_clear_durable_http_bridge_response_anchor` already exists and does exactly the
right fenced write, but it is only reachable from the eventless
`missing_response_created_timeout` path. `session.last_completed_response_id` is
only cleared on an account rebind. Neither reacts to upstream saying the id does
not exist.

Injecting the anchor also enables store-context input trimming, so the damage is
larger than one lost anchor: a client payload carrying the whole conversation is
trimmed against a response upstream no longer has. A representative live line is
`store_context_input_trimmed original_items=192 trimmed_to=4` on a turn whose
anchor upstream had already rejected twice.

### Live evidence

Measured on a deployment running `main` plus the first commit of this branch:

| | container lifetime (35 min) | one 15-min window inside it |
| --- | --- | --- |
| `continuity_fail_closed reason=previous_response_not_found` | 102 | 93 |
| of those, `previous_response_source=proxy_injected` | 51 | 46 |
| distinct `previous_response_id` behind all of them | 4 | 4 |
| `event=fresh_reattach_anchor_injected` | - | 48 |
| `session_anchor_injected` | - | 52 |
| `store_context_input_trimmed` | - | 100 |
| `event=durable_anchor_invalidated` | **0** | **0** |
| `event=terminal_error detail=stream_incomplete` | - | 93 |

Four response ids account for every one of those failures, and the anchor
invalidation path never fired once. A read-only SQLite check against the same
deployment found those ids still sitting in `http_bridge_sessions.latest_response_id`
on `state=active` rows minutes after upstream had rejected them. Clearing that
column is what lets such a session continue, and no code path does it.

### Fix

When upstream rejects an anchor this proxy injected, invalidate both carriers.
Three conditions keep the invalidation to exactly the id that is proven gone:
the anchor must be proxy-injected; it must be the id upstream named, when the
error message carries one; and the durable write, which reuses the existing
fenced clear, must still find that id on the row. So a concurrent turn that
already stored a newer anchor keeps it, and a differently anchored request that
merely matched the same anonymous error event keeps its own. The in-memory
anchor is cleared under the same equality check, and its dependent state (input
item count, prefix fingerprint, pending tool-call manifest) goes with it.

A client-supplied anchor is the client's own state and is left alone: it still
returns the standard error, and this proxy simply stops storing and replaying
it on the client's behalf.

`ambiguous_continuation_recovery_mode` is not the lever for this.
`_HTTP_BRIDGE_AMBIGUOUS_RECOVERY_ERROR_CODES` covers outcomes where upstream
acceptance is genuinely unknown; this outcome is a definitive answer, and the
`server_*` modes replay from the anchor that was just rejected.
`client_full_history_once` cannot apply either, because it requires a
client-supplied `previous_response_id` (`request_submit.py`).

### Proof of oracle

`test_previous_response_not_found_drops_only_the_proxy_injected_anchor` drives a
real `previous_response_not_found` event through
`_process_http_bridge_upstream_text` for a session holding the anchor in both
carriers, and
`test_durable_bridge_clear_response_anchor_keeps_a_newer_anchor` exercises the
fenced write directly. With the two production hunks reverted and the tests
unchanged:

```
FAILED test_previous_response_not_found_drops_only_the_proxy_injected_anchor[True-True]
   AssertionError: assert 0 == 1                 (the durable clear is never called)
FAILED test_durable_bridge_clear_response_anchor_keeps_a_newer_anchor
   AssertionError: assert None == 'resp_newer'   (an unrelated newer anchor is wiped)
2 failed, 2 passed, 724 deselected
```

The other two parameters are the negative controls, and both pass in either
tree by design:

- `[False-True]` is a client-supplied anchor. No clear may happen and the
  session must be untouched.
- `[True-False]` is a proxy-injected anchor while upstream's message names a
  different response id. Removing only the named-id check makes this one fail
  with `assert 1 == 0`.

## Scope

On the retry circuit, only its accounting of server-side continuity loss
changes. Ownership, reclaim, takeover, quarantine, and the open threshold are
untouched. Genuine upstream `stream_incomplete`, `stream_idle_timeout`, and
`clean_close` failures keep tripping the circuit as before.

On the anchor, only `previous_response_not_found` against an anchor this proxy
injected triggers the invalidation. The eventless-timeout clear keeps its
existing behaviour, alias rows and `latest_turn_state` stay in place so the
durable session remains reattachable, and the fenced write still refuses when
the row no longer holds the rejected id. `clear_latest_response_anchor` and
`_execute_fenced_session_update` gain an optional argument each and are
unchanged for existing callers.

Not covered: the native WebSocket transport keeps its own
`_WebSocketContinuityState.last_completed_response_id`, which is likewise only
rewritten on a completion and never invalidated by a rejection. Every failure
measured here is on the HTTP bridge, so that surface is left alone rather than
changed without evidence of the same wedge.

## Why this is a separate PR

Checked against the open PRs that touch these files; none owns either surface:

- **#1633** changes account-health neutrality in `_fail_pending_websocket_requests`.
  On the retry circuit it moves the other way, adding
  `HTTP_BRIDGE_EVENTLESS_TIMEOUT_CODE` as a fourth charged detail. It touches no
  anchor-invalidation code.
- **#1384** touches no `retry_circuit` code at all.
- **#1649** reads `session.last_completed_response_id` as a liveness baseline
  before retiring a stale pending session. It reads that field and never clears
  it, so it neither fixes nor conflicts with the anchor invalidation here.
- **#1860**, **#1647**, and **#1183** share files with this branch but touch
  neither the retry circuit nor the anchor carriers.
- **#1730**, **#1841**, and **#1619** have merged; this branch is rebased on top
  of all three.

## Validation

```
uv run ruff format --check .                      ->  955 files already formatted
uv run ruff check app/ tests/                     ->  All checks passed!
uv run ty check                                   ->  All checks passed!
uv run python scripts/check_proxy_architecture.py ->  proxy architecture checks passed
uv run pytest tests/unit -q -p no:randomly        ->  6443 passed, 3 skipped, 12 warnings in 136.23s
```

`http_bridge/mixin.py` ends at 2434 lines and `streaming/mixin.py` at 1100,
within the 2436 and 1100 ratchets in `scripts/check_proxy_architecture.py`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recovery when response ownership or continuity becomes unavailable.
  * Prevented stale response anchors from removing newer valid anchors.
  * Improved handling of rejected previous-response references.
  * Added safer, bounded retries during model-transition conflicts.
  * Refined retry suppression and cooldown behavior for continuity failures.

* **Tests**
  * Added coverage for recovery flows, retry limits, stale anchors, cooldowns, and durable session state.


Komzpa:
@codex review

Komzpa:
@codex review

Soju06:
Thanks — the diagnosis of both wedges is outstanding. I verified that current main still has the elapsed-cooldown defect (`retry_circuit.py` computes `persisted_cooldown_until = now_monotonic + cooldown_remaining` unconditionally, lines 260-261), the live evidence is convincing, and the proof-of-oracle tests with negative controls are exactly the kind of coverage we want. CI is fully green. That said, this can't merge in its current shape:

1. **The branch re-lands the five reverted PRs.** The merge-base diff against current main contains the replayed squash commits of #1849, #1706, #1847, #1730 and #1619 (b4e156e00, 846520a2c, 11dee3aaf, 05b7bf52d, 86be7fa71), which #1858 reverted and which #1862 re-offers as its own reviewable branch. Merging #1857 as-is would land that entire ~4,600-line set as a side effect of a wedge fix, bypassing the maintainer review #1862 exists for. Please either rebase this branch onto current main with only the wedge-fix commits, or explicitly mark it as stacked on #1862 and rebase once that lands. Note the PR body's claim that "#1730 ... and #1619 have merged" is no longer true post-revert.

2. **Duplicate migration with #1862.** `app/db/alembic/versions/20260820_000000_repair_retired_identity_and_warmup_stamp.py` (from #1847) exists in both branches, both revising `20260816_000000_add_model_source_embeddings` — whichever merges second breaks the revision chain. The edit to the shipped `20260813_000000_add_file_account_pins.py` also belongs to that reverted set, not to this fix.

3. **Open codex findings on the current head (fa2ef6a, 2026-08-22 review):** (a) `request_submit.py:3270` — a hard-key request with no explicit anchor/file but `candidate_portable=False` (e.g. it carries a durable `operation_id`) sets `account_bound_replay=True` yet still passes `require_same_account=False`, letting selection resend account-scoped state on a different account; the fresh-account exception should apply only to portable/account-neutral requests. (b) `upstream_events.py:292` — the quarantine generation re-check after `renew_live_session`'s awaited durable write can only return False, it cannot undo the persisted response-ID/continuity mutation, so a stale generation-N recovery can overwrite the durable owner/anchor after N+1 is installed; the generation/snapshot needs to be part of an atomic durable CAS or serialized with generation changes. (c) `ring_membership.py:274` — the derived owner hostname check misses `?`, `#`, and whitespace, so a malformed instance ID becomes a query/fragment and the session forwards to the wrong host/port. Notably (b) and (c) are in the bundled quarantine-fencing / blue-green commits rather than the two described fixes — which is itself an argument for splitting.

4. **Scope.** Beyond the five re-landed PRs, the branch bundles blue-green owner forwarding, quarantine recovery fencing, compact state-anchor fixes, websocket health-penalty changes, and a silent-account replay change — nine openspec change directories in one PR whose body describes two defects. Each of those deserves its own PR with its own evidence, like the excellent write-up here.

Also heads-up on overlap so work isn't duplicated: #1908 attacks the same expired-durable-cooldown → leaked half-open lease root cause, and #1863/#1867/#1902/#1879/#1891 all touch rejected/stale-anchor recovery. Worth cross-referencing when you split this up.

Soju06:
Closing this one out after triage review. The diagnosis of both wedges was excellent — the maintainer confirmed the elapsed-cooldown defect on main (`retry_circuit.py` computes `persisted_cooldown_until = now_monotonic + cooldown_remaining` unconditionally) and the live evidence and proof-of-oracle tests set a high bar. But the four blockers from the 2026-08-26 review comment remain unaddressed after several days, the branch is now conflicting with main, and in particular it still re-lands the five reverted PRs (#1849/#1706/#1847/#1730/#1619) that #1862 exists to re-offer for review, along with the duplicate `20260820_000000_repair_retired_identity_and_warmup_stamp.py` migration and the three open codex findings on fa2ef6a2 (`request_submit.py:3270`, `upstream_events.py:292`, `ring_membership.py:274`).

Meanwhile both core fixes have focused successor vehicles on current main: #1962 carries the elapsed-cooldown normalization and exclusive half-open probe return (superseding #1908 as well), and #1902 — together with the already-merged #1863 — retires proxy-injected anchors after an explicit `previous_response_not_found`, with contributor attribution for the diagnosis here. Rather than keep a conflicting 31-commit branch open in parallel, we'll track those.

If you'd like to re-offer any of the remaining bundled changes (quarantine-recovery fencing, blue-green owner forwarding, compact state-anchor handling, websocket health-penalty scoping), please open focused PRs against current main per the 2026-08-26 comment — each with its own evidence, like the write-up here.

PR #1925, opened 2026-08-26
Fix invalid refresh tokens not surfacing re-auth action

plastictaste:
## Summary

- classify the upstream `invalid_refresh_token` error as a permanent authentication failure
- map it to `REAUTH_REQUIRED` so active-looking accounts are removed from routing
- expose the existing Re-authenticate action in the dashboard after refresh-token failure
- add regression coverage for classification and status mapping

## User impact

When CodexLB receives `invalid_refresh_token`, the account will now be marked “Re-auth required” instead of remaining “Active” and retrying a dead token. The Accounts page can then use the existing Re-authenticate action to start a fresh OAuth flow.

## Verification

- `uv run ruff check app/core/balancer/logic.py tests/unit/test_auth_refresh.py`
- `uv run ruff format --check app/core/balancer/logic.py tests/unit/test_auth_refresh.py`
- `uv run pytest tests/unit/test_auth_refresh.py -q` (12 passed)

## Summary by CodeRabbit

* **Bug Fixes**
  * Accounts with invalid refresh tokens are now marked as requiring re-authentication.
  * Affected accounts are excluded from routing until they sign in again or provide fresh credentials.
  * Existing authentication state is preserved while re-authentication is required.
  * Users receive clearer guidance when a refresh token can no longer be used.

* **Tests**
  * Added coverage for refresh-token failures, routing exclusion, and re-authentication status handling.

### DIFF
--- a/app/core/balancer/logic.py
+++ b/app/core/balancer/logic.py
@@ -29,6 +29,10 @@ PERMANENT_FAILURE_CODES = {
     "app_session_terminated": "ChatGPT session ended - re-login required",
     "account_session_expired": "ChatGPT session ended - re-login required",
     "account_auth_invalidated": "Authentication failed after token refresh - re-login required",
+    # The OAuth endpoint uses this code for an invalid/revoked refresh token.
+    # Keep it in the permanent set so the account is surfaced for re-auth
+    # instead of remaining active while every request retries the dead token.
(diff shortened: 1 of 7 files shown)


Komzpa:
@codex review

Komzpa:
@codex review

PR #1947, opened 2026-08-28
fix(proxy): retire cooldown-suppressed HTTP bridge sessions

### LINKED ISSUE #1943: bug: half-open retry probe reuses an undispatched WebSocket pre-created during cooldown
### Pre-flight checklist

- [x] I searched existing issues and PRs.
- [x] I confirmed this behavior is still present on current main.
- [x] The production trigger is intermittent, but the lifecycle defect and a deterministic regression test are identifiable.

Closest reports are #1942 (initial zero-event HTTP-bridge stream failure), #1830/#1841 (misclassified recovery frames and poisoned anchors), #1857/#1908 (leaked or expired half-open lease), #1891 (poisoned-anchor quarantine), and #1876 (long-lived ghost operations). None covers a WebSocket created during cooldown, never dispatched to, and later reused as the half-open probe.

### codex-lb version

Production observation: downstream build based on the 1.24.1-beta.1 line.

Root cause re-verified against current upstream main at 2268f8caf1fe9d74a8734bd3f9cd8bd5152b5d3f (2026-08-28).

### Deployment method

Kubernetes, PostgreSQL durable HTTP bridge, one codex-lb workload replica, separate TCP/TLS relay. The workload and relay processes did not restart during the incident window.

Codex Desktop using native streaming /v1/responses.

### Model(s) involved

gpt-5.6-sol

### What happened?

Two genuine eventless upstream WebSocket closures opened the hard-key HTTP-bridge retry circuit. Codex then retried during the 60-second cooldown.

Each retry first created or selected an upstream WebSocket and only later reached the retry-circuit admission check in _submit_http_bridge_request_with_handoff. The check returned the expected 503, but the undispatched session was not retired or quarantined.

One such connection was created near the end of the cooldown and remained registered while idle. About 90 seconds later, the next request reused that connection as the exclusive half-open probe. It failed almost immediately with another zero-event stream_incomplete, after which subsequent Codex retries returned:

    503 upstream_request_timeout
    HTTP responses session bridge is cooling down after repeated upstream timeouts; retry shortly.

The initial no-close-frame transport event is intermittent. The amplification is deterministic: a cooldown-suppressed request can leave a WebSocket in the reuse pool even though no response.create was dispatched on it.

### Sanitized event sequence

All identifiers and absolute timestamps are omitted. The same hard bridge key, durable session, account, and model were used throughout.

    T+0.0s
    http_bridge_event event=reader_failure
    detail=stream_incomplete
    response_events_seen=0
    transport_classification=websocket_transport_error
    error_message="... no close frame received or sent"

    T+8.4s
    http_bridge_event event=reader_failure
    detail=stream_incomplete
    response_events_seen=0
    transport_classification=websocket_transport_error

    T+8.4s
    http_bridge_retry_circuit event=opened
    failures=2
    cooldown_seconds=60.0
    detail=stream_incomplete

    # Several client retries during cooldown:
    http_bridge_event event=create request_stage=reattach
    http_bridge_retry_circuit event=suppressed retry_after_seconds=...
    http_bridge_event event=close

    # The last cooldown-era connection was not closed:
    T+57s
    http_bridge_event event=create request_stage=reattach

    # About 90 seconds after that connection was created:
    T+147s
    http_bridge_event event=reuse
    http_bridge_retry_circuit event=half_open failures=2

    T+149s
    http_bridge_event event=terminal_error
    detail=stream_incomplete

The durable operation evidence rules out an unbounded replay as the cause:

- the ambiguous zero-output operation stayed unknown;
- it had no response id, zero event bytes, incomplete spool;
- recovery_dispatch_count was exactly 1;
- the later half-open client turn was a different operation and stored only a locally synthesized response.failed event.

So the safe recovery dispatch was already bounded. The failing half-open attempt was caused by connection freshness/lifecycle, not permission to replay the ambiguous operation repeatedly.

### Current code path

On current main, the retry admission check occurs after a _HTTPBridgeSession has already been obtained.

app/modules/proxy/_service/http_bridge/request_submit.py around lines 977-1022:

    retry_allowed = await self._http_bridge_precreated_retry_allowed(...)

    if not retry_allowed:
        retry_after_seconds = ...
        _log_http_bridge_event("submit_retry_circuit_suppressed", ...)
        raise ProxyResponseError(503, ...)

This branch does not:

- set session.closed;
- set reconnect_requested / retire_after_drain;
- call _retire_http_bridge_after_drain_if_ready;
- remove the undispatched session from the reuse pool.

A similar startup-cooldown terminal path in http_bridge/streaming.py returns before submit without retiring the selected/pre-created session.

The comment says the gate is "before any reconnect/send", but it is only before send. Production logs show event=create before cooldown suppression.

### Why related work does not fully cover this

- #1942 covers the initial zero-event close/hanging SSE symptom, not cooldown-created connection reuse.
- #1841 fixes upstream error classification and same-anchor poison accounting.
- #1857 and #1908 address stale/expired half-open lease state and lease release, not WebSocket freshness.
- #1891 improves poison quarantine and terminal failure accounting. Even if merged, the late cooldown admission branch still leaves an undispatched connection reusable.
- #1886 can fall back to raw HTTP only when the suppressed request is provably replay-safe. A continuity-bound request that is not safe for that fallback still follows this path.

### Expected behavior

A request suppressed by the hard-key retry circuit must not leave a newly created, undispatched upstream WebSocket eligible for later reuse.

Preferably, admission should happen before opening an upstream connection whenever no safe bypass can apply. If the late check is structurally required for durable-operation arbitration, the selected session should be retired/quarantined before returning 503 when it has no visible pending work.

After cooldown expiry, the exclusive half-open probe should use a genuinely fresh upstream WebSocket, not one created by a previously suppressed request.

This should not:

- lower or disable the circuit threshold;
- allow a second ambiguous operation recovery dispatch;
- replay tool calls or user input without a durable fence;
- admit generic incremental requests during cooldown.

### Suggested regression coverage

1. Open a hard-key circuit with two zero-event stream_incomplete failures.
2. During cooldown, make a client request that creates a bridge session but is suppressed before response.create.
3. Assert that session is closed/retiring and cannot be returned by _get_or_create_http_bridge_session.
4. Advance past cooldown and submit concurrent requests.
5. Assert exactly one half-open probe is admitted and it opens a new upstream WebSocket.
6. Assert the ambiguous operation recovery count remains bounded at one and no tool/output transcript is replayed twice.

### Additional context

No durable session, alias, anchor, operation, or retry-circuit row was deleted or manually settled during the investigation. The state was inspected read-only.

hongzexin:
## Summary

- Retire an HTTP bridge session when late retry-circuit cooldown suppression returns 503.
- Retire the startup pre-submit cooldown terminal path as well.
- Preserve the existing 503 envelope, Retry-After header, and no-upstream-send behavior.
- Add regression coverage and an OpenSpec change.

## Fable5 analysis

Fable5 (claude-fable-5, response `msg_l4vLQ0i8RzQiFjCGE2W7cOiE`) evaluated the issue and found Strategy A (mark the session for retirement and use the existing bounded drain-retirement path after cooldown suppression) feasible with minimal risk. Strategy B would require broader request-admission plumbing and was not selected.

## Validation

- Targeted unit tests: 4 passed
- `tests/unit/test_proxy_http_bridge.py`: 681 passed; one unrelated fixture/environment failure (`sqlite3.OperationalError: no such table: file_account_pins`)
- Related integration tests: 3 passed
- Ruff format/check: passed
- `make architecture-check`: passed
- `make typecheck`: passed
- Strict OpenSpec validation for this change and affected specs: passed

Fixes #1943

Source sidechat: 01a04718-7322-72d0-b4b3-8f5bb4581157
Recovered from: 2268f8caf1fe9d74a8734bd3f9cd8bd5152b5d3f

## Summary by CodeRabbit

* **Bug Fixes**
  * HTTP bridge sessions suppressed by retry cooldown are now retired safely instead of being reused.
  * Requests continue to return the expected 503 timeout response without being sent upstream.
  * Shared sessions can finish pending work before closing, while authorized retries remain unaffected.

* **Tests**
  * Added coverage for session retirement, reconnect behavior, non-reusability, and ordinary failure handling.

### DIFF
--- a/app/modules/proxy/_service/http_bridge/request_submit.py
+++ b/app/modules/proxy/_service/http_bridge/request_submit.py
@@ -999,6 +999,14 @@ class _HTTPBridgeRequestSubmitMixin:
                 1,
                 math.ceil(await self._http_bridge_precreated_retry_cooldown_seconds(session)),
             )
+            # The session may have been created before this late admission
+            # check.  Do not leave an unsubmitted socket reusable during the
+            # cooldown: mark it for retirement before surfacing the 503 and
+            # let the normal bounded drain path handle registry detach,
+            # aliases, leases, and close ownership.
+            session.upstream_control.reconnect_requested = True
+            session.upstream_control.retire_after_drain = True
+            await self._retire_http_bridge_after_drain_if_ready(session)
             _log_http_bridge_event(
                 "submit_retry_circuit_suppressed",
                 session.key,
--- a/app/modules/proxy/_service/http_bridge/streaming.py
+++ b/app/modules/proxy/_service/http_bridge/streaming.py
@@ -3999,6 +3999,14 @@ class _HTTPBridgeStreamingMixin:
                 request_state.request_id,
                 retry_cooldown_seconds,
             )
+            # Session creation precedes this startup admission check.  Mark
+            # the opened bridge as retiring before returning the terminal
+            # cooldown result so no later request can reuse its socket.  The
+            # bounded helper closes immediately when idle and otherwise waits
+            # for existing owners to drain.
+            session.upstream_control.reconnect_requested = True
+            session.upstream_control.retire_after_drain = True
+            await self._retire_http_bridge_after_drain_if_ready(session)
             # This path returns before the request is submitted, so the normal
             # detach/finally cleanup cannot settle an API-key reservation.
             # Release it before handing the synthetic terminal event to the
--- a/openspec/changes/retire-cooldown-suppressed-http-bridge-session/context.md
+++ b/openspec/changes/retire-cooldown-suppressed-http-bridge-session/context.md
@@ -1,0 +1,24 @@
+# Context
+
+The bridge opens or reuses a session before the request-specific pre-created
+retry decision. That ordering is required for continuity and account routing,
+but it means a cooldown decision can arrive after a WebSocket has already been
+registered. Returning an error alone is insufficient: the session remains in
+the registry and may be selected by a later request.
+
+The existing `_retire_http_bridge_after_drain_if_ready` helper is the correct
+lifecycle owner. Setting both control flags makes `_http_bridge_session_reusable_for_request`
+reject new work immediately. The helper closes only when there is no visible
+pending request, queue count, unanchored reservation, or competing close, and
+otherwise leaves the session marked for retirement until the remaining owner
+settles it.
+
+The startup terminal path is equivalent to late suppression from a lifecycle
+perspective: it has an already-created session but no upstream `response.create`
+dispatch. It must use the same flags and helper so both paths are idempotent
+and share registry detachment, alias cleanup, lease settlement, and bounded
+socket close behavior.
+
+Replay exceptions remain intentionally unchanged. A proof-gated full resend or
+an operation-fenced continuity replay is an explicitly authorized dispatch;
+those paths must not be retired by the generic cooldown suppression branch.
--- a/openspec/changes/retire-cooldown-suppressed-http-bridge-session/proposal.md
+++ b/openspec/changes/retire-cooldown-suppressed-http-bridge-session/proposal.md
@@ -1,0 +1,41 @@
+# Retire Cooldown-Suppressed HTTP Bridge Sessions
+
+## Why
+
+Issue #1943 exposes a lifecycle leak in the HTTP Responses bridge. A request
+can create and register an upstream WebSocket before the late retry-circuit
+admission check runs. When hard-key cooldown suppresses that request, the proxy
+returns the expected 503 but leaves the newly created session reusable. The
+orphaned socket can then be selected by a later half-open probe and keep the
+key in a repeated failure loop.
+
+The same leak exists in the startup pre-submit cooldown path used for
+continuity-bound requests: it returns a synthetic 503 before submit without
+retiring the session that was already opened.
+
+## What Changes
+
+- Mark a session as `reconnect_requested` and `retire_after_drain` whenever a
+  hard-key cooldown rejects a request before upstream dispatch.
+- Invoke the existing bounded drain-retirement helper immediately so an idle
+  session is detached and closed; sessions with other pending work remain
+  non-reusable until that work drains.
+- Cover both late submit suppression and startup pre-submit terminal paths.
+- Preserve proof-gated and operation-fenced replay bypasses, durable retry
+  circuit state, reservation settlement, and the existing 503 envelope.
+
+## Impact
+
+- Cooldown-suppressed bridge sessions cannot be reused for later requests.
+- Newly opened but never-dispatched WebSockets are closed and removed from the
+  registry when no other work owns the session.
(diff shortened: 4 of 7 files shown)


Soju06:
Triage update (round 16): this PR was reviewed and adversarially verified this round and is **approved for merge** — the change itself is sound.

Two things blocked the merge attempt today:

1. ~~`Tests (pytest, integration-bridge)` hang~~ — resolved. The 180s pytest-timeout in `tests/integration/test_proxy_websocket_responses.py::test_backend_responses_websocket_fails_over_confirmed_proxy_connect_before_dispatch` did **not** reproduce on a same-snapshot re-run (attempt 3 is green), so it was the known bridge-shard flake, not a lifecycle regression from this PR.
2. `Contributors attribution` — `hongzexin` is missing from `.all-contributorsrc` (first-time-contributor gate). Please add yourself on this branch (`npx all-contributors add hongzexin code`) and push; that's the only remaining red.

Once the attribution entry lands and CI is green on the new head, this merges immediately next pass.

Soju06:
DEMOTED IN MERGE PLANNING: semantic overlap with #1891, which has now merged. The PR is currently CONFLICTING against main — the conflict in `app/modules/proxy/_service/http_bridge/request_submit.py` sits precisely in the cooldown/retry-block region this PR changes: #1891 replaced the cooldown computation with the `_http_bridge_precreated_retry_block` primitive (now at `request_submit.py:1044` and `streaming.py:4437`/`4620` on main), while this PR's retirement flags are anchored to the old `_http_bridge_precreated_retry_cooldown_seconds`-only suppression branches.

Needs an author rebase plus semantic re-verification against #1891's new primitive: confirm the session-retirement flags (`reconnect_requested` + `retire_after_drain` + `_retire_http_bridge_after_drain_if_ready`) are applied on the retry-block rejection paths as they now exist on main, and that #1891's key quarantine does not already retire/close the same sessions (which would change what this PR still needs to cover). Happy to re-review after the rebase.

Soju06:
Taking this over in #2072 — thank you @hongzexin for the fix and the analysis; your commit (`fix(proxy): retire cooldown-suppressed HTTP bridge sessions`) is preserved verbatim with your authorship as the first commit of the new branch, and you are now listed in `.all-contributorsrc`.

What #2072 adds on top of this branch:

1. The two owner blockers: the missing Contributors attribution entry, and the rebase onto main after #1891 (the retirement now hangs off `_http_bridge_precreated_retry_block` in both suppression branches; re-verified that #1891's quarantine does not already retire these sessions — your tests still fail on main and pass on the new branch).
2. A concurrency fix found during adversarial verification: flagging the *shared* session unconditionally could fence off a half-open probe admitted on the same session between its gate and its dispatch registration (502 "bridge is retiring" for the one submission the cooldown permits). #2072 makes the submit count as an admission waiter from entry and retires the session only when no other turn owns it, with a deterministic interleaving test.

Closing this PR to avoid duplicate review; please follow along on #2072.

PR #1952, opened 2026-08-28
fix(proxy): preserve tool search pairs through replay

### LINKED ISSUE #1530: bug: client-fault upstream 400 penalizes the serving account and 502s unrelated hard-sticky sessions (plus an unclassified 'tool search call' missing-tool-output variant)

Komzpa:
## Summary

Codex replay histories can contain completed `tool_search_call` / `tool_search_output` pairs. On `main` the owner-unavailable projection drops them, so a replacement account never sees the discovered tools and the next turn fails with `No tool output found for tool search call`. This PR keeps completed, client-owned pairs in the account-neutral replay when every discovered tool declaration is itself account-neutral, sanitizes their response-owned ids on both transports, and closes a V1 compact-trigger gap.

- **Replay safety (`app/modules/proxy/replay_safety.py`).** `tool_search_call` / `tool_search_output` are no longer omitted from the projection. `tool_search_output` is validated in the shape Codex actually serializes (`codex-rs/protocol/src/models.rs` `ResponseItem::ToolSearchOutput { id?, call_id, status, execution, tools }`): `tools` is a list of `LoadableToolSpec` values, each run through the existing declared-tool rules (`_tool_declaration_is_account_neutral` for `function` / `custom`, with `defer_loading` accepted only as a bool) or a `namespace` whose `tools` hold such declarations. The generic scoped-state walk treats a validated `tools` list like `additional_tools`, so JSON-schema property names such as `parameters.properties.file_id` are not read as live references. String `output`, missing `tools`, `status: failed`, `execution: server`, MCP / hosted / container-bound declarations, unknown fields, and nested namespaces fail closed. The string-`output` branch and its `"output"` field entry are gone.
- **Transports.** HTTP bridge and WebSocket strip replayed tool-search ids before fresh retries and compare trimmed content instead of length; the WebSocket client full-resend body is built from the sanitized *untrimmed* list while prefix trimming stays on the anchored upstream body; the `*_previous_response_input_trimmed` log line fires only when the item count actually changed.
- **Compact triggers (`api.py`).** `/v1/responses/compact` captures the raw input before `V1ResponsesCompactRequest` hoists trailing developer/system messages and returns 400 when a `compaction_trigger` exists but is not the last top-level item. Duplicate terminal triggers keep their normalization. This is a client-visible V1 400 for a previously accepted shape (`[..., compaction_trigger, developer]`). The unreachable `validate_compaction_trigger_shape` block in `_compact_responses` was removed rather than carried over.
- **Fixture sanitiser (`scripts/traffic_analysis/codex_body_sanitize.py`).** The traffic-parity drift guard (`test_the_item_and_content_key_allowlists_cover_what_production_validates`, from #2401) requires every item type in `_ACCOUNT_NEUTRAL_INPUT_ITEM_FIELDS` to have a sanitiser rule, so `tool_search_call` / `tool_search_output` rules were added with the same upstream shape.
- **OpenSpec.** `openspec/changes/preserve-tool-search-replay-semantics/` now carries the ADDED tool-search / compact-trigger requirements with the upstream shape, plus a MODIFIED "Verified full resend can recover from selection-time owner loss" requirement so the main spec stops promising that the pair is omitted. `proposal.md` states the reach explicitly.

## Reach

The pair is retained only when the rest of the request is already account-neutral. Responses-Lite requests that declare `tool_search` / `namespace` tools in `additional_tools`, and non-Lite requests declaring `{"type": "tool_search"}` or `defer_loading` function tools at top level, still fail the declared-tool allowlists and stay owner-bound. Extending those allowlists is a separate, spec-bearing change, so this PR does not claim `Fixes #1530`.

## Review history

- 09-08 / 09-14 blocker (`_ACCOUNT_NEUTRAL_TOOL_SEARCH_OUTPUT_TOOL_FIELDS = {"name"}` matched no Codex shape, so real pairs failed closed where `main` proceeded): fixed with option (a). `tests/unit/test_replay_safety_tool_search.py` uses the upstream JSON (`function` with `strict`/`parameters`, `namespace` with a deferred child from `responses_api_tests.rs::loadable_tool_spec_namespace_serializes_with_deferred_child_tools`) and `test_backend_responses_websocket_replays_tool_search_pair_to_replacement_account` delivers that shape to the replacement account end to end.
- String-`output` branch dropped together with the `"output"` field entry; the outdated Codex P1 thread (string `output` + malformed `tools`) is moot: an `output` field is now an unknown field and the item fails closed regardless of `tools`.
- Dead `validate_compaction_trigger_shape` block removed; `_capture_raw_v1_compaction_trigger_error` docstring now states the actual rule.
- P3 nits: trim log gated on a real length change; `defer_loading` decision documented in code and spec.
- Rebased onto current `main` as three commits; the 54 `openspec/specs/**` trailing-newline deletions are gone and the `compact_pin_denial` resolution from the 09-10 merge was dropped (removed on `main` by #2416).

## Validation

- `uv run ruff check . && uv run ruff format --check .` — All checks passed! / 1287 files already formatted
- `uv run ty check` — All checks passed!
- `npx @fission-ai/openspec@1.11.0 validate --type change --strict preserve-tool-search-replay-semantics` — Change 'preserve-tool-search-replay-semantics' is valid
- `uv run pytest tests/unit/test_replay_safety.py tests/unit/test_replay_safety_tool_search.py` — 232 passed
- `uv run pytest tests/unit/test_proxy_utils.py tests/unit/test_proxy_http_bridge.py` — 2480 passed
- `uv run pytest tests/integration/test_http_responses_bridge.py` — 197 passed
- `uv run pytest tests/integration/test_proxy_websocket_responses.py tests/integration/test_proxy_compact.py tests/integration/test_proxy_compact_triggers.py` — 216 passed
- `uv run pytest tests/unit/test_codex_body_fixtures.py tests/unit/test_replay_safety_tool_search.py` — 179 passed
- `uv run pytest tests/unit` (remaining files) — 7682 passed, 4 skipped; the single failure (`test_codex_body_fixtures` drift guard) was fixed by the sanitiser rules above and re-verified


Komzpa:
@codex review

Komzpa:
@codex review

Soju06:
CI is red on the current head (590a1d2), so this needs another pass before it can land:

1. **Failing unit tests (blocking).** `tests/unit/test_proxy_http_bridge.py::test_stream_via_http_bridge_projects_plaintext_durable_full_resend_when_owner_is_unavailable` fails on the `[None-False-None]`, `[None-False-gpt-5.3]`, and `[None-True-None]` parametrizations. Removing `tool_search_call`/`tool_search_output` from `_ACCOUNT_NEUTRAL_REPLAY_OMITTED_ITEM_TYPES` (app/modules/proxy/replay_safety.py:21) changes the durable full-resend projection contract that this test pins: the `completed_search_bookkeeping` fixture (tests/unit/test_proxy_http_bridge.py:24593) is now retained (ids stripped) instead of omitted. If preserving the pair is the intended contract — your openspec delta says it is — please update the expected projected input for those parametrizations accordingly, and run the full unit suite rather than the targeted selection listed in the PR description; that's what caught this.

2. **Open Codex P1 from the c9c65ec review (replayed tool-search call ids).** `responses_input_items_are_self_contained_fresh_replay` still rejects any item with a nonempty `id` (app/modules/proxy/replay_safety.py:281), while canonical replayed `tool_search_call` items carry a server-assigned id (see the existing fixtures cited in that review). The new trim helpers only strip `tool_search_output` ids (`_strip_websocket_replayed_tool_search_output_id`, app/modules/proxy/_service/websocket/helpers.py), and `test_websocket_client_previous_response_full_resend_retry_allows_tool_search_history` uses id-less fixtures, so the direct-WebSocket full-resend path likely still fails closed for real replay histories — i.e., the nightly `No tool output found for tool search call` failure may persist on that path. Please either handle the server-assigned call id (strip before validation, mirroring the output-id treatment) with a regression test using an id-bearing fixture, or explain why that path is out of scope for this PR.

The compact-trigger validation (duplicate rejection + V1 pre-hoist check via `_capture_raw_compaction_trigger_error`) and the server-executed/ambiguous-output rejections look correct against main and are well covered.

Komzpa:
@codex review

Soju06:
Thanks for the quick rework — this pass fixed the failing `test_proxy_http_bridge` parametrizations, added the V1 pre-hoist trigger validation, and the OpenSpec artifact validates strict-clean when I run it locally (`openspec validate --strict preserve-tool-search-replay-semantics` => valid at 28ed90f), so task 3.2 can simply be checked off. CI is fully green on the current head. Four items remain before this can land:

1. **Blocking — replayed tool-search call ids on the direct-WebSocket full-resend path (carried over from my previous comment, still open).** `responses_input_items_are_self_contained_fresh_replay` still rejects every nonempty `id` (app/modules/proxy/replay_safety.py:281), while canonical replayed `tool_search_call` items carry a server-assigned id. Your new `test_websocket_client_previous_response_full_resend_retry_allows_tool_search_history` uses id-less fixtures, so real replay histories still fail closed on that path and the nightly `No tool output found for tool search call` failure likely persists there. The projection path now strips ids correctly — please either give this path the same treatment (with an id-bearing regression fixture) or state explicitly that the WS full-resend path is out of scope for this PR.

2. **Blocking — the `first_output_index == 0` sanitization is dead code.** The new branches in `_trim_http_bridge_previous_response_input_items` (app/modules/proxy/_service/http_bridge/helpers.py:1136-1138) and `_trim_websocket_previous_response_input_items` (app/modules/proxy/_service/websocket/helpers.py:1987-1989) return a sanitized list of the same length, but both callers only install the result when the length changed (app/modules/proxy/_service/http_bridge/streaming.py:1793, app/modules/proxy/_service/websocket/mixin.py:3128), so an anchored continuation that already starts with an id-bearing `tool_search_output` forwards the response-owned id unchanged. Either compare content instead of length at the call sites, or apply the id stripping independently of prefix trimming — and cover it with a test that starts the input at the output item.

3. **Spec/impl mismatch on failed outputs.** `_tool_output_is_self_contained` (app/modules/proxy/replay_safety.py:698-704) admits `status: "failed"` tool-search pairs as portable, but your OpenSpec requirement scopes account-neutral portability to completed pairs. Pick one: require completed status for `tool_search_output` specifically, or extend the spec to cover failed pairs and add a fixture.

4. **Trivial:** check off tasks.md 3.2 — strict validation passes in CI-equivalent tooling, the local-blocked note is now stale.

The compact-trigger work (duplicate rejection, V1 pre-hoist dependency, single-terminal-trigger preservation) and the server-executed/nested-compaction rejections all verify correctly against main and are well covered. This is close.

Soju06:
Thanks for the 08-31 push — it closes items 1–3 from my previous comment: both trim helpers now strip `tool_search_call`/`tool_search_output` ids, both call sites compare content instead of length, and failed `tool_search_output` is rejected. `openspec validate --strict preserve-tool-search-replay-semantics` passes, the branch test-merges cleanly onto current main (18 merges since your base, incl. #1995), and the targeted suites pass on the merged tree. I also confirmed the defect on main: your new `test_replay_safety` cases, the trim-helper cases and `test_v1_proxy_compact_rejects_non_terminal_compaction_trigger_before_instruction_hoist` all fail on pristine `origin/main`.

One new blocker surfaced while tracing the WebSocket change, plus housekeeping:

1. **Blocking — the fresh replay body is now the prefix-trimmed list (`app/modules/proxy/_service/websocket/mixin.py:3154`).** `client_full_resend_retry_safe` is computed at :3142 against the *untrimmed* input, but `client_full_resend_input_items = trimmed_input_items` now builds `client_full_resend_payload` (:3168) from the *trimmed* list. For any anchored delta whose prefix is trimmable — this is not tool-search specific; `[function_call, function_call_output, user]` with an anchor this connection has not tracked — the prepared `fresh_upstream_request_text` becomes `[function_call_output, user]` with no `previous_response_id` while `fresh_upstream_request_is_retry_safe` stays `True`. That body is installed at mixin.py:5828 (`previous_response_not_found` retry, `require_account_neutral=False`, owner-bound), so the retry sends a dangling tool output and draws an upstream 400 ("No tool call found for … output"). Repro via `_prepare_websocket_response_create_request(previous_response_id="resp_untracked", input=[function_call(id), function_call_output, user], continuity_state=_WebSocketContinuityState())`: on main the fresh input is the full 3-item pair; on this branch it is the 2-item dangling output. Note the HTTP bridge deliberately does the opposite (`streaming.py:1996` builds the client full-resend fresh body from `untrimmed_effective_payload`). Please strip the tool-search ids on the *untrimmed* list for `client_full_resend_input_items` (e.g. expose a `_sanitize_websocket_replayed_tool_search_ids(items)` helper used by both the trim and this assignment) and keep prefix trimming only for the anchored upstream body, with a regression test covering a trimmable delta + untracked anchor that asserts the fresh body retains the call/output pair.

2. **Duplicate tests.** `test_proxy_compact_rejects_duplicate_compaction_trigger_before_upstream` and `test_proxy_compact_preserves_single_terminal_compaction_trigger` in `tests/integration/test_proxy_compact.py:355/390` are verbatim copies (same names, same account ids) of the tests already in `tests/integration/test_proxy_compact_triggers.py:16/126` from #1809 — the backend-route duplicate rejection was already on main via the raw `Depends` on `responses_compact`; both copies pass on pristine main. Please drop them. The genuinely new compact behavior here is the V1 pre-hoist `Depends` (api.py:6714), which is correct and now covered.

3. **Housekeeping (carried over):** check off `tasks.md` 3.2 and remove the "blocked locally" note — strict validation passes. The PR body also still says the change allows "compaction context items" through neutral replay (that exemption was removed in 6d42ccf; the nested-compaction test now asserts rejection) and that "no spec files are changed" — please refresh it.

4. **CI / threads.** `CI Required` is red on 0ad1cc3 (Playwright smoke flake; the CodeQL rust job fails only because the branch predates the #1995 crates), so a fresh push/rebase onto main is needed anyway. There are also 9 unresolved Codex/CodeRabbit threads; as far as I can trace, every one is addressed by the current code, but please leave a short reply on each so they can be resolved.

The replay-safety predicate changes (client-owned execution, `tools` XOR string `output`, failed-status and server-execution rejection, nested compaction) and the trim/id-sanitization on the anchored path verify correctly against main. Once (1) lands this should be ready.

Komzpa:
@codex review

Soju06:
Re-reviewed on head `65a73c8` (pushed 09-04 23:56 UTC). The 09-02 blocker is closed: `client_full_resend_input_items` is now built from `_sanitize_websocket_previous_response_input_items(previous_response_input_items)` on the untrimmed list (app/modules/proxy/_service/websocket/mixin.py:3141), prefix trimming stays on the anchored body only, and the trimmable-delta regression is covered. Duplicate `test_proxy_compact` copies are gone, tasks.md 3.2 is checked, PR body refreshed. Newest non-cancelled runs on this head are green (`CI Required` success at 00:20 UTC; the 00:05 integration-core failure was superseded by the rerun), Contributors attribution passes, and the branch test-merges cleanly onto current main including this round's merges (#1967 #1989 #2014 #2082 #2087 #2012 #2017 #2091 #2025 #2066 #2067 — no file conflicts). No unresolved review threads remain. Verdict: merge-ready; I intend to merge as-is, but one fail-open below is worth a quick follow-up (or a small push before merge if you prefer).

**P2 — discovered tool declarations inside `tool_search_output.tools` bypass the tool-declaration rules.** `_tool_output_is_self_contained` (app/modules/proxy/replay_safety.py:697-703) only checks `isinstance(item.get("tools"), list)`; the elements never go through `_tools_are_account_neutral` / `_contains_account_scoped_tool_state`, and the generic `_contains_account_scoped_input_state` walk (:1034) does not know `container`, `server_url`, `headers`, or the `mcp` / `code_interpreter` declaration types. Reproduced on this head using your own fixture shape (`[tool_search_call, tool_search_output, user]`):

- `tools=[{"type":"code_interpreter","container":"cntr_owner_a"}]` inside the output -> `responses_payload_is_account_neutral_fresh_replay` returns **True**; the identical declaration at top-level `tools` returns False.
- `tools=[{"type":"mcp","server_label":"x","server_url":"https://x","headers":{"Authorization":"Bearer s"}}]` -> **True** inside the output, False at top level.

On main the pair is omitted from the owner-unavailable projection, so these declarations never reached account B; this PR now forwards them in the projected replay, and on the WS path `_install_verified_fresh_replay` clears `replay_required_account_id` for such a body. It is bounded in practice — real Codex Lite traffic is still fail-closed upstream of this by the `additional_tools` bundle (`tool_search` / `namespace` types are not in `_ACCOUNT_NEUTRAL_TOOL_TYPES`, :24) and by server-assigned ids on the direct predicate — so it needs an id-less non-Lite payload, hence P2 rather than P1. But it is fail-open relative to main and to the spec delta's "account-neutral only when the pair is self-contained" requirement. Fix: run every element of `tools` through the existing declaration rules (`_tools_are_account_neutral`), and decide explicitly how `defer_loading` function tools should pass, since that field is not in `_ACCOUNT_NEUTRAL_TOOL_DECLARATION_FIELDS` (:25-30).

**P3 — the inverse asymmetry: JSON-schema property names are read as live references.** With the pair retained, `_mapping_has_account_scoped_reference` (:1063) sees `{"file_id": {"type": "string"}}` inside a discovered function's `parameters.properties` as a non-empty `file_id`. Reproduced: `tools=[{"type":"function","name":"read","parameters":{"type":"object","properties":{"file_id":{"type":"string"}}}}]` inside the output -> False on this head (main recovered it via the projection path since the pair was omitted), while the same tool at top-level `tools` -> True because `_contains_account_scoped_tool_state` skips `parameters`. Applying the declaration rules per the P2 fix resolves this too.

**P3 — headline benefit is unreachable for real Codex traffic today.** Responses-Lite requests carry `tool_search` / `namespace` declarations in `additional_tools`, which fail `_tools_are_account_neutral`; non-Lite `{"type":"tool_search"}` fails `_ACCOUNT_NEUTRAL_TOOL_TYPES`; `defer_loading` function tools fail the field allowlist. So the retention only changes outcomes for id-less / non-Lite payloads. Fine to land as groundwork, but the PR body and proposal.md should say so rather than implying the nightly `No tool output found for tool search call` failure is fixed end-to-end; extending the allowlists is a separate, spec-bearing change.

**P3 — dead code.** The `validate_compaction_trigger_shape` block in `_compact_responses` (app/modules/proxy/api.py:6821-6826) is unreachable for both callers: the backend route already rejects via `Depends(_capture_raw_compaction_trigger_error)` (:6741) and the V1 route passes `validate_compaction_trigger_shape=False` (:6795). Drop it.

**P3 — V1 trigger rule is narrower than its docstring.** `_capture_raw_v1_compaction_trigger_error` (api.py:758-777) rejects `[.., compaction_trigger, developer]` but accepts and normalises `[compaction_trigger, user, compaction_trigger]`, which also contains a hidden non-terminal trigger. Either state the actual rule ("reject when a trigger exists and the last item is not a trigger") or reject any non-terminal trigger. Note this is a client-visible V1 400 for a previously accepted shape, bundled into a replay PR — acceptable, but it should be named in the PR body.

**P3 — observability / transport asymmetry nits.** (a) When only tool-search ids are stripped and no prefix is trimmed, both call sites still emit `*_previous_response_input_trimmed original_items=N trimmed_to=N` (streaming.py:1867, mixin.py:3327) — gate the log on an actual length change or rename it. (b) The WS owner-bound client full-resend body is id-sanitised (mixin.py:3141) while the HTTP bridge builds the equivalent body from `untrimmed_effective_payload` with tool-search ids intact (streaming.py:2004), so the two transports send different owner-bound retry bodies for the same history. Harmless for the owner account, but worth aligning.

None of the P3s change the verdict. If you want to fold the P2 in before merge, ping me and I will re-verify on the new head; otherwise I will merge and track it as a follow-up.

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Soju06:
Re-reviewed on head `124fbe2b`. CI is green on this exact head, the branch test-merges cleanly onto current `main` (e987c56c8, after this round's merges — no conflicts introduced by #2078/#2079/#2092/#2114 or the dependabot batch), all four lint-job guards pass on the merged tree, and the targeted suites pass. The 09-05 P2 (discovered `tools` bypassing the declaration rules) and the schema-property P3 are both closed, and the outdated Codex P1 thread (string `output` + malformed `tools`) is refuted on this head. One new blocker from checking the fix against the actual upstream wire shape, plus carried-over items.

**Blocking — the `tools` allowlist matches a shape codex never emits, so the retained-pair path is unreachable and real pairs now fail closed where main proceeded.** `_tool_search_output_tools_are_account_neutral` (app/modules/proxy/replay_safety.py:849-856, `_ACCOUNT_NEUTRAL_TOOL_SEARCH_OUTPUT_TOOL_FIELDS = {"name"}` at :35) accepts only `{"name": "<str>"}` elements. Upstream `codex-rs/protocol/src/models.rs` defines `ResponseItem::ToolSearchOutput { id?, call_id, status: String, execution: String, tools: Vec<serde_json::Value>, internal_chat_message_metadata_passthrough? }` — `execution` and `tools` are required and there is no `output` field — and `codex-rs/core/src/tools/context.rs` fills `tools` by serializing `LoadableToolSpec` values, which are `{"type":"function","name":...,"description":...,"strict":false,"parameters":{...}[,"defer_loading":true]}` or `{"type":"namespace","name":...,"description":...,"tools":[...]}` (see `loadable_tool_spec_namespace_serializes_with_deferred_child_tools` in `codex-rs/tools/src/responses_api_tests.rs`). Feeding those exact JSON shapes into `responses_payload_is_account_neutral_fresh_replay` on this head with your `[user, tool_search_call, tool_search_output, user]` fixture returns `False` for both; only the synthetic `{"name": "workspace-search"}` fixture and `tools: []` return `True`. Because the pair was also removed from `_ACCOUNT_NEUTRAL_REPLAY_OMITTED_ITEM_TYPES`, a real history flips from account-neutral on `main` (pair omitted, replay proceeds) to fail-closed here. That is the safe direction, and reach is narrow (requests still declaring `tool_search`/`namespace`/`additional_tools` already fail neutrality), but the ADDED requirement's scenario "matching completed `tool_search_output` with a `tools` list → accepted as portable" and the PR body's "replacement account receives the tool-search result" are not true for any payload codex produces. Please either (a) run each element through the existing declaration rules — `_tool_declaration_is_account_neutral` for `type: function` (and decide explicitly how `strict` / `defer_loading` pass, since they are not in `_ACCOUNT_NEUTRAL_TOOL_DECLARATION_FIELDS`), plus a `type: namespace` branch that recurses into its `tools` — and add a regression fixture using the upstream JSON above; or (b) keep the current conservative allowlist but restore omission of the pair when the predicate fails (main's behavior) rather than failing the whole replay, and rewrite the spec scenario/proposal/PR body to say the retention is groundwork that real Codex traffic does not reach yet.

**Related — drop the string-`output` branch.** replay_safety.py:713-717 accepts a `tool_search_output` carrying string `output` and no `tools`. Upstream has no such field and requires `tools` + `execution`; accepting a shape that cannot originate from codex is fail-open by default. Require `tools` (list) and reject otherwise; the `"output"` entry in `_ACCOUNT_NEUTRAL_INPUT_ITEM_FIELDS["tool_search_output"]` can go with it.

**Carried over from 09-05, still unaddressed (non-blocking):** the `validate_compaction_trigger_shape` block in `_compact_responses` (app/modules/proxy/api.py:6823-6828) is unreachable for both callers (the V1 caller passes `False` at :6797, the other path validates earlier); `_capture_raw_v1_compaction_trigger_error` (api.py:758) still accepts `[compaction_trigger, user, compaction_trigger]` while its docstring says hidden non-terminal triggers are rejected — pick one and, since this is a client-visible V1 400 for a previously accepted shape, name it in the PR body; `*_previous_response_input_trimmed original_items=N trimmed_to=N` still logs when only ids were stripped (app/modules/proxy/_service/http_bridge/streaming.py:1867, app/modules/proxy/_service/websocket/mixin.py:3329); the HTTP bridge owner-bound full-resend body is built from `untrimmed_effective_payload` with tool-search ids intact (streaming.py:2004) while the WS path sanitizes them via `_sanitize_websocket_previous_response_input_items` (mixin.py:3141).

The prefix-trim/id-strip changes on both transports and the V1 pre-hoist trigger validation still verify correctly against main. If you go with option (a), a `Fixes #1530` (its unclassified "tool search call" missing-output variant) would be appropriate once the end-to-end path is actually reachable; otherwise leave it out.

Soju06:
Integrated current main at `21ef8531`. Compact-trigger shape validation now runs before the current-main compact source-pin guard, preserving both contracts instead of dropping either side of the conflict.

Validation: 61 compact/compact-trigger integration tests pass; full Ruff/typing, architecture/cancellation/timing/settings guards and strict OpenSpec pass. Awaiting this head's CI and review.

@codex review

Soju06:
Re-triaged at head `21ef8531` against `main` `ca243dec4`. The 09-08 blocker is still open at head (the only commit since is the `origin/main` merge), and the branch has gone conflicting again. Everything below was verified by reading the code on both sides, not the description.

**1. Blocking (unchanged from 09-08): the retained-pair path is unreachable, and real pairs now fail closed.** `_ACCOUNT_NEUTRAL_TOOL_SEARCH_OUTPUT_TOOL_FIELDS = frozenset({"name"})` (`app/modules/proxy/replay_safety.py:36`) and `_tool_search_output_tools_are_account_neutral` (`:881`) require `set(tool) <= {"name"}` for every element of `tool_search_output.tools`. Codex serialises `LoadableToolSpec` values — `{"type":"function","name":...,"description":...,"strict":false,"parameters":{...}}` and `{"type":"namespace","name":...,"tools":[...]}` — every one of which carries at least `type`, so `_tool_output_is_self_contained` (`:725`) returns `False`. Because this branch also removed `tool_search_call`/`tool_search_output` from `_ACCOUNT_NEUTRAL_REPLAY_OMITTED_ITEM_TYPES` (`:22`; on main they are still omitted at `replay_safety.py:20-21`, with `_project_account_neutral_replay_item:226-247` dropping completed pairs before the neutrality predicate at `:705` runs), the whole replay is refused where main projected the pair away and proceeded. Concretely: an owner-loss failover of a thread whose history is `[tool_search_call, tool_search_output(tools=[{"type":"function",...}]), user]` is refused on this branch and replayed on main. So the production delta is fewer cross-account replays, with no compensating gain, because the "replacement account receives the tool-search result" outcome asserted by the spec delta and the PR body is not reachable for any payload Codex emits. Please pick one of the two options from the 09-08 review: (a) run each `tools` element through the declaration rules (`_tool_declaration_is_account_neutral` for `type: function`, an explicit decision on `strict` / `defer_loading`, and a `type: namespace` branch that recurses into the nested `tools`), with a regression fixture built from the upstream JSON shape; or (b) keep the conservative allowlist but restore main's omission when the predicate fails, and rewrite the spec scenario, `proposal.md` and the PR body to say the retention is groundwork.

**2. Drop the string-`output` branch.** `_tool_output_is_self_contained` (`replay_safety.py:739`) still accepts a `tool_search_output` carrying a string `output` and no `tools`. Upstream requires `tools` + `execution` and has no `output` field, so this admits a shape that cannot originate from Codex; remove it together with the `"output"` entry in `_ACCOUNT_NEUTRAL_INPUT_ITEM_FIELDS["tool_search_output"]` (`:102-114`).

**3. Dead code (carried over).** The `validate_compaction_trigger_shape` block in `_compact_responses` is unreachable: the backend route already rejects via `Depends(_capture_raw_compaction_trigger_error)` and the V1 route passes `False`.

**4. Conflicts — one is substantive.** `git merge-tree origin/main <head>` conflicts in `app/modules/proxy/api.py`, `tests/unit/test_proxy_utils.py` and 8 `openspec/specs/*/spec.md` files. The `api.py` conflict is not mechanical: the `compact_pin_denial` call added by the 09-10 merge no longer exists on main (removed by #2416), so that part of the resolution must be dropped rather than re-applied. The spec conflicts are self-inflicted — this branch deletes the trailing newline of 54 `openspec/specs/*/spec.md` files (54 files, one deletion each) with no relation to the change; please revert that churn so the branch stops colliding with every spec-touching PR.

What does verify correctly against main and should survive: the V1 pre-hoist trigger validation closes a real gap (`v1_responses_compact` on main, `api.py:6981`, has no raw-trigger dependency), and the prefix-trim / id-sanitisation work on both transports is sound. Note also that #2099 rewrites `_ACCOUNT_NEUTRAL_INPUT_ITEM_FIELDS` and `responses_input_items_are_self_contained_fresh_replay` in this same file, so whichever lands second needs a semantic rebase, not just a textual one.

PR #2031, opened 2026-09-02
fix(proxy): normalize SDK transcription headers

kostazol:
## Problem

OpenChamber uses the OpenAI Node SDK for `/v1/audio/transcriptions`. The subscription-backed transcription path forwarded its `OpenAI/JS` user agent and `x-openai-client-*` headers to ChatGPT `/transcribe`, triggering an upstream Cloudflare/WAF challenge. Its minimal allowlist did not forward `x-stainless-*` headers. Equivalent curl multipart requests work because they do not carry the forwarded SDK identity.

## Solution

- Normalize non-native transcription requests to canonical `codex_cli_rs` `User-Agent`, `originator`, and `version` values.
- Remove forwarded OpenAI SDK fingerprint headers before subscription-backed `/transcribe` forwarding.
- Preserve native Codex behavior, account authorization/account ID, multipart forwarding, failover, timeouts, accounting, and model-source routing.

## Validation

- Added fake-upstream regression coverage for an OpenAI Node SDK-shaped `/v1/audio/transcriptions` request, including multipart bytes, selected account headers, canonical fingerprint, and `{ "text": ... }` response.
- Added native Codex transcription fingerprint coverage.
- `uv run pytest tests/unit/test_proxy_upstream_fingerprint.py tests/integration/test_proxy_transcriptions.py -q` (`53 passed` after rebase).
- Ruff format/check and `ty check` passed.
- `openspec validate fix-transcription-sdk-fingerprint --strict` passed.

## OpenSpec

`openspec/changes/fix-transcription-sdk-fingerprint/`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved audio transcription compatibility for non-native SDK clients.
  * Non-native requests now use the canonical Codex CLI identity when forwarded upstream.
  * OpenAI SDK-specific fingerprint headers are no longer forwarded for these requests.
  * Native Codex requests retain their existing fingerprint behavior.
  * Cancelled transcription requests now correctly release reserved API key usage.

* **Tests**
  * Added coverage for fingerprint normalization, native requests, multipart uploads, upstream header handling, and cancellation cleanup.


Soju06:
Thanks - I reproduced the defect on main and the fix is the right shape. Confirmed with your tests copied onto pristine `main` (665e58e31): `test_non_native_transcription_request_is_rewritten_to_codex_cli_fingerprint` and `test_v1_audio_transcriptions_normalizes_openai_sdk_fingerprint_upstream` both fail with `'OpenAI/JS 4.104.0'` forwarded verbatim from `app/core/clients/proxy.py:1014-1015`; on your head all 51 tests in the two touched files pass, `openspec validate --strict` passes, and ruff is clean. Reusing `_is_native_codex_request` / `_normalize_non_native_upstream_fingerprint` (proxy.py:1018-1019) instead of adding a transcribe-specific normalizer is what we want.

Two items before this can merge:

1. **Attribution (CI-blocking).** You are not in `.all-contributorsrc`, and `.github/scripts/check_all_contributors.py` includes the PR author login in the expected set, so the required `Contributors attribution` job will fail once fork CI runs. Please append an entry (login `kostazol`, contributions `["code", "test"]`) following the existing format at the end of the `contributors` array.

2. **Description accuracy (openspec/proposal.md:3 and PR body).** On main the transcribe builder only forwards `User-Agent` plus `x-openai-*` / `x-codex-*` (`_TRANSCRIBE_FORWARD_HEADER_PREFIXES`, proxy.py:997), so `x-stainless-*` never reached `/transcribe`; what leaked was the `OpenAI/JS` UA and `x-openai-client-*` keys. Please reword "preserves that SDK fingerprint ... x-stainless-*" so the proposal describes the actual pre-fix behavior. The spec's "MUST NOT forward x-stainless-*" requirement is fine to keep as a normative invariant.

One non-blocking question: `_normalize_non_native_upstream_fingerprint` also injects `version` (proxy.py:879). Codex CLI's shared client sends `originator` and `User-Agent` as default headers on every request (`codex-rs/login/src/auth/default_client.rs` `default_headers()`), so `originator` matches native `/transcribe` traffic, but `version` is set per-request on the Responses path and is not part of the `voice.rs` fingerprint that #199 matched. If you have a way to exercise a real OpenChamber -> `/v1/audio/transcriptions` -> ChatGPT round trip with this branch, a one-line confirmation in the PR body that the WAF challenge is gone with the extra `version` header present would close that gap; otherwise we can accept it on the basis that bare curl (no Codex headers at all) succeeds.

FYI #1966 (merged today) appended a test at the same location in `tests/integration/test_proxy_transcriptions.py`, so please rebase onto current `main`; expect a trivial conflict there.

kostazol:
Thanks — addressed both merge blockers and rebased onto current `main` (`575c2087`).

- Added `kostazol` to `.all-contributorsrc` with `code` and `test`; local contributor check passes.
- Corrected proposal and PR description: pre-fix `/transcribe` forwarded `OpenAI/JS` and `x-openai-client-*`; `x-stainless-*` was not forwarded by its allowlist.

Kept canonical `version`: this change intentionally reuses the established non-native fingerprint normalizer, and the compatibility requirement requires its canonical value. No live OpenChamber-to-ChatGPT round trip was run: regression coverage uses a fake upstream and no real credentials are used.

Validation after rebase: `53 passed` for touched suites, Ruff/ty clean, strict OpenSpec valid.

Soju06:
Re-verified on current `main` (dd28d7dff): the two new tests still fail on pristine main — `test_non_native_transcription_request_is_rewritten_to_codex_cli_fingerprint` (unit) and `test_v1_audio_transcriptions_normalizes_openai_sdk_fingerprint_upstream` (integration) — because `_build_upstream_transcribe_headers` (`app/core/clients/proxy.py:1005-1023`) still forwards the inbound SDK `User-Agent`/`x-stainless-*` headers verbatim without the `_normalize_non_native_upstream_fingerprint` step that the Responses path already applies. On a local test-merge of this branch with main, the touched suites pass (`tests/unit/test_proxy_upstream_fingerprint.py` + `tests/integration/test_proxy_transcriptions.py`, 53 passed; one isolation-order error in `test_transcription_model_restriction_uses_fixed_model` reproduces identically on pristine main and passes standalone, so it is not yours). The authoritative CI run on `e285b69` is green (run 33649016667); the other run on the same SHA (33649164594) is a cancelled concurrency duplicate whose "failure" jobs are just the aggregation of cancelled shards. Both items from the 09-02 review are addressed.

The only thing standing between this and a merge is the `needs rebase` label: `.all-contributorsrc` now conflicts because main appended another entry (`hongzexin`) at the same tail position. `app/core/clients/proxy.py` and the openspec files merge cleanly, and none of the PRs merged in this round touch your files. Please rebase onto current `main` keeping both entries — no code changes expected. If you would rather we resolve that one-entry conflict ourselves (`maintainerCanModify` is on), say so and we will push it to your branch.

PR #2055, opened 2026-09-03
fix(usage): deactivate accounts only on explicit terminal signals, not bare 404/402

Soju06:
## Summary

Background usage refresh (`app/modules/usage/updater.py`) treated a bare HTTP `402`/`404` from `/backend-api/wham/usage` as proof that the account no longer exists and moved it to `deactivated` (`_DEACTIVATING_USAGE_STATUS_CODES = {402, 404}`). A bare status code is not an account-specific signal, and during an upstream outage it is fleet-wide.

This PR removes the status-code branch. An account now changes status during usage refresh **only** on an explicit terminal signal:

- an error `code` in `PERMANENT_FAILURE_CODES` (mapped through `account_status_for_permanent_failure`, unchanged), or
- an error message that explicitly says the account is deactivated (`_DEACTIVATING_USAGE_MESSAGE_HINTS`, unchanged).

Everything else, including bare `402`/`404`, is now a plain refresh failure: logged, counted, status and routing availability untouched, retried on the next cycle (the same path `403` already took).

## Incident

2026-09-03 14:43–15:23 UTC: OpenAI returned HTTP 404 for every `chatgpt.com/backend-api/*` path for ~40 minutes (status.openai.com "Elevated errors across ChatGPT and Codex"). The 60s refresh loop deactivated all 17 active accounts within 3 minutes (`deactivation_reason = "Usage API error: HTTP 404 - Usage fetch failed (404)"`), the proxy entered `no_accounts` 503 degraded mode, and because `refresh_accounts()` skips `DEACTIVATED` accounts nothing recovered when upstream came back; every account had to be reactivated by hand. Two further accounts had been deactivated with the same reason on 09-01 and 09-02 and were healthy once reactivated.

## Changes

- `app/modules/usage/updater.py`: drop `_DEACTIVATING_USAGE_STATUS_CODES` and the status-only branch in `_should_deactivate_for_usage_error`. First-attempt and post-auth-refresh retry both go through the same classifier, so both reflect the new rule.
- `tests/unit/test_usage_updater.py`: invert the previous bare-402 ⇒ deactivated expectation and parameterize it with bare-404; assert status unchanged, no status write, no routing-unavailable mark, failure logged, later cycle retries; add a post-auth-refresh-retry 404 case; keep permanent-code mapping and explicit-message coverage.
[...]

PR #2057, opened 2026-09-03
fix(usage): ignore ambiguous 404 refresh errors

### LINKED ISSUE #2056: fix: prevent mass deactivation on ambiguous usage 404
## Summary

On 2026-09-03, the background usage refresh classified a cohort-wide upstream HTTP 404 as permanent account deactivation. Eight active Plus accounts were marked `deactivated` within roughly two minutes with the identical reason `Usage API error: HTTP 404 - None`.

The stored access/refresh/id-token ciphertexts were unchanged, so this was a status-classification failure rather than credential loss or actual account deletion. A bare 404 can represent an unavailable usage endpoint, route, or upstream rollout and must not remove otherwise healthy accounts from the pool.

## Expected behavior

- A bare or ambiguous usage HTTP 404 keeps the account active and eligible for later refresh/proxy selection.
- Explicit account-deactivation codes/messages remain permanent.
- Credential/session invalidation remains `reauth_required`.
- Existing payment-required behavior remains unchanged.

## Proposed fix

The usage updater should only deactivate on an unambiguous permanent signal. Add regression coverage for the bare 404 shape and document the policy in OpenSpec.

## Evidence

- Incident deactivation reason: `Usage API error: HTTP 404 - None`
- Affected accounts: 8
- Token ciphertext changes for affected rows: 0
- Reproduction before the fix: mass deactivation detected
- Regression after the fix: bare 404 classification returns non-permanent; explicit deactivation and 402 remain permanent

msmahdinejad:
## Summary

The usage refresh worker treated every HTTP 404 as proof that an account was permanently deactivated. During the incident, an upstream bare 404 affected a cohort of healthy Plus accounts and removed them from the pool even though their stored token material was unchanged.

This PR makes the deactivation decision evidence-based: a bare/ambiguous 404 is a transient refresh failure, while explicit deactivation signals and the existing payment-required behavior remain permanent. Session invalidation codes continue to map to `reauth_required`.

## Type of change

- [ ] **Breaking change**

Linked issue: Fixes #2056

## OpenSpec

Change directory: `openspec/changes/ignore-ambiguous-usage-404/`

## Changes

- Remove bare HTTP 404 from the permanent usage-deactivation status set.
- Preserve explicit deactivation code/message handling and HTTP 402 handling.
- Add a regression test reproducing the incident error shape.
- Document the conservative status policy in OpenSpec.

## Test plan

- `uv run pytest tests/unit/test_usage_updater.py -q` — 121 passed
- `uv run ruff check app/modules/usage/updater.py tests/unit/test_usage_updater.py` — passed
- Candidate Docker image self-test — passed
- Local runtime verification — `/health/ready` 200; live `POST /backend-api/codex/responses` 200; bare 404 remains non-deactivating

OpenSpec CLI is not installed in the local environment; the strict validation command was attempted via the pinned `npx` package but could not complete in this environment.

## Checklist

- [x] Ran the relevant backend validation locally.
- [x] If touching specs: added the OpenSpec change.
- [x] Simplicity gates reviewed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Ambiguous HTTP 404 responses during usage refresh no longer deactivate accounts.
  * Accounts remain active unless a permanent deactivation or payment-required signal is received.
  * Credential or session invalidation signals continue to require re-authentication and remove accounts from proxy selection.

* **Documentation**
  * Updated contributor acknowledgments in the project documentation.


Soju06:
Thanks for the incident report in #2056 and for the regression test — the defect is real. Your test fails against current main, where `app/modules/usage/updater.py:1357` still has `_DEACTIVATING_USAGE_STATUS_CODES = {402, 404}` and `_should_deactivate_for_usage_error` (line 1364) returns `True` on the status-code match before any other classification.

Closing this one as superseded by the maintainer PR #2055, which carries the same fix in a broader form and is the canonical vehicle for this bug:

- `_should_deactivate_for_usage_error` drops the status-code branch entirely, so bare 404 **and** bare 402 are both treated as transient refresh failures. A bare status code is not account-specific and is fleet-wide during an upstream outage — keeping 402 as a deactivation trigger is where this PR diverges from that design decision.
- Tests cover bare 402/404 on the first attempt (`test_usage_updater_keeps_account_active_on_bare_402_or_404`) and bare 404 on the post-auth-refresh retry path (`test_usage_updater_retry_keeps_account_active_on_bare_404`), including the assertion that no routing-unavailable mark is written. Your `UsageFetchError(404, "None")` shape is covered: after the change the classifier only inspects `exc.code` against `PERMANENT_FAILURE_CODES` and the explicit deactivation message hints, so the literal message text does not matter.
- OpenSpec: #2055 carries a MODIFIED delta for the same `usage-refresh-policy` › "Usage refresh deactivates on clear deactivation signals" requirement (`openspec/changes/usage-refresh-explicit-deactivation-only/`).

This branch is also currently CONFLICTING (`.all-contributorsrc`, after the #1877 / #1948 / #1996 / #2072 merges touched it). We will link #2056 from #2055 so your issue is closed by the canonical fix. If you find a scenario #2055 does not cover once it merges, please open a follow-up issue against it.

Issue #2058, opened 2026-09-03
bug: transient 404 from /wham/usage during an OpenAI incident permanently deactivates every account (pool never self-heals)


kvz:
### Pre-flight checklist

- [x] I searched existing issues and confirmed this is not a duplicate (#1092 is pool exhaustion by rate limits; #353 is the 401 case; nothing covers transient 404 → permanent deactivation).
- [x] I am running a supported codex-lb version.
- [x] I can reproduce the problem with the steps below.

### codex-lb version

main @ `eeab46a5` (Docker, built from source). The relevant code is unchanged on current main (`575c2087`).

### Deployment method

Docker, image built from source

### Client used against codex-lb

Codex CLI 0.153.0

### ChatGPT account plan(s) involved

Pro (5 accounts)

### Model(s) involved

gpt-5.6-sol

### What happened?

During today's OpenAI incident (status.openai.com: "Elevated errors across ChatGPT and Codex", Codex API degraded, 2026-09-03 ~15:00–16:00 UTC), `GET https://chatgpt.com/backend-api/wham/usage` transiently returned **404 with an empty body**. The usage refresher treated that as a permanent client error and deactivated **all five accounts** within one refresh cycle:

```
accounts.status = 'deactivated'
accounts.deactivation_reason = 'Usage API error: HTTP 404 - None'
```

From then on every request failed with `503 No available accounts. Service is operating in degraded mode: all upstream accounts are unavailable` — and kept failing long after OpenAI recovered, because `deactivated` never rejoins the pool (unlike `rate_limited`, it does not self-heal).

Timeline (UTC, from `request_logs`):
- 14:50:47–14:51:40 — five proxied requests get `Upstream error: HTTP 404 Not Found` (early tremors of the incident)
- 15:18:39 — last successful proxied request
- shortly after — usage refresh 404s → all accounts `deactivated`
- ~16:25 — probing the same endpoint from inside the container with the same (decrypted) access tokens + `chatgpt-account-id` header: **200**, `plan_type: "pro"`, tokens valid for weeks. `codex/models` also 200. Upstream was fine again; only codex-lb's state was stuck.
- Manual `UPDATE accounts SET status='active', deactivation_reason=NULL WHERE status='deactivated'` restored service immediately; the statuses held through subsequent refresh cycles.

### Where

`app/modules/usage/updater.py`:

```python
_DEACTIVATING_USAGE_STATUS_CODES = {402, 404}
...
[...]

PR #2062, opened 2026-09-03
fix(accounts): keep ambiguous usage 404s retryable

### LINKED ISSUE #2061: Usage refresh should not deactivate accounts on ambiguous HTTP 404
## Problem\n\nA bare HTTP 404 from the upstream usage endpoint is currently treated as a permanent account deactivation. The account is then excluded from routing and future usage refreshes, so it cannot automatically recover when its usage window resets or when the endpoint becomes available again.\n\nA 404 without a permanent error code or an explicit deactivation/deletion message is ambiguous: it may be endpoint-specific or transient rather than proof that the account itself is gone.\n\n## Expected behavior\n\n- Keep ambiguous usage 404 responses retryable and apply the existing refresh-failure cooldown.\n- Continue deactivating when the upstream error code or message explicitly identifies account deactivation, suspension, or deletion.\n- Preserve the existing handling for 402 and permanent credential/session codes.\n\n## Regression coverage\n\nAdd an updater test proving a bare 404 leaves the account active and suppresses repeated refreshes during the cooldown, plus coverage that explicit permanent signals still deactivate.

MIlhamDedi:
## Summary

Keep an account active when the upstream usage endpoint returns an ambiguous
HTTP 404, allowing normal usage polling and quota recovery to resume after the
existing cooldown. Explicit permanent account codes and deactivation messages
still produce a terminal account state.

## Type of change

- [ ] **Breaking change**

Linked issue: Fixes #2061

## OpenSpec

- [ ] This PR touches a codex-faithful path and preserves upstream-equivalent behavior

Change directory: `openspec/changes/keep-ambiguous-usage-404-retryable/`

## Changes

- Remove bare HTTP 404 from status-only permanent usage failures.
- Put ambiguous usage 404 responses into the existing process-local refresh
  cooldown so the scheduler retries without hammering upstream.
- Preserve terminal handling for HTTP 402, known permanent error codes, and
  explicit account-deactivation messages.
- Add regression coverage for both ambiguous and explicit-permanent 404 paths.

## Simplicity

- [x] No new setting or required setup step.
- [x] No README, `.env.example`, dashboard navigation, dependency, or schema change.

## Test plan

```text
uv run pytest tests/unit/test_usage_updater.py -q
122 passed

make lint
proxy architecture checks passed
All checks passed
1003 files already formatted

uv run ty check
All checks passed

openspec validate keep-ambiguous-usage-404-retryable --type change --strict
Change 'keep-ambiguous-usage-404-retryable' is valid

openspec validate usage-refresh-policy --type spec --strict
Specification 'usage-refresh-policy' is valid
```

Repository-wide `openspec validate --specs` still reports the pre-existing
`model-source-routing` specification failure on current `main`; this PR does
not modify that capability.

## Screenshots / output

No dashboard rendering changes.

```text
HTTP 404, no permanent code/message -> account remains active; refresh cooldown applies
HTTP 404, code=account_deactivated -> account becomes deactivated
```

## Checklist

- [x] Linked the related issue.
- [x] Ran the relevant local test, lint, type, and OpenSpec checks.
- [x] Simplicity gates reviewed.

## Summary by CodeRabbit

- **Bug Fixes**
  - Ambiguous usage-service `404` responses no longer permanently deactivate accounts.
  - These failures now enter a temporary cooldown, preventing repeated requests while allowing usage polling to resume later.
  - Explicit deactivation signals and permanent error codes continue to be handled as account deactivation.
  - Accounts remain eligible for routing when a `404` does not include a recognized permanent failure reason.

- **Documentation**
  - Added usage-refresh policy documentation covering temporary failures, cooldown behavior, recovery, and permanent deactivation handling.

### DIFF
--- a/app/modules/usage/updater.py
+++ b/app/modules/usage/updater.py
@@ -1351,10 +1351,11 @@ def _reset_at(reset_at: int | None, reset_after_seconds: int | None, now_epoch: 
     return now_epoch + max(0, int(reset_after_seconds))
 
 
-# The usage endpoint can return 403 for accounts that are still otherwise usable
-# for proxy traffic, so treat it as a refresh failure instead of a permanent
-# account-level deactivation signal.
-_DEACTIVATING_USAGE_STATUS_CODES = {402, 404}
+# The usage endpoint can return 403 or 404 for accounts that are still otherwise
+# usable for proxy traffic, so treat those bare statuses as refresh failures
+# instead of permanent account-level deactivation signals. Explicit permanent
+# error codes and deactivation messages are still handled below.
+_DEACTIVATING_USAGE_STATUS_CODES = {402}
 _DEACTIVATING_USAGE_MESSAGE_HINTS = (
     "your openai account has been deactivated",
     "account has been deactivated",
@@ -1381,7 +1382,7 @@ async def _resolve_upstream_route_for_account(account: Account, *, operation: st
 
 
 def _mark_usage_refresh_auth_cooldown(account_id: str, status_code: int) -> None:
-    if status_code not in {401, 403}:
+    if status_code not in {401, 403, 404}:
         return
     cooldown_seconds = max(0.0, float(get_settings().usage_refresh_auth_failure_cooldown_seconds))
     if cooldown_seconds <= 0:
--- a/openspec/changes/keep-ambiguous-usage-404-retryable/context.md
+++ b/openspec/changes/keep-ambiguous-usage-404-retryable/context.md
@@ -1,0 +1,31 @@
+# Ambiguous usage 404 handling
+
+## Purpose
+
+Usage refresh should remove an account from service only when upstream sends
+account-level evidence that the account is permanently unavailable. A status
+code that can also describe the usage route itself is not enough evidence for
+a terminal account transition.
+
+## Decision
+
+A bare usage `404` follows the same temporary cooldown path as ambiguous usage
+authentication failures. The existing permanent-code and message classifiers
+still run first, so `account_deactivated`, `account_suspended`,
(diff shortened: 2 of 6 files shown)


Soju06:
Thanks for the PR and for the regression test. I confirmed the defect on current main (`app/modules/usage/updater.py:1357` still lists 404 in `_DEACTIVATING_USAGE_STATUS_CODES`, and `test_usage_updater_cools_down_ambiguous_404_without_deactivation` fails against main with the account flipped to `deactivated`). The fix itself is correct and the OpenSpec change validates.

However, this is superseded by the owner's #2055, which was opened a few hours earlier for the same defect and is the canonical fix. Two points where this PR diverges from #2055 are product decisions the owner has already made there:

1. `app/modules/usage/updater.py:1358` — this PR keeps `{402}` as a status-only deactivation trigger. #2055 removes the status-code set entirely, and its `usage-refresh-policy` spec delta states that a bare HTTP status, including 402 or 404, MUST NOT change account status. The spec delta in `openspec/changes/keep-ambiguous-usage-404-retryable/specs/usage-refresh-policy/spec.md` would conflict with that requirement once #2055 lands.
2. `app/modules/usage/updater.py:1385` — adding 404 to the auth-failure cooldown set is a reasonable idea, but #2055 deliberately leaves a bare 404 as a plain refresh failure retried on the next (already staggered) scheduler cycle. If you think the cooldown is needed for the 404-storm case (#2058), please raise it as a comment on #2055 so it can be decided in one place.

Closing this as superseded by #2055; the related incident reports (#2056, #2058) are covered there. If you plan to contribute further, please also add yourself to `.all-contributorsrc` in your next PR — the attribution check is the only red CI job on this head.

PR #1700, opened 2026-08-12
fix(accounts): recover Free accounts after reset

HulianBuligon:
## Summary

Recover Free accounts that remain `rate_limited` behind a stale future cooldown after persisted monthly history proves that the blocked quota window reset. Recovery now completes atomically before ordinary active-only warm-up, preserving generic 429 and Retry-After protection.

## Type of change

- [x] `fix:` — bug fix (no behavior change beyond the bug)
- [ ] **Breaking change**

Linked issue: Related to #1261 and #1267

## OpenSpec

- [x] This PR includes / updates an OpenSpec change
- [ ] This PR touches a codex-faithful path

Change directory: `openspec/changes/recover-rate-limited-on-confirmed-reset/`

## Changes

- Bind early recovery to a post-block Free monthly baseline matching the persisted cooldown marker, then scan only adjacent history pairs for a confirmed reset.
- Compare-and-set status, reason, `reset_at`, and `blocked_at` before clearing markers and making the account active.
- Keep candidate evaluation and sender preflight active-only, while reusing the confirmed monthly tuple for durable warm-up deduplication.
- Trigger reset warm-up after a real selected-window reset regardless of prior usage, while retaining jitter and availability gates.

## Simplicity

- [x] Works with zero configuration
- [x] Adds no required setup step, setting, README section, environment variable, or dashboard navigation
- [x] Simplicity budgets are unchanged

## Test plan

- `uv run pytest -q tests/unit/test_limit_warmup.py tests/unit/test_usage_refresh_scheduler_recovery.py tests/integration/test_usage_refresh_scheduler_scope.py tests/integration/test_usage_repository.py`
  - 109 passed, 7 PostgreSQL-only skips
- `uv run ruff format --check .`
- `uv run ruff check .`
- `uv run ty check`
- `openspec validate recover-rate-limited-on-confirmed-reset --strict`
- `openspec validate --specs`
  - 49 passed, 0 failed
- CodeRabbit CLI review
  - 0 issues
- Independent review
  - one post-block baseline finding fixed; re-review clean

## Screenshots / output

No dashboard-visible change.

Production-scale validation against an existing SQLite history with sliding monthly reset deadlines:

- four stale Free accounts transitioned from `rate_limited` to `active`
- `reset_at` and `blocked_at` were cleared
- all six Free accounts ended active
- three genuinely exhausted Plus accounts remained `rate_limited`
- no duplicate warm-up attempt was created after deployment
- no usage-refresh scheduler error signature was logged
[...]

PR #2070, opened 2026-09-04
fix(accounts): recover paid accounts after long-window reset

Komzpa:
Paid accounts can remain `rate_limited` after a weekly reset even when fresh usage is available. Extend the existing reset-confirmed recovery to each plan's applicable long window and require matching persisted reset evidence before clearing the block. Primary exhaustion, generic Retry-After cooldowns, unknown plans and concurrent credential/status changes remain protected.

Recovery uses one account update guarded by the latest primary, secondary and monthly usage IDs. A stale recovery cannot undo an operator's reactivation through a later compensating write. Select the newest qualifying adjacent reset pair, with exact microsecond boundary comparisons on SQLite, so older or false candidates cannot hide a real reset.

The scheduler reuses the resolved reset for recovery and deduplicated warm-up. A reset-leading usage index includes PostgreSQL invalid-index repair; forward-only merge revisions preserve the current migration graph. This extends Soju06/codex-lb#1700 and complements Soju06/codex-lb#2048.

OpenSpec: `openspec/changes/archive/2026-09-09-recover-rate-limited-on-confirmed-reset/` and the updated account-routing and usage-refresh specs.

Validation at `06e18b988822c70e5eeaeb66a31628906a955b2d`, rebased onto `e040053b0b9ec07dd46805660c18e6ffa1bc6876`:

- 83 recovery, account/usage repository and populated-parent migration tests pass; 11 PostgreSQL-only cases are excluded from that SQLite run.
- Exact-reset cases cover −1, 0 and +1 microsecond; a sample at the new reset boundary remains invalid. Newer usage and operator reactivation controls pass.
- The current Alembic graph has one head, with successful SQLite upgrade and no schema drift. Earlier PostgreSQL index-repair, guarded-update and reset-query checks pass.
- 89 load-balancer refresh tests pass, with 3 existing skipped cases. Repository lint and type checks pass. Strict validation of the two main specs reports four identical requirement-keyword errors on both the base and this branch; the migration contract validates.

[Hosted checks](https://github.com/Soju06/codex-lb/pull/2070/checks) track the published head.


Komzpa:
@codex review

Komzpa:
@codex review

Soju06:
Thanks — I traced the premise on main and it holds: `_confirmed_free_monthly_reset_recovery` hard-gates on Free and `background_recovery_state_from_account` only recovers once `reset_at <= now`, so a paid account 429'd on its weekly window sits behind upstream's `resets_at` even after usage history proves an early reset. The generalization, the primary-window gate, the plan-type CAS, and the scheduler-level Team test all look right, and CI is green on `aa35aa6`. The branch also still merges cleanly against current main after this round's merges. A few things before this can merge:

1. **Baseline ordering in the in-cycle merge** — `app/core/usage/refresh_scheduler.py:570-577`. When `usage_reset_confirmed(before, after)` already holds for the current pair and `reset_transition_candidate` returns a marker row, the code replaces `baseline` with `history[0]` without checking that it precedes `current.before`. Your own spec text (both `specs/*/spec.md` deltas) requires the adjacent pair to be "at or after that baseline", and this branch also drops the `before.recorded_at > blocked_at` property that main's `baseline=before` gave for free. It only bites when the persisted marker matches the *new* window's deadline, but the direction is fail-open (clears a cooldown), so please guard it: merge only when `(history[0].recorded_at, history[0].id) <= (current.before.recorded_at, current.before.id)`, otherwise fall through to `_latest_confirmed_reset_transition_after_baseline`. This is the still-open Codex thread at L576.

2. **Migration coverage** — `app/db/alembic/versions/20260904_000000_add_usage_reset_transition_index.py`. Every index migration in the repo has an upgrade/downgrade test (`tests/integration/test_migrations.py:1202`, plus the PostgreSQL invalid-leftover repair case at `:1246`, and `:1797` for the parent revision `20260830_000000`). Please add the same pair for this revision so the `_drop_invalid_postgres_index` path is exercised; AGENTS.md treats that as expected coverage for new revisions.

3. **Open review threads** — five non-outdated Codex threads are unresolved on the current head (`refresh_scheduler.py:432` and `:576`, `repository.py:1010` and `:1038`, migration `:41`). The perf (L1010) and TOCTOU (L432) concerns look adequately addressed by the new `reset_at`-leading index and the `_recovery_usage_watermark_is_current` re-read; a short reply and resolve on each would clear them. Per CONTRIBUTING, unresolved actionable threads block readiness.

4. **Optional (P3)** — `app/modules/usage/repository.py:1022-1041`. `after_stmt` returns only the *first* row with `reset_at >= baseline + 60`. If that pair fails `usage_reset_confirmed`'s temporal predicate, the caller gets `None` and — since the same row is returned every cycle — later valid pairs are never examined. Main's `history_since` full scan did not have this ceiling for Free. Either iterate the first N candidates or order candidates so a predicate-passing pair is preferred. Not blocking, but worth a note in `design.md` if you keep the single-candidate shape.

5. **Overlap with #2048** — that PR adds `resolve_capacity_plan_type` in `app/core/balancer/logic.py` and edits the same `_select_long_window_entry` body and import line in `refresh_scheduler.py` that you change with `normalize_capacity_plan_type`. Whichever lands second will need a rebase; converging on one helper would be ideal.

Komzpa:
@codex review

Komzpa:
@codex review

Soju06:
Re-checked on head `0b02d4b` against current `main` (`a0d94a46a`). The R19 points on baseline ordering (`_usage_history_at_or_before` fall-through), migration coverage, and the single-candidate scan are addressed, the guards and openspec pass on the stacked tree, and the scheduler-level Team test is the right shape. Three things still block:

1. **Stale against #2103 — unit tests fail on the merged tree.** `4b951e590` (09-07) removed `import time` from `app/modules/proxy/load_balancer.py` (it now reads the clock through `REAL_CLOCK`/`utcnow`) and rewrote this same test module to patch `app.modules.proxy.load_balancer.utcnow`. Your branch still has 27 `monkeypatch.setattr("app.modules.proxy.load_balancer.time.time", lambda: now)` sites in `tests/unit/test_usage_refresh_scheduler_recovery.py` (first at line 504, last at 2051). `git merge-tree` is textually clean, so GitHub shows MERGEABLE and the 09-06 CI run predates the change, but on the merged tree those tests (`test_reconcile_recovers_team_after_confirmed_weekly_reset_before_legacy_deadline[*]`, `test_reconcile_skips_recovery_when_latest_usage_changes_before_cas`, `test_reconcile_reblocks_*`, `test_reconcile_preserves_replacement_credentials_when_rollback_loses_cas`, `test_reconcile_keeps_team_blocked_when_primary_window_is_exhausted`, ...) raise `ImportError: 'app.modules.proxy.load_balancer' is not a package`. Replacing those sites with `monkeypatch.setattr("app.modules.proxy.load_balancer.utcnow", lambda: _epoch_to_naive_utc(now))` (the helper already exists at line 64) makes the module pass locally. Please rebase onto current main and re-run.

2. **OpenSpec artifacts contradict the migration** (the open Codex P1 on the revision file). `openspec/changes/recover-rate-limited-on-confirmed-reset/proposal.md:28` says "No API, schema, migration, setting, dependency, dashboard, or deployment contract changes", `design.md:21` lists schema/migration under non-goals, and `design.md:75` says "No data or configuration migration is required" — while the PR adds `20260904_000000_add_usage_reset_transition_index` and `idx_usage_window_account_reset_time` on `usage_history` (CONCURRENTLY on PostgreSQL). Please update the impact/non-goals/migration-plan text to describe the index migration and its rollback. Migration chain note: the branch must stay parented on `20260830_000000_add_quota_warmup_claim_expiry`, and owner PR #2165 also adds a revision — whichever lands second must re-parent onto the new head.

3. **Remaining Codex threads on the head** — please reply and resolve or fix:
   - `app/modules/usage/repository.py:332` — `strftime('%s', recorded_at)` truncates fractional seconds, so `_usage_reset_confirmed_clause` can accept a pair on SQLite that `usage_reset_confirmed` then rejects in Python, and the ascending `LIMIT 1` returns that same pair every cycle. It is fail-closed (Python re-validates), but it is a drift between the SQL mirror and the authoritative predicate; `(julianday(recorded_at) - 2440587.5) * 86400.0` gives a fractional epoch and closes it cheaply (same pattern at line 1338).
   - `app/modules/usage/repository.py:1109` — `ORDER BY candidate_after.recorded_at ASC, id ASC ... LIMIT 1` returns the oldest confirmed pair after the baseline, while `design.md` still says "the most recent pair that passes the temporal reset predicate becomes the transition evidence". Either order newest-first or amend the design text; the practical impact is which reset tuple keys the warm-up claim.
   - `app/core/usage/refresh_scheduler.py:499` — the rollback vs concurrent `reactivate_account` window is between the CAS and the watermark re-read (milliseconds) and the direction is fail-closed; I would treat this as accepted risk with a one-line comment rather than add a version column.

Also noting #2048 still edits the same `_select_long_window_entry` body and import line in `refresh_scheduler.py` plus `CAPACITY_PLAN_ALIASES` in `logic.py`; whichever lands second needs a rebase. None of the PRs merged this round (#2078, #2079, #2092, #2114, #2143, #2148, dependabot batch) touch these files, so no new conflicts from them.

Soju06:
main now has alembic head `20260908_000000_add_subscription_overflow` (merged in #2165). Please re-parent your migration's `down_revision` onto it so `alembic heads` stays single; the PR cannot merge until then.

Soju06:
Oldest-20 triage at `014d17d6`: keep open and blocked on correctness as well as integration. The current Codex findings still require an implementation: rollback must not undo a concurrent operator reactivation, persisted reset lookup must choose the newest qualifying transition, and SQLite fractional timestamps must not hide a later valid reset. OpenSpec must describe the added migration. Current CI has typing/unit/integration/PostgreSQL/migration failures and the scheduler conflicts with main. This overlaps #2048's recovery surface but adds durable reset evidence; do not merge or close either as interchangeable.

Soju06:
Closing as superseded by #2405 (`3d501ac5e`, merged 2026-09-13). Issue #2404 — the same defect this PR fixes — is closed as of 2026-09-13T07:56Z.

What I verified on `origin/main` (`ca243dec4`), not just from the PR descriptions:

- **The premise is gone.** `_confirmed_free_monthly_reset_recovery` no longer exists. `app/core/usage/refresh_scheduler.py:520-568` gates recovery on a baseline whose `reset_at` matches the persisted block deadline (`abs(baseline.reset_at - account.reset_at) <= _BLOCK_RESET_MATCH_TOLERANCE_SECONDS`) for *any* window, and documents the reasoning explicitly: "That anchoring -- not the account's plan -- is what keeps a paid account with an exhausted short window from being released by unrelated long-window availability."
- **The durable reset evidence that the 09-10 hold cited as the reason not to close is now on main.** `_resolve_reset_evidence` iterates `for window in _RESET_EVIDENCE_WINDOWS`, calling `usage_repo.history_since(account.id, window, since)` with `since = blocked_at`, so recovery survives a restart and is not limited to the in-cycle monthly pair. It additionally requires the matching baseline to be unique across slots (`matching_slots == 1`) and keeps warm-up evidence in a separate map so an unvalidated transition can never override a cooldown — strictly stronger than this branch's single applicable-long-window selection.
- **The safety property this branch implemented as a plan-type CAS is also on main**, at `refresh_scheduler.py:564`: `if window != "primary" and _short_window_blocks_recovery(...)` withholds a long-window recovery while the account's own short window is exhausted and unelapsed.
- **The new index is not needed by main's query shape.** `app/db/models.py:2794` (`idx_usage_window_account_time`) and `:2818` (`idx_usage_window_account_time_covering`) already lead with `(window, account_id, recorded_at)`, which is exactly `history_since`'s filter and order. #2405's commit message records the opposite lesson from production — "a bounded lookback would have missed a baseline ~1,700 rows back" — which is precisely the failure mode of the `LIMIT 1` reset-transition query in `app/modules/usage/repository.py`.

Why this is a close rather than a rebase: the head `06e18b98` is `dirty` against `ca243dec4` and was last rebased onto `e040053b` (#2204, 09-11), which predates #2405. It rewrites `app/core/usage/refresh_scheduler.py` (+180/-46) — the same body #2405 rewrote (+267/-66) — plus both `openspec/specs/account-routing/spec.md` and `openspec/specs/usage-refresh-policy/spec.md`, the already-archived `openspec/changes/archive/2026-09-09-recover-rate-limited-on-confirmed-reset/` folder, and three Alembic revisions (one index + two merge revisions) whose parents no longer match main's head. Re-deriving it on top of #2405 would be a rewrite, not a rebase.

Thank you — the diagnosis was right, the production behaviour it describes is fixed, and the CAS/watermark discipline in this branch informed the shipped design. If a residual gap remains after #2405 (for example a measured slow `history_since` scan on a large `usage_history`), please open a focused follow-up with the `EXPLAIN` output rather than reviving this branch.

PR #2082, opened 2026-09-04
fix(proxy): report duplicate tool-call replay terminal

Komzpa:
References original PR #1706 and the maintainer's close request on #1862 to re-offer this piece standalone.

## Summary
- emit `duplicate_tool_call_replay_suppressed` instead of `stream_incomplete` when a replayed side-effecting tool call is intentionally suppressed
- keep the intentionally fenced duplicate terminal out of the HTTP bridge retry circuit
- preserve non-success settlement while avoiding account-health penalties for the suppressed replay

## Validation
- `uv run ty check`
- `uv run ruff check .`
- `uv run ruff format --check .`
- `uv run pytest -q --tb=short --disable-warnings tests/unit/test_proxy_utils.py::test_stream_responses_suppresses_contiguous_side_effect_replay_across_response_ids tests/unit/test_proxy_utils.py::test_http_bridge_duplicate_tool_call_replay_emits_dedicated_terminal_failure tests/unit/test_proxy_utils.py::test_websocket_duplicate_tool_call_completion_uses_dedicated_terminal_code tests/unit/test_proxy_utils.py::test_websocket_suppressed_duplicate_does_not_mask_genuine_terminal_failure tests/unit/test_proxy_http_bridge.py::test_http_bridge_retry_circuit_ignores_suppressed_duplicate_terminal`
- `uv run openspec validate --strict report-suppressed-duplicate-tool-call-terminal`

## Summary by CodeRabbit

- **Bug Fixes**
  - Suppressed duplicate tool-call replays now report the dedicated `duplicate_tool_call_replay_suppressed` terminal error instead of the generic `stream_incomplete` code.
  - Requests remain unsuccessful without triggering account-health penalties or retry-circuit failures.
  - Consistent behavior is now applied across direct, HTTP bridge, and WebSocket responses.

- **Tests**
  - Added regression coverage for terminal events, logging, retry handling, and genuine stream failures.

PR #2067, opened 2026-09-04
test(proxy-responses): fence leaked recovery settlement tasks

JustYannicc:
## Summary

The bridge integration suite can leave a recovery-settlement retry alive after a stale-owner replay. That retry keeps touching the shared SQLite database while the next fixture resets the schema or starts the app, producing `database is locked` at `runtime_sentinels` startup.

## Type of change

- [x] `test:` — test-only change
- [ ] `fix:` — bug fix (no behavior change beyond the bug)

Linked issue: None provided. This is a CI fixture-isolation fix for the bridge integration failure. Hosted CI is being rerun after one unrelated dashboard smoke-test flake.

## OpenSpec

- [x] Not applicable — test-only fixture cleanup

## Changes

- Add a bridge-scoped fixture helper that cancels and awaits leaked `http-bridge-recovery-settlement-*` tasks.
- Run it before lifespan shutdown and again at app fixture teardown.
- Reap only tasks owned by the active event loop; leave closed `TestClient` portal-loop tasks alone.
- Keep production retry behavior, fingerprint logic, and SQLite configuration unchanged.

## Simplicity

No settings, migrations, README sections, dashboard navigation, or production behavior are added.

## Test plan

- `uv run pytest -q tests/integration/test_http_responses_bridge.py tests/integration/test_proxy_websocket_responses.py` → 286 passed
- `uv run pytest -q tests/integration/test_detached_persistence.py tests/unit/test_graceful_shutdown.py tests/unit/test_otel.py` → 115 passed
- `uv run pytest -q tests/unit/test_proxy_http_bridge.py -k 'settlement or background_cleanup'` → 21 passed, 942 deselected
- `uv run ruff check tests/conftest.py`
- `uv run ruff format --check tests/conftest.py`
- `uv run ty check`
- `git diff --check`

## Screenshots / output

Not applicable. This PR changes test fixture cleanup only.

## Checklist

- [x] Title is in Conventional Commits format.
- [x] Ran the relevant integration and static checks locally.
- [x] Simplicity gates reviewed.
- [x] CHANGELOG is not edited by hand.

PR #2066, opened 2026-09-04
test(otel): isolate startup database paths

JustYannicc:
## Problem

The OTEL lifespan test mocks the database lifecycle but still executes unrelated startup database work. In the full suite, that work can write to shared SQLite state and contend with another test’s writer, producing a `database is locked` failure or startup stall.

## Why this PR exists

This test should exercise bridge-membership staleness, not the independent startup seed and routing-cache database paths. The existing mocks left those paths live and made the test sensitive to unrelated suite-wide SQLite contention.

## How this PR solves it

The test replaces the startup accounts repository seed and routing-availability cache refresh with async-safe stubs. The mocked lifespan now remains scoped to the behavior under test and cannot issue those unrelated SQLite writes.

## Scope

- Test-only change in `tests/unit/test_otel.py`.
- No application/runtime behavior, schema, or production SQLite configuration changes.

## Relationship to related work

- This PR fixes the unit-test OTEL lifespan startup lock observed on [#1962](https://github.com/Soju06/codex-lb/pull/1962).
- It does not fix the separate integration-bridge lock in `verify_encryption_key_fingerprint()` observed on [#1954](https://github.com/Soju06/codex-lb/pull/1954) (run 33690549740, job 100448103028) and [#1962](https://github.com/Soju06/codex-lb/pull/1962) (run 33721422129, job 100541208170).
- The separate integration-startup fixture repair is tracked in [#2067](https://github.com/Soju06/codex-lb/pull/2067), which handles leaked recovery-settlement tasks in bridge integration fixtures. #2066 neither includes that fix nor makes #2067 green by itself.

## Verification

- `uv run pytest tests/unit/test_otel.py -q` — 49 passed on the reconciled head.
- `uv run ruff check tests/unit/test_otel.py` — passed.
- `git diff --check` — passed.
- The branch includes live `upstream/main` at `dd28d7dff94cdd4919067c1986fd9606b9bbc6b9`; the PR diff remains limited to `tests/unit/test_otel.py`.

## Current status

Ready for maintainer review after hosted CI completes for the reconciled head. No application changes are included.

## Summary by CodeRabbit

* **Tests**
  * Improved test isolation by mocking startup dependencies, preventing database and routing-cache operations from interfering with other tests.

PR #2083, opened 2026-09-04
fix(proxy): fork model-transition owner conflicts

Komzpa:
References original PR #1619 and the maintainer's close request on #1862 to re-offer this piece standalone.

## Summary
- reproduce the current `continuity_owner_conflict` failure on an eligible model-transition request
- retry the eligible local conflict once on a new hard account-neutral child lane
- clear parent affinity, hard-continuity, and reused turn-state identity before submitting the child lane

## Validation
- `uv run ty check`
- `uv run ruff check .`
- `uv run ruff format --check .`
- `uv run pytest -q --tb=short --disable-warnings tests/unit/test_proxy_http_bridge.py::test_stream_via_http_bridge_forks_account_neutral_model_transition_after_owner_conflict tests/unit/test_proxy_http_bridge.py::test_stream_via_http_bridge_limits_model_transition_owner_conflict_fork_to_one_retry tests/unit/test_proxy_http_bridge.py::test_stream_via_http_bridge_model_transition_owner_conflict_fork_does_not_rebind_parent_turn_alias`
- `uv run openspec validate --strict fork-model-transition-owner-conflict`

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved recovery when switching models encounters a session ownership conflict.
  - Eligible requests can now retry through a fresh, account-neutral connection instead of failing immediately.
  - Retry handling prevents previous session or turn state from being carried into the new connection.
  - Conflicts remain safely limited to a single retry, preserving the original error when recovery is unsuccessful.

- **Tests**
  - Added coverage for successful recovery, conflict handling, and isolation of prior session state.

### DIFF
--- a/app/modules/proxy/_service/http_bridge/streaming.py
+++ b/app/modules/proxy/_service/http_bridge/streaming.py
@@ -2016,6 +2016,7 @@ class _HTTPBridgeStreamingMixin:
             else dict(headers)
         )
         fresh_replay_excluded_account_ids: set[str] = set()
+        model_transition_owner_conflict_fork_attempted = False
         unanchored_fork_spill_attempted = False
         verified_stale_anchor_generation_captured = False
         verified_stale_anchor_circuit_key: _HTTPBridgeSessionKey | None = None
@@ -2068,6 +2069,70 @@ class _HTTPBridgeStreamingMixin:
                     durable_full_resend_fresh_payload
                 )
             return durable_full_resend_is_account_neutral
+
+        def switch_model_transition_to_account_neutral_fork(exc: ProxyResponseError) -> bool:
+            nonlocal account_neutral_recovery
+            nonlocal affinity
+            nonlocal bridge_session_key
+            nonlocal downstream_turn_state
+            nonlocal force_local_recovery_creation
+            nonlocal incoming_turn_state_header
+            nonlocal model_transition_owner_conflict_fork_attempted
+            nonlocal preferred_account_has_continuity_provenance
+            nonlocal request_state
+            nonlocal session_creation_headers
+            nonlocal session_header_fallback_key
+
+            error_code, _error_message = _proxy_error_code_message(exc)
+            if (
+                durable_model_transition_lookup is None
+                or error_code != "continuity_owner_conflict"
+                or model_transition_owner_conflict_fork_attempted
+                or forwarded_request
+                or not _http_bridge_payload_is_account_neutral_fresh_replay(effective_payload)
+                or request_state.previous_response_id is not None
+                or rewritten_file_account_id is not None
+            ):
+                return False
+            reused_parent_turn_state = (
+                incoming_turn_state_header is not None and downstream_turn_state == incoming_turn_state_header
+            )
+            failed_owner_id = request_state.preferred_account_id
+            _log_http_bridge_event(
+                "model_transition_owner_conflict_fork",
+                bridge_session_key,
+                account_id=failed_owner_id,
+                model=effective_payload.model,
+                detail="outcome=retry_without_previous_model_owner",
+                cache_key_family=bridge_session_key.affinity_kind,
+                model_class=_extract_model_class(effective_payload.model) if effective_payload.model else None,
+                owner_check_applied=True,
+            )
+            if failed_owner_id is not None:
+                fresh_replay_excluded_account_ids.add(failed_owner_id)
+            session_creation_headers = without_http_bridge_session_affinity_headers(session_creation_headers)
+            incoming_turn_state_header = None
+            session_header_fallback_key = None
+            affinity = _AffinityPolicy()
+            replay_kind, replay_key = make_http_bridge_account_neutral_replay_key(uuid4().hex)
+            bridge_session_key = _HTTPBridgeSessionKey(
+                replay_kind,
+                replay_key,
+                bridge_session_key.api_key_id,
+                strength="hard",
+            )
+            account_neutral_recovery = True
+            force_local_recovery_creation = True
+            model_transition_owner_conflict_fork_attempted = True
+            request_state.affinity_policy = affinity
+            request_state.hard_continuity_anchor = False
+            request_state.preferred_account_id = None
+            request_state.excluded_account_ids.update(fresh_replay_excluded_account_ids)
+            preferred_account_has_continuity_provenance = False
+            if reused_parent_turn_state:
+                request_state.session_id = None
+                downstream_turn_state = None
+            return True
 
         def owner_unavailable_allows_account_neutral_replay(exc: ProxyResponseError) -> bool:
             return (
@@ -2279,6 +2344,8 @@ class _HTTPBridgeStreamingMixin:
                 # carries its own continuity is a safe fresh raw-HTTP replay.
                 if effective_payload.previous_response_id != payload.previous_response_id:
                     setattr(exc, _HTTP_BRIDGE_PREPARED_ANCHOR_ATTR, True)
+                if switch_model_transition_to_account_neutral_fork(exc):
+                    continue
                 if not owner_unavailable_allows_account_neutral_replay(exc):
                     exc_code, _exc_message = _proxy_error_code_message(exc)
                     if not unanchored_fork_spill_attempted and _http_bridge_unanchored_fork_can_spill_on_cap(
--- a/openspec/changes/fork-model-transition-owner-conflict/.openspec.yaml
+++ b/openspec/changes/fork-model-transition-owner-conflict/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-09-05
--- a/openspec/changes/fork-model-transition-owner-conflict/.openspec.yaml
+++ b/openspec/changes/fork-model-transition-owner-conflict/.openspec.yaml
@@ -1,2 +1,2 @@
 schema: spec-driven
-created: 2026-09-05
+created: 2026-09-04
--- a/openspec/changes/fork-model-transition-owner-conflict/proposal.md
+++ b/openspec/changes/fork-model-transition-owner-conflict/proposal.md
@@ -1,0 +1,18 @@
+# Fork Model-Transition Owner Conflicts
+
+## Why
+
+A durable hard continuity lookup can still identify the old model owner's
+account when the client starts a fresh request on an incompatible model. If
+bridge creation reports `continuity_owner_conflict`, the current path surfaces
+the 502 even when the payload is account-neutral and can safely start a new
+local lane.
+
+## What Changes
+
+- Fork eligible local model-transition owner conflicts onto one new
+  account-neutral HTTP bridge lane.
+- Exclude the conflicting owner, force local creation, and keep the child lane
+  hard once selected.
+- Clear parent-derived request state so the child lane cannot rebind the old
+  turn alias.
--- a/openspec/changes/fork-model-transition-owner-conflict/specs/responses-api-compat/spec.md
+++ b/openspec/changes/fork-model-transition-owner-conflict/specs/responses-api-compat/spec.md
@@ -1,0 +1,28 @@
+# responses-api-compat Delta
+
+## ADDED Requirements
+
+### Requirement: HTTP bridge model-transition owner conflicts fork a guarded child lane
+
+When a hard continuity lookup identifies a durable owner for an incompatible model and HTTP bridge creation returns `continuity_owner_conflict`, the service MUST retry at most once on a new account-neutral child lane when the request is local, has no `previous_response_id`, has no resolved file owner, and the effective payload passes the account-neutral fresh-replay validator. The child lane MUST exclude the conflicting owner, force local creation, use hard affinity, and clear parent-derived affinity, hard-continuity, and reused turn-state identity before submitting the request.
+
+#### Scenario: Eligible model transition conflict forks once
+
+- **GIVEN** a durable hard continuity key belongs to account A for the old model
+- **AND** the new request uses an incompatible model with account-neutral input
+- **WHEN** HTTP bridge creation returns `continuity_owner_conflict`
+- **THEN** the service retries once on a new hard account-neutral lane
+- **AND** the retry excludes account A and does not forward the request
+
+#### Scenario: Child lane does not inherit parent identity
+
+- **GIVEN** the model-transition request reuses the parent turn-state header
+- **WHEN** the service forks to the account-neutral child lane
+- **THEN** the submitted child request has no parent session id
+- **AND** it has no parent affinity policy or hard-continuity anchor
+
+#### Scenario: Second conflict remains fail-closed
+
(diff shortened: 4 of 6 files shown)


Komzpa:
@codex review

Komzpa:
@codex review

Soju06:
Triage (round 19, head `7ffbb80`): **merge-ready**, with one follow-up that I'd like tracked because it limits how much of the claimed fix actually reaches real clients.

Verified on the current head: CI green on the exact SHA (unit / integration-core 1-3 / integration-bridge / e2e / PostgreSQL / ty / ruff / simplicity budgets), clean merge-tree against fresh `origin/main` (none of this round's merges — #2082, #2087, #2091, #2066, #2067 etc. — touch the fork path), openspec change folder present, and the fork guard (`streaming.py:2087-2096`: one attempt, account-neutral payload only, no `previous_response_id`, no forwarded request, no rewritten file account) is tight.

### P2 — follow-up: the reused-turn-state fork does not converge for Codex CLI-style clients

For any client that sends `x-codex-turn-state` (Codex CLI default), `api.py:6131-6141` reuses the incoming token as `downstream_turn_state` and builds the *response* header before the service runs (`affinity.py:647-651`). The fork's `reused_parent_turn_state` branch (`streaming.py:2130-2132`) then nulls `request_state.session_id` / `downstream_turn_state`, so the child lane never claims the turn-state alias (in-memory registration is skipped at `streaming.py:4728` because `account_neutral_recovery` is set; durable `latest_turn_state` stays `None`). The parent row — live, healthy, on account A, old model — keeps the alias. Your new test `..._does_not_rebind_parent_turn_alias` asserts exactly this (`stream_downstream_turn_states == [None]`), but it is single-turn.

The child *does* durably publish its response ids as `previous_response` aliases (`session_registry.py:382+`, `_register_http_bridge_previous_response_id_impl`, not gated on turn state). So:

- Chaining client (store=true): turn 2 arrives with parent token + child's `previous_response_id`. `durable_bridge_coordinator.lookup_request_targets` resolves turn-state → parent row (A) and previous_response → child row (C): two identities on two accounts, not same-account, not the session-header carve-out (`durable_bridge_coordinator.py:137-161`) → **502 `continuity_owner_conflict`**, raised pre-loop where the fork guard (`previous_response_id is None`) cannot help. That is the same error the PR sets out to remove, one turn later.
- Non-chaining client (store=false): the parent token re-resolves to the old-model parent every turn → `durable_model_transition_lookup` → preferred A → same conflict → re-fork on every turn. Never converges: one fresh nonce lane per turn until idle TTL, no prompt-cache continuity, plus a wasted failed selection and `owner_conflict` log per request.

This differs from main's `switch_to_account_neutral_replay`, where the owner is dead/abandoned so the parent alias goes away; here the parent stays live, so it is specific to this fork. `codex review --base origin/main` flagged the same thing independently (its P1). Remedy options: mint a child turn-state and propagate it into the response header (needs `api.py` plumbing since the header is built up front), or let the child claim the parent turn-state alias durably (the coordinator already has an account-neutral-wins carve-out for the session-header alias; a turn-state equivalent would be needed). Either way a two-turn regression test (fork, then follow-up with the returned header) would have caught this. Happy to take this as a follow-up PR if you prefer to land the current scope first.

### P3s (non-blocking)

1. `request_state.request_stage` stays `"follow_up"` on the forked child even though it is a first turn on a fresh lane; `switch_to_account_neutral_replay` recomputes it via `_http_bridge_request_stage(durable_lookup=None)`. Only affects the first-turn timeout/log selection keyed on `request_stage == "follow_up"` (`streaming.py:1891`, `:1951`).
2. The docs/tests model the conflict as a 502, but the production raise from account selection (`sticky_selection.py:459-463`, `load_balancer.py:769-778`) reaches the streaming loop via `selection_failure_response` (`selection_errors.py:41`) as a **503** with code `continuity_owner_conflict`; only the coordinator/`continuity.py` variants are 502. The guard matches on code, so behavior is fine; `proposal.md:8` ("surfaces the 502") and the test fixtures should say "`continuity_owner_conflict` (502/503)".
3. `strength="hard"` on the child key (`streaming.py:2122`) is redundant: `internal_unanchored_parallel` is already in `_HARD_HTTP_BRIDGE_AFFINITY_KINDS` (`support.py:85-93`), so `_HTTPBridgeSessionKey.__post_init__` (`support.py:1260`) yields `hard` anyway, and `creation_keys[1].strength == "hard"` is tautological. Harmless.
4. Excluding the durable owner A from the child (`fresh_replay_excluded_account_ids.add(failed_owner_id)`, `streaming.py:2110-2111`) is conservative: when A is the only account with capacity/model access, the fork ends in `no_accounts` / `hard_affinity_saturated` instead of the original conflict — an error-shape change for the client. Acceptable, but the openspec should state it.

Soju06:
Triage (round 20, head `7ffbb80`, unchanged since 09-04): holding this one — clarifying the round-19 note, which read as "merge-ready with a follow-up". Two things need to land in this PR before it can merge.

**1. The fork does not converge for real clients (round-19 P2, still open on head).** Verified again on current `main` (`a0d94a46a`): `app/modules/proxy/api.py:6142-6150` reuses the incoming `x-codex-turn-state` as `downstream_turn_state` and builds the response header before the service runs; the fork's `reused_parent_turn_state` branch (`app/modules/proxy/_service/http_bridge/streaming.py:2132-2134`) nulls `downstream_turn_state`, so `streaming.py:4729` skips `_register_http_bridge_turn_state` for the child, and `_service/http_bridge/session_registry.py:734-739` only registers the `session_header` alias when `session.key.affinity_kind == "session_header"` (the child key is `internal_unanchored_parallel`). The parent row therefore stays the resolved identity for every later turn while the client keeps sending the parent token:

- chaining client (turn 2 with parent token + child `previous_response_id`): `durable_bridge_coordinator.py:163-190` sees two specific identities on two accounts and raises the same `continuity_owner_conflict` this PR sets out to remove, pre-loop where the fork guard (`previous_response_id is None`) cannot apply;
- non-chaining Codex CLI client: every turn re-resolves the old-model parent, hits the conflict, and forks a fresh nonce lane again (failed selection + new durable row + new upstream lane per turn, no convergence).

Please add a two-turn regression (fork, then a follow-up carrying the returned header) and pick a convergence mechanism: mint a child turn state and propagate it into the response header (needs the `api.py` header plumbing to accept a late override), or let the child durably claim the parent turn-state alias (a turn-state equivalent of the coordinator's account-neutral session-header carve-out at `durable_bridge_coordinator.py:176-181`).

**2. Repro on current `main`.** When #1862 was closed the ask was to re-offer this piece "with the repro" if the defect still reproduces on current `main`. The `streaming.py` hunk here is the reverted #1619 code minus two comments, and all three tests monkeypatch `_get_or_create_http_bridge_session` to raise `continuity_owner_conflict`, so nothing exercises the actual raise sites (`app/modules/proxy/_load_balancer/sticky_selection.py:472-476`, `app/modules/proxy/load_balancer.py:780-785`, or the coordinator). tasks.md 1.1 says the failure was reproduced — please turn that into a test that sets up the durable bridge row / sticky `CODEX_SESSION` row state that produces the conflict on `main` for an account-neutral model-transition request and only passes with the fork. That also tells us whether the conflict is deterministic for a given state (which decides how bad the per-turn re-fork in (1) is).

The round-19 P3s (child keeps `request_stage == "follow_up"` where `switch_to_account_neutral_replay` recomputes it; 502 vs 503 wording in `proposal.md:8` and the fixtures; redundant `strength="hard"`; openspec should state the `no_accounts`/`hard_affinity_saturated` error-shape change when the excluded owner is the only capable account) are unchanged and remain non-blocking.

Mechanically the branch is fine: CI green on the head SHA, clean merge against fresh `origin/main` (re-checked after this round's merges, none of which touch these files), architecture/timing-seam/cancellation guards and `openspec validate --strict` pass on the merged tree.

Soju06:
Oldest-20 triage at unchanged `7ffbb804`: keep open and blocked despite green CI and an old clean Codex review. The round-20 maintainer findings still apply: the fork clears downstream turn state after the API has already chosen the parent response header, so a follow-up can resolve the parent again. The existing three tests inject `continuity_owner_conflict` rather than constructing the durable/sticky state that produces it. Required proof is a real conflict followed by a second request using the returned header, both chaining and non-chaining, with convergence to the child. This is partial overlap with #1905, not a duplicate.

Soju06:
Reviewed from two independent angles plus a local `codex review`, with each finding verified before posting.

**Merging this.** The fork is mechanically sound and strictly better than main on every path I could construct: without it the model-transition owner conflict is a hard 502 and the conversation is dead. All three new tests fail at the merge base, `openspec validate --strict` passes, and the delta text is honest — it governs a single forking request and says explicitly that the child lane must not rebind the old turn alias.

One real incompleteness and one test-shape note, neither of which should hold a fix that turns a hard 502 into a served turn:


---

### MAJOR — `app/modules/proxy/_service/http_bridge/streaming.py`

**The fork (streaming.py:2097-2135) neither rebinds nor tombstones the conflicting parent durable row and clears downstream_turn_state, so the child lane registers no alias at all and every subsequent turn re-derives the same conflict — burning a fresh account and cold upstream session per turn, or returning the original 502 once the resent history stops being account-neutral.**

_How it fails:_ Client on the old model with x-codex-turn-state: T; ensure_http_downstream_turn_state echoes T, so reused_parent_turn_state is True and the fork sets downstream_turn_state = None (streaming.py:2134). That disables both alias-publication paths — _register_http_bridge_turn_state is skipped at streaming.py:4728 (guarded on downstream_turn_state is not None and not account_neutral_recovery) and recovery_turn_state is None at streaming.py:4569 — and the fork body makes no durable-bridge call, so the parent row keeps latest_turn_state=T, account_id=A and the old model. Turn N+1 therefore re-derives the identical durable_model_transition_lookup and conflicts again: if the resent history is still plain text it forks onto yet another account (fresh websocket, full history resend, no prompt-cache anchor), and if it carries a `reasoning` or `web_search_call` item — the normal shape once the assistant has answered — responses_payload_is_account_neutral_fresh_replay returns False, the fork guard fails, and the client gets the original 502 back. I also confirmed the fork creates a newly reachable alias split: building the post-fork state against the real DurableBridgeSessionCoordinator (parent session_header row with turn-state alias http_turn_model_parent on acc-model-owner/old model, child internal_unanchored_parallel account-neutral-replay row with previous_response alias resp_child on acc-model-alternate/new model), lookup_request_targets gives turn_state+previous_response -> continuity_owner_conflict; previous_response only -> acc-model-alternate (recovers); turn_state only -> acc-model-owner on the old model (same conflict again). The conflicting-alias raise happens inside lookup_request_targets at streaming.py:1435, which is re-raised unconditionally at streaming.py:1445-1448 before the `while True:` the fork lives in, so the fork structurally cannot recover from the state it created. That case needs a client that starts supplying previous_response_id, which the fork's own guard implies it does not; the realistic recurrence is the turn-state-only one.

_Suggested fix:_ Either state the per-turn recurrence and the account-neutral-payload-only scope explicitly in proposal.md and the spec delta, or make the fork converge: after the child lane settles, repoint or tombstone the parent's turn-state and session-header aliases onto the child the way switch_to_account_neutral_replay does through recovery_turn_state, guarded so a still-live old-model parent lane is not stolen. Whichever ships, add a turn N+1 regression — a second _stream_via_http_bridge call carrying the echoed turn-state header, with and without the child's previous_response_id, after a fork.

---

### MINOR — `tests/unit/test_proxy_http_bridge.py`

**All three regression tests inject the conflict from a monkeypatched _get_or_create_http_bridge_session using the one error message that only DurableBridgeCoordinator.lookup_request_targets produces — a site whose exception is deliberately re-raised before the creation loop — so the tests model a producer that cannot reach the patched call site.**

_How it fails:_ The three new tests (tests/unit/test_proxy_http_bridge.py:27044, :27163, :27256) raise ProxyResponseError carrying 'Durable continuity aliases resolve to conflicting upstream owners', a string whose only app-side producer is durable_bridge_coordinator.py:162. That call sits at streaming.py:1435 inside `except ProxyResponseError: raise` (streaming.py:1445-1448, 'Conflicting durable aliases are a continuity decision... Never soften that fail-closed result'), outside and before the `while True:` creation loop the fork hooks into. The fork guard matches on error CODE only, so it does fire for the real producers — sticky_selection.py:463, load_balancer.py:776 and request_submit.py:4620, whose AccountSelection failures map through selection_failure_response to a 503 carrying error code continuity_owner_conflict inside _get_or_create_http_bridge_session — but nothing in the PR demonstrates that, which is what the maintainer asked for when closing #1862.

_Suggested fix:_ Add one regression that produces continuity_owner_conflict from a real selection path rather than a mock — e.g. a hard sticky row that disagrees with required_account_id (load_balancer.py:766-779) or a live turn-state alias on an account-neutral lane owned by a third account reaching bind_account_neutral_recovery_owner / resolve_required_account_id inside _get_or_create_http_bridge_session. At minimum, use the message the selection path actually emits ('Account-owned continuity sources conflict; retry the logical turn'). While there, restore the negative controls PR #1619 carried and this PR dropped (forwarded request, resolved file owner, non-neutral payload); nothing currently pins the four guard clauses at streaming.py:2087-2096.

---

### Findings raised and then refuted

Recorded so the next round does not re-litigate them:


- Reviewer 2's claim that the patched call site is unreachable on current main — that 'the only code path in the repo that emits the exact string is lookup_request_targets', whose exception is re-raised before the loop — is refuted as a reachability argument. The fork guard tests `error_code != "continuity_owner_conflict"`, not the message, and there are several producers inside _get_or_create_http_bridge_session: _load_balancer/sticky_selection.py:463 and load_balancer.py:776 return AccountSelection(error_code='continuity_owner_conflict'), which selection_errors.py:27-41 maps to a 503 whose payload code is continuity_owner_conflict, plus request_submit.py:4620 raises it directly. The defect is reachable; only the test fidelity is at issue, which is what I kept.

- Reviewer 1's framing of turn N+1 scenario (a) — client echoes the turn state AND sends previous_response_id: resp_child, producing a hard 502 — as the expected follow-up is refuted as the dominant case. The fork's own guard requires request_state.previous_response_id is None on the forking turn, so the client demonstrably does not supply an anchor on this path; the realistic turn N+1 is the turn-state-only case, which I verified resolves back to the parent row and re-conflicts. I confirmed case (a) is reachable (probe: turn_state + previous_response -> continuity_owner_conflict) and folded it into the confirmed finding as a secondary consequence rather than the headline.

- Reviewer 1's report that the bridge suite shows one failure at head is accurate but the failure is not attributable to this PR: test_stream_via_http_bridge_fails_closed_before_file_affinity_when_previous_response_owner_misses fails identically with app/ checked out at the merge base ('no such table: file_account_pins'), so it is pre-existing environment breakage as Reviewer 2 concluded.

Soju06:
Tracking the incompleteness above as #2389 so it does not get lost, and merging this — a fix that turns a hard 502 into a served turn should not wait on making the fork converge.

PR #1881, opened 2026-08-22
fix(http-bridge): expire stale inflight creations

Komzpa:
This bundled PR is being split into focused current-main PRs.

Base for the split: `Soju06/codex-lb@dd28d7dff94cdd4919067c1986fd9606b9bbc6b9`

Original bundled head: `2d0a50c99e4b471d9508fcb7cceb2de98e8ea631`

## Split Map

- Soju06/codex-lb#2084 - `fix(http-bridge): wait for aborted bridge owner`
  - Slice 1 of Soju06/codex-lb#1881.
  - Marker lifecycle: owner-task provenance on inflight markers, snapshot-time cleanup of completed markers, ownership-based capacity counting, and the `wait-for-aborted-bridge-owner` OpenSpec delta.
  - Owned-wait decision: bounded by configured admission wait, per the spec text.
- Soju06/codex-lb#2088 - `fix(http-bridge): recover from draining owner rejection`
  - Slice 2 of Soju06/codex-lb#1881.
  - Blue-green drain recovery: `bridge_drain_active` owner-forward classification, pre-dispatch bootstrap-rebind bounding, and the `recover-draining-owner-forward` OpenSpec delta.
- Soju06/codex-lb#2086 - `fix(proxy): allow prefix-settled replay outputs`
  - Slice 3 of Soju06/codex-lb#1881.
  - Replay safety certification: allow replay suffixes that settle pending prefix tool calls, with the `allow-prefix-settled-tool-output-replay` OpenSpec delta.
- Soju06/codex-lb#2087 - `fix(http-bridge): reject unsafe forwarded headers`
  - Slice 4 of Soju06/codex-lb#1881.
  - Header safety: reject unsafe reconstructed bridge metadata, drop unsafe ordinary client headers, and add the `reject-illegal-bridge-forward-headers` OpenSpec delta.

Issue #1695, opened 2026-08-11
flaky: test_v1_responses_http_bridge_reconnects_after_clean_upstream_close intermittently fails with 409 on CI


Soju06:
`tests/integration/test_http_responses_bridge.py::test_v1_responses_http_bridge_reconnects_after_clean_upstream_close` has failed with `assert 409 == 200` on the second request in three unrelated CI runs over the past week (observed on PRs #1615, #1653, and one more takeover branch), each time passing on a plain rerun and passing locally (3/3 and 56/56 in the containing module).

The 409 suggests the reconnect race occasionally resolves to a conflict (the bridge still considers the previous owner live when the second request lands) under CI runner timing. Likely needs either a deterministic synchronization point in the test (wait for the close to be fully settled before the second request) or a widened acceptance if 409-then-retry is legitimate behavior for this window.

Filing so the next flake doesn't get papered over with reruns — happy to take it in a future triage round if nobody claims it.

Soju06:
Narrowed the candidate set: the only 409 producers reachable from `POST /v1/responses` are the three `bridge_instance_mismatch` sites in `http_bridge/mixin.py` (turn-state scope conflict ~1169, generated-turn-state continuity ~1188, durable claim rejection ~1800). The test client sends no turn-state header, so the durable claim at ~1800 is the prime suspect.

Proposed mechanism: the fake upstream queues the clean close(1000) behind `response.completed`, so the reader teardown (sets `closed`, pops the session via `_detach_http_bridge_session_locked`, then a scheduled `_close_http_bridge_session` releases the durable lease) races the second request's acquisition. That path can detach the retiring session (`registry_detach`, which sets `force_durable_takeover_after_detach`) and immediately claim the durable row while `release_live_session` is mid-flight. `claim_session`'s read-modify-write in `durable_bridge_repository.py` leans on `with_for_update()`, which is a no-op on SQLite (CI), so an interleaved release/claim can return a lookup that fails the owner check and surfaces the 409.

Two notes before picking a fix: (1) capturing the 409 body in the test assertion (or bridge event logs in CI) would confirm which site fires; (2) since a single-instance deployment can 409 a real client at this boundary, widening the test to accept 409-then-retry would mask a product bug — the settlement ordering (or a same-instance claim retry after settle) looks like the right fix, with a deterministic test barrier (wait for the retiring session to leave the registry before the second POST) as an interim de-flake. Likely shares a root boundary with #1304.

[...]

PR #2084, opened 2026-09-04
fix(http-bridge): wait for aborted bridge owner

Komzpa:
Consolidated into [Soju06/codex-lb PR 2088](https://github.com/Soju06/codex-lb/pull/2088), which carries the owner lifecycle changes together with their draining-owner recovery consumer. The survivor preserves all 17 original owner-lifecycle regression tests and the original change authorship; its full bridge/forwarding contract suite passes. Closing this duplicate review surface.

### DIFF
--- a/app/modules/proxy/_service/http_bridge/helpers.py
+++ b/app/modules/proxy/_service/http_bridge/helpers.py
@@ -909,6 +909,8 @@ async def _wait_for_http_bridge_retained_owner(
     try:
         await wait_on_shared_future(owner_task, timeout=timeout)
     except TimeoutError:
+        if owner_task.done():
+            return True
         return False
     except asyncio.CancelledError:
         current_task = asyncio.current_task()
--- a/app/modules/proxy/_service/http_bridge/helpers.py
+++ b/app/modules/proxy/_service/http_bridge/helpers.py
@@ -707,6 +707,8 @@ _HTTP_BRIDGE_LOCAL_RESET_MESSAGE = HTTP_BRIDGE_LOCAL_RESET_MESSAGE
 T = TypeVar("T")
 
 _HTTP_BRIDGE_INFLIGHT_STARTED_AT_ATTR = "_codex_lb_started_at"
+_HTTP_BRIDGE_INFLIGHT_OWNER_TASK_ATTR = "_codex_lb_owner_task"
+_HTTP_BRIDGE_INFLIGHT_ABORT_ERROR_ATTR = "_codex_lb_abort_error"
 # Provenance marker for bridge failures raised strictly before the current
 # request dispatched upstream. Only exceptions carrying this attribute are
 # safe for the streaming wrapper's raw-HTTP replay.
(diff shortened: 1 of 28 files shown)


Komzpa:
@codex review

Komzpa:
@codex review

Soju06:
Thanks for carving this out per the #1881 split and bounding the owned wait to one extra configured interval -- that resolves the 08-26/08-31 blockers, and the retained-marker/snapshot/observer-cancellation fixes from the Codex threads check out. Head `7475b81d6` still merges cleanly onto current `main` (including the 09-05 merges of #2014/#2082/#2087 that also touch `tests/unit/test_proxy_http_bridge.py`). Two remaining defects, both reproduced against `7475b81d6` with a scratch test:

1. **Capacity waiters fast-fail on a retained aborted marker** (`app/modules/proxy/_service/http_bridge/mixin.py:1304-1308` selection, `:1389-1390` `except ProxyResponseError: raise`). `capacity_wait_future = next(f for f in inflight.values() if not handoff)` does not skip markers that this PR now deliberately keeps registered after abort (done, exception = the 429). `wait_on_shared_future` returns `shared.result()` synchronously for a done future, so the stored `ProxyResponseError` is re-raised to an unrelated-key waiter before it waits at all. Repro: owner for key A cancellation-resistant, waiter B times out and aborts it (marker retained), a new request for key C at `max_sessions=1` is rejected after 0.4 ms with the configured wait at 200 ms. On `main` this cannot happen because settled markers are popped at abort; with retention it holds for the whole owner-unwind window (and indefinitely for a cancellation-resistant owner). This contradicts the modified requirement ("when the configured proxy admission wait timeout expires the proxy MUST reject"). Suggested: prefer pending non-handoff markers in the selection; when only aborted-retained markers remain, observe the owner via `_wait_for_http_bridge_aborted_owner` (bounded by the admission interval) rather than inheriting its 429. Please add a regression for the cross-key case. The same instant-429 applies to same-key arrivals during the unwind window -- if that is intended, please state it in the spec delta.

2. **Codex P1 "Finalize markers when owner cancellation lands before creation" was resolved without a code change and still reproduces.** The `await self._enforce_http_bridge_capacity_after_planned_closes(...)` call (`mixin.py:1341-1345`) is outside the `try/except BaseException` that settles the marker. A timing-out waiter now actively calls `owner_task.cancel()` while holding `_http_bridge_lock`; if the owner is parked on that lock acquire, `CancelledError` exits `_get_or_create_http_bridge_session` without `_settle_and_close_failed_http_bridge_creation`, leaving a done marker with a dead owner registered. Repro: owner with a planned LRU close, abort under the lock exactly as `_evict_http_bridge_inflight_waiter` does -> owner dies before `create_session`, marker remains in `_http_bridge_inflight_sessions`, next request at `max_sessions=1` is rejected in 0.2 ms, and only `http_bridge_activity_snapshot_nowait` (reached solely from the health endpoint) reclaims it. Fix: extend the failure-cleanup coverage to every await after marker publication (wrap the enforce call in the same `except BaseException` -> `_fail_http_bridge_inflight_session_creation` path), and add the regression Codex asked for.

Minor: `test_rejected_creator_does_not_release_the_registered_winners_durable_row` (#1695) was reworked into an unrelated test and the `late_owner_after_inflight_evict` test was deleted; `_raise_if_http_bridge_creation_superseded` (`mixin.py:1550`) and `release_durable_session=not superseded` are still reachable (e.g. `_close_all_http_bridge_sessions` failing inflight waiters), so please keep a #1695 regression or note why the scenario is now unreachable.

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Soju06:
Re-checked at `3b5815d213b76a05915bbd0addcb2a348b6f71f2` against current `main` (`e987c56c8`, which now includes #2127/#2150 on `mixin.py`/`helpers.py` and this round's merges #2078/#2143/#2079/#2092/#2114/#2148 plus dependabot; the branch still merges cleanly, no conflicts). The PR's own 18 new tests pass locally, but the head is not mergeable yet:

**CI is red on the head itself (not stacking noise)**
- `Lint (ruff)` / `test_repository_proxy_architecture_passes`: `http_bridge/mixin.py has 2439 lines; limit is 2436`. `main` is at 2427, so the branch adds +12 net; move the new admission-recovery branches into `helpers.py` or trim.
- `test_repository_timing_seams_pass`: `app/modules/proxy/_service/http_bridge/helpers.py:927` `wait_on_shared_future(owner_task, timeout=timeout)` is a raw-timeout site with allowance 0 for `helpers.py`. `_wait_for_http_bridge_retained_owner` needs the service threaded through so it can pass `scheduler=scheduler_for(service)` (the same way `_await_http_bridge_registry_wait` does). `test_repository_allowances_are_exact` and `test_repository_report_reproduces_committed_block` fail only as a consequence; please fix the seam rather than editing the allowance table.

**Both items from the 09-05 comment were resolved without a code change and still reproduce on this head** (scratch tests, same fixture style as `test_inflight_waiter_returns_429_when_aborted_owner_resists_cancellation`):

1. Cross-key fast-fail on a retained aborted marker. `_service/http_bridge/mixin.py:1310` still picks any non-handoff future and `:1388` `except ProxyResponseError: raise` re-raises the stored 429 synchronously (`wait_on_shared_future` returns `shared.result()` for a done future). Repro: owner A cancellation-resistant, waiter B times out (0.404 s = 2 intervals, expected), then a request for unrelated key C at `max_sessions=1` is rejected after **0.4 ms** with `proxy_admission_wait_timeout_seconds=0.2`. The new `except Exception` branch (`_evict_http_bridge_retained_capacity_waiter_after_error`) never sees this case because the abort error is a `ProxyResponseError`. The same-key path (`:1427` `except Exception: raise`) has the same instant-429 behavior. Fix: prefer pending non-handoff markers in the selection and, when only retained aborted markers remain, observe the owner within the admission budget instead of inheriting its 429; add the cross-key regression. If instant 429 for same-key arrivals during the unwind window is intended, the delta must say so -- `specs/proxy-admission-control/spec.md:30-36` currently only describes the timing-out request.

2. Owner cancelled while parked on `_http_bridge_lock` inside `_enforce_http_bridge_capacity_after_planned_closes` (`_service/http_bridge/mixin.py:1347`, still outside the `except BaseException` cleanup). Repro: one idle LRU session, `max_sessions=1`, hold the lock, release the bounded close so the owner blocks on the lock acquire, then call `_abort_http_bridge_inflight_creation_locked` exactly as a timing-out waiter does. Result: owner exits with `CancelledError`, `_create_http_bridge_session` never runs, the done marker with a dead owner stays in `_http_bridge_inflight_sessions`, and the next request is rejected in **0.27 ms**; only `http_bridge_activity_snapshot_nowait` reclaims it. This contradicts the delta's own text (`spec.md:44-46`: "the proxy MUST remove or settle that in-flight marker"). CodeRabbit flagged the same lines (1343-1348) on 09-05. Fix: cover every await after marker publication with the `_fail_http_bridge_inflight_session_creation` path, and add the regression.

**Smaller items**
- `test_rejected_creator_does_not_release_the_registered_winners_durable_row` (#1695) is deleted but `_raise_if_http_bridge_creation_superseded` (`helpers.py:1686`, called from `mixin.py:1547`) and `release_durable_session=not superseded` (`helpers.py:1814`) are still reachable; keep a #1695 regression or state why the scenario is unreachable now.
- `helpers.py:939` `except BaseException: return True` in `_wait_for_http_bridge_retained_owner` also swallows `KeyboardInterrupt`/`SystemExit` propagated from the owner task; `CancelledError` is already handled above, so `except Exception` is sufficient.
- Scope: the synthesized turn-state provenance plumbing now spans `api.py`, `streaming.py`, `session_registry.py`, `protocol.py`, `support.py` (new `_HTTPBridgeSessionKey.synthesized_turn_state` with `compare=False`, new `synthesized_downstream_turn_state_aliases` set, new `downstream_turn_state_synthesized` request-state field). That is well past the "marker-lifecycle slice" this PR was carved out as; consider splitting it so the admission-recovery change can be reviewed on its own.

Soju06:
Integrated current main at `1fcd9399`. Fixed the removed admission-setting test seam and preserved the late cancellation-resistant owner regression. Threaded the service scheduler through retained/aborted owner waits instead of raising timing allowances. Preserved the existing OpenSpec scenario identity during reconciliation.

Validation: 29 focused owner/cancellation/capacity tests pass, plus full lint/typing, architecture/cancellation/timing/settings guards and strict/canonical specs.

Still held for the two substantive cases in the previous maintainer review: a later cross-key request can inherit a retained aborted marker's 429 immediately, and cancellation while reacquiring capacity after planned closes needs marker cleanup proof. The integration/seam fixes above do not establish those invariants. #1881 remains closed for its already-landed work; this PR is the outstanding lifecycle follow-up, not a reason to close it as a duplicate.

@codex review

PR #2086, opened 2026-09-04
fix(proxy): allow prefix-settled replay outputs

Komzpa:
Consolidated into [Soju06/codex-lb PR 2088](https://github.com/Soju06/codex-lb/pull/2088), which includes the prefix-settled tool-output replay change and its tests/spec alongside the bridge recovery that consumes it. The original patch content is retained. Closing this duplicate review surface.

### DIFF
--- a/app/modules/proxy/_service/http_bridge/streaming.py
+++ b/app/modules/proxy/_service/http_bridge/streaming.py
@@ -265,6 +265,7 @@ from app.modules.proxy.helpers import (
 from app.modules.proxy.replay_safety import (
     AccountNeutralReplayProjection,
     project_responses_input_for_account_neutral_fresh_replay,
+    responses_input_suffix_has_response_owned_prefix_settling_output_ids,
     responses_input_suffix_matches_pending_tool_calls,
     responses_input_suffix_retains_prior_output,
     responses_payload_is_account_neutral_fresh_replay,
@@ -397,6 +398,13 @@ class _VerifiedDurableFullResend:
         ):
             return None
         input_items = cast(list[JsonValue], payload.input)
+        pending_tool_calls = durable_lookup.latest_pending_tool_calls
+        if pending_tool_calls is not None and responses_input_suffix_has_response_owned_prefix_settling_output_ids(
+            input_items,
+            stored_count=stored_count,
+            pending_tool_calls=pending_tool_calls,
+        ):
+            return None
         replay_projection = project_responses_input_for_account_neutral_fresh_replay(
             input_items,
             stored_count=stored_count,
@@ -405,7 +413,6 @@ class _VerifiedDurableFullResend:
             # the exact-manifest check rejects response-owned messages.
             preserve_developer_message_ids=True,
         )
-        pending_tool_calls = durable_lookup.latest_pending_tool_calls
         if replay_projection is None:
             return None
         safe_fresh_context = responses_input_suffix_retains_prior_output(
@@ -1563,8 +1570,16 @@ class _HTTPBridgeStreamingMixin:
                 or not isinstance(payload.input, list)
             ):
                 return None, None, False
+            raw_input_items = cast(list[JsonValue], payload.input)
+            pending_tool_calls = lookup.latest_pending_tool_calls
+            if pending_tool_calls is not None and responses_input_suffix_has_response_owned_prefix_settling_output_ids(
+                raw_input_items,
+                stored_count=stored_count,
+                pending_tool_calls=pending_tool_calls,
+            ):
+                return stored_count, lookup.latest_input_full_fingerprint, False
             replay_projection = project_responses_input_for_account_neutral_fresh_replay(
-                cast(list[JsonValue], payload.input),
+                raw_input_items,
                 stored_count=stored_count,
                 # Classification only: inline Responses-Lite developer IDs
                 # must remain visible until the exact-manifest check rejects
--- a/app/modules/proxy/replay_safety.py
+++ b/app/modules/proxy/replay_safety.py
@@ -270,7 +270,7 @@ def responses_input_items_are_self_contained_fresh_replay(input_items: list[Json
         if not _input_item_has_only_known_fields(item, item_type):
             return False
         call_id_value = item.get("call_id")
-        call_id = call_id_value if isinstance(call_id_value, str) and call_id_value else None
+        call_id = cast(str, call_id_value) if _is_nonblank_string(call_id_value) else None
         if item_type in _TOOL_CALL_TYPES:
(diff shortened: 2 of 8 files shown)


Komzpa:
@codex review

Komzpa:
@codex review

Soju06:
Thanks for splitting this out with a spec delta as requested on #1881. CI is green on 52d3b5b, the Codex/CodeRabbit threads are addressed, and the branch still merges cleanly against current main (no overlap with the R19 merges so far). Three things block merge:

1. **The raw response-owned-ID guard is bypassable for Responses-Lite prefixes.** `responses_input_suffix_has_response_owned_prefix_settling_output_ids` (`app/modules/proxy/replay_safety.py:471`) calls `_direct_tool_call_prefix_state` without the projection's `canonical_lite_developer_index`; both call sites (`app/modules/proxy/_service/http_bridge/streaming.py:402` and `:1575`) pass nothing. For a stored prefix `[additional_tools bundle, canonical developer message, user, function_call call-1]` the prefix walk returns `None` on the developer message, the guard returns `False`, and both callers fall through to the projected path where `project_responses_input_for_account_neutral_fresh_replay` strips the output `id` before the prefix-settled proof runs. Reproduced on this head: `_verify_durable_full_resend` returns a proof for suffix `[{"type": "function_call_output", "id": "fc_output_response_owned", "call_id": "call-1", "output": "result"}]` with that Lite prefix and manifest `{"call-1": "function_call"}`. That is exactly the shape the new requirement says MUST fail, and `test_verified_durable_full_resend_rejects_response_owned_prefix_settling_output_id` (`tests/unit/test_proxy_http_bridge.py:14106`) only covers the non-Lite prefix. Either thread `canonical_lite_developer_index` from the classification projection into the raw guard (compute the projection first, then run the guard on the raw items with its index), or check the IDs on the raw suffix items directly without re-walking the prefix. Please add the Lite-prefix regression in both `test_replay_safety.py` and the `_verify_durable_full_resend` test.

2. **Reachability of the admitted shape is not shown.** The durable manifest is built from response *output* items (`_durable_pending_tool_call_manifest`, `app/modules/proxy/_service/http_bridge/upstream_events.py:976`, registered at `:3704`), while `latest_input_item_count`/`latest_input_full_fingerprint` are that same request's *input*, and all three are written together per response id in `register_owned_alias` (`app/modules/proxy/durable_bridge_repository.py:3625-3661`, replaced whenever the response id changes). Under normal upstream behavior a manifest call therefore cannot already be pending inside its own fingerprint-verified prefix; the proposal's "can force owner-bound recovery" has no trigger attached. Since this loosens a fail-closed certification (flagged as risk-sensitive on #1881), please document the concrete client sequence or log that produces `prefix_pending == manifest`, ideally as a bridge-level test that fails on main for that sequence, not only the helper-level `responses_input_suffix_matches_pending_tool_calls` cases.

3. **Spec delta should be MODIFIED, not ADDED.** `openspec/specs/responses-api-compat/spec.md:4782` ("Historical output remains mandatory": matching output missing -> exact manifest proof fails), under the requirement at `:4717`, conflicts with the new requirement when the missing output arrives in the suffix. Please express this as a MODIFIED delta on that requirement so the two readings are reconciled instead of coexisting.

Minor: the exact-manifest path on main still relies on projection to strip output `id`s, so the new raw guard enforces the property only in prefix-settled mode. If the property matters, it should apply to both modes; if it does not, the ~40-line guard plus two call sites can go.

Soju06:
Status check on 52d3b5b (unchanged since 09-04): the three points from the 09-05 comment are still open, and I re-verified them against current main (e987c56c8, which includes this round's merges). The branch still merges cleanly (0 conflicts on the merge tree) and lint/architecture guards pass, so no rebase is needed — the fixes below are all that stands between this and merge. The "codex: ok" label does not clear them; Codex reviewed the non-Lite shape only.

1. **Lite-prefix bypass, reproduced again on the merged tree.** With stored prefix `[additional_tools bundle, {"type":"message","role":"developer",...}, user, function_call call-1]`, manifest `{"call-1": "function_call"}` and suffix `[{"type":"function_call_output","id":"fc_output_response_owned","call_id":"call-1","output":"result"}]`, `_verify_durable_full_resend` returns a proof, while the same suffix on the plain prefix is correctly rejected. Cause: the new `responses_input_suffix_has_response_owned_prefix_settling_output_ids` (`app/modules/proxy/replay_safety.py`) accepts `canonical_lite_developer_index` but both call sites in `app/modules/proxy/_service/http_bridge/streaming.py` (`_VerifiedDurableFullResend`, ~L402, and the classification path, ~L1575) call it with only `stored_count`/`pending_tool_calls`. `_direct_tool_call_prefix_state` then sees `occupies_canonical_lite_position=False` at index 1, `historical_interleave_is_bounded` needs `len(pending_calls) == 1` while pending is still empty, the walk returns `None`, the guard returns `False`, and control falls through to the projected input where `project_responses_input_for_account_neutral_fresh_replay` has already stripped the `id`. Concrete fix: compute the projection first (as main already does) and pass `replay_projection.canonical_lite_developer_index` into the raw guard — the index is anchored on the original position (bundle at 0 and developer message at 1 both survive projection), so it is valid against the raw items. Alternatively drop the prefix re-walk from the guard and simply check `"id" in item` on raw suffix outputs whose `call_id` is in the manifest. Please add the Lite-prefix regression next to `test_verified_durable_full_resend_rejects_response_owned_prefix_settling_output_id` in `tests/unit/test_proxy_http_bridge.py` and a matching case in `tests/unit/test_replay_safety.py`.

2. **Reachability is still unshown.** `_durable_pending_tool_call_manifest` (`app/modules/proxy/_service/http_bridge/upstream_events.py:972`) is built from the response's output items and is written together with `latest_input_item_count`/`latest_input_full_fingerprint` for the same response id in `register_owned_alias` (`app/modules/proxy/durable_bridge_repository.py`), so a manifest call cannot normally already sit inside its own fingerprint-verified prefix. Since this loosens a fail-closed certification, please attach the concrete client sequence (or a bridge-level test that fails on main) that produces `prefix_pending == manifest`.

3. `openspec/changes/allow-prefix-settled-tool-output-replay/specs/responses-api-compat/spec.md` is still `## ADDED Requirements`; it conflicts with the existing scenario "Historical output remains mandatory" at `openspec/specs/responses-api-compat/spec.md:4801` on current main. Please express it as a MODIFIED delta on that requirement.

Note: #2084 and #2088 (both still open) touch the same `streaming.py`/`test_proxy_http_bridge.py` regions; whichever lands first will need the other rebased. None of the PRs merged this round conflict with this branch.

PR #2096, opened 2026-09-05
fix(proxy): preserve anonymous output response ownership

nurazem:
## Summary

When response A is streaming and request B awaits `response.created` on the same upstream connection, A's output events may omit `response_id`. The anonymous matcher assigns those events to B because B is the sole unresolved request. A then receives its correctly identified terminal event without its answer.

Prefer the unique started response for anonymous output events in both the HTTP bridge and WebSocket paths. Preserve existing pre-created output handling when no response has started, and leave metadata/error matching unchanged.

## Type of change

- [x] `fix:` — bug fix

No linked issue; reproduced directly against current `main`.

## OpenSpec

- [x] Includes an OpenSpec change.
- [x] Preserves upstream event formats; only request ownership changes.

Change: `openspec/changes/archive/2026-09-04-fix-anonymous-output-ownership/`.

## Reproduction

```text
pending: A(response_id=resp_A), B(response_id=None)
upstream: response.output_text.delta(item_id=msg_A, delta="answer A")
before:  A receives 0 events; B receives 1
with fix: A receives 1 event; B receives 0
```

Eight bridge-level regressions cover message items, text, content parts, function-call arguments, reasoning summaries, and refusals. Additional controls cover draining owners and multiple started responses.

## Validation

- Eight new output-delivery regressions fail on unmodified main (`b2c5ffc`).
- Affected suites: **1,174 passed, 1 failed**. The failure, `test_stream_via_http_bridge_fails_closed_before_file_affinity_when_previous_response_owner_misses`, also fails on unmodified main with `no such table: file_account_pins`.
- Ruff and targeted `ty` checks pass.
- OpenSpec validation: **58 passed, 0 failed**; change validated and archived.
- Full local CI passed frontend checks, Python lint/type checks, and Rust clippy, then stopped at `rust-audit` because `cargo deny` is not installed. Later full-CI stages were not run.

```bash
uv run pytest tests/unit/test_http_bridge_cancel_drain.py \
  tests/unit/test_proxy_http_bridge.py \
  tests/unit/test_websocket_terminal_cancellation.py \
  tests/integration/test_proxy_websocket_responses.py -q
```

No new settings, dependencies, database changes, or dashboard changes.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed routing of anonymous upstream output events so they remain associated with the uniquely identified started response, including responses that are draining.
  - Prevented output from being assigned to requests still awaiting response initialization.
  - Left ownership unresolved when multiple started responses could match, avoiding incorrect delivery.

- **Tests**
  - Added coverage for text, tool, content, reasoning, refusal, and function-call output events, including concurrent and draining scenarios.

### DIFF
--- a/app/modules/proxy/_service/http_bridge/upstream_events.py
+++ b/app/modules/proxy/_service/http_bridge/upstream_events.py
@@ -2599,6 +2599,7 @@ class _HTTPBridgeUpstreamEventsMixin:
             elif response_id is None:
                 matched_request_state = _match_websocket_request_state_for_anonymous_event(
                     session.pending_requests,
+                    event_type=event_type,
                     prefer_previous_response_not_found=is_previous_response_not_found_event
                     or is_missing_tool_output_event,
                     previous_response_id_hint=previous_response_id_hint,
--- a/app/modules/proxy/_service/websocket/helpers.py
+++ b/app/modules/proxy/_service/websocket/helpers.py
@@ -1661,6 +1661,7 @@ def _match_websocket_request_state_for_anonymous_event(
     pending_requests: deque[_WebSocketRequestState],
     *,
     prefer_previous_response_not_found: bool,
+    event_type: str | None = None,
     previous_response_id_hint: str | None = None,
     error_message: str | None = None,
     allow_unanchored_previous_response_error: bool = False,
@@ -1673,6 +1674,33 @@ def _match_websocket_request_state_for_anonymous_event(
             error_message=error_message,
             allow_unanchored_previous_response_error=allow_unanchored_previous_response_error,
         )
+
+    # Output belongs to an already-created response. A younger pipelined
+    # request may still lack its response ID while the active response emits
+    # item/text/tool events without a response_id field.
+    if (
+        event_type is not None
+        and event_type.startswith("response.")
+        and event_type
+        not in {
+            "response.created",
+            "response.queued",
+            "response.in_progress",
+            "response.completed",
+            "response.failed",
+            "response.incomplete",
+        }
+    ):
+        started_requests = [
+            request_state
+            for request_state in pending_requests
+            if request_state.response_id is not None
+            and (_http_bridge_request_counts_against_queue(request_state) or request_state.draining_until_terminal)
+        ]
+        if started_requests:
+            return started_requests[0] if len(started_requests) == 1 else None
+        # Preserve supported pre-created output/reasoning preludes when no
+        # started response can own the event.
 
     visible_requests = [
         request_state for request_state in pending_requests if _http_bridge_request_counts_against_queue(request_state)
--- a/app/modules/proxy/_service/websocket/mixin.py
+++ b/app/modules/proxy/_service/websocket/mixin.py
@@ -5374,6 +5374,7 @@ class _WebSocketMixin:
             elif response_id is None:
                 request_state = _match_websocket_request_state_for_anonymous_event(
                     pending_requests,
+                    event_type=event_type,
                     prefer_previous_response_not_found=is_previous_response_not_found_matching_event
                     or is_missing_tool_output_event,
                     previous_response_id_hint=previous_response_id_hint,
--- a/openspec/changes/archive/2026-09-04-fix-anonymous-output-ownership/context.md
+++ b/openspec/changes/archive/2026-09-04-fix-anonymous-output-ownership/context.md
@@ -1,0 +1,9 @@
+## Reproduction
+
+Two requests share an upstream connection. Response A has an ID; request B awaits response.created. Output events for A carry item_id but no response_id. The anonymous matcher chooses B because it is the only unresolved request. The final response.completed event still resolves to A by response ID, so A completes without its streamed output.
+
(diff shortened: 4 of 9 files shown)


Soju06:
Verified the defect on current main (a0d94a46a): with `A(response_id=resp_A)` streaming and `B(response_id=None, awaiting_response_created=True)` on the same bridge session, `_match_websocket_request_state_for_anonymous_event` (`app/modules/proxy/_service/websocket/helpers.py:1660` on your head) returns B via the single-unresolved-visible branch, so A's anonymous output deltas are delivered to B. That window is exactly what the `Semaphore(1)` create gate allows, since it is released on `response.created` (`http_bridge/upstream_events.py:4033`), and the draining variant leaks a cancelled response's tail into a same-prompt retry. Your tests union-merged onto main fail with `assert 0 == 1` on unmodified main and pass with the fix; `test_http_bridge_cancel_drain.py`, `test_proxy_http_bridge.py`, `test_websocket_terminal_cancellation.py`, `tests/integration/test_proxy_websocket_responses.py` and `tests/simulation` pass on the merged tree (1324 passed), and the proxy architecture / timing-seam / cancellation-safety guards pass. The fix itself looks right and scoped.

Four things before this can merge:

1. **Rebase.** The only conflict is `tests/unit/test_http_bridge_cancel_drain.py` against 4b951e590 (#2103), which also appended tests at EOF. Keep both blocks; no semantic interaction. None of the PRs merged in this round (#2078, #2079, #2092, #2114, #2143, ...) touch the files you change. Base onto a0d94a46a or newer.

2. **Archive attribution on the WebSocket transport (CodeRabbit's outside-diff finding is correct).** `_websocket_archive_request_state_for_payload` (`websocket/mixin.py:680`, call at :733) still calls the matcher without `event_type`, while the delivery path at `mixin.py:5375` now passes it. On the WS transport, archive attribution is computed independently via `_websocket_archive_request_id_for_message` (`mixin.py:767`, used at :1003 and :1216), so after this PR A's deltas are delivered to A but archived under B's `archive_request_id`. Pass `event_type=event_type` there and add an assertion that the archived request id for an anonymous `response.output_text.delta` is A's when B is awaiting `response.created`. (The HTTP bridge path is fine: `_archive_http_bridge_upstream_text` reuses `matched_request_state`.)

3. **Contributors attribution.** The only red job in the non-cancelled CI run is `Contributors attribution`; add a `nurazem` entry to `.all-contributorsrc` (`code`, `bug`, `test`).

4. **Spec placement.** Two parts:
   - `openspec/specs/proxy-architecture/spec.md` is the structural fitness-gates spec (see its Purpose). Event-ownership semantics for the bridge live in `openspec/specs/responses-api-compat/spec.md` alongside the other ownership requirements (e.g. "Hard continuity owner lookup fails closed", "Compact requests preserve scoped turn-state ownership"). Please move the new Requirement and the change's spec delta there; `openspec validate --specs` still passes either way, so this is about keeping the SSOT navigable.
   - The change folder is committed directly under `openspec/changes/archive/2026-09-04-fix-anonymous-output-ownership/`. Active changes belong in `openspec/changes/<name>/`; archiving happens after merge. Please move it to `openspec/changes/fix-anonymous-output-ownership/` so `openspec validate --strict` picks it up as an in-flight change.

Optional: `_pop_terminal_websocket_request_state` (`helpers.py:1892`, matcher call at :1928) also reaches the anonymous matcher, but only for terminal events your filter excludes, so leaving it without `event_type` is fine — a one-line comment saying so would prevent the next reader from "fixing" it.

Soju06:
Closing as superseded — thank you, the diagnosis was correct and it was verified before this call.

Re-checked against current `origin/main` (`ca243dec4`): the defect you described is fixed by the same mechanism in #2357 (commit `976d20b71`, merged 2026-09-11).

- `app/modules/proxy/_service/websocket/helpers.py:1860` now defines `_is_response_output_event()`, and the matcher at `:1886-1895` returns the single pending request that already has a `response_id` ("visible or draining") for anonymous `response.*` output frames — placed ahead of the `unresolved_visible_requests` branch at `:1914` that used to hand A's deltas to B.
- Both call sites you patched pass the event type on main: `http_bridge/upstream_events.py:2703-2711` and `websocket/mixin.py:5495`.
- The archive-attribution gap that was blocker #2 in the 2026-09-08 review of this PR is also closed: `websocket/mixin.py:750-757` (`_websocket_archive_request_state_for_payload`) now passes `event_type=event_type`, so delivery and archive select the same owner.
- Your parametrized control `test_anonymous_output_owner_excludes_waiting_requests` passes against main's matcher as written, for both `draining=True` and `draining=False`: with A(`response_id=resp_A`) and B(`response_id=None`) there is exactly one created request and A is returned.

Two deliberate differences are worth knowing about, in case either still matters to you. Your version returned `None` whenever more than one started response existed, and excluded `response.queued` / `response.in_progress` from the output class. #2357 was narrowed during its review to the opposite rule — exactly one created owner wins, every other count falls through to the pre-existing visible/draining rules — and its output class excludes only `response.completed` / `response.failed` / `response.incomplete`. If you can produce a case where the multi-started or `queued` / `in_progress` shape still misroutes on current main, please open a focused follow-up with a test that fails there; we would take it.

The eight-case regression matrix in this PR was an accurate, independent reproduction of #2350 and it held up under verification. Sorry it got overtaken.

PR #2101, opened 2026-09-05
fix(proxy): add native history and notes routes with account-local ownership

Komzpa:
## Summary

Native Codex history and notes calls (`alpha/history/v2/*`, `alpha/notes/v2/*`) have no route through codex-lb, so the extension fails at the proxy. This PR registers the ten authenticated POST routes on the existing Codex control proxy, forwards bodies and native headers unchanged, and keeps each session's account-local history/notes state on one upstream account. Ordinary Responses routing is not changed.

## Type of change

- [x] `fix:` — bug fix (missing routes; no behavior change for existing traffic)

Linked issue: none. Related alternative: #2102 (durable ownership tables + cross-account history fan-out + proxy ciphertext envelope). The two designs are mutually exclusive; this one adds no schema, no migration, no wire-format change, and no new setting.

## OpenSpec

- [x] This PR touches a codex-faithful path (native request/response shape) and preserves upstream-equivalent behavior: the body, `x-openai-encrypted-tool-arguments`, `x-openai-tool-output-truncation-policy`, and the upstream response are relayed byte-for-byte.

Change directory: `openspec/changes/proxy-native-history-notes-routes/` (new capability `native-history-notes-proxy`, delta on `upstream-proxy-routing`).

## Changes

- `app/modules/proxy/api.py`: ten `POST /backend-api/codex/alpha/{history,notes}/v2/*` routes on `_codex_control_proxy`. For these paths the plaintext body `context.session_id` is read as the affinity identity and cross-account retry is disabled. A body that is not a JSON object gets `400` (same as thread-goal bodies); a JSON object without a usable identity is forwarded once, unpinned — the upstream protocol stays authoritative for payload validation.
- `app/modules/proxy/affinity.py`: `_sticky_key_for_codex_control_request(..., body_session_id=)` builds a hard `CODEX_SESSION` policy with source `history_session` in its own namespaced key (`_history_session_selection_key`), no legacy raw-row lookup, no cap spillover, seeded from the soft process-session owner (`_codex_session_selection_key(session_id)`). `_CodexSessionSource` now aliases the repository literal instead of duplicating it.
- `app/modules/proxy/sticky_repository.py`: `_ContinuitySource` gains `"history_session"`; it is the single owner of the literal (the test double in `tests/unit/test_proxy_load_balancer_refresh.py` now imports it too).
- `app/modules/proxy/_service/codex_control.py`: `codex_control_request(body_session_id=, allow_cross_account_retry=)`; with retry disabled the selected account's upstream error is returned as-is and health is recorded on that account.
- `app/modules/proxy/affinity_observation.py`: request logs label the source as `history_session`.

## Review points from the previous round, and how each was settled

1. **`"body_session"` had no producer but granted cap-spillover mobility.** Removed entirely (literal, `legacy_selection_key`, `preferred_owner_sticky_inputs`, and the `bare_session_key` mobility set in `sticky_selection.py` are back to `main`). The only producer is the control route, and it emits `history_session`, which is hard and never spills.
2. **Marked Responses/compact were reclassified into hard fail-closed ownership.** Dropped. `_history_session_affinity` and the turn-metadata parser are gone; `_sticky_key_for_responses_request`, `_sticky_key_for_compact_request`, and `compact.py` are identical to `main`. Only the ten native routes pin. Quota rotation and #2166 overload rebinding keep working for every session; only the account-local history/notes calls fail loudly when their owner is unavailable. The first native call seeds from the soft process-session owner, so notes land where that session's Responses currently run without changing how Responses are routed. Spec scenarios: "History-marked Responses keep ordinary soft affinity", "First native call follows the process-session owner", "History owner is unavailable". Known limit, stated in the proposal's Non-goals: when Responses rotate across accounts, a native history query sees only the pinned account's store; recording ingest partitions and fanning out is #2102's design, not this one.
3. **`400` on a body without plaintext `context.session_id` vs today's `405`.** No live capture exists (checked the operator's request logs, container logs, and state directories: the extension has never reached this proxy because it is gated by the model catalog's `use_history_notes_extension`). The shape is settled by the client source instead: codex-rs `ext/history-notes/src/backend.rs` (rust-v0.153.x, same tree as the 0.155 build) always inserts a plaintext `"context": {"session_id", "current_agent_name"}` object into the JSON body and sends `x-openai-encrypted-tool-arguments: true` as a flag header on `history/v2/search_contents`, `notes/v2/search_contents`, `notes/v2/append_to_file`, `notes/v2/write_file` only — its test is literally `marks_encrypted_history_and_notes_arguments_without_changing_the_json_body`. The proxy therefore no longer rejects a JSON object that lacks the identity: it is forwarded once, unpinned, exactly like a thread-goal body without `threadId`. Only a non-JSON-object body gets `400`, matching `_thread_goal_payload_from_request`. The client maps every non-2xx to one generic tool error, so the status choice is invisible to real clients.
4. **ty red.** `_ContinuitySource` is the single literal owner; `affinity._CodexSessionSource = _ContinuitySource`. ty: 0 diagnostics.
5. **Overlap with #2102.** Compared in detail. #2102 delivers the same ten routes as a side effect of a pooled-history architecture (2 tables, 5 migrations, a ledger-recovery hook, a 4096-entry process cache, up-to-32-account fan-out, a proxy-minted `codex-lb-context-v1:` ciphertext in the client's `encrypted_output` slot that is rewritten into N native parts on the next Responses turn, replay-safety widening so marked inference rotates, ~60 files / ~4,750 diff lines). Its current head has CI failures (integration-core-3, Rust workspace, contributors attribution). This PR is the minimal design the maintainer said is "much smaller, which is in its favor if points 3 and 4 are resolved"; both are resolved above.

## Simplicity

No new setting, README section, `.env.example` entry, dashboard item, migration, or default change. The feature is zero-config: routes exist, and the only state is a sticky-session row in the existing table.

## Test plan

```
uv run pytest tests/unit/test_proxy_utils.py -q                          # 1415 passed
uv run pytest tests/unit/test_proxy_load_balancer_refresh.py tests/unit/test_affinity_observation_source.py -q   # 103 passed, 3 skipped
uv run pytest tests/integration/test_proxy_api_extended.py -q -k 'native_history or history_session or codex_control_json'   # 16 passed
uv run pytest tests/integration/test_daybreak_capability_routes.py -q   # 133 passed
uv run pytest tests/integration/test_proxy_compact.py -q                # 55 passed
uv run ruff check . && uv run ruff format --check .                      # clean
uv run ty check                                                          # 0 diagnostics (was 5 on the rebased head)
make architecture-check                                                  # all five guards pass
openspec validate --type change --strict --no-interactive -- proxy-native-history-notes-routes   # valid (openspec 1.11.0)
openspec validate --specs                                                # 65 passed, 0 failed
```

Coverage: all ten routes preserve body and both native headers (parametrized); missing identity is forwarded once with no session affinity and no failover; malformed body 400; unknown `alpha/notes/v2/*` name stays 405; owner upstream error returns without touching account B; first native call follows the process-session owner and persists the `history_session` row; a history-marked Responses turn still rotates off a paused owner (ownership class unchanged); a native call with a paused owner fails without selecting B and without replacing the stored owner; auth is checked before routing for these paths; unit canaries prove the `history_ingest_requested` marker (header or `client_metadata` carrier) yields the same policy shape as an unmarked turn for both Responses and compact.

## Screenshots / output

Request shape (from codex-rs `ext/history-notes/src/backend.rs`, forwarded unchanged):

```
POST /backend-api/codex/alpha/notes/v2/write_file
x-openai-tool-output-truncation-policy: {"Bytes":4000}
x-openai-encrypted-tool-arguments: true
content-type: application/json

{"path":"notes.md","contents":"...","context":{"session_id":"<thread-uuid>","current_agent_name":"<agent>"}}
```


Soju06:
Checked bc9dfedf2 against current main (a0d94a46a). The premise holds: main has no `alpha/history/v2/*` or `alpha/notes/v2/*` routes, and codex-cli 0.153.4 does carry all ten paths, the `history_ingest_requested` turn-metadata field, and the two `x-openai-*` headers. It still merges cleanly onto main (none of this round's merges — #2078, #2079, #2092, #2114, #2143, #2148 etc. — touch these files), and the architecture/timing-seam/cancellation guards pass on the merged tree. A few things before this can leave draft:

1. **CI is red on `Type check (ty)`** (a real failure on the current head, not a cancelled run). `_CodexSessionSource` was widened in `app/modules/proxy/affinity.py:32`, but `_ContinuitySource` in `app/modules/proxy/sticky_repository.py:28` (parameter type of `get_account_id_and_abandonment`, `:176`/`:192`) is still `Literal["session_header", "thread_header", "turn_state"]`, so ty rejects `continuity_source=sticky_source` at `app/modules/proxy/_load_balancer/sticky_selection.py:383` and `app/modules/proxy/load_balancer.py:799`. Either import the affinity alias there or widen it in lockstep.

2. **`"body_session"` is dead.** Nothing constructs an `_AffinityPolicy` with `codex_session_source="body_session"` — the control path in `_sticky_key_for_codex_control_request` (`affinity.py:576`) emits `"history_session"`. Yet `"body_session"` is added to the Literal, to the `legacy_selection_key` / migration checks (`affinity.py:93`, `:107`, `:157`), and to the `bare_session_key` mobility set in `sticky_selection.py:451`. That last one grants cap-spillover mobility to a source with no producer. Please remove it or wire a real producer.

3. **This is a routing behavior change, not just a route add.** `_history_session_affinity` (`affinity.py:468`) re-classifies every Responses/compact request carrying `history_ingest_requested:true` from soft thread/process locality (`thread_header` → PROMPT_CACHE) into a hard `CODEX_SESSION` owner keyed by the client-controlled session id, taking precedence over `turn_state`, with no cap spillover and fail-closed `hard_affinity_saturated` when the owner is paused or exhausted — your own test asserts `response.failed` after pausing the owner. For opted-in sessions that disables the balancer's quota rotation and the #2166 overload rebinding for the whole Codex process lifetime. That needs an explicit owner call and a spec section explaining why hard fail-closed is preferable to soft-pinning Responses and letting only the account-local history/notes calls fail loudly. Please state it in the PR body rather than under a "preserve ownership" title.

4. **Body shape.** `_native_history_notes_session_id` (`app/modules/proxy/api.py:960`) hard-requires plaintext `context.session_id` and returns 400 otherwise. codex sends these with `x-openai-encrypted-tool-arguments`; if the context envelope is not plaintext in practice, every call becomes a 400 instead of today's 405. A captured real request (headers + body skeleton) from codex-cli 0.153.4 in the PR body would settle it.

5. **Overlap with #2102** (leosanxyz), which registers the same ten routes via a dedicated `codex_context_request` path with a required API key plus a DB ownership table + migration. The two designs are mutually exclusive; the owner needs to pick one before either proceeds. This one is much smaller, which is in its favor if points 3 and 4 are resolved.

Issue #1919, opened 2026-08-25
bug: Need to reauthenticate accounts with Advanced Account Security every single day


vitobotta:
### Pre-flight checklist

- [x] I searched existing issues and confirmed this is not a duplicate.
- [x] I am running a supported codex-lb version (see Releases).
- [x] I can reproduce the problem with the steps below.

### codex-lb version

1.23.0

### Deployment method

Docker (ghcr.io/Soju06/codex-lb)

### Client used against codex-lb

Codex CLI, Codex app (desktop / web)

### ChatGPT account plan(s) involved

Pro

### Model(s) involved

gpt-5.6-sol, gpt-5.6-luna

### What happened?

Hi! First of all, huge thanks for this awesome project! It makes it easier to use two ChatGPT subscriptions without having to log out and log back in to change the subscription.

I am both a coder and a security researcher so I had to enroll in the Advanced Account Security otherwise OpenAI would have restricted me from using the models from security work.

Codex-lb works great while both accounts are authenticated, but I basically need to reauthenticate both account once a day or so. Is there a setting or something that I am missing or how can I solve this?

Thanks!

### What did you expect to happen?

I would expect the accounts to stay authenticated in Codex-lb.

### Steps to reproduce

```markdown
1. Start Codex-lb with Docker on my VPS 
2. Authenticate the ChatGPT accounts in Codex-lb
3. Configure Codex and Hermes to connect to Codex-lb via Tailscale
4. Use the models normally
5. After some time, around after a day or perhaps after some idle time, Codex-lb requires both accounts to be authenticated again
```

### Relevant logs

```shell

```

### Configuration / environment

```yaml

```

Soju06:
Thanks for the report and the kind words.

Here is what codex-lb does with the credentials on current `main`, so we can pin down where the daily re-login comes from:

- Every account's refresh token is exercised at least once every ~12-18h regardless of traffic: the Auth Guardian (enabled by default since 1.23.0, #1662) force-refreshes any active/paused account whose last refresh is older than 12h, on a 6h cycle (`app/core/auth/guardian.py`). A request that hits an upstream 401 also triggers a forced refresh and retry (`app/modules/proxy/_service/refresh.py`).
[...]

PR #2120, opened 2026-09-06
fix(auth): retain unexpired access on refresh preflight failure

Irvinwop:
## Summary

Preserve the current ordinary request when a freshness preflight discovers a
permanently invalid refresh credential but the persisted access token remains
unexpired. The actual upstream request still authenticates that token normally.

Refs #1919 — partial mitigation, not a claim to override Advanced Account
Security session expiry or eliminate all daily reauthentication. Builds on the
already-merged #1877; that change preserves subsequent request routing, while
this fixes the first request that discovers the refresh failure. The new
Responses regression fails on the unmodified current base before dispatch.

## Type of change

- [x] `fix:` — bug fix

## OpenSpec

- [x] Includes a verified, archived OpenSpec change and synchronized main spec.
- [x] Preserves upstream request/response and OAuth wire formats.

Change: `openspec/changes/archive/2026-09-06-preserve-access-on-refresh-preflight/`

## Changes

- Recover only an ordinary active-account preflight, never forced refresh.
- Require an explicit refresh-credential-only failure, fresh `reauth_required`
  row, matching canonical warning, and known future access-token expiry.
- Keep credentials, refresh timestamp and warning unchanged. No new probe.
- Keep recovery outside shared refresh work so forced callers still fail.
- Regression coverage includes native/OpenAI Responses paths, actual access
  rejection, expired/unknown tokens, account/session invalidation, and both
  ordinary-first and forced-first concurrent callers with real database sessions.

## Simplicity

- [x] Zero configuration; no new setting, dependency, schema or migration.
- [x] No required setup, README, dashboard layout or navigation changes.

## Test plan

```text
uv run pytest tests/unit/test_auth_manager.py tests/unit/test_auth_guardian.py tests/unit/test_auth_refresh.py tests/unit/test_auth.py tests/unit/test_proxy_load_balancer_refresh.py tests/unit/test_load_balancer.py tests/integration/test_token_refresh_claims.py tests/integration/test_auth_preflight.py tests/integration/test_proxy_responses.py tests/integration/test_auth_guardian_multi_replica.py -q --tb=short --show-capture=no --timeout=60
578 passed, 3 pre-existing skips

make lint
uv run ty check
git diff --check
passed (including architecture and cancellation checks)

npx --yes @fission-ai/openspec@1.11.0 validate preserve-access-on-refresh-preflight --strict
passed before archiving

npx --yes @fission-ai/openspec@1.11.0 validate --specs
58 passed
```

Global `validate --specs --strict` reports pre-existing placeholder-purpose
warnings in unrelated capabilities; the changed usage-refresh capability and
change validate strictly. The affected suites above were run, not the full
multi-platform/container CI matrix. No live account credentials or requests
were used in these tests, and no running installation was changed.

## Behavioral output

```text
Before: stale ACTIVE -> refresh_token_invalidated -> response.failed (no dispatch)
After:  stale ACTIVE -> same persisted warning -> stored access token -> response.completed
        (only when the upstream resource endpoint accepts the token)
Unchanged: actual access rejection -> failed forced refresh / existing failover
```

## Checklist

- [x] Conventional Commits title; linked related issue without overstating scope.
- [x] Regression reproduced before implementation; tests added.
- [x] Relevant local CI subsets, type checks, strict change and all-spec checks run.
- [x] Simplicity gates reviewed; CHANGELOG left unchanged.

## Summary by CodeRabbit

- **Bug Fixes**
  - Requests can continue using an unexpired access token when refresh credentials fail permanently, reducing unnecessary first-request failures.
  - Accounts are marked as requiring reauthentication while preserving the existing access and refresh credentials.
  - Forced refreshes, expired or unavailable access tokens, transient errors, and upstream authentication rejections continue to fail as expected.
  - Added coverage for Responses endpoints and concurrent refresh scenarios to ensure consistent behavior.

### DIFF
--- a/app/modules/accounts/auth_manager.py
+++ b/app/modules/accounts/auth_manager.py
@@ -36,6 +36,7 @@ from app.db.models import Account, AccountProxyBinding, AccountStatus
 from app.db.session import get_background_session
 from app.modules.accounts.refresh_claims import RefreshClaimCoordinatorPort, get_refresh_claim_coordinator
 from app.modules.proxy.account_cache import get_account_selection_cache, mark_account_routing_unavailable
+from app.modules.proxy.account_eligibility import account_access_token_expires_at
 
 
 class AccountsRepositoryPort(Protocol):
@@ -164,6 +165,19 @@ _TOKEN_REFRESH_CLAIM_POLL_SECONDS = 0.25
 # token; it must fail closed and surface the terminal state instead.
 _TERMINAL_REFRESH_STATUSES = frozenset({AccountStatus.REAUTH_REQUIRED, AccountStatus.DEACTIVATED})
 
+# These describe the refresh credential, not an upstream rejection of the
+# access token or termination of the account/session. Only ordinary preflight
+# callers may try a still-unexpired access token after these failures.
+_REFRESH_CREDENTIAL_FAILURE_CODES = frozenset(
+    {
+        "refresh_token_expired",
+        "refresh_token_reused",
+        "refresh_token_invalidated",
+        "invalid_refresh_token",
+        "invalid_grant",
+    }
+)
+
 
 _RefreshSingleflightKey: TypeAlias = tuple[str, str]
 
@@ -288,10 +302,34 @@ class AuthManager:
 
     async def ensure_fresh(self, account: Account, *, force: bool = False) -> Account:
         if force or (account.status != AccountStatus.REAUTH_REQUIRED and should_refresh(account.last_refresh)):
-            account = await _REFRESH_SINGLEFLIGHT.run(
-                _refresh_singleflight_key(self._encryptor, account),
-                lambda: self._run_refresh(account),
-            )
+            ordinary_active_preflight = not force and account.status == AccountStatus.ACTIVE
+            try:
+                account = await _REFRESH_SINGLEFLIGHT.run(
+                    _refresh_singleflight_key(self._encryptor, account),
+                    lambda: self._run_refresh(account),
+                )
+            except RefreshError as exc:
+                if (
+                    not ordinary_active_preflight
+                    or not exc.is_permanent
+                    or exc.code not in _REFRESH_CREDENTIAL_FAILURE_CODES
+                ):
+                    raise
+                # Keep recovery outside singleflight: a forced caller sharing
+                # this exchange must still fail after an upstream auth rejection.
+                # Re-read because the owned refresh task (or a peer) may have
+                # updated a different ORM instance. Never clear its warning.
+                latest = await self._repo.get_by_id_fresh(account.id)
+                if (
+                    latest is None
+                    or latest.status != AccountStatus.REAUTH_REQUIRED
+                    or latest.deactivation_reason != PERMANENT_FAILURE_CODES[exc.code]
+                ):
+                    raise
+                expires_at = account_access_token_expires_at(latest, self._encryptor)
+                if expires_at is None or expires_at <= time.time():
+                    raise
+                account = _adopt_account_row(account, latest)
         return await self._ensure_chatgpt_account_id(account)
 
     async def _run_refresh(self, account: Account) -> Account:
--- a/openspec/changes/archive/2026-09-06-preserve-access-on-refresh-preflight/context.md
+++ b/openspec/changes/archive/2026-09-06-preserve-access-on-refresh-preflight/context.md
@@ -1,0 +1,45 @@
+## Context and limits
+
+An OAuth refresh failure and an access-token rejection are distinct events.
+Existing upstream routing already permits warning-state accounts to serve
+ordinary requests until known access-token expiry. This change closes the
+first-request preflight gap without changing that policy or hiding the warning.
+
+For example, a stale imported credential can have a future access-token expiry
+while its refresh token returns `refresh_token_invalidated`. The first native
+Responses request should reach upstream with the stored access token, just as
+the next request would once it observes `reauth_required`.
+
+The recovery belongs outside the shared refresh task: an ordinary caller and
+a forced-refresh caller may share the same exchange but must not share a
+successful fallback. The forced caller may have already observed an upstream
+401, so its error must remain an error. The fresh database read also avoids
+depending on mutations to another caller's ORM snapshot.
+
+The persisted warning must match the refresh error's canonical reason. The
+terminal-status guard can synthesize a refresh error for an already-invalidated
+session; checking the stored reason prevents that synthesized error from
+accidentally authorizing this narrower preflight recovery.
+
+Only explicit refresh-credential failures qualify: `refresh_token_expired`,
+`refresh_token_reused`, `refresh_token_invalidated`, `invalid_refresh_token`,
+and `invalid_grant`. Session/account invalidation and transport or persistence
+failures retain their existing handling. No extra usage or model probe is sent.
+The actual upstream request remains the authority on token acceptance; a JWT
+expiry is only a local pre-dispatch check, not proof of authentication.
+
+This does not establish that Advanced Account Security caused the reported
+refresh rejection, nor that any particular previously rejected credential
+remained valid. It cannot prevent reauthentication when OpenAI actually expires
+or revokes the access token.
+
+## Verification
+
+- The native Responses regression failed on the unmodified base with
+  `response.failed` before dispatch and passes with this change.
+- The affected auth, guardian, refresh-claim, balancer and Responses suites
+  passed: 578 passed, 3 pre-existing skips.
+- Repository lint, formatting, type, architecture and cancellation checks passed.
+- Strict change validation and ordinary all-spec validation passed (58 specs).
+  Global strict validation reports existing placeholder-purpose warnings in
+  unrelated capabilities; they are not changed by this patch.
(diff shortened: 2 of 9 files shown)


Soju06:
Triage (round 20, head `e4718ddef`, base `origin/main` @ `a0d94a46a`). Merges cleanly against current main; none of this round's merges (#2078, #2079, #2092, #2114, #2143, #2148, dependabot batch) touch `auth_manager.py`, so no conflict. Verdict: **merge-ready once the P2 below is addressed and attribution is added**. The Responses-preflight fix itself is right: an ordinary ACTIVE caller whose refresh credential died should not throw away a still-valid access token, and keeping recovery outside the singleflight so forced sharers still fail (`app/modules/accounts/auth_manager.py:317-319`) is the correct fencing.

**P2 — recovery re-read autoflushes the caller's dirty attached `Account` as an unguarded status write, then decides on the row it just overwrote**

`ensure_fresh` passes the caller's `account` object into the refresh. On a permanent failure `_handle_permanent_refresh_failure` CAS-writes REAUTH_REQUIRED through the owned session and then mutates the caller's object (`auth_manager.py:1100-1102`, `account.status = status; account.deactivation_reason = reason`). The new re-read at `auth_manager.py:322` goes through `self._repo.get_by_id_fresh`, which is a plain `session.execute(select(...))` (`repository.py:213-215`) — so when `account` is attached to that same session (every non-proxy caller: `get_accounts_context` -> `AccountsService` probe/reset-credit/consume, `/v1/reset-credit`, `rate_limit_reset_credits`, `limit_warmup`, automations), autoflush emits `UPDATE accounts SET status=?, deactivation_reason=? WHERE id=?` with none of the `expected_*` CAS predicates that `update_status_if_current` (`auth_manager.py:1091-1098`) exists to enforce, then SELECTs back its own write.

Reproduced in the merged worktree with an attached account and a peer writing `REAUTH_REQUIRED` + `app_session_terminated` between the CAS and the re-read: outcome `('recovered', REAUTH_REQUIRED, 'Refresh token was revoked - re-login required')` — the DB reason was clobbered to the refresh-credential string and recovery proceeded against a session-invalidated account, which the PR's own spec forbids (`openspec/specs/usage-refresh-policy/spec.md`, "session invalidation MUST NOT qualify" / "MUST NOT mask a different session-invalidated warning"). The same window clobbers a concurrent dashboard re-auth (ACTIVE + rotated tokens) back to REAUTH_REQUIRED on status/reason, leaving a healthy account stuck in warning (REAUTH_REQUIRED skips ordinary refresh; the Guardian's forced refresh then hits `_TERMINAL_REFRESH_STATUSES`). On PostgreSQL the autoflushed UPDATE also holds the row lock for the rest of the request, blocking peers' CAS on that account.

This is PR-introduced: on main `ensure_fresh` raises here and `get_session`/`get_background_session` roll back on exception, so the dirty attributes were discarded. The PR both flushes them and continues to a success path callers commit. Proxy stream/bridge callers are unaffected (detached `_clone_account` snapshots into a fresh repo session), so the motivating Responses case is fine — but the helper is on the shared `AuthManager` surface.

Fix options: do the re-read without flushing caller state — `session.expire(account)` before the re-read (discarding the pending status/reason), or wrap it in `sync_session.no_autoflush`, or read via `self._refresh_repo_factory()` and stop mutating the caller's object in the failure path. Please add a test with an attached account plus an intervening peer status change (both the session-terminated and re-auth variants). `codex review --base origin/main` independently flagged the same path.

**P3 (design/scope, no blocker)**

- Scope creep beyond the body/spec: non-proxy callers (dashboard reset-credit read/consume, probe, `/v1/reset-credit`, `rate_limit_reset_credits`, automations ping, `limit_warmup`) that previously surfaced "Account credentials could not be refreshed" now silently proceed on the stored access token of a REAUTH_REQUIRED row. Untested and undocumented, and these are exactly the attached-session callers where the P2 bites. Either gate recovery to the proxy request path or document + test these callers.
- Recovery depends on which code the *original* failure had, not on account state: a claim waiter / stale-ACTIVE caller hitting `_terminal_status_refresh_error` gets synthesized `refresh_token_invalidated` (`auth_manager.py:1371`), and the equality gate at `:326` (`latest.deactivation_reason != PERMANENT_FAILURE_CODES[exc.code]`) then recovers only if the prior holder also failed with that code; for `invalid_grant`/`refresh_token_expired`/`refresh_token_reused`/`invalid_refresh_token` the same account fails the request. Conservative but ordering-dependent — compare against the set of refresh-credential reasons rather than one code's string.
- Snapshot must be ACTIVE (`:305`), so a RATE_LIMITED/QUOTA_EXCEEDED snapshot with a dead refresh token still fails its first request; `token_expired` is excluded from `_REFRESH_CREDENTIAL_FAILURE_CODES` even though `PERMANENT_FAILURE_CODES`' comment says ordinary requests may continue on the stored access token for it. Both conservative, not regressions — noting for the record.
- Unresolved CodeRabbit thread on `spec.md:1689-1693` ("forced refresh recovery MUST NOT return that token as a successful refresh"): technically accurate — the post-401 retry (`app/modules/proxy/service.py:1609`, `force=True`) must keep failing — but the phrasing implies forced callers have a recovery path. Adopting the suggested wording ("ordinary preflight recovery MUST propagate the authentication failure or use the existing safe failover path") is a harmless clarity fix.

**Merge gate**

- `Contributors attribution` is red on `e4718ddef`: no `Irvinwop` entry in `.all-contributorsrc`. Every other check (pytest slices, Lint, ty, OpenSpec, Docker, Playwright) is green on the current head; "CI Required" is red solely from this one. Please add yourself (or we can push it — `maintainerCanModify` is on).

Soju06:
Re-triaged at round 22 against `origin/main` @ `ca243dec4`. The premise still holds — the only commit to touch `app/modules/accounts/auth_manager.py` since your branch point is #2261 (settings constantization), so an ordinary ACTIVE preflight whose refresh credential just died still throws away a still-unexpired access token. The round-20 P2 is still unfixed, and reading the merged code gives it a sharper consequence than was reported:

**The recovery gate reads back its own uncommitted write, so for every session-attached caller it is self-confirming rather than a check on committed state.**

`ensure_fresh` re-reads via `self._repo.get_by_id_fresh(account.id)`, which is `session.execute(select(Account)...execution_options(populate_existing=True))` (`app/modules/accounts/repository.py:206-217`). `SessionLocal` is built without `autoflush=False` (`app/db/session.py:373`), so that SELECT autoflushes first. By then `_handle_permanent_refresh_failure` has already mutated the caller's ORM object in place (`app/modules/accounts/auth_manager.py:1078-1082`, `account.status = status; account.deactivation_reason = reason`). When the caller's `account` is attached to the same session as `self._repo`, the autoflush emits `UPDATE accounts SET status=?, deactivation_reason=? WHERE id=?` carrying **none** of the `expected_status` / `expected_deactivation_reason` / `expected_reset_at` / `expected_refresh_token_encrypted` predicates that `update_status_if_current` exists to enforce. `populate_existing=True` then hands back the identity-map object — `latest is account`, and `_adopt_account_row` short-circuits on `target is source` (`auth_manager.py:1323-1336`) — so your `latest.status != REAUTH_REQUIRED` and `latest.deactivation_reason != PERMANENT_FAILURE_CODES[exc.code]` checks are comparing the row against what this request just flushed.

Failing scenario: account A is ACTIVE with stale `last_refresh`; a peer commits REAUTH_REQUIRED + `app_session_terminated` while this request's refresh is in flight; the request is `AccountsService` reset-credit (`app/modules/accounts/service.py:274`, account attached via `_get_visible_account`). The autoflush overwrites the peer's `app_session_terminated` with `"Refresh token was revoked - re-login required"`, the re-read returns that clobbered object, the gate passes, and recovery proceeds on a session-invalidated account — which your own spec forbids ("session invalidation MUST NOT qualify", "MUST NOT mask a different session-invalidated warning"). The mirror case clobbers a concurrent dashboard re-auth (ACTIVE + rotated tokens) back to REAUTH_REQUIRED, leaving a healthy account wedged: REAUTH_REQUIRED skips ordinary refresh and the Guardian's forced refresh then hits `_TERMINAL_REFRESH_STATUSES`. On PostgreSQL the autoflushed UPDATE also holds the row lock for the rest of the request.

This is introduced by the PR: on main `ensure_fresh` re-raises here and `get_session`/`get_background_session` roll back, so the dirty attributes are discarded. The attached callers are `AccountsService` reset-credit read/consume (`service.py:274`, `:337`), probe (`service.py:711`), `/v1/reset-credit` (`app/modules/rate_limit_reset_credits/api.py:330`), automations (`app/modules/automations/service.py:1198`) and limit-warmup (`app/modules/limit_warmup/service.py:293`). Proxy stream/bridge callers use detached clones in a separate repo session, so the motivating Responses path is genuinely fine — but the helper lives on the shared `AuthManager` surface.

To land this:

1. Make the re-read non-mutating: expire the caller's object on the repo session before it, run it inside `no_autoflush`, or read through `self._refresh_repo_factory()` and stop mutating the caller's object in the failure path. Whichever you choose, assert in a test that the gate sees committed state and not `latest is account`.
2. Add regressions with an *attached* account plus an intervening peer write, in both variants: peer writes `app_session_terminated` (recovery MUST refuse) and peer writes ACTIVE + rotated tokens (MUST NOT be clobbered back to REAUTH_REQUIRED).
3. Either gate recovery to the proxy request path or document and test the dashboard/automation callers, which today would start silently serving from a REAUTH_REQUIRED row.
4. Rebase — three conflicts against `ca243dec4`: `app/modules/accounts/auth_manager.py`, `openspec/specs/usage-refresh-policy/spec.md`, `tests/integration/test_proxy_responses.py`.
5. Add yourself to `.all-contributorsrc`; every pytest shard and `ty` are green and "Contributors attribution" is the only real red.

If you would rather not carry this, say so and we will re-land it with `Co-authored-by:` attribution.

Issue #1707, opened 2026-08-13
bug: existing Codex thread can remain unusable on dead hard-affinity owner while fresh/side chat works


myudak:
### Pre-flight checklist

- [x] I searched existing issues and confirmed this is not a duplicate of the already-fixed/older-version behavior.
- [x] I am running a supported codex-lb version.
- [x] I can reproduce the problem with the steps below.

### codex-lb version

`1.23.0`

### Deployment method

`uvx (codex-lb)`

### Client used against codex-lb

Codex app (desktop / web)

### ChatGPT account plan(s) involved

Plus

### Model(s) involved

`gpt-5.6-sol`

### Summary

An existing/long-running Codex thread can become permanently unusable when its upstream owner/account becomes unavailable, even though other healthy accounts are still available in the pool.

The strongest signal is that **the exact same Codex client works immediately when I use `/side chat` or `Continue in new chat`, while continuing the original thread keeps failing**.

This makes the failure look specific to that thread's continuation / hard-affinity state rather than the model, client, endpoint, or global account pool.

This still reproduces on `1.23.0`.

Related older report: #1505 describes `hard_affinity_saturated` when a `codex_session` owner becomes unavailable. The issue here is that I am still seeing effectively the same user-visible failure on `1.23.0`, including after clearing sticky-session mappings.

### What happened?

The original thread repeatedly fails with:

```text
No account selected strategy=capacity_weighted sticky=True model=gpt-5.6-sol error=No available accounts
proxy_error_response ... status=503 code="hard_affinity_saturated" message="No available accounts"
```

and sometimes degrades into:

```text
No account selected strategy=capacity_weighted sticky=True model=gpt-5.6-sol error=No available accounts
Operating in degraded mode: all upstream accounts are unavailable
proxy_error_response ... method=POST path=/backend-api/codex/responses status=404 code="upstream_error" message="Not Found"
INFO: 127.0.0.1:xxxxx - "POST /backend-api/codex/responses HTTP/1.1" 404 Not Found
```

Remote compaction in that thread also fails:

```text
Error running remote compact task: unexpected status 404 Not Found: Not Found,
url: http://127.0.0.1:2455/backend-api/codex/responses
```

At the same time, the dashboard showed **two other active accounts with substantial weekly capacity remaining (~75% and ~91%)**. One exhausted/paused account was at 0%.

So the user-visible state is effectively:

```text
old thread -> unavailable owner / continuation anchor -> No available accounts
[...]

PR #2121, opened 2026-09-06
fix(proxy): recover tool-complete goal followups from unavailable owners

Irvinwop:
## Summary

Allow a complete, durable-manifest-verified tool batch to be followed by fresh user input when proving that a full resend can recover from an unavailable response owner.

Refs #1707 — **partial fix** for replayable tool-complete goal restarts. This does not make opaque encrypted compaction checkpoints portable between accounts.

## Type of change

- [x] `fix:` — bug fix

## OpenSpec

- [x] This PR includes / updates an OpenSpec change.
- [x] This PR touches a codex-faithful path and preserves request/response payloads.

Archived change: `openspec/changes/archive/2026-09-06-recover-tool-complete-goal-followup/`

## Changes

- Prove the exact prior-response call/result batch independently of a trailing self-contained user-input suffix. The helper previously rejected that suffix even when the complete batch matched its durable pending-call manifest.
- Keep the existing complete-body projection, durable-prefix fingerprint, account-neutrality, account scope, file ownership, and pre-dispatch checks. No input is dropped by this proof helper.
- Preserve rejection of missing parallel results, duplicate or mismatched calls, interleaved user input, and later tool/developer/assistant items. The existing three-item developer-interleave exception is not broadened.
- Exercise both HTTP endpoints with the real database and account selector: paused or quota-exhausted owner, fresh replay to a replacement, then anchored continuation on that replacement. Account-owned files remain pinned and cannot use this recovery.

No settings, migrations, defaults, or dashboard changes.

## Test plan

Base: current upstream `main`, `5ad638b6a4c9c094bcc8866b1d7487173fe3b54e`.

With the production helper patch removed, the new goal-recovery regression failed on **both HTTP endpoints with 502**. With the patch, it recovers without the unavailable owner's anchor and retains the complete input.

```text
uv run pytest tests/unit/test_replay_safety.py \
  tests/unit/test_durable_bridge_sessions.py \
  tests/unit/test_durable_bridge_transcript_codec.py \
  -q --tb=short --show-capture=no --timeout=120
585 passed

# Final focused run, including added file-owner rejection cases:
uv run pytest tests/unit/test_replay_safety.py \
  -q -k goal_followup --tb=short --show-capture=no --timeout=90
22 passed, 360 deselected

make lint
uv run ty check
openspec validate responses-api-compat --type spec --strict
openspec validate --specs
git diff --check
All passed
```

An earlier broad run with a 60-second per-test timeout timed out in the existing idle-reader replacement test; the 120-second run above passed. These are local test results, not a claim about hosted CI or production deployment.

## Example behavior

The replay input contains a verified stored prefix, the prior response's complete tool call and result, and a new goal instruction. There is no client-supplied `previous_response_id`.

- Before: context proof fails; bridge restores the old anchor and returns `previous_response_owner_unavailable` when its owner cannot serve.
- After: existing account-neutral recovery removes that anchor, forwards the full projected history to an eligible replacement, and completes. The next anchored request stays on that replacement.
- If an account-owned file or opaque encrypted checkpoint is involved, this patch does not authorize cross-account recovery.

## Checklist

- [x] Conventional Commits title and related issue linked.
- [x] Regression, integration, rejection, and continuation tests added.
- [x] Relevant lint/type/test targets run locally.
- [x] OpenSpec delta verified, synchronized, archived; specs validated.
- [x] Simplicity gates reviewed; no additional setup or configuration.
- [x] CHANGELOG not edited.

## Summary by CodeRabbit

* **New Features**
  * Added support for continuing a conversation with a fresh user message after a completed tool-call batch.
  * Full tool-call and result history is preserved when resuming interrupted goals.

* **Bug Fixes**
  * Improved recovery when the previous account is unavailable, allowing eligible conversations to continue on a replacement account.
  * Added safeguards to reject incomplete, reordered, duplicated, or interleaved tool-call sequences.

### DIFF
--- a/app/modules/proxy/replay_safety.py
+++ b/app/modules/proxy/replay_safety.py
@@ -422,6 +422,16 @@ def responses_input_suffix_matches_pending_tool_calls(
         and _fresh_developer_interleave_is_bounded(suffix, index=1)
     ):
         suffix = [suffix[0], suffix[2]]
+    # A goal restart or steer can append user input after a fully settled
+    # prior-response tool batch. Prove that batch independently: new input
+    # must not stand in for a missing parallel call or interrupt its results.
+    first_followup = next(
+        (index for index, item in enumerate(suffix) if isinstance(item, dict) and _is_fresh_followup_input(item)),
+        len(suffix),
+    )
+    if not all(isinstance(item, dict) and _is_fresh_followup_input(item) for item in suffix[first_followup:]):
+        return False
+    suffix = suffix[:first_followup]
     if not all(
         isinstance(item, dict)
         and isinstance(item.get("type"), str)
@@ -612,7 +622,7 @@ def _is_retained_response_message(item: Mapping[str, JsonValue]) -> bool:
 
 def _is_fresh_followup_input(item: Mapping[str, JsonValue]) -> bool:
     item_type = item.get("type")
-    if item_type in {"input_file", "input_image", "input_text"}:
+    if item_type in ("input_file", "input_image", "input_text"):
         return _input_content_part_is_self_contained(item, allow_output=False)
     return (
         item_type in (None, "message")
--- a/openspec/changes/archive/2026-09-06-recover-tool-complete-goal-followup/proposal.md
+++ b/openspec/changes/archive/2026-09-06-recover-tool-complete-goal-followup/proposal.md
@@ -1,0 +1,31 @@
+## Why
+
+A client can resend a complete response's tool calls and matching results, then
+append a user instruction when continuing a stopped goal. The durable pending
+call manifest proves that no parallel call was omitted, but the existing proof
+rejects every user follow-up after that complete batch. The HTTP bridge can then
+inject its old response anchor and fail on an unavailable owner, despite having
+a replayable full request. This is a bounded part of issue #1707, not a claim
+that opaque compaction checkpoints can move between accounts.
+
+## What Changes
+
+- Accept a trailing self-contained user-input suffix after exact settlement of
+  the durable prior-response tool manifest.
+- Keep all retained calls, results, and fresh input in the replay body.
+- Preserve existing durable-prefix, account-neutrality, file ownership, account
+  scope, and pre-dispatch recovery checks.
+- Preserve the tighter existing developer-interleave contract; this change does
+  not combine that exception with a new trailing input suffix.
+
+## Capabilities
+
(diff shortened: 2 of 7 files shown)


Soju06:
Verified on main `a0d94a46a` (round baseline): with your tests applied against main's `app/modules/proxy/replay_safety.py`, `test_full_resend_tool_manifest_accepts_goal_followup_after_all_results` fails and all four account-neutral `test_http_bridge_goal_followup_after_complete_tool_batch_leaves_unavailable_owner` cases return 502; with the patch, 22/22 pass, and the wider replay/durable-full-resend suites (344 tests), the proxy architecture/timing-seam/cancellation guards and `openspec validate --specs` are green on the merged tree. The change is scoped correctly: the follow-up slice at `replay_safety.py:425-434` runs after the three-item developer exception and before the exact-manifest equality, so a trailing user message cannot stand in for a missing parallel call.

Two things stand between this and merge:

1. Rebase onto current main (`e987c56c8`). The only conflict is `openspec/specs/responses-api-compat/spec.md`: since your branch point, #2168 (`8c6467d97`, Rust SSE framing), #2171, #2114 and #2184 all appended Requirements at the end of that file, and your "Exact tool-manifest replay preserves fresh user follow-up" requirement lands at the same spot. Keep all blocks (union); there is no code conflict (`replay_safety.py` and `streaming.py` are untouched on main since then).
2. Add yourself to `.all-contributorsrc` (the Contributors attribution job is the only red on `f5e044d8c` and it is what fails the "CI Required" aggregate).

On the Copilot thread at `replay_safety.py:434`: I traced it and do not think it is a hole. The proof only sets `safe_fresh_context`; the actual cross-account dispatch is still gated by `_http_bridge_payload_is_account_neutral_fresh_replay` on the projected payload (`app/modules/proxy/_service/http_bridge/streaming.py:1611` and `:2078` on current main), which runs `responses_input_items_are_self_contained_fresh_replay` over the whole input, so a trailing user message with foreign `internal_chat_message_metadata_passthrough`, unknown fields or a non-completed status is rejected before replay (I probed each case: proof=True, payload-neutral=False). This is the same contract `responses_input_suffix_retains_prior_output` has always used for `_is_fresh_followup_input`. A one-line reply on the thread noting that gate would let it be resolved.

Heads-up: #2086 rewrites the body of the same function (prefix-settled outputs); whichever lands second will need a small textual rebase.

PR #2218, opened 2026-09-09
feat(auth): add dashboard user tables and migrate the shared admin into a user

Soju06:
## Summary

Stacked on the roles-schema PR (`feat/dashboard-roles-schema`, #2217). Second Phase-1 PR of the dashboard RBAC plan: the **per-user account tables**, plus the expand/contract bridge that turns today's shared admin password into an `admin` user without any visible migration step — same password, same TOTP, no banner. **Authentication still reads the legacy `dashboard_settings` columns in this release**; every credential write is mirrored onto the user row so the next PR can switch authority atomically.

## Type of change

- [x] `feat:` — new capability (schema + compat projection)

Linked issue: groundwork for #673 (user management); no issue closed.

## OpenSpec

- [x] This PR includes an OpenSpec change

Change directory: `openspec/changes/add-dashboard-users-schema/` (new capability `dashboard-users`; `openspec validate --strict` passes)

## Changes

**Schema** (`app/db/models.py`, revision `20260909_010000_add_dashboard_users`):
- `dashboard_users`: `id`, `username` (unique), `display_name`, `email` (unique, nullable), `role_id` FK → `dashboard_roles` RESTRICT, `role_source` (`manual|mapping|scim`), `status` (`active|disabled|invited`), `password_hash`, `totp_secret_encrypted`, `totp_last_verified_step`, `session_generation`, `must_change_password`, `is_break_glass`, `created_by_user_id` self-FK SET NULL, `last_login_at`, timestamps.
- `dashboard_identities`: external identities (`provider`, `provider_key`, `subject` unique together, `groups_json`, `last_seen_at`) FK → users CASCADE. Column-only; no login path yet.
- `api_keys` gains `owner_user_id` / `created_by_user_id` (FK SET NULL) and `deactivated_reason`, all nullable and unused for now (existing keys stay "service keys"). Added here so the later self-service change needs no second SQLite table rebuild.
- Backfill: when a legacy password exists, the revision inserts the `admin` user (deterministic UUIDv5 id, admin preset, `is_break_glass=true`, TOTP secret and replay step copied). Identifiers are frozen literals in the revision (a unit test keeps them in sync with the runtime constants). Idempotent; downgrade drops the tables and columns.

[...]

Issue #673, opened 2026-05-16
feat: user-management


geoHeil:
### Pre-flight checklist

- [x] I searched existing issues and discussions for similar proposals.
- [x] This is a concrete proposal, not an open-ended question.

### Problem / motivation

Currently the dashboard is single user. you have to login with the password - but then everything is possible.

However, seeing the status of the various subscriptions is really useful.

I want to expose them to other users.

### Proposed change

Can codex-lb either:

- have a readonly deployment which is coupled to the same db but only displays things
- user management/permissions so some users can only view vs. the admin can configure codex-lb

### Area

Account management & quota tracking

Komzpa:
@geoHeil how about: users can use their api key as password and only see stats on their api key?

Or do you have other "users" in mind?

geoHeil:
I think this is almost there.

I would want for users to be able to see the remaining stats.

However, users may be ill defined i.e. in my setup I front codex-lb with litellm. So I have 1 user (the service account for litellm) and then users configured there.

I would love to have readonly / somehow shareable version of the dashbaord to show the usage stats/ remaining tokens to some user who should not have permission to administer codex-lb

Komzpa:
Will a simple "guest mode password" be a reasonable solution for your scenario?

If you're looking to share statistics, will something like prometheus-compatible endpoint that will get pushed into something like pre-existing grafana be more useful?

geoHeil:
The prometheus endpoint sounds sensible - but is overkill for my current needs.

A guest passrod with readonly permissions would orpbably be the easiest first step

geoHeil:
@Komzpa are you fine with guest mode password which is readonly access? should I send a PR? do you want to do it yourself?

Komzpa:
@geoHeil I think it should be fine, if internally you abstract it enough so that in future it's possible to convert it to multi-user system some day. Please send the PR if you have the time to do it.

geoHeil:
Draft PR is up for the read-only guest access path discussed here: https://github.com/Soju06/codex-lb/pull/703

geoHeil:
awesome! when is the plan for a next release?

PR #2223, opened 2026-09-09
feat(auth): make dashboard users the source of truth for login and sessions

Soju06:
## Summary

Stacked on the users-schema PR (`feat/dashboard-users-schema`, #2218). Third Phase-1 PR of the dashboard RBAC plan and the **authority flip**: `dashboard_users` becomes the source of truth for dashboard sign-in and sessions. The shared password becomes the `admin` account's password with no visible migration step: same password, same TOTP secret, no banner. Everyone signs in once more after the upgrade because the session cookie format changed (deliberately fail-closed across the rolling window).

This is the largest PR of the stack (app code ≈ +1.1k net). Cookie codec, session validation, login, TOTP, bootstrap state, and the repository all switch from the legacy columns to the user row **together** — splitting them would leave a release where two sources of truth disagree (the exact hazard PLAN §H12 calls out). Tests and OpenSpec make up the rest of the diff.

## Type of change

- [x] `feat:` — new capability (per-user login and sessions)

Linked issue: groundwork for #673 (user management); no issue closed.

## OpenSpec

Change directory: `openspec/changes/user-login-and-session-v2/` (`admin-auth` ADDED/MODIFIED, `dashboard-users` ADDED; `openspec validate --strict` passes)

## Changes

**Users cache + local auth state** (`app/core/auth/dashboard_users_cache.py`, `app/modules/dashboard_users/repository.py`): 5 s TTL process cache over `dashboard_users`, invalidated through the existing cross-replica bus (`dashboard_users` namespace). `local_auth_state()` replaces every `dashboard_settings.password_hash is None` predicate: an install requires authentication when an active account holds a password or an external identity. The legacy credential columns are never read again on any request path.

**Session cookie v2** (`DashboardSessionStore`): user sessions carry `{v:2, exp, iat, uid, sg, pv, tp, am}` (account id, the account's `session_generation`, password/TOTP verified flags, auth method), guest sessions `{v:2, exp, iat, guest:true, gg}`. Anything else, including every v1 payload, is rejected. Old replicas reject v2 cookies for the same reason (the keys they need are absent), so a mixed fleet never accepts a cookie it cannot validate.

**Validation** (`validate_dashboard_session`): loads the account by id (must be `active`, generation must match), builds the principal from the account's role via `resolve_role_grants`, keeps `role: admin|guest` on the wire, adds `user_id/username/role_slug/auth_method` to the principal. TOTP policy applies per account: an account without a secret under "require TOTP" is held at `403 totp_enrollment_required` on everything except the self-service routes (session, TOTP setup/verify, logout, logout-all, me, password change).

**Login** (`POST /api/dashboard-auth/password/login {username?, password}`): the username may be omitted while exactly one active account holds a password (the session response says so via `login.username_field: hidden`); with two or more the API answers `422 username_required`. Unknown username, disabled account, and wrong password produce the identical `401 invalid_credentials` and the same bcrypt cost. Per-IP limiter unchanged.

**Account-bound sessions**: `POST /password/change` rotates the hash and bumps `session_generation` atomically in one transaction (database-side increment) and re-issues the cookie to the caller with a lifetime no longer than the replaced session; `POST /api/dashboard-auth/logout-all` signs the account out everywhere; disabling an account or bumping its generation ends its sessions on the next request. `DELETE /password` is allowed only on a solo install (`409 other_users_exist` otherwise) and still regenerates the bootstrap token. `GET /api/dashboard-auth/me` returns the signed-in account.

**Session response additions** (all optional, defaulted): `user`, `authMethod`, `mustChangePassword`, `totpEnrollmentRequired`, `login {usernameField, providers, localLogin}`, `accessSummary` (team facts, **only** for authenticated accounts holding `users:manage`), `assignableRoleIds` (admin, operator, viewer; the member preset becomes assignable once own-scoped views ship). `permissions` keeps `read`/`write` first and appends every grant as `<permission>:<scope>`. `role` stays `admin|guest`.

**Compat and migration**: the compat `admin` row is now primary; its writes are mirrored onto the legacy columns in the same transaction (and re-applied on the optimistic-conflict retry) so a replica of the previous release keeps working during the rolling window. Revision `20260909_020000_reproject_compat_admin_credentials` copies legacy → `admin` once before the first read, covering credentials an old replica wrote legacy-only. Identifiers are frozen literals in the revision. Downgrade is a no-op.

**Behaviour preserved** (PRINCIPLES P1): passwordless install → implicit local admin and remote bootstrap token exactly as before; upgraded install → `admin` signs in with just the password; guest flows byte-identical apart from the extra scoped `permissions` entries.

**Tests**: session store (v2 payloads, v1 rejected), dependency branches, login resolution (sole user, 422, identical 401 bodies, single bcrypt call for unknown accounts), per-account TOTP incl. replay refusal and compat mirroring, enrollment gate allow-list, password change atomicity + cookie re-issue, atomic generation bump under stale objects, logout-all, disabled user / generation mismatch → 401, remove-password solo rule, `/me`, session response shape (`login`, `accessSummary` null for guests and unauthenticated callers, exact `assignableRoleIds`), legacy column ignored when no account holds a password, non-admin accounts never mirrored, TTL rules, re-projection migration cases, guest suites unchanged, route permission matrix, e2e auth flow.

**Docs**: `docs/authentication.md` "Signing in" section + roles paragraph.

## Simplicity

- [x] Zero config; no new setting, env var, README section, or nav item.
- New setting(s): none
- [x] Budgets unchanged. Frontend untouched (login form changes land in the next PR).

## Test plan

```
uv run ruff check . && uv run ruff format --check . && uv run ty check        # clean
uv run pytest tests/unit/test_dashboard_*.py tests/unit/test_bootstrap.py \
  tests/integration/test_dashboard_password_auth.py tests/integration/test_dashboard_totp_auth.py \
  tests/integration/test_dashboard_bootstrap.py tests/integration/test_dashboard_users_schema.py \
  tests/integration/test_auth_middleware.py tests/integration/test_guest_session_generation.py \
  tests/integration/test_dashboard_route_permission_matrix.py tests/integration/test_dashboard_permission_gates.py \
  tests/e2e/test_auth_flow.py -q
uv run pytest tests/integration/test_migrations.py::test_postgresql_migration_contract_policy_and_drift_match -q
openspec validate user-login-and-session-v2 --strict                          # valid
codex review --base feat/dashboard-users-schema                                # round 1: 2 P1 + 2 P2 fixed; round 2: 3 P1 fixed
```


Soju06:
Note on the GitGuardian status: the only new literals in this diff are test fixture passwords (e.g. `password123`/`ops-password` pairs next to usernames in the login tests), the two deterministic UUIDv5 identifiers already present in #2218, and a lazily generated dummy bcrypt hash computed at runtime. No credential or token is committed; if the dashboard shows a "username/password" match on a test file it can be marked as a false positive.

PR #2234, opened 2026-09-09
fix(load-balancer): keep soft sticky owner when it is only request-locally unavailable

Soju06:
## Why

Native Codex conversations over HTTP downstream (`native_codex_http`) re-run account selection every turn, while the WebSocket bridge selects once and keeps its upstream socket. Both derive the same soft TTL-bounded `prompt_cache` thread row, so the difference is how that row behaves under per-turn pressure. Today the row is preserved when its owner is *visible but unhealthy* (rate-limited, in backoff), but when the owner is *absent from the request's candidates* — because this request's retry loop excluded it after one transient failure (code-less burst 429, response-create cap) or the per-account stream-cap filter dropped it — selection treats the owner as removed and **permanently rebinds** the conversation to the fallback (`sticky_selection.py` `persist_fallback` path).

Production (2026-09-09, request_logs 24h): codex_exec conversations over HTTP touch >1 account in 74.6% of conversations (avg 3.38) vs 16.4% over WS; a switched turn at ≥60k input drops prompt-cache hit 94.7%→53.7% and raises TTFT p50 6.2s→7.9s, p90 16.8s→21.6s. Switch rate is 0.4–1.8%/h quiet but 13–40%/h under load / upstream overload (09-08 07–11Z).

## What changes

- A soft TTL-bounded (`prompt_cache`) owner missing from the request's selectable candidates *only* because of request-local pressure (per-account cap filter, or excluded by this request's own retry loop while its persisted status is still recoverable) keeps its mapping: the alternate serves this request, nothing is upserted/deleted, the existing `internal_soft_affinity_spillover` diagnostic fires (now honouring `redact_sensitive_details`).
- Permanent removal is unchanged: overload-isolated, `reallocate_sticky`, paused/deactivated, or out-of-continuity-scope owners are still rebound. Hard rows (`codex_session` turn-state, legacy raw keys) keep their rules.
- `StickySelectionRequest.exclude_account_ids` (defaulted) lets the predicate distinguish "excluded by this request" from "not in the pool"; `LoadBalancer.select_account` forwards the set it already receives.
- No new setting (same rule the bare process-session cap spillover already applies, generalised to the TTL-bounded soft kind). Ceilinged files: `load_balancer.py` +3 lines (2982/3021); `service.py`, `retry.py`, bridge/streaming mixins untouched.

## OpenSpec

`openspec/changes/preserve-soft-sticky-owner-on-request-local-unavailability/` — ADDED requirement "Request-local unavailability of a soft TTL-bounded owner is non-mutating" under `sticky-session-operations`. `openspec validate --specs` and `--strict` pass.

## Tests

- `tests/unit/test_proxy_soft_sticky_spillover.py` (new): owner dropped by cap filter → preserved; owner excluded by retry loop with recoverable status → preserved; owner paused/deactivated/out of scope → rebound; hard rows unchanged.
- `tests/unit/test_load_balancer_concurrency.py`: retry-loop plumbing boundary (`exclude_account_ids` → `StickySelectionRequest`) covered end-to-end through `select_account`.
- Gates run locally: ruff, `check_proxy_architecture`, `check_cancellation_safety`, `check_proxy_timing_seams`, `check_settings_tiers`, `ty check`, `tests/unit` full run green. Local `codex review --base c38e4de15`: no actionable regressions.

## Ops

Monitor `internal_soft_affinity_spillover` rate and the per-conversation distinct-account count after deploy; expected: HTTP conversations converge toward the WS figure (~1.3 accounts/conversation).


Komzpa:
@codex review

PR #2235, opened 2026-09-09
perf(dashboard): bound the raw tail of the conversation presence union to constant index ranges

Soju06:
## Why

The two most expensive statements in production `pg_stat_statements` (2026-09-09) are the dashboard's distinct-conversation reads: `conversation_count`/`conversation_request_count` from `GET /api/dashboard/overview` (mean 3.07s, 276 calls) and the hourly conversation trend buckets (mean 1.5s). While a dashboard tab is open, postgres on the 4 vCPU prod host runs at 100–186% CPU from these alone.

Both are built on `conversation_presence_union`, whose raw `request_logs` branch LEFT OUTER JOINs the single `account_usage_rollup_state` row and filters with `watermark IS NULL OR requested_at < lo OR requested_at >= least(watermark, hi)`. Because the OR references a joined column and has an `IS NULL` disjunct, PostgreSQL can only apply it as a post-scan filter: the only Index Cond is the outer `requested_at` window, so every poll walks the whole 7-day window (14 days per overview poll) although the un-folded tail is ~2–3 hours.

## What changes

- `app/modules/accounts/usage_time_rollup_read.py` only: the conversation-presence primitives read the watermark as an uncorrelated scalar subquery inside the same statement and express the raw complement as two constant half-open ranges `[since, lo)` and `[greatest(lo, least(coalesce(W, lo), hi)), until)`. PostgreSQL plans a BitmapOr of two index ranges (dashboard union) / two parameterized Index Cond arms (reports' labeled union); SQLite plans a scalar subquery plus index search. Missing or epoch watermark degrades to exactly the legacy raw read.
- Single-snapshot contract unchanged (still one statement). No caller signature change, no schema change, no setting.
- Out of scope (follow-ups): `COUNT(DISTINCT)` → `GROUP BY`, weekly trailing-demand memo (#see sibling PR), facet-skip partial indexes (sibling PR).

## OpenSpec

`openspec/changes/bound-conversation-raw-tail-index-range/` — MODIFIED requirement "Distinct-conversation reads combine the presence rollup with a raw live tail in one statement" under `query-caching` (mechanism clause + 2 scenarios). Validates `--strict`.

## Tests

- `tests/integration/test_conversation_presence_union.py` (new, added to `POSTGRES_PYTEST_TARGETS`): union aggregates vs raw-only oracle for every watermark position (missing/epoch/before/inside/after) over aligned, unaligned, open-ended windows with/without buckets and soft-deleted rows; labeled union row-for-row (`COUNT(*)` per label); statement shape assertions; on PostgreSQL `EXPLAIN (ANALYZE)` asserts every `request_logs` node is an index/bitmap scan with Index Cond on `requested_at` and zero rows removed by filter.
- Existing parity harness `tests/integration/test_request_usage_rollup_parity.py` unchanged and green; 102 related dashboard/repository tests green. Local `codex review --base c38e4de15`: no actionable regressions.


Soju06:
Rebased onto main (11ab09ba3). Since #2227 moved every reports read onto the permanent `report-aggregation` source, `conversation_labeled_presence_union` had no callers left; this PR now **deletes** it instead of rewriting it, and the query-caching delta is rebased on main's current requirement text (reports paragraph kept verbatim). The dashboard union rewrite (scalar-subquery watermark, constant-window raw tail, one statement/one snapshot) is unchanged. Re-verified: SQLite 111 passed; PostgreSQL 16 112 passed incl. the index-bounded plan test.

Komzpa:
@codex review

PR #2244, opened 2026-09-09
feat(proxy): trace terminal-frame delivery on Responses SSE streams

Soju06:
## Why

During the 2026-09-09 codex-lb vs direct A/B, 1 of 13 LB Responses SSE streams closed after the full body but **before `response.completed` reached the client**, while `request_logs.status` recorded `success`. Status is settled when the proxy *parses* the upstream terminal frame (`terminal_event_seen = True` is set before the terminal `yield`), not when the frame is *written* downstream, and nothing records whether the write happened. Two write paths lose the frame silently: uvicorn's `RequestResponseCycle.send` returns without writing once the cycle is marked disconnected (peer FIN/RST racing the terminal write), and an exception in an `api.py` stream wrapper while formatting the terminal block. The truncation rate is currently unmeasurable; Codex clients treat a missing terminal as a stream error and retry.

## What changes (observational only, zero-config)

- **Protocol-side disconnect stamp** (`app/core/http_protocol.py`, `http_protocol_httptools.py`): `connection_lost` stamps the loss kind into the in-flight request's ASGI `scope["state"]`. Under HTTP/1.1 pipelining uvicorn's httptools impl swaps `self.cycle` to the newest queued request while the earlier response streams, so the httptools subclass tracks the cycle whose ASGI task is running and stamps every open request (active, newest, queued).
- **`DeliveryTracedStreamingResponse`** (new `app/modules/proxy/downstream_delivery.py`): wraps ASGI `send`, counts chunks/bytes, detects the terminal SSE event type across a rolling tail buffer (frames split across chunks handled), and commits the terminal **only once the complete block through the blank-line terminator was handed to the server and `send` returned**. Exactly one bounded outcome per stream: `terminal_written`, `terminal_after_disconnect`, `exception_before_terminal`, `cancelled_before_terminal`, `ended_without_terminal` → one `responses_stream_terminal_delivery` log line (DEBUG on success, WARNING otherwise) + counter `codex_lb_stream_terminal_delivery_total{surface,outcome}`.
- **Wiring**: `_stream_responses` in `api.py` returns the traced response (`surface="responses"`), covering native codex HTTP, the HTTP session bridge, and the per-request WS-upstream variant.
- Forwarded bytes, request-log status and settlement are untouched. No new `CODEX_LB_*` setting, no migration, no ceiling-guarded file growth. No-ops without `prometheus_client`.

Join key for ops: the log's `request_id` is the ingress id, stored in `request_logs.archive_request_id` (`request_logs.request_id` is overwritten with the upstream response id).

## Deferred (documented in the proposal)

Settlement reordering is *not* a safe fix (would flip rows to cancelled and under-count consumed usage) — needs its own responses-api-compat change once the trace shows outcome frequencies. Wrapper-exception hardening and other SSE surfaces deferred likewise.

## OpenSpec

`openspec/changes/add-responses-terminal-delivery-trace/` — ADDED under `proxy-runtime-observability` (trace metric/log) and `http-ingress-limits` (disconnect stamp), MODIFIED settlement wording under `responses-api-compat`. Validates `--specs` and `--strict`.

## Tests

- `tests/unit/test_downstream_delivery.py` (new): outcome matrix incl. split-terminal-block commitment and disconnect/cancellation inside a split frame.
- `tests/integration/test_http_disconnect_stamp.py` (new): fake-transport and live-socket regressions, including pipelined requests where the queued request had replaced `self.cycle`.
- `tests/integration/test_proxy_responses.py`: wiring of `_stream_responses` and `scope["state"]` identity through the full middleware stack.
- Gates green: ruff, architecture, cancellation-safety, timing-seams, settings-tiers, ty, openspec. Local `codex review --base c38e4de15` (2 rounds; round-1 P2s on pipelining stamp and split-frame commitment fixed in the second commit): no actionable regressions.


Komzpa:
@codex review

PR #2253, opened 2026-09-09
perf(request-logs): speed up SQLite filter options

### LINKED ISSUE #2267: bug: dashboard filter options take seconds to load on SQLite
## What happens

Dashboard filter options took 4.7–6.2 seconds to load on SQLite during the September 9 investigation, using roughly one CPU core. The request rows loaded in 20 ms. `GET /api/request-logs/options` also runs every 30 seconds while the dashboard is active.

Cleanup had finished, but the delay remained.

Without SQLite planner statistics, the recursive facet query repeatedly scans live request-log rows instead of advancing through indexed facet values. Adding the partial indexes proposed in #2246 did not fix the reproduced query plan.

The query introduced in #1239 is still present on main. The [investigation and regression results in #2253](https://github.com/Soju06/codex-lb/pull/2253) include the measured timings and SQLite query plans.

## Expected behavior

The options query should work efficiently with the standard SQLite indexes, without requiring operators to run `ANALYZE`.

It must return the same options and ordering, with the same soft-deletion filtering, empty or missing reasoning values, and supported statuses. Filtered requests and PostgreSQL should keep their existing behavior.

## Related work

#2253 changes the SQLite query and tests the real HTTP endpoint with repeated synthetic rows. It measures SQLite work instead of relying on a timing threshold, and covers both the standard schema and #2246's proposed indexes. Those indexes are complementary, not a prerequisite. The proposed fix has not yet been measured as a deployed endpoint improvement.

JustYannicc:
## Summary

Dashboard filter options took 4.7–6.2 seconds to load on a populated SQLite database. The recursive query repeatedly scanned live rows instead of advancing through indexed values.

Fixes #2267.

## Changes

Enumerate indexed facet values first, then check row visibility through an equality `EXISTS`. Preserve paired-facet boundaries, ordering, empty/NULL values, soft deletion and supported statuses.

Only unfiltered SQLite options use the new query. Filtered requests and PostgreSQL keep their existing paths. No indexes, settings, migrations or operator statistics-maintenance step are required.

The partial indexes in #2246 are complementary; this query works with or without them. OpenSpec: [SQLite request-log facets](https://github.com/Soju06/codex-lb/tree/5c6b46a9249f1fc6988ee497cf6150becc79a8bd/openspec/changes/archive/2026-09-09-sqlite-request-log-facets).

## Test plan

The current-main composition passed 20 targeted tests. The real HTTP regression uses 20,006 synthetic rows and verifies exact JSON, with deterministic SQLite work counts:

| Schema without planner statistics | Before | After |
|---|---:|---:|
| Standard indexes | 1,717,000 operations | 1,200 operations |
| With proposed partial indexes | 1,681,200 operations | 1,000 operations |

Lint, typing and strict validation of all 64 specs pass. Tests used an isolated temporary SQLite database. [Hosted checks](https://github.com/Soju06/codex-lb/pull/2253/checks) remain required; full local CI and PostgreSQL/container tests were outside this verification.

Dead-only value groups can still require scanning. The SQL improvement has not yet been measured as a deployed endpoint improvement or a fix for every dashboard delay.


Komzpa:
@codex review

Soju06:
CI follow-up: the current integration-core-3 failure is the combined #2246/#2253 fixture collision (`idx_logs_live_api_key already exists`), reproduced on unchanged main. #2296 repairs both explicit indexed/unindexed test layouts; local verification passes both tests with the original response and VM-operation assertions intact. Keeping that fixture fix separate from this PR. This follow-up is needed because both independently reviewed changes have now landed.

Issue #1852, opened 2026-08-20
bug: anchor-poison threshold (default 7) is unreachable because the retry circuit trips at a hardcoded 2 — dead previous_response_id is never abandoned on 1.24.0-beta.3


nickwilson159:
### Pre-flight checklist

- [x] I searched open and closed issues and confirmed this is not a duplicate.
- [x] I am running a supported codex-lb version.
- [x] I can reproduce the problem with the steps below.

Closest prior art, and why this is distinct:

- **#1830** (closed, fixed by #1841 in `1.24.0-beta.3`) — same 503 `cooling down` surface. **This report reproduces on b3, with #1841 in the build.** #1841 bounded the wedge (it now self-clears after a cooldown instead of persisting) but did not remove the mechanism, and the misclassification it targeted is still observable (see step 2 below).
- **#1633** (open) — also about a hardcoded ~60s and honest classification, but for the *pre-`response.created` silence watchdog*. That is a different timer from the retry-circuit backoff described here; the two are complementary.
- **#1816** (closed, #1817) — `previous_response_id` continuation loops on 1.23.0, about recognising the terse upstream rejection. This report is about threshold **arithmetic**, not classification recognition.
- No issue mentions `anchor_poison_failure_threshold`.

### codex-lb version

`1.24.0-beta.3` (confirmed via `x-app-version` response header)

### Deployment method

uvx (codex-lb)

### Client used against codex-lb

Codex app (desktop / web) — `Codex Desktop/0.148.0-alpha.15 (Windows 10.0.26200; x86_64)`

### ChatGPT account plan(s) involved

Pro (pool of 2 accounts, both `status=active` throughout)

### Model(s) involved

`gpt-5.6-sol`

### What happened?

**Summary: `http_responses_session_bridge_anchor_poison_failure_threshold` defaults to `7`, but the retry circuit that gates the same counter trips at a hardcoded `2`. The poison threshold is therefore effectively unreachable, so a dead `previous_response_id` is never abandoned and the bridge re-injects it indefinitely.**

Both values are compared against the *same* counter. In `app/modules/proxy/_service/http_bridge/upstream_events.py`:

```python
consecutive_failures = await self._record_http_bridge_retry_circuit_failure_for_attempt_selection(
    session, detail=retry_circuit_detail, selection=retry_circuit_attempt_selection,
)
poison_candidate_detail = _http_bridge_anchor_poison_detail(retry_circuit_detail)
if (
    poison_candidate_detail is not None
[...]

PR #2276, opened 2026-09-09
fix(proxy): preserve newer quarantine during response cleanup

### LINKED ISSUE #2268: bug: old bridge requests can clear a newer session's quarantine
## What happens

An old HTTP bridge request can finish late and clear a replacement session's quarantine. Later requests can then reuse a session that should still be blocked after recent failures.

The race is:

1. Session A is quarantined, and a completing request records its generation.
2. Cleanup removes that entry.
3. Session B reuses the same key and is quarantined with the same generation number.
4. A finishes its delayed cleanup and clears B's quarantine.

Main increments generations on each entry and checks the generation during cleanup, without verifying which session owns it. Cleanup can also erase an initial failure recorded while completion was awaiting settlement.

The maintainer [reproduced the key-reuse failure while reviewing #1953](https://github.com/Soju06/codex-lb/pull/1953#issuecomment-5541052199). The relevant code remains on main.

## Expected behavior

A completion should clear only the quarantine it observed for its own session. Replacement entries and failures recorded during settlement must survive, including after pruning or reset.

Quarantine must remain bounded without silently discarding protection for active failures. #2276 proposes overflow handling and delayed clearing for some failures loaded from the database; those tradeoffs still need maintainer review.

## Related work

#2276 contains the quarantine fix extracted from #1953. The input-shape and forwarding fix is tracked separately in #2269 and implemented by #2277. This issue covers quarantine lifetime and cleanup, not the broader recovery work in #1867.

JustYannicc:
## Summary

An old bridge response can finish after its session key is reused and clear the replacement's quarantine. Cleanup can also erase a first failure recorded during settlement.

Closes #2268.

## Changes

Allocate generations for the service lifetime, verify the session that owns cleanup, and preserve evidence recorded after completion began. Keep admission bounded without evicting active poison evidence.

A completion that discovers durable poison during settlement keeps that quarantine. After successful settlement and fresh-anchor registration, the next request's first-touch load can revoke it from a zero-failure, non-tombstone row. Stale-load and newer-failure guards still apply.

This replaces the quarantine portion of #1953 and starts directly from main. It does not require #2277's input-classification fix. OpenSpec: [quarantine evidence lifetime](https://github.com/Soju06/codex-lb/tree/2e419652e20f31eda223470484f27814a1a523af/openspec/changes/fix-http-bridge-quarantine-evidence-lifetime).

## Test plan

The new HTTP completion-race regression fails on main and passes here. Bridge, quarantine and adjacent lifecycle suites pass locally, including a rerun with locked dependencies.

Reconciled with main `efe0f581` after the dashboard prewarm and bridge-budget changes. On this candidate, 1,342 bridge, quarantine and HTTP route tests pass using isolated temporary SQLite. The unchanged idle-handoff test is excluded from this standalone run because its timeout correction belongs to #2277. Ruff, typing, strict main specs and the active change validate.

The original split also passed 1,499 combined tests with #2277 and composed identically in both orders. [Hosted checks](https://github.com/Soju06/codex-lb/pull/2276/checks) and review must cover the updated head.

## Open decisions

Owner acceptance of bounded poison overflow, changed durable-only cleanup timing and the full extracted scope remains pending. These tasks stay unchecked in the active change.


Soju06:
Triage scope review on 2e419652: this is broader than the stale-completion identity/generation fix. When the poison registry is full, `_remember_http_bridge_quarantine_poison_overflow` installs a service-wide deadline and unknown keys are treated as poisoned; the candidate also changes durable-only poison cleanup timing. Those policies are explicitly unresolved in the active change, so a CLEAN GitHub state is not yet an accepted whole-scope review.

Please keep #2268 linked to the stale-cleanup regression and make the overflow tradeoff independently reviewable: the acceptance evidence should show unrelated fresh keys during all-poison saturation, recovery after the bounded window, and why the failure scope cannot be limited to the rejected key/session. The identity/generation race fix can be separated if that broader policy decision would hold it up. This comment is a scope review, not approval of the full lifecycle implementation.

Soju06:
Reviewed from two independent angles plus a local `codex review`, with every finding verified against the code before posting.

**The #2268 fix itself is sound and genuinely covered**: service-lifetime generations plus the canonical-session / weak-owner identity fence do close the cleanup race, the bridge surface is green at head (1152 passed), and the new `tests/integration/test_http_promotion_quarantine.py` carries a real product-path regression — with `app/` at the merge base the during-completion-first-strike parameterisation fails while the other five pass.

**What blocks it is the poison-overflow policy bundled with it**, and your own `tasks.md` leaves 2.1 (owner acceptance for the bounded window) unticked — so this is partly a decision for me and partly a defect.


---

### BLOCKER — `app/modules/proxy/_service/http_bridge/quarantine.py`

**One rejected poison arm sets a service-global deadline (quarantine.py:187, consulted at :405) that makes _http_bridge_session_key_poison_quarantined return True for every key in the process — including never-seen soft keys that can never carry poison evidence — 404ing unrelated healthy conversations. main has no such flag; it let the registry exceed the cap instead.**

_How it fails:_ During an upstream incident the 1024-slot registry saturates with active poison fences (each lives 600s+cooldown, and _load_http_bridge_retry_circuit arms one locally for every durable poison row written by any replica). The 1025th arm is refused by _admit_http_bridge_quarantine_key and _remember_http_bridge_quarantine_poison_overflow installs a service-level deadline. From that instant two production gates reject unrelated traffic: request_submit.py:1183-1222 raises HTTP 404 bridge_previous_response_not_found for any request carrying a proxy-injected previous_response_id — i.e. the normal sticky continuation — and streaming.py:2957-2988 raises the same 404 for any anchorless delta-only payload, including soft http-history keys whose retry circuit is skipped entirely at retry_circuit.py:558, and it short-circuits before the `await self._load_http_bridge_retry_circuit(session)` that could disprove it. Reproduced against the real ProxyService: a never-seen soft key http-history:unrelated-conversation goes from poison_quarantined False to True the moment the 1025th arm is refused; a brand-new hard prompt_cache key likewise. On the merge base the same script prints False. The merge base's own comment states the cap 'holds as a correctness bound, not an unconditional one' — so this converts bounded memory growth during an incident into service-wide request rejection at exactly the moment you least want it.

_Suggested fix:_ Scope the fail-closed state to what was actually rejected instead of to the service: keep a small bounded set/ring of rejected keys with their own deadlines and have _http_bridge_session_key_poison_quarantined consult that set, or fail/retire the rejected session's own request; and never let the fallback answer for a key whose strength cannot carry retry-circuit evidence. Alternatively, when every slot holds active poison, evict the entry with the earliest poison_quarantined_until rather than refusing admission. If the policy is wanted as designed, split it out of this PR as already requested on #2268 and land the generation/owner identity fence — the fix #2268 actually describes — on its own; it does not depend on the overflow behaviour.

---

### MAJOR — `app/modules/proxy/_service/http_bridge/quarantine.py`

**The overflow window is not bounded as the spec and design claim: it renews to now+600s on every later poison event — both a rejected arm (via _remember_..., line 202-213) and a successful arm while the flag is live (quarantine.py:570-577) — so it never drains while any poison activity continues, even after the registry has emptied.**

_How it fails:_ The delta states 'a rejected poison arm MUST establish a service-level fail-closed deadline covering its required lifetime and retained active poison deadlines' and responses-api-compat:851 calls it a 'bounded service-level fail-closed deadline'; design.md says 'A single overflow deadline covers rejected poison keys'. Neither authorizes renewal from later unrelated events. I drove the real ProxyService with a virtual clock: saturate, reject at t=1000 -> overflow_until=1600; then a single poison event every 590s with no other traffic gives overflow_until 2190, 2780, 3370, 3960 — and by t=2180 the registry had drained to 1-2 entries, so the renewals are coming from *successful* arms on a nearly empty registry via the max() at lines 570-577, not from saturation. At t=3950 a never-seen soft key is still reported poison-quarantined, 2950s after the single rejection, and clearing the registry entirely does not release the flag. One poison arm per 600s anywhere in the process keeps the whole service fail-closed indefinitely — and continuous poison arms are the defining traffic pattern of the incident that caused the overflow in the first place.

_Suggested fix:_ Drop the renewal at quarantine.py:570-577 so the deadline really is bounded by the rejected arm's own required lifetime plus the poison deadlines retained at establishment, and release the flag as soon as the registry has a free slot rather than only by wall clock. Emit a metric or alarm while it is active — today only a per-rejection warning log exists. Scoping the evidence per-key (previous finding) removes this entirely.

---

### MINOR — `app/modules/proxy/_service/http_bridge/retry_circuit.py`

**While the global overflow window is active, the durable-load arm guard at retry_circuit.py:887 reads the global flag as 'this key is already quarantined' and skips installing genuine per-key poison fences, so the fail-closed mechanism suppresses the very evidence it stands in for.**

_How it fails:_ The guard is `arm_poison_quarantine and (episode_replaced or not _http_bridge_session_key_poison_quarantined(self, key))`. During overflow the second term is False for every key, so unless episode_replaced is True the arm is skipped. A request touching key K that has no registry entry and a free slot — adopting another replica's at-threshold poison row — therefore records no local fence. When the overflow deadline finally lapses K has no poison evidence; if the next planning touch falls inside _ensure_http_bridge_retry_circuit_loaded_for_key's cache window the row is not reloaded either, leaving a half-open probe planned against the anchor the row's failures were recorded on — the #1852 case poison quarantine exists to prevent. I confirmed the guard term by inspection and by observing that _http_bridge_session_key_poison_quarantined returns True for a key with no registry entry during overflow; I did not build the full durable-row plus expiring-cooldown interleaving, so the post-window dispatch is reasoned from the code path rather than executed.

_Suggested fix:_ Have the arm guard consult the per-key registry entry (e.g. `entry is not None and _http_bridge_quarantine_has_active_poison(entry, now)`) rather than _http_bridge_session_key_poison_quarantined, so a service-level overflow suppresses dispatch decisions without suppressing evidence recording.

---

### MINOR — `app/modules/proxy/_service/http_bridge/quarantine.py`

**_clear_http_bridge_quarantine (line 696) silently discards an explicitly supplied fence unless the caller also passes key_generation_captured=True, reinstating exactly the post-await recapture bug this PR removes.**

_How it fails:_ key_generation became optional (default None) and the new key_generation_captured flag gates whether it is honoured. A caller passing key_generation=<pre-await fence> but omitting the flag falls into the `elif not key_generation_captured:` branch, which overwrites the caller's generation, raw generation and strike count with a fresh capture taken after the awaits — i.e. it clears evidence armed during those awaits, the precise defect #2268 is about. The one production caller (upstream_events.py:3947) does pass the flag, so this is latent rather than live, but the trap has already bitten once inside the diff: the pre-existing test_a_completion_clear_spares_a_strike_armed_during_settlement had to be edited to add key_generation_captured=True, otherwise its explicitly passed fence was ignored. The next completion-path caller gets the old bug back with no error and no test failure.

_Suggested fix:_ Drop the boolean and make the capture explicit in the type: take a _HTTPBridgeQuarantineClearFence (or a sentinel) as a required argument for the primary key, so 'I did not capture' cannot be spelled the same way as 'I captured absence'.

---

### MINOR — `tests/unit/test_proxy_http_bridge.py`

**The detached-predecessor / key-reuse half of #2268 — the half the issue leads with — is covered only by helper-level tests driving an entry path the sole production caller never takes; the one product-path regression covers the other half.**

_How it fails:_ Running the new tests/integration/test_http_promotion_quarantine.py with app/ checked out at the merge base gives '1 failed, 5 passed': only during-completion-first-strike fails, so the product-path signal is entirely about evidence recorded during settlement. The key-reuse tests (test_http_bridge_quarantine_clear_does_not_pop_newer_primary_session_entry, ..._owner_fence_survives_reused_session_id, ..._first_strike_survives_detached_predecessor_completion) use a SimpleNamespace stand-in for the service and call _clear_http_bridge_quarantine(service, predecessor) with no fence, routing through the auto-capture branch that the sole production caller (upstream_events.py:3947, which passes key_generation_captured=True) never takes. A future refactor that keeps the is_primary identity check but changes how the completion path supplies or captures its fence — moving the capture back below an await, say — would leave all three green while the externally visible failure returns.

_Suggested fix:_ Add one route-level case alongside the existing promotion test: complete a turn on session A, quarantine it, let a replacement B take the key, then deliver A's late response.completed and assert the follow-up request still opens fresh instead of re-attaching to the wedged session.

---

### Findings raised and then refuted

Recorded so the next round does not re-litigate them:


- Reviewer 2 checked the neighbouring half of the maintainer's earlier P3 — that dropping a wedged-reattach quarantine at saturation is a new regression — and found it does not reproduce: main's prune evicts the freshly added weaker entry the same way. I did not re-derive that but record it here so it is not re-litigated; only the global poison classification is new relative to the merge base, which I did verify directly (base _prune_http_bridge_quarantine_registry has no service-level flag and its comment states the cap 'holds as a correctness bound, not an unconditional one').

- Reviewer 1's framing of the overflow behaviour as an implementation slip ('the spec claims the window is bounded by the rejected arm's lifetime') is only half right and should not be sent back as a bug in isolation: the blanket classification of unknown keys is exactly what the PR's own delta mandates — 'Unknown keys and subsequently admitted weaker entries MUST remain poison-classified during that window' — so it is designed behaviour requiring a policy decision. What genuinely contradicts the spec is the renewal, which I split out as its own confirmed finding.

Komzpa:
@codex review

Soju06:
Maintainer call on this one: **please split it.**

The #2268 half is ready and I would take it today. It verified well — service-lifetime generations plus the canonical-session / weak-owner identity fence do close the cleanup race, the bridge surface is green at head (1152 passed), and `tests/integration/test_http_promotion_quarantine.py` carries a real product-path regression: with `app/` at the merge base the during-completion-first-strike parameterisation fails while the other five pass. That is exactly the shape this repo asks for.

What I cannot take in the same PR is the poison-overflow policy bundled with it, for two reasons.

**It has a confirmed blocker.** One rejected poison arm sets a *service-global* deadline (`quarantine.py:187`, consulted at `:405`) that makes `_http_bridge_session_key_poison_quarantined` return true for **every key in the process**, not just the rejected one. The window is also not bounded the way the design says — it renews to `now+600s` on every later poison event, from a successful arm as well as a rejected one, so under sustained poison pressure it never closes. That is a fleet-wide behaviour change hiding inside a targeted race fix.

**Your own `tasks.md` says it is not ready.** Item 2.1, "owner acceptance for the bounded window", is unticked. You were right to leave it unticked and right not to assume the answer — but it means this PR is waiting on me for one half while the other half is merge-ready, and the two are coupled only by having been written together.

So: land the #2268 fix on its own and I will merge it. Bring the overflow policy back as its own PR, with the global-deadline defect fixed (the fence needs to be per-key, or the arm needs to stop setting a global) and a bounded window that actually terminates. I will give that one a proper decision rather than a merge gate — the question of what codex-lb should do when the poison registry saturates is a real design question and deserves to be argued on its own terms, not as a rider.

Three smaller things from my review comment travel with whichever PR owns the relevant code: the durable-load arm guard reading the global flag as "this key is already quarantined" and skipping genuine per-key entries (`retry_circuit.py:887`), `_clear_http_bridge_quarantine` silently discarding an explicitly supplied fence unless `key_generation_captured=True` (`quarantine.py:696`), and the detached-predecessor half of #2268 — the half the issue leads with — being covered only by helper-level tests driving an entry path the sole production caller never takes.

Soju06:
Re-triaged at head `2e419652` (unchanged since 2026-09-09). The maintainer's split call from 2026-09-12 stands, and I re-verified the blocker directly in the code rather than restating it — plus two things that make it worse than described.

**The #2268 premise is real and still unfixed on main** (`ca243dec4`): `quarantine.py:44` starts `generation` at 0 and `_quarantine_http_bridge_session` bumps it with `entry.generation += 1` (:271), so an entry recreated after its predecessor was popped restarts at 1 and `clear_fenced` (:396-436) — which compares that integer and nothing else — lets session A's captured fence 1 pop replacement B's quarantine. The unconditional `registry.pop` for an inactive entry (:400-404) is the second half: it erases a first eventless strike recorded during settlement. Your fix for both is sound.

**What still blocks this PR is the overflow policy, and it is not bounded in three separate senses:**

1. *It is global.* `quarantine.py:405` makes `_http_bridge_session_key_poison_quarantined` return `True` for every key in the process. Both consumers turn that into a client-visible 404: `request_submit.py:1139-1181` for any proxy-injected `previous_response_id` (the normal sticky continuation) and `streaming.py:3216-3250` for any anchorless delta-only payload. Worse, the streaming gate evaluates the predicate *before* `await self._load_http_bridge_retry_circuit(session)`, so the global flag short-circuits the only lookup that could disprove it — and for soft `http-history` keys that lookup is a no-op regardless, since `retry_circuit.py:540-543` returns early on `key.strength != "hard"`. A key that can never carry poison evidence therefore gets a 404 it can never clear.

2. *It renews.* `quarantine.py:571-577` extends the deadline to `max(overflow_until, entry.poison_quarantined_until)` on every **successful** poison arm while the flag is live, and `_remember_...` (:211-213) takes `max()` with the existing value. Sustained poison arms are the defining traffic of the incident that saturates the registry, so the window never closes.

3. *It is unfalsifiable.* Grepping the head file, the attribute is written only at :187-213 and :571-577 and read at :177/:405/:571. `_revoke_http_bridge_poison_quarantine` does not clear it, and emptying the registry does not clear it. Only wall clock releases it — no amount of contrary evidence can.

**Concrete path forward**

- You already opened the split as #2345 ("preserves the poison-preserving soft cap and adds no service-wide overflow rule"). That is the PR the maintainer said he would take — but it is currently `CONFLICTING` with the `needs rebase` label. Please rebase it onto current `main` and call out explicitly whether its extra scope (local-vs-durable poison authority and replay-origin propagation) is part of the accepted fix or a third concern, because the accepted scope was the service-lifetime generation plus the canonical-session/weak-owner identity fence.
- Re-scope this PR to the overflow policy alone, against current `main`, with: (a) the service attribute replaced by a bounded per-key structure (e.g. a capped `OrderedDict[_HTTPBridgeSessionKey, float]` of rejected keys with their own deadlines) consulted only for the exact rejected key, and never answered for a key whose `strength` cannot carry retry-circuit evidence; (b) the renewal at :571-577 deleted so the window is bounded by the rejected arm's own required lifetime; (c) a release as soon as a registry slot frees; (d) the arm guard at `retry_circuit.py:887` changed to `entry is not None and _http_bridge_quarantine_has_active_poison(entry, now)` so overflow suppresses dispatch decisions without suppressing evidence recording. Also consider the simpler alternative the maintainer floated: when every slot holds active poison, evict the entry with the earliest `poison_quarantined_until` rather than refusing admission — that keeps the hard cap without inventing a new fail-closed state.
- Whichever PR ends up owning `_clear_http_bridge_quarantine`: drop the `key_generation_captured: bool` parameter. At :696 the `elif not key_generation_captured:` branch silently overwrites an explicitly supplied pre-await fence with a post-await recapture, which is precisely the #2268 bug re-armed for the next caller. Make the parameter a required `_HTTPBridgeQuarantineClearFence` so "I did not capture" cannot be spelled the same way as "I captured absence".
- Finally, add one route-level regression for the half #2268 leads with: complete a turn on session A, quarantine it, let replacement B take the key, deliver A's late `response.completed`, and assert the follow-up request opens fresh. Today that half is covered only by `SimpleNamespace` helper tests driving the auto-capture entry path that the sole production caller (`upstream_events.py:3947`) never takes.

CI is green and there is no migration here, so nothing else gates this — only the scope.

PR #2278, opened 2026-09-09
fix(proxy): count incomplete responses toward bridge retries

### LINKED ISSUE #2273: bug: incomplete responses can bypass bridge retry limits
## What happens

The HTTP bridge can receive repeated incomplete responses without counting them toward its retry circuit.

This terminal takes the missed path:

```json
{
  "type": "response.incomplete",
  "response": {
    "incomplete_details": { "reason": "stream_incomplete" }
  }
}
```

The circuit reads `response.error`, so it misses `stream_incomplete` when the reason appears only in `incomplete_details`. An otherwise equivalent terminal with `response.error.code: "stream_incomplete"` does count. This mismatch remains in the current code.

## Expected behavior

An eligible incomplete stream should count once, regardless of which of those two fields identifies it. Duplicate completion handling must not count it again.

An explicit response error should take precedence. Unknown or missing reasons, and reasons unrelated to transport failure, must keep their current treatment. The client should receive the same terminal payload, with unchanged request logging and account-health behavior.

The regression should exercise the actual HTTP bridge reader with both raw and already-interpreted terminal events, preserving the existing eligibility exclusions.

## Related work

#2278 implements this correction independently, using the circuit recorder already on main. The probe ownership changes remain in #1962 and are tracked in #2272. #1219 preserves incomplete reasons in WebSocket logs; it does not add this HTTP bridge accounting.

JustYannicc:
## Summary

Reason-only `response.incomplete` terminals with `incomplete_details.reason: stream_incomplete` now count toward the HTTP bridge retry circuit. Previously, only the equivalent explicit response error counted.

## Type of change

- [x] `fix:` bug fix

Closes #2273.

## OpenSpec

Change directory: `openspec/changes/count-http-bridge-incomplete-reason/`.

## Changes

Use the existing incomplete-reason reader only when no explicit response error exists. Preserve error precedence, eligibility exclusions, duplicate suppression, downstream payloads, logging and account health. The fix uses main APIs and has no dependency on #1962.

## Test plan

Verified against main `0f6a31c56ac30804ca1c0fac27ca02c6f59bf2b0`:

- `uv run pytest tests/unit/test_http_bridge_reason_accounting.py -q`: 19 passed, covering raw/interpreted terminals, neutral reasons, explicit errors and exclusions.
- Four existing HTTP bridge route tests expanded to 6 passing cases: both Responses endpoints, connection reuse, streaming and valid failure SSE.
- `make lint typecheck`: passed, including architecture, cancellation, timing and settings guards.
- OpenSpec 1.11.0 strict validation: all 65 specs and this change passed.

Both database environment variables pointed to the same dedicated disposable SQLite database. Hosted PostgreSQL and full CI results remain separate gates.

## Screenshots / output

The terminal remains unchanged:

```json
{"type":"response.incomplete","response":{"incomplete_details":{"reason":"stream_incomplete"}}}
```

An eligible attempt consumes one circuit strike; duplicate handling does not consume another.

## Checklist

- [x] Ran the relevant local check subset.
- [x] Strict spec validation and local implementation verification pass.
- [x] Simplicity gates reviewed; no settings or defaults added.


JustYannicc:
@coderabbitai review

@codex review

Please review current head `b69abad0b389982fa071a28eb950cf0110a981f4` against upstream `efe0f581a18a36f4d491b92a36ef79b0c432d252`. Check the focused issue scope and current follow-up.

Soju06:
Triaged against current `main` (ca243dec4). The classifier gap you describe is real, but we cannot justify merging it as written.

**Confirmed:** `app/modules/proxy/_service/http_bridge/upstream_events.py:4263-4270` derives `error_code` solely from `settlement_event.response.error`, so a `response.incomplete` carrying only `incomplete_details.reason` leaves `error_code = None` and the strike guard at `:4281` never fires. That part of the report is accurate.

**Blocking question — does the input shape exist?** `stream_incomplete` is codex-lb's *own* internal transport code, not an upstream `incomplete_details.reason` value:

- `app/core/clients/proxy_websocket.py:377` — "Let Responses relay owners map it to `stream_incomplete`"; it is assigned on an incomplete close handshake.
- `_build_stream_incomplete_terminal_event_for_request` (`app/modules/proxy/_service/streaming/helpers.py`) synthesizes a **`response.failed`** with `error.code = stream_incomplete`; `app/core/errors.py:258-271` leaves `incomplete_details` at `None`. So codex-lb never produces the reason-only shape.
- Every attested upstream reason in this repo is `max_output_tokens` or `content_filter` (`openspec/specs/responses-api-compat/spec.md:4850-4855`, `tests/unit/test_chat_response_mapping.py:226,614`, `tests/unit/test_proxy_http_bridge.py:27398`).

#2273 provides a hand-written JSON blob, not an observation. Please attach one of: a `request_logs` row where the HTTP bridge settled a `response.incomplete` with `incomplete_details.reason = "stream_incomplete"` and no `response.error`, or a captured upstream frame. Without that this is a new poison-class trigger keyed on a string the upstream does not send.

**Blast radius is larger than the description.** The new `error_code` is assigned *before* two other consumers:

1. `upstream_events.py:4271` — the `terminal_error` log detail changes from `None` to `stream_incomplete`.
2. `upstream_events.py:4317` — `_http_bridge_effective_anchor_poison_detail(session, error_code)`; `retry_circuit.py:72` maps `"stream_incomplete" -> "repeated_zero_event_stream_incomplete"`, and with `_HTTP_BRIDGE_RETRY_CIRCUIT_FAILURE_THRESHOLD = 2` (`retry_circuit.py:40`) two consecutive reason-only terminals on a hard key authorize a fenced anchor abandonment and quarantine — the conversation loses its anchor.

The PR states "Preserve ... logging and account health" and "downstream payload ... unchanged". Please state the poison/quarantine consequence explicitly in the proposal and add a test that pins the abandonment behaviour at threshold, or gate the new classification so it strikes the circuit without entering the poison map.

**Two smaller items:**

- The body claims "Four existing HTTP bridge route tests expanded to 6 passing cases", but the PR has `deletions: 0` and modifies no existing test file. Please correct it.
- The re-export line added to `app/modules/proxy/service.py` puts that file at exactly 2600 lines against the `service_lines = 2600` ceiling (`openspec/specs/proxy-architecture/spec.md:16`), leaving no headroom for anything else that touches the facade.

The branch is also `CONFLICTING` (`openspec/specs/responses-api-compat/spec.md` and `service.py` both moved). Rebase after the evidence question is settled — if the shape cannot be observed in production we will close this as a fix for an unreachable state.

PR #2283, opened 2026-09-09
test(shutdown): wait for assertions before completing the response

### LINKED ISSUE #2282: test: preStop checks can outlast the fixture response
## What happens

The preStop process test can fail when its drain-status check runs late. The fixture finishes the active WebSocket response 250 ms after draining begins, but the test still expects `in_flight == "1"`. A delayed client sees `"0"` even though the server behaved correctly.

To reproduce on main, delay `_wait_until_draining` by 700 ms in `test_prestop_commits_deadline_before_sigterm_and_cannot_reopen`, then run that test. The `in_flight` assertion fails. The delay is shorter than the configured drain timeout.

This was reproduced from source with Python 3.13 and an isolated temporary SQLite database. It is a test timing defect; no production shutdown defect was demonstrated.

## Expected behavior

The fixture should keep the response active until the test finishes checking drain commitment, refusal to reopen and late WebSocket rejection. The client should then explicitly release the response, rather than racing a 250 ms timer.

The process suite should pass both normally and with delayed test interaction, preserving the existing production shutdown contract.

## Related work

#2283 extracts this test-only correction from #1953. It is independent of the quarantine and forwarding fixes in #2276 and #2277. The injected delay is a reproduction probe, not a proposed production delay.

JustYannicc:
## Summary

The preStop process test can fail when its drain-status check runs late. Its fixture finishes the active response after 250 ms, before the test verifies that the response is still in flight.

Closes #2282.

## Changes

Add a controlled-completion fixture mode. The test releases the response only after checking drain commitment, refusal to reopen and late WebSocket rejection.

This changes two test files and extracts the remaining shutdown-test correction from #1953. It starts directly from main and has no dependency on the proxy fixes. Production behavior and the existing graceful-shutdown contract are unchanged, so no OpenSpec change is needed.

## Test plan

On main, a 700 ms delay before the drain-status check causes `in_flight` to be `"0"` instead of `"1"`. With the handshake, all five process tests pass with that delay, and all five pass normally.

Lint and formatting pass for both files. Tests and subprocesses used the same isolated temporary SQLite database. The injected delay is a regression probe, not part of the patch. Full repository CI has not run locally; [hosted checks](https://github.com/Soju06/codex-lb/pull/2283/checks) remain required.

### DIFF
--- a/tests/fixtures/graceful_websocket_server.py
+++ b/tests/fixtures/graceful_websocket_server.py
@@ -141,6 +141,10 @@ class _LifecycleApp:
 
         while not shutdown_state.is_draining():
             await asyncio.sleep(0.01)
+        if self._mode == "controlled_complete":
+            release_message = await receive()
+            assert release_message["type"] == "websocket.receive"
+            assert release_message.get("text") == "complete"
         await asyncio.sleep(self._completion_delay_seconds)
         await send(
             {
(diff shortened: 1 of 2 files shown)


Komzpa:
@codex review

PR #2295, opened 2026-09-10
fix(settings): stage zero warmup threshold compatibility

HulianBuligon:


HulianBuligon:
Consolidated into PR #2122 so the compatibility migration and the warm-up behavior change land together. Closing this preparatory PR to keep a single review surface.

Issue #1471, opened 2026-07-24
ci(db): gate long-running migrations with a production-scale duration check


Soju06:
## Problem

The existing migration checks (Alembic policy/drift, PostgreSQL upgrade-from-empty, lock serialization) all run against empty or tiny databases, so a revision that is instant in CI can block startup for many minutes at production data volumes. `20260722_000000_backfill_request_log_useragent_families` passed every gate and then ran for ~10 minutes over a ~3.2M-row `request_logs` on a production deployment, during which the app serves nothing (see #1470 for the runtime-side proposal).

## Proposal

A nightly (or migrations-path-triggered) CI job that:

1. Seeds a PostgreSQL service container with a synthetic production-scale fixture (e.g. 3–5M `request_logs` rows with realistic column distributions; generation script checked in).
2. Runs `python -m app.db.migrate upgrade` across the revisions added since the last stable tag and records per-revision wall time.
3. Fails (or requires a `long-migration` label plus a release-notes callout) when any single revision exceeds a threshold (suggest 30s), so "this release needs a maintenance window" is visible at merge/release time instead of during the rollout.

Complements #1470: that issue removes the startup-blocking property for data backfills; this gate catches whatever remains blocking (schema rewrites, index builds without `CONCURRENTLY`, unexpected table scans).

## Notes

- Threshold and fixture size should be config in one place so they can track production growth.
- Per-revision timing output belongs in the job summary for quick triage.

Soju06:
Status audit (2026-07-27) of open PRs carrying new Alembic revisions, as a baseline for what this gate needs to catch — no open PR implements the gate itself yet.

Effectively instant at production scale (guarded, schema-only DDL on PostgreSQL):
- #1491 — new empty `account_plan_downgrade_observations` table.
- #1487 / #1486 — nullable `TEXT` column on `http_bridge_sessions`. Note both PRs ship the identical revision id `20260725_000000_add_http_bridge_pending_tool_calls`; whichever lands second must drop or re-id its copy.
- #1417 — nullable `DateTime` on `sticky_sessions`. Its `down_revision` points at `20260720_000000_add_request_log_conversation_id`, skipping the 20260722 backfill and 20260724 rollup revisions, so it forks a second head — the Migration checks on that PR are already red.
[...]

Issue #1470, opened 2026-07-24
feat(db): make data-backfill migrations non-blocking, resumable, and observable


Soju06:
## Problem

Data-backfill migrations currently run inside the startup `migrate upgrade` path, as single unfiltered statements, with no progress output. Concretely, `20260722_000000_backfill_request_log_useragent_families` is one `UPDATE` over every matching `request_logs` row:

- On a production-scale table (~3.2M rows, ~2.4M matching) it runs for many minutes while the app **blocks startup and serves nothing** — the deployment looks identical to a hang because neither the runner nor the app logs anything during the statement.
- It is **not filtered** (`WHERE useragent IS NOT NULL AND position('/' in useragent) > 0` but no `IS DISTINCT FROM` guard), so a retry — or an operator pre-running it — re-writes every row again instead of converging to a no-op.
- If the deploy is aborted mid-statement, `docker stop` does not end the server-side query; the zombie transaction keeps holding locks until it is terminated manually, and rolling back to the previous image then requires knowing about `migrate stamp` (see below).

Measured today during the v1.22.0 rollout: the identical backfill expressed as a filtered, batched loop (`... AND useragent_group IS DISTINCT FROM <target expression>`, 20k-row batches) touched only the ~190k rows that actually needed changes and completed in ~30s **online**, with the old version still serving.

## Proposal

1. **Two-phase migrations**: keep schema migrations startup-blocking (seconds), but let a revision declare a *data phase* that runs after the app starts serving — batched, filtered to rows that need changing, resumable across restarts, with a watermark so the runner knows completion.
2. **Runner progress logging**: per revision, log start / elapsed / affected-row counts (and for data phases, batch progress). A deploy operator must be able to distinguish "migrating, 40% done" from "hung".
3. **Recovery hint**: `MigrationBootstrapError` for an unknown-future revision already suggests "run an Alembic downgrade"; it should also mention `python -m app.db.migrate stamp <revision>`, which is the actual one-step recovery when rolling back an image against an already-migrated (additive) schema.

(1) is the substantive design change and probably warrants an OpenSpec change; (2) and (3) are small standalone improvements that could land first.

## Acceptance sketch

- A release containing only guarded schema DDL + a declared data phase starts serving in seconds; the backfill completes in the background and reports progress.
- Re-running any data phase is a no-op (converges, filtered).
- A rollback to the previous image after a partial upgrade is a documented two-step: stamp + retag.

JustYannicc:
[...]

PR #2307, opened 2026-09-10
feat(db): log per-revision migration progress

JustYannicc:
## Summary

Startup migrations currently give operators no per-revision elapsed or failure progress. Online Alembic runs now log each revision before execution, then its elapsed time on completion or failure. The migration CLI enables stderr logging while preserving its stdout result.

Refs #1470, partial implementation of runner progress only. Background data phases, resumability, affected-row counts, batch percentages, recovery guidance, and #1471 benchmark policy remain separate.

## Type of change

- [x] `feat:` new operator-visible capability

## OpenSpec

- [x] Includes a verified, archived OpenSpec change

Change: `openspec/changes/archive/2026-09-10-log-migration-revision-progress/`. Main database-migrations spec and context are synced.

## Changes

A small adapter wraps Alembic's migration-step iterator because its after-apply callback cannot report starts or failures. Alembic retains revision selection, execution, version bookkeeping, and transactions. The adapter restores the callback on exit and propagates the original failure.

`executed` means the step finished, not that the upgrade committed. Progress records contain only revision identity, direction, and elapsed time. They omit SQL, parameters, database URLs, revision descriptions, and exception messages. Existing exception handling outside these records is unchanged.

## Test plan

- Regression failed on the original runner's absent progress records; the CLI stderr regression also failed before logging was enabled.
- Migration suites: 118 passed, 10 PostgreSQL-only skips. Strengthened progress suite: 3 passed, including a revision reading its own start record before executing, failure privacy, no later revision after failure, and no-op behavior.
- Dedicated disposable SQLite: `codex-lb-db upgrade head` and `codex-lb-db check` passed.
- `make lint`, `make typecheck`, strict OpenSpec validation, and simplicity budgets passed.
- Independent source review found no concrete defect. PostgreSQL execution awaits hosted CI. No live runtime changes.

## Screenshots / output

```text
Migration started revision=20260213_000000_base_schema direction=upgrade
Migration executed revision=20260213_000000_base_schema direction=upgrade elapsed_seconds=0.016
```

Elapsed time varies. A failing step emits `Migration failed` and rethrows.

## Checklist

- [x] Conventional Commits title; related issue linked as partial coverage.
- [x] Regression coverage and relevant local checks complete.
- [x] Strict specs validated; implementation verified and archived.
- [x] Simplicity gates reviewed. Zero configuration; no new settings or dashboard changes.
- [x] CHANGELOG untouched.

### DIFF
--- a/app/db/alembic/env.py
+++ b/app/db/alembic/env.py
@@ -6,6 +6,7 @@ from alembic import context
 from sqlalchemy import engine_from_config, pool
 
 from app.core.config.settings import get_settings
+from app.db.migration_progress import revision_progress
 from app.db.migration_url import to_sync_database_url
 from app.db.models import Base
 
@@ -60,8 +61,9 @@ def run_migrations_online() -> None:
             render_as_batch=connection.dialect.name == "sqlite",
         )
 
-        with context.begin_transaction():
-            context.run_migrations()
+        with revision_progress(context.get_context()):
+            with context.begin_transaction():
+                context.run_migrations()
 
 
 if context.is_offline_mode():
--- a/app/db/migrate.py
+++ b/app/db/migrate.py
@@ -928,6 +928,7 @@ def _parse_args() -> argparse.Namespace:
 
 def main() -> None:
     args = _parse_args()
+    logging.basicConfig(level=logging.INFO)
     database_url = get_settings().database_url if args.db_url is None else args.db_url
 
     if args.command == "upgrade":
--- a/app/db/migration_progress.py
+++ b/app/db/migration_progress.py
@@ -1,0 +1,55 @@
+from __future__ import annotations
+
+import logging
+from collections.abc import Iterator
+from contextlib import contextmanager
+from time import monotonic
+
+from alembic.runtime.migration import MigrationContext, RevisionStep
+
+logger = logging.getLogger(__name__)
+
+
+@contextmanager
+def revision_progress(context: MigrationContext) -> Iterator[None]:
+    """Report online revision execution while Alembic retains transaction control."""
+    # Alembic exposes only an after-apply callback. Keep the iterator adapter
+    # here so starts and failures can be reported without replacing its runner.
+    original = context._migrations_fn
+    assert original is not None
+    active: RevisionStep | None = None
+    started = 0.0
+
+    def steps(heads: tuple[str, ...], migration_context: MigrationContext) -> Iterator[RevisionStep]:
+        nonlocal active, started
+        for step in original(heads, migration_context):
+            if not isinstance(step, RevisionStep):
+                yield step
+                continue
+            active = step
+            started = monotonic()
+            direction = "upgrade" if step.is_upgrade else "downgrade"
+            logger.info("Migration started revision=%s direction=%s", step.revision.revision, direction)
+            yield step
+            logger.info(
+                "Migration executed revision=%s direction=%s elapsed_seconds=%.3f",
+                step.revision.revision,
+                direction,
+                monotonic() - started,
+            )
+            active = None
+
+    context._migrations_fn = steps
+    try:
+        yield
+    except BaseException:
+        if active is not None:
+            logger.error(
+                "Migration failed revision=%s direction=%s elapsed_seconds=%.3f",
+                active.revision.revision,
+                "upgrade" if active.is_upgrade else "downgrade",
+                monotonic() - started,
+            )
+        raise
+    finally:
+        context._migrations_fn = original
--- a/openspec/changes/archive/2026-09-10-log-migration-revision-progress/.openspec.yaml
+++ b/openspec/changes/archive/2026-09-10-log-migration-revision-progress/.openspec.yaml
@@ -1,0 +1,2 @@
+schema: spec-driven
+created: 2026-09-10
--- a/openspec/changes/archive/2026-09-10-log-migration-revision-progress/design.md
+++ b/openspec/changes/archive/2026-09-10-log-migration-revision-progress/design.md
@@ -1,0 +1,17 @@
+## Context
+
+Alembic owns revision selection, execution, version bookkeeping and transactions. Its completion callback alone cannot report a revision before it begins or identify a failed revision.
+
+## Decisions
+
+Wrap the Alembic migration-step iterator in the online environment. Yield each original step unchanged so Alembic retains execution and transaction control. Keep this compatibility adapter in one small module. Measure monotonic elapsed time around each yielded step. A surrounding exception handler reports the active revision and rethrows the original exception.
+
+Use INFO for start and executed events, ERROR for failure. The CLI configures logging only when invoked as a command. No global logging setup during imports. Completed execution can still be rolled back by a later failure; use `executed`, not `committed`.
+
+## Risks
+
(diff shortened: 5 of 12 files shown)


Komzpa:
@codex review

Soju06:
The premise is real and the adapter is sound. `scripts/docker-entrypoint.sh:5` and `scripts/distroless-entrypoint.py:11` run `python -m app.db.migrate upgrade` as a separate pre-server process that configures no logging at all, so every `logger.info` there is swallowed by `logging.lastResort` and a multi-minute backfill is indistinguishable from a hang — exactly item (2) of #1470. I traced the pinned runner (alembic 1.19.1, `runtime/migration.py:608-634`) and confirmed the wrapper generator resumes only after each step's `begin_transaction(_per_migration=True)` closes, so `elapsed_seconds` genuinely brackets execution, the final step still gets its completion record, `StampStep` passes through untouched, `_migrations_fn` is restored in `finally`, and the abandoned-generator path on failure emits no spurious `executed` record while `active` still names the failing revision. Driving the real runner in the tests rather than mocking it is the right call.

## Blocker

`app/db/migrate.py:931` — `logging.basicConfig(level=logging.INFO)` lowers the **root** level, and the requirement this PR adds to `openspec/specs/database-migrations/spec.md` ("Progress SHALL NOT include SQL, parameters, database URLs, revision descriptions, or exception messages") is false as a result. Two paths, and the second is the one that matters:

**CLI path (new in this PR).** `_build_alembic_config` builds a bare `Config()` with no ini, so `config_file_name is None` and `app/db/alembic/env.py:13-15` skips `fileConfig` — root level is the only gate in that process. Alembic's own runner does `log.info("Running %s", step)` (`runtime/migration.py:616`), and `MigrationStep.__str__` (`:987-996`) appends `RevisionStep.doc` (`:1025-1027` → `Script.doc`, `script/base.py:838-855`), which is the revision module docstring. `codex-lb-db upgrade` on a fresh database now prints:

```
INFO:alembic.runtime.migration:Running upgrade  -> 20260213_000000_base_schema, create base schema
```

**Server path (pre-existing on `main`, but this PR is what makes it a spec violation).** `database_migrate_on_startup` defaults to `True` (`app/core/config/settings.py:314`), so any deployment that is not the docker/distroless entrypoint — pip/uvx, systemd, Helm with the default — runs migrations in-process via `app/db/session.py:1167` → `app/db/migrate.py:1016` `run_startup_migrations`, inside a uvicorn process whose root logger is INFO (`app/cli.py:170` → `app/core/runtime_logging.py:482-484`). Nothing anywhere in `app/` pins a level on the `alembic.*` loggers. So the same description-bearing line is printed there today, and it keeps being printed after any CLI-only fix. Scoping the CLI handler is therefore not sufficient: the requirement you are adding has to be made true on the startup path too, or narrowed.

**Why CI does not catch it.** `tests/unit/test_migration_progress.py` filters `record.name == "app.db.migration_progress"` before asserting `"private-sql-parameter" not in ...` — it asserts the absence of a description on the one logger that structurally cannot carry one, so the privacy scenario passes vacuously. `test_cli_upgrade_prints_progress_to_stderr` only asserts that two substrings are *in* `result.stderr`; it never asserts what else is on the stream. OpenSpec validation cannot see runtime logger levels at all.

## Three consequences of the same line, worth fixing together

- **No timestamp, wrong format under JSON.** `logging.BASIC_FORMAT` is `"%(levelname)s:%(name)s:%(message)s"`. For a feature whose purpose is telling "migrating" from "hung", a `docker logs` line with no clock does not answer the question. Worse, `log_format` is settable (`app/core/config/settings.py:478`): with `CODEX_LB_LOG_FORMAT=json` the server emits JSON via `build_log_config()` while the migrate process — which runs *first* in the container — emits plain text, so a JSON shipper drops or quarantines precisely the lines this feature exists to surface.
- **Unmanaged rendering path.** `basicConfig` installs the stock `logging.Formatter`, not `UtcDefaultFormatter`/`JsonFormatter`, which are the only formatters that run `_redact_record_text` (`app/core/runtime_logging.py:217`). I checked the call sites in this process (`app/db/migrate.py` and `app/db/migration_lock.py:54`) and none currently interpolates a URL or a secret, so this is not a live leak and it is not in breach of the observability requirement's letter (that requirement is scoped to records rendered *by* those formatters). It is still the wrong shape: the one production process whose entire input is a DSN handed in via `--db-url` should not get a second, unmanaged renderer.
- **Blast radius beyond `upgrade`.** `basicConfig` sits before subcommand dispatch in `main()`, so it also changes `check`/`current`/`stamp`/`wait-for-*`. `codex-lb-db check` goes through `check_schema_drift` (`app/db/migrate.py:645`) → `MigrationContext.configure` and `compare_metadata`, which now dump `Context impl %s.` / `Will assume %s DDL.` (`runtime/migration.py:205-208`) and `Detected added column '%s.%s'` / `Detected removed index %r on %r` (`autogenerate/compare/tables.py:266`, `constraints.py:305`) to stderr, duplicating the stdout drift report on every `make` DB target. The spec text you add describes only the upgrade path.

## What would make this mergeable

1. Replace the bare `basicConfig` with a dedicated stderr handler that uses the product's formatter (`UtcDefaultFormatter` / `JsonFormatter`, selected off `log_format` — `logging.config.dictConfig(build_log_config())` is the cheapest way to inherit both the timestamp and the redaction pass), and keep root at `WARNING`.
2. Raise only the loggers you want to surface: `app.db.migrate`, `app.db.migration_progress`, and `app.db.migration_lock`. Include the lock one deliberately — its `"Waiting for %s held by another process elapsed=%.1fs timeout=%.1fs"` (`app/db/migration_lock.py:54`) is the other half of the "looks hung" story (a peer replica holding the advisory lock) and is invisible today for the same reason. Do not scope so narrowly that it stays dark.
3. Make the requirement true on the startup path as well: `logging.getLogger("alembic.runtime.migration").setLevel(logging.WARNING)` applied on both the CLI path and the in-process migration path — your own records now carry the revision identity, so nothing is lost. If you would rather keep alembic's line, then the spec sentence has to be narrowed to name only the product's own `app.db.migration_progress` records, and that is a maintainer wording call, not something to decide silently in the implementation.
4. Pin both in tests: add `assert "create base schema" not in result.stderr` to `test_cli_upgrade_prints_progress_to_stderr`, and add an equivalent assertion on the startup path (`run_startup_migrations` under a root-INFO config) so the requirement is enforced where it is actually at risk. Drop or supplement the `record.name == "app.db.migration_progress"` filter in the privacy test — as written it cannot fail.

Two non-blocking notes:

- `context._migrations_fn` is private alembic API and `pyproject.toml` pins `alembic>=1.16.5` with no upper bound. The risk is bounded (the image runs `uv sync --frozen`, and the real-runner tests give CI a chance to go red on a bump), and `design.md` acknowledges the tradeoff — but a short comment in `migration_progress.py` naming the alembic version range the attribute was verified against (1.19.1 here) would help whoever does the next dependency bump.
- FYI, the one thing that would have made root-INFO much worse is safe: `sqlalchemy/log.py` pins the `sqlalchemy` logger to WARNING at import, so statement/parameter echo does not switch on.

Sequencing: #2322 appends to the tail of the same `openspec/specs/database-migrations/spec.md` and `context.md`, so whichever lands second needs a rebase.

PR #2308, opened 2026-09-10
fix(proxy): project pooled quotas into Codex rate-limit events

Soju06:
## Summary

Codex CLI can warn that less than 25% of its weekly quota remains when the selected account is nearly exhausted but the pool still has capacity. Native SSE and WebSocket streams currently forward that account's `codex.rate_limits` event, including when API-key quota privacy hides the HTTP headers.

Rebuild these events from the visible pooled quota snapshot so stream-driven warnings and status use the same quota scope as downstream headers.

## Type of change

- [x] `fix:` — bug fix

## OpenSpec

- [x] Includes an OpenSpec change, verified, synced, and archived.

Change: [project-pooled-codex-rate-limits](openspec/changes/archive/2026-09-10-project-pooled-codex-rate-limits/).

Intentional upstream divergence: account-specific `codex.rate_limits` events are projected or omitted, as recorded in [Native Codex stream quotas describe the pool](openspec/specs/responses-api-compat/spec.md#requirement-native-codex-stream-quotas-describe-the-pool). Other native response events retain their existing behavior.

## Changes

- Share the existing quota visibility policy across HTTP, source SSE, HTTP-to-WebSocket bridges, and direct WebSockets.
- Build fresh quota events from pooled windows, resets, and credits; retain original account telemetry for ingestion and attribution.
- Omit events when quota privacy is enabled, no pooled window is known, or the limit family is unrepresented. WebSocket quota-cache failures omit telemetry without interrupting inference; cancellation still propagates.
- Keep public SSE filtering and `/api/codex/usage` API-key budget semantics unchanged. No new configuration or default changes.

With `hide_upstream_quota_from_api_keys=true`, this suppresses account warnings; showing pooled percentages requires the existing visibility setting to allow them.

## Test plan

```text
make lint typecheck
  Architecture, cancellation, timing, settings tiers, Ruff, formatting, ty: passed
openspec validate --specs --strict
  65 capability specs passed; change also passed strict validation before archival
```

Isolated pytest runs covered:

- Responses normalization and API regression suites: 293 passed.
- Quota/privacy, API-key usage, and raw ingestion suites: 67 passed, 4 existing PostgreSQL-only transaction/row-lock tests skipped under SQLite.
- Additional transport and projection checks: 20 passed (includes unit tests from the first run).

New tests exercise real two-account aggregation and API-key privacy, direct/bridged HTTP including trailing slashes, both WebSocket routes, split source SSE, unknown quota families, cache failure, and cancellation. [Verification details](openspec/changes/archive/2026-09-10-project-pooled-codex-rate-limits/verification.md).

## Example output

For native `POST /backend-api/codex/responses`, two equal-capacity Pro accounts with 97% and 36% weekly usage produce 66.5% pooled usage. Previously the selected account's 97% triggered the low-quota warning. With quota visibility enabled, the downstream event instead contains the pooled value (illustrative snapshot without a known reset):

```text
data: {"type":"codex.rate_limits","rate_limits":{"primary":null,"secondary":{"used_percent":66.5,"window_minutes":10080,"reset_at":null}}}
```

With privacy enabled, the quota event is omitted while response deltas and completion continue.

## Checklist

- [x] Conventional Commits title.
- [x] Tests added and existing expectations updated.
- [x] Relevant `make` targets and focused regression suites passed locally.
- [x] Strict OpenSpec validation passed and implementation verified against the archived change.
- [x] Simplicity gates reviewed; no new setup, settings, or dashboard surface.
- [x] CHANGELOG not edited by hand.

### DIFF
--- a/app/modules/proxy/_service/websocket/mixin.py
+++ b/app/modules/proxy/_service/websocket/mixin.py
@@ -502,6 +502,7 @@ from app.modules.proxy.http_bridge_forwarding import (
     OwnerForwardRelayFailure as OwnerForwardRelayFailure,
 )
 from app.modules.proxy.load_balancer import AccountLease, effective_account_concurrency_caps
+from app.modules.proxy.rate_limit_events import project_codex_rate_limit_event, rate_limit_headers_for_client
 from app.modules.proxy.request_policy import (
     apply_api_key_enforcement,
     apply_enforced_service_tier_model_fallback,
@@ -1081,6 +1082,19 @@ async def _process_and_forward_upstream_websocket_text(
         codex_session_affinity=codex_session_affinity,
         clock=clock,
     )
+    if parsed_frame.event_type == "codex.rate_limits":
+        # Raw account telemetry has already been processed/archived. Only the
+        # client projection changes; quota lookups must not fail inference.
+        try:
+            quota_headers = await rate_limit_headers_for_client(api_key, proxy.rate_limit_headers)
+        except Exception as exc:
+            _facade().logger.warning("Unable to load pooled websocket quota metadata: %s", type(exc).__name__)
+            quota_headers = {}
+        pooled_event = project_codex_rate_limit_event(parsed_frame.payload or {}, quota_headers)
+        if pooled_event is None:
+            upstream_control.suppress_downstream_event = True
+        else:
+            downstream_text = json.dumps(pooled_event, ensure_ascii=True, separators=(",", ":"))
     suppress_downstream_event = upstream_control.suppress_downstream_event
     downstream_texts = upstream_control.downstream_texts
     downstream_sequence_request_state = upstream_control.downstream_sequence_request_state
--- a/app/modules/proxy/_service/websocket/protocol.py
+++ b/app/modules/proxy/_service/websocket/protocol.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from collections.abc import Awaitable, Callable
 from typing import Any, Protocol
 
 from app.core.clock import Clock, Scheduler
@@ -72,3 +73,6 @@ class _WebSocketServiceProtocol(Protocol):
     _write_request_log: Any
     _write_websocket_connect_failure: Any
     proxy_responses_websocket: Any
+
+    @property
+    def rate_limit_headers(self) -> Callable[[], Awaitable[dict[str, str]]]: ...
--- a/app/modules/proxy/api.py
+++ b/app/modules/proxy/api.py
@@ -295,6 +295,13 @@ from app.modules.proxy.overflow import (
     resolve_subscription_overflow,
     restore_client_store,
 )
+from app.modules.proxy.rate_limit_events import (
+    hide_upstream_quota_for_api_key_clients as _hide_upstream_quota_for_api_key_clients,
+)
+from app.modules.proxy.rate_limit_events import (
+    project_codex_rate_limit_event,
+    rate_limit_headers_for_client,
+)
 from app.modules.proxy.request_policy import (
     apply_api_key_enforcement,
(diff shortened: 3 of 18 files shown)


Soju06:
Triage (read-only review, round 22). Verified against `origin/main` at `ca243dec4`.

**Premise confirmed.** Native `/backend-api/codex/*` requests run `_normalize_public_responses_stream` with `enforce_openai_sdk_contract=False`, which forwards `codex.rate_limits` verbatim today, so the Codex CLI does see the selected account's weekly usage rather than the pool's. Nothing merged since 2026-09-10 changes that.

**One real defect: the instance-to-instance forward hop silently deletes the event.**

`api.py` (main, lines 1521-1552) forwards a request from the owner instance to the origin instance with:

```
include_rate_limit_headers=False,
...
enforce_openai_sdk_contract=False,
```

and the comment there states the contract explicitly: *"Forward verbatim and let the origin run its own normalization."* `include_rate_limit_headers=False` makes `rate_limit_headers = {}` (main `api.py:6410-6419`). With this PR the owner instance still runs `_normalize_public_responses_stream`, hits the new branch with `enforce_openai_sdk_contract=False` and `codex_rate_limit_headers or {}` == `{}`, `parse_rate_limit_headers({})` returns `None`, `project_codex_rate_limit_event` returns `None`, and the `continue` drops the frame. The origin instance then has nothing left to project, so a multi-instance deployment loses *every* quota event on forwarded turns — including the pooled one this PR exists to deliver. Same class of defect as the multi-instance hop provenance loss caught during #2386's review.

The parameter is already typed `Mapping[str, str] | None`, so the fix is to stop collapsing `None` into `{}` and let the two states mean different things — `None` = "this hop is not the origin, pass through", `{}` = "origin, but hidden or no pooled window, suppress":

1. In `_normalize_public_responses_stream`, gate the new block on the sentinel:

```python
if raw_event_type == "codex.rate_limits" and codex_rate_limit_headers is not None:
    if enforce_openai_sdk_contract:
        continue
    pooled_event = project_codex_rate_limit_event(payload, codex_rate_limit_headers)
    if pooled_event is not None:
        yield format_sse_event(pooled_event)
    continue
```

2. At the `_stream_responses` call site, pass `codex_rate_limit_headers=rate_limit_headers if include_rate_limit_headers else None`.
3. Add a regression test that drives `_stream_responses` with `include_rate_limit_headers=False, enforce_openai_sdk_contract=False` and asserts the upstream `codex.rate_limits` frame reaches the caller unchanged.

**Things I checked that are fine:**

- `credits.balance` as a string matches the repo's own model (`CreditStatusDetailsData.balance: str | None` in `proxy/types.py`; `helpers.py:310` emits `x-codex-credits-balance` as a formatted string), so forwarding the header value is consistent, not a type regression.
- The limit-family discriminator loop mirrors `live_snapshots.parse_rate_limit_event` exactly.
- Projection happens strictly after ingestion: live usage is parsed at `app/core/clients/proxy.py:3666` and `app/modules/proxy/_service/http_bridge/upstream_events.py:2302`, both upstream of this normalization layer, so account attribution and routing state are untouched.
- The WebSocket path is wired correctly (`downstream_text` is consumed by the `elif not suppress_downstream_event` branch below it), and `rate_limit_headers()` is served from `get_rate_limit_headers_cache()`, so the per-event lookup is not a DB round trip.

**Also blocking: rebase.** The only code conflict is the import block in `app/modules/proxy/api.py` — main dropped the `app.modules.proxy.overflow` import group (#2416 reverted the overflow model-source fallback), so keep main's import layout and re-add only the two `app.modules.proxy.rate_limit_events` imports. The rest are the usual append points in `openspec/specs/responses-api-compat/{spec.md,context.md}`; note #2303 appends to the same file.

PR #2319, opened 2026-09-10
fix(proxy): contain post-terminal health write failures

JustYannicc:
## Summary

A failing account-health write after a Responses terminal can emit a second `response.failed` event or abort stream completion. Catch and log ordinary post-terminal health exceptions while preserving the original response, reservation settlement order, and cancellation propagation.

## Type of change

- [x] `fix:` — bug fix (no behavior change beyond the bug)

Linked issue: Fixes #2033

## OpenSpec

- [x] This PR includes / updates an OpenSpec change
- [x] This PR touches a codex-faithful path and preserves upstream-equivalent behavior

Change directory: `openspec/changes/archive/2026-09-10-fix-post-terminal-health-failures/`

## Changes

- Use one exception-logging helper at the six post-terminal error-health writes. Existing settlement gates, queued-account deduplication, and HTTP status forwarding stay intact.
- Add a real ASGI `/v1/responses` regression and four stream-handler cases covering keyed owner continuations and an unkeyed stream with visible text followed by a quota error.

Pre-terminal health writes, settlement failures, and success recording keep their existing behavior. This is independent of unmerged PRs, including #1905.

## Test plan

Both database environment variables pointed to the same dedicated disposable SQLite database before app imports and tests.

```bash
uv run pytest tests/unit/test_proxy_utils.py \
  -k 'stream and (health or settlement or terminal or owner_rewrite)' -q
# 72 passed
uv run pytest tests/integration/test_proxy_responses.py \
  -k test_v1_responses_health_write_failure_keeps_one_terminal -q
# 1 passed
make lint typecheck
# Passed, including architecture and cancellation checks
pnpm dlx @fission-ai/openspec validate --specs --strict
# 65 passed
```

The four new stream-handler cases and the public HTTP test fail against pinned main `0f6a31c56` and pass with this patch. The active change passed strict validation before archive; main specs passed again after sync. Independent review found no actionable defects. Hosted CI and maintainer review remain pending.

## Screenshots / output

Public request: `POST /v1/responses` with `{"model":"gpt-5.1","instructions":"hi","input":"hello","stream":true}`. A synthetic upstream emits text and then raises a quota error; the injected health write raises `RuntimeError`.

- Before: `response.failed/usage_limit_reached`, then `response.failed/upstream_error`.
- After: one `response.failed/usage_limit_reached`; the health failure is logged with exception information.

## Checklist

- [x] Title is in Conventional Commits format.
- [x] Linked the related issue above.
- [x] Added tests covering the change.
[...]

PR #2322, opened 2026-09-10
feat(db): explain guarded migration recovery stamping

JustYannicc:
## Summary

Explain guarded metadata-only stamping when an image rollback encounters an unknown migration revision. The hint requires schema and data compatibility, ended migration transactions, a recoverable backup and encryption key, and a build containing both revisions.

## Type of change

Linked issue: Partial #1470, recovery guidance only. Data phases and progress reporting remain separate; this does not close the issue.

## OpenSpec

Change directory: `openspec/changes/archive/2026-09-10-guarded-migration-recovery-hints/`, verified and synced to `database-migrations`.

## Changes

- Preserve the failed upgrade and all existing migration execution and stamp behavior.
- Document recovery preconditions and checks before resuming traffic. Stamping does not undo schema or data. The older build cannot stamp a current revision absent from its migration scripts.

## Test plan

- CLI regression failed on the missing hint before the change and passed afterward. Failed upgrade and stamp commands leave disposable SQLite fixtures byte-for-byte unchanged.
- `uv run pytest tests/integration/test_migration_recovery_hints.py tests/unit/test_db_migrate.py -q`: 61 passed.
- Disposable SQLite `python -m app.db.migrate upgrade head` then `check`: `migration_policy=ok`, `schema_drift=none`.
- `make lint typecheck`: passed.
- Strict OpenSpec change validation and `openspec validate --specs --strict`: passed, 65 specs.

Both database environment variables used dedicated disposable SQLite targets. No PostgreSQL execution or live rollout is claimed. Independent read-only review found no actionable findings.

## Screenshots / output

The existing unknown-revision error now includes this conditional guidance:

```text
For an image rollback only after verifying schema and data compatibility with the rollback image, ensuring migration transactions have ended, and preserving a recoverable database backup and encryption key: run `python -m app.db.migrate stamp <revision>` from a build containing both the recorded and target revisions, against the same database. Stamping only changes migration metadata; it does not roll back schema or data.
```

## Checklist

- [x] Ran `uv run pre-commit run local-ci --hook-stage manual --all-files` or the relevant `make <target>` subset locally.
- [x] If touching specs: `openspec validate --specs` passes and `/opsx:verify` is clean.
- [x] Simplicity gates reviewed: the six simplicity rules (PRINCIPLES.md P1-P6).

### DIFF
--- a/app/db/migrate.py
+++ b/app/db/migrate.py
@@ -673,7 +673,12 @@ def _schema_ahead_error(state: MigrationState) -> MigrationBootstrapError:
     return MigrationBootstrapError(
         f"Database schema revision(s) {','.join(state.unknown_revisions)} are not known to this build "
         f"(head={state.head_revision}); the schema was likely migrated by a newer version. "
-        "Deploy a matching or newer image, or run an Alembic downgrade to a revision this build knows."
+        "Deploy a matching or newer image, or use a build containing the required migrations for a reviewed "
+        "Alembic downgrade. For an image rollback only after verifying schema and data compatibility with the "
+        "rollback image, ensuring migration transactions have ended, and preserving a recoverable database backup "
+        "and encryption key: run `python -m app.db.migrate stamp <revision>` from a build containing both the "
+        "recorded and target revisions, against the same database. Stamping only changes migration metadata; "
+        "it does not roll back schema or data. See docs/database.md for recovery preconditions and verification."
     )
 
 
--- a/docs/database.md
+++ b/docs/database.md
@@ -42,6 +42,23 @@ Compose volume. If that guard fires, run the `postgres-upgrade` profile above be
 It also refuses nested `/var/lib/postgresql/data` directories that still report a pre-18 major version, because those
 layouts need an explicit pg_upgrade before the Postgres 18 container can safely open them.
 
+## Recovering an image rollback with an unknown revision
+
+If startup reports a revision unknown to the image, first deploy a matching or newer image. An older image cannot infer compatibility from the revision label.
+
+Stamping changes migration metadata only. It does not undo schema changes, restore data or cancel database transactions. Even additive changes can break old readers or writes through new constraints and defaults.
+
+Use a metadata-only stamp for an image rollback only after these checks:
+
+1. Stop application writers and migration jobs. Confirm their database transactions have ended, including server-side queries from an interrupted deployment. Stopping a container alone is not proof.
+2. Preserve a consistent, recoverable backup of the current database, its encryption key and required data files. Record the current image, database target and recorded revisions. Do not replace newer writes with an old backup.
+3. On a disposable copy, verify that the rollback image can read and write the existing schema and data. Review every intervening schema and data migration, including incomplete backfills. If compatibility is unknown or incompatible, keep the matching image or plan a reviewed downgrade using the required migration scripts.
+4. Select the exact revision expected by the rollback image. Use a build containing both the recorded and target revisions. The old build cannot stamp an unknown current revision; do not delete or purge migration metadata to bypass that error.
+5. Only after those checks, run `python -m app.db.migrate stamp <revision>` in that migration-capable build, replacing `<revision>` with the verified target. Keep the same explicit database configuration. Do not use the newer build's `head`, which can name a different revision.
+6. Switch to the rollback image while writers remain stopped. Run `python -m app.db.migrate current` and `python -m app.db.migrate check` against the same database. Retain the encryption key, verify startup and the rehearsed read/write behavior, then resume traffic. A successful stamp or schema check alone does not prove application compatibility.
+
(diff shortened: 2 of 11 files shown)


JustYannicc:
@coderabbitai review

Please review current head `dbdd591c500bb1a568aa458ae51e2b2d0ca522a8`. The prior coverage pass found missing substantive coverage for this head. The previously requested #2319 review has completed and confirmed available included capacity. This is one incremental review request; preserve earlier applicable findings and review the remaining changes.

JustYannicc:
@coderabbitai review

Please review exact head `dbdd591c500bb1a568aa458ae51e2b2d0ca522a8`. The latest attempt was rate limited at 13:09 UTC with a 13-minute cooldown. That boundary has passed; no review is in progress and no completed coverage is recorded for this head. This is one request after the changed cooldown condition.

Komzpa:
@codex review

Soju06:
Thanks — the gap is real (`_schema_ahead_error` on `main` still offers only "matching/newer image or Alembic downgrade", and #1470 item 3 asks for exactly this hint). But the runbook as written dead-ends under this repo's own drift guard, so I can't recommend merging it yet.

**Blocking — the documented sequence cannot succeed with default settings.**

Step 6 of the new `docs/database.md` section says to run `python -m app.db.migrate check` against the rollback image and then "verify startup ... then resume traffic". After a metadata-only stamp back, neither works:

- `check_schema_drift` (`app/db/migrate.py:645-674`) compares the live schema against the *rollback image's* `Base.metadata`. Every table/column the newer revisions added is physically present and absent from that metadata, so `compare_metadata` returns `remove_table` / `remove_column` diffs.
- `_is_ignored_schema_drift` (`app/db/migrate.py:608-628`) only excuses a fixed `_LEGACY_EXTRA_COLUMNS` allowlist and one SQLite `REAL`/`FLOAT` case. Your own worked example — "a new release might add an optional column that the previous release ignores" — lands directly in the *unignored* `remove_column` path.
- The CLI therefore prints `schema_drift_detected` and `raise SystemExit(1)` (`app/db/migrate.py:1211-1215`).
- Startup then raises `Schema drift detected after startup migrations: ...` (`app/db/session.py:1231`) and re-raises it, because `database_migrations_fail_fast` defaults to `True` (`app/core/config/settings.py:399`, tier T0).

Failing scenario, concretely: release N+1 adds one nullable optional column; the operator does everything in steps 1-5, including proving read/write compatibility on a disposable copy; at step 6 `check` exits 1 and the rollback container never becomes ready, with no guidance in the runbook about what they are looking at.

This isn't incidental — `openspec/specs/database-migrations/spec.md:24-45` ("Startup schema drift guard") *requires* fail-closed on drift. The requirement this PR appends therefore contradicts an existing requirement in the same spec file. Please pick one and make it explicit:

- **(a)** Keep the stamp route and finish it: add a step stating that `check` will report `schema_drift_detected` listing exactly the newer objects, that this is expected, and that the rollback image must be started with `CODEX_LB_DATABASE_MIGRATIONS_FAIL_FAST=false` (`docs/reference/settings.md:93`) — plus an explicit instruction to restore the default the moment a matching image is redeployed. Then amend the new spec requirement to reference the drift guard so the two requirements stop contradicting each other. Note this is a T0 setting and disabling it also suppresses migration-failure fail-fast, so the docs should say that out loud.
- **(b)** Drop the stamp route from the runbook and document the reviewed physical downgrade as the only supported rollback.

I'd take (a), but it needs the override named, not implied.

**Also blocking (small):**

1. `app/db/session.py:1167-1181` carries a second, independently hard-coded schema-ahead diagnostic for `CODEX_LB_DATABASE_MIGRATE_ON_STARTUP=false`. It still says only "deploy a matching or newer image, or run an Alembic downgrade to a revision this build knows." Deployments running a dedicated migration job — the population most likely to roll an image back — never see the new guidance. Extract a shared formatter and use it in both places.
2. The error text ends with `See docs/database.md`. That path does not exist in the published image: `Dockerfile:71-76` copies only `/opt/venv`, the egress binary, `app`, `config`, `scripts`, and `app/static`. Point at the published page instead: `https://soju06.github.io/codex-lb/database/#recovering-an-image-rollback-with-an-unknown-revision` (`mkdocs.yml:3`, `mkdocs.yml:69`).

**Non-blocking:**

3. `tests/integration/test_migration_recovery_hints.py:62` asserts Alembic's internal `"Can't locate revision identified by '20990101_000000_future'"`. That string belongs to Alembic, not to any contract of ours, and will break on an Alembic bump. Assert non-zero exit + the revision id appearing in stderr + unchanged file bytes, which is what the scenario actually claims.

**Merge order:** #2307 appends to `openspec/specs/database-migrations/context.md` at `@@ -72,3` and `openspec/specs/database-migrations/spec.md` at `@@ -511,3` — the same anchors this PR uses. Whichever lands second will need a rebase.

PR #2326, opened 2026-09-10
fix(warmup): preserve live reset evidence across skipped polls

### LINKED ISSUE #1975: bug(warmup): live usage ingestion can consume reset evidence without invoking limit warm-up
### Pre-flight checklist

- [x] I searched open and closed issues and confirmed this is not a duplicate.
- [x] I am running a supported codex-lb version (see version note below).

### codex-lb version

`v1.24.0`

The production observation was made on `v1.24.0-beta.4`. I verified that these three relevant files are byte-identical in `v1.24.0-beta.4`, the supported `v1.24.0` tag, and current `main` at `02113fd`:

- `app/modules/usage/live_ingest.py`
- `app/core/usage/refresh_scheduler.py`
- `app/modules/limit_warmup/service.py`

### Deployment method

Other: native arm64, uv-managed local installation, one process, SQLite.

Codex app (desktop / web)

Pro. The race does not appear plan-specific.

### Model(s) involved

Normal proxied traffic used a Codex model. The configured limit warm-up model was `auto`.

### What happened?

A reset transition persisted by live usage ingestion can be consumed without ever invoking reset-confirmed limit warm-up.

The two persistence/trigger paths currently have different behavior:

1. `LiveUsageIngestor._ingest()` writes the same `UsageHistory` row shape as the poller, but does not invoke `LimitWarmupService` or persist a durable reset event.
2. `UsageRefreshScheduler._refresh_as_leader()` reads `before_*`, calls `UsageUpdater.refresh_accounts()`, and invokes limit warm-up only inside `if usage_written:`.
3. If live ingestion has just written the reset row, the updater's freshness gate skips the poll. `usage_written` is therefore false and the warm-up evaluator is not called.
4. On a later tick, the reset row is already the `before` sample, so the exhausted-to-available adjacent pair has been lost.

In an anonymized production trace, consecutive usage rows showed a real `51% -> 0%` transition and a forward `reset_at` jump. Calling the shipped `usage_reset_confirmed(before, after)` on that exact pair returns `True`, but there was no corresponding row in `account_limit_warmups`. A different account whose `82% -> 0%` transition was captured by the controlled poller seconds later did create and successfully complete a limit warm-up attempt.

This is the same race class already identified for quota-planner probes in PR #1273:

https://github.com/Soju06/codex-lb/pull/1273#discussion_r3570062571

That review was addressed with `suppress_live_usage` for quota-planner probes only:

https://github.com/Soju06/codex-lb/pull/1273#discussion_r3571661960

The reset-confirmed `LimitWarmupService` path remains uncovered.

An exhausted-to-available transition accepted from any authoritative usage source should trigger the same reset-confirmed warm-up state machine exactly once.

Live ingestion should either evaluate/enqueue the reset transition itself, or the scheduler should retain and consume unprocessed reset evidence even when the next poll is skipped for freshness.

### Steps to reproduce

```markdown
1. Enable global and per-account limit warm-up and leave live usage ingestion enabled.
2. Persist an exhausted usage sample for an eligible account.
3. Send normal proxied traffic whose response headers or `codex.rate_limits`
   event publish an available sample with a later `reset_at`.
4. Let `LiveUsageIngestor` persist that sample immediately before the account's
   background refresh tick.
5. Observe that `UsageUpdater.refresh_accounts()` skips the fresh account and
   returns `usage_written = false`.
6. Verify that `usage_history` contains the reset pair but
   `account_limit_warmups` contains no attempt for that reset.
7. Run `usage_reset_confirmed()` directly on the adjacent rows; it returns true.
```

### Relevant logs

No exception is logged. This is a silent missed trigger: the live usage row is written successfully and the subsequent poll is legitimately skipped as fresh.

```yaml
database: sqlite
replicas: 1
architecture: arm64
limit_warmup_enabled: true
limit_warmup_windows: secondary
limit_warmup_model: auto
per_account_limit_warmup_enabled: true
live_usage_ingestion_enabled: true
usage_refresh_interval_seconds: 60
```

### Additional context

A regression test should cover a live-ingested exhausted-to-available transition followed by a freshness-skipped poll and assert exactly one warm-up attempt. Duplicate live snapshots and the next normal poll should not create a second attempt.

A durable "last reset evidence consumed" marker or a small reset-event queue may be safer than relying on whichever writer happens to create the latest adjacent pair.

JustYannicc:
## Summary

Live usage can persist a confirmed reset before the scheduler polls. A freshness skip then loses the warm-up trigger. This change recovers the retained reset pair and consumes it through the existing durable attempt claim.

## Type of change

- [ ] **Breaking change**

Linked issue: Fixes #1975

## OpenSpec

- [ ] This PR touches a codex-faithful path and preserves upstream-equivalent behavior

Change directory: `openspec/changes/archive/2026-09-10-preserve-live-reset-evidence/`

## Changes

- Evaluate selected-account reset evidence even when polling writes nothing.
- Anchor history at the first retained observation of the current reset identity, including its predecessor. This survives restart and delayed deadlines.
- Apply current quota availability and opt-in state to historical reset proof. Keep bootstrap, paid-to-Free and idle triggers gated on actual poll writes.
- Add the ingestion/scheduler regression suite to hosted PostgreSQL coverage.

The existing atomic claim prevents duplicate attempts after repeated ingestion, scheduler restarts and concurrent workers. Already-claimed windows skip history lookup. The first-reset lookup can scan retained account/window history, and the following span is not row-capped. No new ledger, migration, setting or deployment is included.

## Test plan

All local tests used equal main/test URLs pointing to a dedicated disposable SQLite database.

```text
uv run pytest tests/unit tests/simulation tests/test_request_logs_options_api.py -q --tb=short --maxfail=3
9680 passed, 101 skipped, 1 known xfailed

Affected ingestion, scheduler, warm-up, atomic-claim and usage-repository suite:
171 passed, 19 PostgreSQL-only checks skipped

make lint typecheck: passed
OpenSpec change validation --strict: passed
OpenSpec canonical validation --specs --strict after sync/archive: 65 passed
```

The reported path was red on the pinned base using real ingestion, the real updater freshness decision and scheduler lifecycle. Current tests cover duplicate snapshots, later polling, restart, concurrent schedulers, all prior-attempt states and current eligibility. Independent standards and correctness review found no remaining actionable findings on the final source tree. PostgreSQL execution remains a hosted check.

## Screenshots / output

```text
Live 51% -> 0% reset, freshness-skipped poll:
  before: no warm-up send
  after: one warm-up send
Duplicate ingestion + restart + later normal poll:
  still one warm-up send
```

## Checklist

- [x] Ran the relevant local make targets and affected test subset.
- [x] Strict OpenSpec validation passes and verification is complete.
- [x] Simplicity gates reviewed: no new setup, setting, README or dashboard surface.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@ INTEGRATION_CORE_SHARD_COUNT := 3
 POSTGRES_PYTEST_TARGETS := \
 	tests/integration/test_cost_backfill.py \
 	tests/integration/test_atomic_quota_warmup_claims.py \
+	tests/integration/test_live_reset_warmup.py \
 	tests/integration/test_report_rollup.py \
 	tests/integration/test_reports_performance_api.py \
 	tests/integration/test_migrations.py::test_postgresql_migration_contract_policy_and_drift_match \
--- a/app/core/usage/refresh_scheduler.py
+++ b/app/core/usage/refresh_scheduler.py
@@ -21,6 +21,7 @@ from app.db.session import detach_session_objects, get_background_session
 from app.modules.accounts.background_repository import BackgroundAccountsRepository
 from app.modules.accounts.repository import AccountsRepository
 from app.modules.limit_warmup.repository import LimitWarmupRepository
+from app.modules.limit_warmup.reset_evidence import recover_current_reset_evidence
 from app.modules.limit_warmup.service import (
     LimitWarmupService,
     StreamingLimitWarmupSender,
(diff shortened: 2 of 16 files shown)


Komzpa:
@codex review

Soju06:
Triage note (read-only review; no code changed on our side).

Verified on `origin/main` @ ca243dec4: **the bug is real and still present.** `UsageRefreshScheduler._refresh_as_leader` (app/core/usage/refresh_scheduler.py:240-320) still gates all reset-evidence resolution and `run_after_usage_refresh` behind `if usage_written:`, and live ingestion does not invoke warm-up, so the #1975 sequence still drops the trigger permanently. The general approach here is sound. Four things need to change first.

**1. Rebase — #2405 rewrote this exact block.** 3d501ac5e (merged 09-13) changed refresh_scheduler.py +267/-66 and tests/unit/test_usage_refresh_scheduler_recovery.py +783. `_resolve_monthly_reset_evidence` (the name in your diff) is now `_resolve_reset_evidence`, returning `.warmup` and `.recovery` (refresh_scheduler.py:271-290, definition at :572), and it already does an anchored search over retained `usage_history` per quota-window slot. Please rebase, and where possible fold `recover_current_reset_evidence` into `_resolve_reset_evidence` rather than adding a second, parallel history scan.

**2. The Codex P1 is real — I confirmed it in the source, it is not a false positive.** `UsageUpdater.refresh_accounts` (app/modules/usage/updater.py:289-390) returns a single bool that collapses at least three different outcomes into `False`: a freshness skip, a usage-refresh auth cooldown (`_is_usage_refresh_in_cooldown` → `continue`), and a genuinely failed fetch (`AccountRefreshResult(usage_written=False, fetch_succeeded=False)`; the comments right there deliberately separate `fetch_succeeded` from `usage_written`). Your diff calls `recover_current_reset_evidence` and `run_after_usage_refresh` unconditionally, and the reset-confirmed branch in `_build_candidate` is the one branch you did **not** guard with the new `usage_written` flag (paid-to-free, initial-free-quota and staggered-idle all were). So on a tick where the poll failed outright, or where the account is sitting in an auth cooldown, an account with retained evidence still creates and consumes its one durable warm-up claim — and since the new `has_attempt()` matches any attempt row regardless of status, a warm-up that fails for the same transient reason can never be retried by a later healthy poll. Gate the recovery path on the genuine "skipped as fresh / succeeded with no write" outcome; the scheduler passes exactly one account, so returning the per-account `AccountRefreshResult` instead of a bool is the cheapest fix.

**3. Per-tick cost (Codex P2, partly real).** Hoisting the three `latest_by_account` calls, `settings_repo.get_or_create()` and `accounts_repo.list_accounts(refresh_existing=True)` out of `if usage_written:` makes them run on every scheduler tick rather than only on ticks that wrote usage. `list_accounts` is already called at the top of the same function, so the roster is now loaded twice per tick.

**4. Unbounded history scan.** `first_reset_observed_at` filters `account_id + window + reset_at BETWEEN ...` ordered `recorded_at ASC LIMIT 1`. There is no index on `reset_at` (app/db/models.py:2789-2830), and the row you want — the *first* observation of the *current* reset — sits at the far end of the ascending range, so the planner walks the account's whole retained window history. On PostgreSQL `idx_usage_window_account_time_covering` carries `reset_at` in its INCLUDE payload so it can stay index-only; SQLite has no INCLUDE. And because `has_attempt` is false for every reset that never produced a warm-up (the common case), this now runs on essentially every tick. Please bound the lookback the way #2405 anchors its search, or add a covering index.

For the record: dropping the `before is None` early return in `_build_candidate` is fine — `usage_reset_confirmed` (limit_warmup/service.py:751-754) handles `None` — and terminal-status accounts are already excluded by `_account_is_safe_candidate`, so those are not concerns. The `🤖 codex: needs work` label is still set and neither finding has a reply. Happy to re-review once the rebase and (2) land.

PR #2251, opened 2026-09-09
feat(frontend): fold the auth cards into one Access card that grows with the team

Soju06:
## Summary

Stacked on the login/header PR (`feat/login-and-header-tiering`, #2237). Last Phase-1 PR of the dashboard RBAC plan: today's four auth cards in Settings (Guest access, Password, Session length, Two-factor) fold into **one Access card** that grows with the install. One account → the same four controls under a single heading plus one sentence and an "Invite a teammate" button. Two or more accounts → a People tab (accounts table, row actions, invite dialog, pending invites, read-only roles) next to a "My sign-in" tab. `/settings/access` is the full-page escape hatch for larger teams. No new nav item.

## Type of change

- [x] `feat:` — new user-facing capability (frontend)

Linked issue: completes the UI half of #673 (user management). `Closes #673` is left to the maintainer's judgement since SSO/self-service phases follow.

## OpenSpec

- [x] This PR includes an OpenSpec change

Change directory: `openspec/changes/access-settings-card/` (`frontend-architecture` 6 ADDED + MODIFIED "Settings page" with the full block; `openspec validate --strict` passes)

## Changes

**Access card** (`features/settings/components/access/`): `access-card.tsx` shell (`#access` / `#access-people` deep links via the existing `advanced-settings-deeplink.ts` convention), `access-solo-body.tsx` (individual tier: one sentence, "Invite a teammate", or "set a password first" → the existing password setup dialog; the four existing control components render below in the same order, files untouched), `access-my-sign-in-tab.tsx`, `access-people-tab.tsx` + `people-row-actions.tsx` + `role-badge.tsx` (permissions on hover), `invite-dialog.tsx` (role select limited to the server's `assignableRoleIds`, default Operator, admin warning, first-invite note, link shown once with copy + 24 h notice), `pending-invites-sheet.tsx` (resend rotates, revoke), `roles-sheet.tsx` (five preset cards, read-only, member greyed "coming later"), `access-page.tsx` (`/settings/access`, redirects without `users:manage`).

**Tiering**: individual body only when the tier is individual; People tab only when the account holds `users:manage`; a teammate without it sees the card with the "My sign-in" controls only. Row actions hide what the server refuses (self row, migrated `admin` row); server 409s (`last_admin_protected`, `insufficient_delegation`, `compat_user_locked`, `invite_pending`) surface inline. Every mutation invalidates the users/invites/roles queries and refreshes the session so the card falls back to the solo body when an install returns to one account.

[...]

PR #2328, opened 2026-09-10
feat(auth): enable the operator and viewer presets end to end, require TOTP for admin-level accounts

Soju06:
## Summary

Stacked on the Access-card PR (`feat/access-settings-card`, #2251). First Phase-2 PR of the dashboard RBAC plan: the Operator and Viewer presets become real end to end. The backend permission matrix is verified for both roles through the product path (invite → accept → sign in), the frontend gates every control on the exact permission its route requires instead of the coarse write alias, and administrators can be required to use two-factor independently of the global toggle (`totp_required_for_admin_role`, D9).

## Type of change

- [x] `feat:` — new capability

Linked issue: part of #673; not closed.

## OpenSpec

Change directory: `openspec/changes/enable-operator-viewer-presets/` (`admin-auth`, `frontend-architecture`, settings deltas; `openspec validate --strict` passes)

## Changes

**Admin-level two-factor** (`dashboard_access.py`, `dependencies.py`, settings): `is_admin_level(grants)` = admin preset or any privileged permission (`security:write`, `users:manage`, `roles:manage`, `accounts:export`, `conversations:read`, `audit:read`); `totp_policy_applies(...)` is the single predicate behind session validation, the session response and the management-session checks. New setting `totp_required_for_admin_role` (DB column, migration `20260910_000000`, default off; edited from the People tab with `security:write`). Enabling either two-factor toggle requires the acting account to have a secret, read from the account row (the legacy settings column is no longer consulted); the response reports how many accounts / administrators still have to enrol. Turning on the global toggle while the migrated `admin` has a password but no secret is refused (`409 compat_user_locked`) because a previous-release replica cannot handle that state.

**Operator / Viewer separation**: integration tests drive both roles through every dashboard route class and pin the matrix (operator: accounts, keys, ops; 403 on conversations, audit, credential export, security settings, user management; viewer: reads only, masked account identities). No route gate needed changing. Frontend controls are gated on the permission their route checks (`conversations:read`, `accounts:export`, `api_keys:*`, `security:write`, `ops:write`, `users:manage`); mutation controls that still sit behind the coarse write alias stay gated on it so custom roles never see dead ends. Personal password/two-factor controls are available to every signed-in account (incl. Viewer), not only to write-capable ones.

**People tab**: "Require two-factor for administrators" switch with the enrolment-count hint, inline handling of the guard and of settings conflicts; the statement line reads the configured policy.

**Tests**: `tests/integration/test_operator_viewer_presets.py` (matrices, policy binding incl. a custom role holding `audit:read`, guards, counts, audit, migration), unit tests for the predicates, vitest for every gating change with Operator and Viewer sessions and for the new switch.

## Simplicity

- [x] Zero config; no new env var, README section, or nav item (the flag is a DB setting edited in the UI).
- New setting(s): `totp_required_for_admin_role` (DB, default off)
- [x] Budgets unchanged.

## Screenshots / output

Before/after PNGs for the People tab sign-in requirements row and the Settings page as Operator and as Viewer: `~/work/codex-lb/rbac-plan-0908/screenshots/pr-2a/{before,after}/`. Attached to the PR thread on request.

## Test plan

```
uv run ruff check . && uv run ruff format --check . && uv run ty check
uv run pytest tests/integration/test_operator_viewer_presets.py tests/integration/test_dashboard_route_permission_matrix.py \
  tests/integration/test_dashboard_permission_gates.py tests/integration/test_settings_api.py tests/unit/test_settings_service.py -q
cd frontend && bun run lint && bun run typecheck && bun run test
openspec validate enable-operator-viewer-presets --strict
codex review --base feat/access-settings-card        # 5 P2 fixed (+ 6 adversarial findings)
```


Soju06:
`Contributors attribution` fails here for a reason unrelated to this PR: the check reports two GitHub contributors (`nikitamgrimm`, `victorstatko`) missing from the all-contributors list. It passed on #2251 on 09-09, so the list needs updating on main; until then every new PR fails this check (and `CI Required` with it). Everything else on this PR is green.

PR #2331, opened 2026-09-10
feat(auth): resolve trusted-header identities to accounts through sign-in providers

Soju06:
## Summary

Stacked on the operator/viewer PR (`feat/enable-operator-viewer-presets`). Phase-2 SSO skeleton: sign-in methods become **providers** (a `dashboard_auth_providers` table with a Password row and a Trusted-header row), every external identity goes through one `IdentityResolver`, and trusted-header requests now resolve to **real accounts** (created on first sight, "JIT") instead of an anonymous implicit admin. Existing reverse-proxy installs keep their behaviour: unknown identities still become administrators by default (D10), now as accounts that appear in the People tab and in the audit log. Administrators can pre-create an account for a proxy identity so the person gets the intended role the moment the proxy first sends that value.

## Type of change

- [x] `feat:` — new capability

Linked issue: part of #673; not closed.

## OpenSpec

Change directory: `openspec/changes/auth-provider-abstraction/` (new capability `identity-providers`; `admin-auth`, `dashboard-users`, `frontend-architecture` deltas; `openspec validate --strict` passes)

## Changes

**Schema** (revision `20260910_010000`): `dashboard_auth_providers` (`kind`, `provider_key`, `enabled`, `label`, encrypted config, `unknown_identity_role_id`, `no_match_role_id`, `link_by_email`, `skip_role_sync`, `idp_mfa_enforced`), seeded with `password/default` and `trusted_header/default` (unknown → admin preset, no-match → viewer). `dashboard_user_invites` gains `expected_provider/provider_key/subject` with a partial unique index on live rows.

**Providers** (`app/core/auth/providers/`): `AuthProvider` protocol, `PasswordProvider`, `TrustedHeaderProvider` (configured header, e-mail parsed from the subject), a registry derived from the table plus `dashboard_auth_mode` (unchanged values), invalidated across replicas.

**IdentityResolver** (`app/modules/dashboard_users/identity_resolver.py`): identity row → account; else a pre-created account whose live invite expects exactly this identity → linked and activated; else e-mail link when the provider allows it; else JIT with the provider's default role (`role_source=mapping`; NULL role → refused, `401 identity_not_provisioned`, audited with subject/e-mail/groups). Usernames are slugs of the subject (`admin` is reserved → `admin-2`), display names bounded, collisions suffixed. Existing accounts are never re-evaluated while the provider has no mappings; writes happen at most once per cache window.

**Request path**: trusted-header requests yield an account principal (`auth_method=trusted_header`); a refused identity falls back to a verified password session when one exists so the break-glass admin stays reachable; an inactive provider answers unauthenticated. `POST /password/setup` keeps its authorisation once any account exists.

**API**: `GET/PATCH /api/auth-providers` (`security:write`; PATCH needs an account and validates roles through delegation). `POST /api/dashboard-users` accepts `ssoOnly` + `expectedIdentity` when a non-password provider is active; SSO-only invites have no link, never expire, cannot be resent, and are shown as "awaiting first sign-in". Session response: `login.providers`, `login.pendingIdentity`, `localPasswordConfigured`.

**Frontend**: `/auth/pending` public route ("your account is not ready yet", with the local sign-in form when a password account exists), invite dialog "Add without a password" with the identity field, SSO-only rows in the pending list, Access-card gates keyed on "has an account".

**Tests**: resolver unit tests (slug rules, steps), integration through the app in trusted-header mode (JIT admin by default, reserved `admin`, refused identity + audit, disabled account, pre-created account linking, duplicate expected identity, e-mail link on/off, no re-evaluation, throttle, break-glass fallback, inactive provider, provider API, migration), PostgreSQL job covers the resolver (naive/aware datetime regression), vitest for the pending screen and invite dialog.

## Simplicity

- [x] Zero config; no new env var, README section, or nav item. `dashboard_auth_mode` values unchanged.
- [x] No new required setup step; existing proxy installs behave as before (D10).
- New setting(s): none (provider options live in the table, edited via the API; the settings card lands in the next PR)
- [x] Budgets unchanged.

App code ≈ +1.5k net against the plan's ~700: table + providers + resolver + request path + API + pending screen are one capability (the identity path cannot ship half-switched).

## Test plan

```
uv run ruff check . && uv run ruff format --check . && uv run ty check
uv run pytest tests/integration/test_auth_provider_abstraction.py tests/integration/test_auth_middleware.py \
  tests/integration/test_dashboard_users_api.py tests/integration/test_dashboard_route_permission_matrix.py -q
make test-postgres   # includes test_auth_provider_abstraction.py
cd frontend && bun run lint && bun run typecheck && bun run test
openspec validate auth-provider-abstraction --strict
codex review --base feat/enable-operator-viewer-presets   # 1 P1 + 6 P2 fixed (+ 7 adversarial findings)
```


Soju06:
GitGuardian note: the only credential-shaped strings this PR adds are test fixtures — a `PASSWORD = "password"` constant and a literal `leaked-token` used as an invalid invite token in the identity tests. Nothing is a real secret; the finding can be marked as a false positive.

Both this PR and #2328 also fail `Contributors attribution`, which is unrelated to these changes: the check reports two GitHub contributors (`nikitamgrimm`, `victorstatko`) that are not in the all-contributors list yet. It passed on #2251 on 09-09, so the list needs a refresh on main; every new PR will fail it until then.

Issue #1949, opened 2026-08-28
bug: encryption-key fingerprint stamp flakes CI on SQLite lock contention


dpearson2699:
### Pre-flight checklist

- [x] I searched existing issues and confirmed this is not a duplicate.
- [x] I am running a supported codex-lb version (see Releases).
- [x] I can reproduce the problem with the steps below.

### codex-lb version

main (c7916fb3 and later; observed on CI for PR #1891 heads that merged current main)

### Deployment method

Other (GitHub Actions CI, ubuntu-24.04, SQLite)

### Client used against codex-lb

Other (pytest integration suite)

### ChatGPT account plan(s) involved

Not applicable (CI test fixtures)

### Model(s) involved

Not applicable

### What happened?

`verify_encryption_key_fingerprint()` runs at every app startup and stamps the fingerprint sentinel with `_stamp_if_absent()`. Its SQLite lock retry budget is three attempts totaling 0.35 seconds (`_SQLITE_LOCK_RETRY_DELAYS_SECONDS = (0.05, 0.1, 0.2)`). On loaded CI runners, where the integration suite boots hundreds of TestClient app instances while tests write usage and bridge state concurrently, that budget is not always enough, and the failure shows up in two different costumes:

1. `Tests (pytest, integration-bridge)` failed with `sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) database is locked` surfacing through the middleware stack into `app/core/config/key_fingerprint.py:57 _stamp_if_absent` during `test_backend_responses_websocket_transparently_retries_precreated_usage_limit_reached` (run 33191378626).
2. An earlier run of the same slice timed out with `test_backend_responses_websocket_preserves_image_generation_tool_advertisement` blocked in `TestClient.__enter__ -> wait_startup`; the faulthandler dump shows every thread idle, which is consistent with the startup stamp waiting out lock contention inside app startup (run 33190444434).

Both failures are on trees whose only delta from main is HTTP-bridge proxy code that does not touch `key_fingerprint.py` (`git diff origin/main...HEAD -- app/core/config/key_fingerprint.py` is empty), and the same slices pass locally on the same commits.

### What did you expect to happen?

The fingerprint stamp should tolerate transient SQLite lock contention: either a longer, jittered retry budget, a `busy_timeout` on the connection it uses, or routing the write through `sqlite_writer_section()` like other SQLite writers in the codebase, so a busy runner cannot turn the startup check into a request-path 500 or a startup hang.

[...]

PR #2342, opened 2026-09-10
test: stub ambient account deletion scheduler

### LINKED ISSUE #2338: test: AccountDeletionScheduler keeps lifespan unit tests parked for 30 s
### Summary

Every unit test that enters `main.lifespan(app)` can be parked for ~30 s by `AccountDeletionScheduler`. The scheduler is deliberately left running in the test harness, so on shutdown the lifespan waits out its full `DELETION_INTERVAL_SECONDS = 30` interval wait instead of stopping promptly. Measured at 7.8 % of runs of `tests/unit/test_otel.py`.

This came out of re-verifying #1949 (evidence in https://github.com/Soju06/codex-lb/issues/1949#issuecomment-5617859952) and is a **separate** finding from the SQLite lock flakes tracked there. Filing it on its own because it is a harness decision, not a product bug, and because the "30 s" it produces has been misread as a `busy_timeout` expiry.

### Evidence

`--durations` attributes every one of 18 stalls in a 230-run loop to a single test:

```
30.32s call  tests/unit/test_otel.py::test_lifespan_shutdown_fails_bridge_capacity_waiter_and_cancels_usage_singleflight
```

A faulthandler dump 12 s into a stall shows only pytest's main thread parked in `asyncio.base_events._run_once` plus the pytest-timeout thread — **no aiosqlite worker threads at all**, i.e. no SQLite statement is in flight. An asyncio task dump at 14.5 s in:

```
##### tasks=2 threads=3 names=['MainThread', 'Thread-1 (_watch)', 'pytest_timeout ...']
##### TASK Task-75  coro=<test_lifespan_shutdown_fails_bridge_capacity_waiter_and_cancels_usage_singleflight>
       tests/unit/test_otel.py:1049 in test_lifespan_...   # `async with main.lifespan(app):`
##### TASK Task-84  coro=<AccountDeletionScheduler._run_loop>
       app/modules/accounts/deletion.py:482 in _run_loop
##### scheduled_timers=1 next_in=15.95s
       when=+15.95s cb=<bound method Timeout._on_timeout of <Timeout [active]>>
```

`app/modules/accounts/deletion.py:482` is `await asyncio.wait_for(self._wake.wait(), timeout=self.interval_seconds)` with `DELETION_INTERVAL_SECONDS = 30` (`app/modules/accounts/deletion.py:94`). That timer is the **only** scheduled callback on the loop, and the test finishes ~30 s after it was armed.

### A/B

Same host, same load generator, interleaved arms:

| arm | runs | ~31 s stalls |
|---|---|---|
| control (unmodified `main`) | 40 | 2 |
| `build_account_deletion_scheduler` stubbed with a no-op | 40 | **0** |

Pooling every control-arm run in that session: 18 / 230 = 7.8 %. `P(0 of 40)` at that rate ≈ 0.04, and the task dump gives the mechanism directly.

### Why it is a harness decision

`build_account_deletion_scheduler` is one of the three entries in `LIVE_MAINTENANCE_LOOP_BUILDERS` (`tests/unit/test_background_loop_harness.py:26-31`), deliberately left running on the grounds that these loops "only touch the isolated per-test database". It is nonetheless a real SQLite reader in every lifespan-entering test (`_any_pending_deletion()`, and a full `run_account_deletion_pass()` on a leader tick) — precisely the stray-writer class that `tests/conftest.py:54-60` says "shows up as SQLite lock flakes and cross-test poisoning".

Options, in the harness rather than the product:

- move `build_account_deletion_scheduler` from `LIVE_MAINTENANCE_LOOP_BUILDERS` into the stubbed `BACKGROUND_LOOP_BUILDERS`; or
- make `AccountDeletionScheduler.stop()` reachable before the lifespan can park on the 30 s interval wait.

### Please do not misread the 30 s

These stalls are a **timer wait, not a lock wait**. The `test_otel` shard is where 7 of the 9 CI failures listed in #1949 were recorded, so it is tempting to read "~30 s" there as SQLite `busy_timeout` expiry (which is also 30 s). It is not: no statement is in flight during the stall. Anyone triaging a 30 s duration in this shard should confirm which of the two it is — the startup sentinel writes now log `sqlite_errorname` and elapsed time for exactly that reason.

Refs #1949 (do not close it — the lock class there is a different, still-open problem).

JustYannicc:
## Summary

Stub the ambient account-deletion scheduler through the existing test fixture so unrelated lifespans cannot start its real database work or interval timer. Explicit deletion tests can still restore the real builder.

## Type of change

Linked issue: Fixes #2338

## OpenSpec

- [x] Not applicable — docs / CI / chore only

Test-harness isolation only. The account-deletion production requirements, scheduler, startup retries, and prewarm behavior are unchanged.

## Changes

- Move the deletion builder from the live-maintenance allowance into the shared no-op fixture and mirror it in the browser-smoke backend.
- Pin its ambient no-op contract with a deterministic regression.
- Exercise an explicit real-builder lifespan that drains a pending account and stops its task. Existing deletion-pass tests and bridge-capacity/singleflight shutdown assertions remain unchanged.

## Test plan

Both database variables pointed to the same dedicated disposable SQLite file before imports. Main, background, and fixture engines were checked against it.

- Before the fixture fix, the harness tests reported **2 failed, 4 passed**. The new regression received a real `AccountDeletionScheduler(interval_seconds=30)` instead of the no-op.
- After the fix, the harness tests plus the named bridge-capacity/singleflight shutdown test reported **7 passed**.
- The bounded named shutdown control took 0.17 s before and 0.10 s after. This run did **not** reproduce the natural timer stall; the deterministic ambient-builder contract is the regression proof. No conclusion about #1949's separate SQLite lock failures.

```bash
uv run pytest tests/integration/test_account_deletion_lifespan.py \
  tests/integration/test_account_deletion_background.py \
  tests/integration/test_accounts_api_extended.py tests/unit/test_otel.py \
  tests/unit/test_background_loop_harness.py -q --timeout=60 --durations=10
# 117 passed, 3 PostgreSQL-only tests skipped
make lint typecheck
# passed
pnpm dlx @fission-ai/openspec@1.11.0 validate --specs --strict
# 65 passed
.venv/bin/python .github/scripts/check_simplicity_budgets.py
# passed
```

Independent review found no actionable findings against base `3b4222a08561f2e948e212535680cf3f1773252d`. Validation used the relevant local subset; full CI and PostgreSQL checks remain hosted gates.

The first hosted unit run caught a missing browser-smoke tuple update. The unchanged parity test reproduced locally as 1 failure. After aligning that builder list, the smoke harness tests, background harness tests, original shutdown control, and explicit real-deletion lifespan test passed: 16 passed. `make lint typecheck` and independent repair review also passed. All parity and shutdown assertions remain intact.

## Checklist

- [x] Ran the relevant local check subset.
- [x] Simplicity gates reviewed; no settings, product defaults, navigation, or README additions.

### DIFF
--- a/scripts/run_dashboard_browser_smoke.py
+++ b/scripts/run_dashboard_browser_smoke.py
@@ -40,6 +40,7 @@ BACKGROUND_LOOP_BUILDERS: tuple[str, ...] = (
     "build_account_usage_rollup_scheduler",
     "build_data_retention_scheduler",
     "build_telemetry_scheduler",
+    "build_account_deletion_scheduler",
 )
 
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -75,6 +75,7 @@ BACKGROUND_LOOP_BUILDERS: tuple[str, ...] = (
     "build_account_usage_rollup_scheduler",
     "build_data_retention_scheduler",
     "build_telemetry_scheduler",
+    "build_account_deletion_scheduler",
 )
 
 
--- a/tests/integration/test_account_deletion_lifespan.py
+++ b/tests/integration/test_account_deletion_lifespan.py
@@ -1,0 +1,45 @@
+"""Explicit opt-in to real deletion work through the app lifespan."""
+
+from __future__ import annotations
+
+import asyncio
+
+import pytest
+
+import app.main as main_module
+import app.modules.accounts.deletion as deletion_module
+from app.db.models import Account
+from app.db.session import SessionLocal
+from app.modules.accounts.repository import AccountsRepository
+from tests.integration.test_account_deletion_background import _seed_account
+
+pytestmark = pytest.mark.integration
+
+
+@pytest.mark.asyncio
+async def test_lifespan_can_opt_into_real_account_deletion(app_instance, monkeypatch):
+    # The real builder publishes a wake target. Restore it after this test too.
+    monkeypatch.setattr(deletion_module, "_scheduler", None)
+    schedulers: list[deletion_module.AccountDeletionScheduler] = []
+
+    def build_scheduler() -> deletion_module.AccountDeletionScheduler:
+        scheduler = deletion_module.build_account_deletion_scheduler()
+        schedulers.append(scheduler)
+        return scheduler
+
+    monkeypatch.setattr(main_module, "build_account_deletion_scheduler", build_scheduler)
+    account_id = "acc_lifespan_deletion"
+    await _seed_account(account_id, log_count=1, usage_count=1)
+    async with SessionLocal() as session:
+        assert await AccountsRepository(session).begin_delete(account_id)
+
+    async with app_instance.router.lifespan_context(app_instance):
+        assert len(schedulers) == 1
+        async with asyncio.timeout(5):
+            while True:
+                async with SessionLocal() as session:
+                    if await session.get(Account, account_id) is None:
+                        break
+                await asyncio.sleep(0.01)
+
+    assert schedulers[0]._task is None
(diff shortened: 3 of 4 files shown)


Komzpa:
@codex review

PR #2345, opened 2026-09-10
fix(proxy): preserve local quarantine evidence during completion

JustYannicc:
## Summary

A successful response can erase a first strike or replacement quarantine recorded while completion loads durable state. Separate local-failure authority from durable adoption so newer evidence survives while adopted durable-only poison still clears after successful settlement and anchor registration.

This separates the accepted cleanup concern from #2276. It preserves the poison-preserving soft cap and adds no service-wide overflow rule. Start with the two cleanup authorities in `quarantine.py`, then their propagation through durable loading and replay completion. The 946 net lines cover one coupled cleanup invariant, including 57 regression/control cases and required OpenSpec records.

## Type of change

Linked issue: Refs #2268. Resolves its stale-cleanup portion; unconditional memory bounds and overflow policy remain open.

## OpenSpec

Change directory: `openspec/changes/archive/2026-09-10-separate-local-quarantine-provenance/`. Requirements are synced to `responses-api-compat`. No wire-format changes.

## Changes

- Capture local failure authority before completion awaits, with service-lifetime generations and replacement-session identity checks.
- Retain local poison's own deadline through durable adoption, and carry replay-origin authority through loading and cleanup.
- Add 57 public-route regression/control cases with shared setup. Distinct race hooks and literal deadline/count assertions remain explicit.

## Test plan

Dedicated disposable SQLite database, with both database variables and all three engines verified before imports/reset.

```sh
uv run pytest tests/unit/test_proxy_http_bridge.py \
  tests/integration/test_http_promotion_quarantine.py \
  tests/integration/test_http_quarantine_provenance.py \
  tests/integration/test_http_quarantine_provenance_races.py \
  tests/integration/test_http_quarantine_recovery_origin.py -q
# 1128 passed
make lint
uv run ty check
pnpm dlx @fission-ai/openspec validate --specs --strict
```

The original public diagnostic fails four cases on main. The retained proof also covers failed-load retry, durable disproof, replacement identity, replay-origin propagation and same-key completion metadata. Transport and race placement are controlled; this does not measure production incidence. Independent Medium review and targeted execution found no remaining actionable findings in the implementation.

Full local `make ci` recipes passed: 1,307 frontend tests, 9,782 unit tests, 2,611 core integration tests, 339 bridge tests, 27 end-to-end tests and 215 PostgreSQL tests, plus Rust audit, migrations, packaging, image scan and both Helm smoke modes. Task-owned database and container names isolated the run. The pre-commit wrapper returned 1 because staging unchanged files during execution changed its Git diff comparison; the coordinator accepted the unchanged-tree recipe evidence. The wrapper itself did not pass.

## Screenshots / output

| Completion interleaving | Main | Fixed |
| --- | --- | --- |
| First timeout recorded during durable lookup | Strike disappears | Count stays 1; next timeout quarantines |
| New replacement entry under the same key | Old completion can clear it | Replacement evidence survives |
| Durable-only poison, successful settlement and registration | Clears immediately | Still clears immediately |

## Checklist

- [x] Ran the relevant `make` subset locally.
- [x] OpenSpec validation and implementation verification passed.
- [x] Simplicity gates reviewed; no settings, defaults, README or navigation changes.

### DIFF
--- a/app/modules/proxy/_service/http_bridge/quarantine.py
+++ b/app/modules/proxy/_service/http_bridge/quarantine.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import logging
+import weakref
 from dataclasses import dataclass
 from typing import Any
 
@@ -44,6 +45,10 @@ _HTTP_BRIDGE_QUARANTINE_POISONED_ANCHOR_REASON = "retry_circuit_poisoned_anchor"
 @dataclass(slots=True)
 class _HTTPBridgeQuarantineEntry:
     generation: int = 0
+    owner_ref: weakref.ReferenceType[_HTTPBridgeSession] | None = None
+    local_failure_generation: int = 0
+    local_poison_generation: int = 0
+    local_poison_until: float = 0.0
     quarantined_until: float = 0.0
     consecutive_eventless_timeouts: int = 0
     last_touched_monotonic: float = 0.0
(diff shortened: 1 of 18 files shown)


JustYannicc:
@coderabbitai review

Soju06:
Triage review, read-only, verified against `origin/main` @ `ca243dec4`.

**The bug is real** — issue #2268 is open and cites a maintainer repro of the key-reuse half while reviewing #1953 — so this is not an invented premise. But the PR cannot be evaluated for merge in its current state, and four things need to change before another review pass is worth anyone's time.

**1. CI Required is red.** Run `34482386287` fails `tests/integration/test_proxy_websocket_responses.py::test_v1_responses_websocket_reuses_upstream_for_sequential_requests` with `AssertionError: assert [(None, None)] == [(None, None)...proxy_owner')]`. That is a WebSocket path this diff does not obviously touch (`support.py` is +2/-1), so it may be base drift — but it has to be green after a rebase before we can tell.

**2. The OpenSpec change is committed straight into `archive/`, which skips the gate.** `.github/scripts/validate_changed_openspec.py:35` only collects change folders whose third path segment is not `archive`, so `openspec/changes/archive/2026-09-10-separate-local-quarantine-provenance/` was never strict-validated by the `Strictly validate changed OpenSpec folders` step. Please move it to an active `openspec/changes/<slug>/` folder; archiving happens after the change ships, not in the PR that introduces it.

**3. Pick a base with #2276.** #2276 (yours, currently MERGEABLE) rewrites `quarantine.py` with 426+/63- and 1325 added test lines; this PR is 113+/16- on the same file and describes itself as an extraction from it. Two open PRs restructuring the same fencing module will semantically conflict whichever lands first. Tell us which one is the base and rebase the other on top of it.

**4. Two design points we want answered in the proposal, not the diff.**

- The strike-counter erasure you are fixing is documented on main as intentional: `_clear_http_bridge_quarantine`'s `clear_fenced` says "An inactive entry still carries the eventless strike counter; a healthy completion resets it regardless of the fence", and `_http_bridge_quarantine_clear_fence` (`quarantine.py:229-240`) returns `None` for an inactive entry specifically so that happens. The counter is named `consecutive_eventless_timeouts`; making a strike survive an intervening successful completion changes what "consecutive" means. Your ordering argument (evidence recorded *after* the proof of health is newer) is defensible, but it should be argued against that comment explicitly.

- `_http_bridge_quarantine_owned_by` resolves ownership through `service._http_bridge_sessions.get(session.key)` and falls back to `entry.owner_ref`. When a completion lands after the session was deregistered (drain/retire, or the idle sweep at `app/main.py:194`) and the entry was armed by a predecessor session on the same key, `clear_fenced` now returns `False` where main would have cleared — the key stays suppressed for the full TTL. That is a capacity-affecting change well beyond "newer evidence survives"; please either scope it out or add a case that pins the intended behavior.

One process note: making `entry.generation` a draw from a new service-wide `_http_bridge_quarantine_sequence` changes the meaning of every existing generation comparison in the module. That is probably the right shape for a Lamport-style cutoff, but it deserves its own commit and its own paragraph in `design.md` rather than being folded into the fix.

Finally — the PR body states the change "does not measure production incidence". For ~950 lines in the most fence-dense module in the tree, please include either an observed incidence figure or the smallest diff that closes the ordering hole, so the review can weigh the blast radius against the payoff.

PR #2360, opened 2026-09-11
fix(balancer): count bare server_error terminals toward overload backoff

### LINKED ISSUE #2348: bug(balancer): upstream `server_error` rejections are excluded from overload isolation accounting
## Summary

`overload_backoff` only counts `server_is_overloaded` / `overloaded_error` toward the trip window. When upstream starts rejecting an account with a bare `server_error` instead, the account is never deprioritized and never reaches the isolation stage, so fresh admissions and soft sticky owners keep landing on it.

On this deployment `server_error` went from a background level of tens per day to **2,347 in one day**, of which 1,483 arrived in a single 6-hour window. None of them contributed to isolation.

## Where

`app/modules/proxy/_load_balancer/overload_backoff.py`

```python
UPSTREAM_OVERLOAD_CODES: frozenset[str] = frozenset({"server_is_overloaded", "overloaded_error"})
```

`record_upstream_overload` is invoked from the `_handle_stream_error` funnel only when `code in UPSTREAM_OVERLOAD_CODES` (`app/modules/proxy/_service/streaming/helpers.py`).

## Evidence

Same deployment, daily counts of the two codes:

| Day | `server_is_overloaded` | `server_error` |
|---|---|---|
| D-4 | 76 | 27 |
| D-3 | 518 | 22 |
| D-2 | 5,077 | 268 |
| D-1 | 684 | 31 |
| D (incident) | 8,333 | **2,347** |

Both codes arrive the same way: HTTP 200, a normal `response.created` SSE event, then the stream dies 1 to 4 seconds later with a bare upstream error and no `code`/`type` beyond the message. From the balancer's point of view they are the same failure class: upstream accepted the turn and then refused to run it.

Observed side effect during the incident: accounts whose rejections were mostly `server_error` kept receiving fresh admissions while their `server_is_overloaded` siblings were correctly isolated. Isolation logs showed 17 accounts engaged, yet `overload_free_candidates=12` in the sticky reroute log lines, i.e. the balancer still considered accounts overload-free that were in practice rejecting almost everything.

## Impact

- Fresh admissions keep selecting an account that upstream is refusing.
- Soft sticky owners are not rerouted, because the reroute predicate is gated on the isolation stage.
- Clients see the raw upstream failure instead of a retry on a healthy account.

## Suggested direction

1. Add `server_error` to the trip accounting. It can be a separate, more conservative counter if the concern is that `server_error` also covers genuine one-off faults: for example require a higher count inside the window, or only count it when it arrives after `response.created` with no output.
2. Alternatively key the accounting on the observable shape rather than the code string: an admitted turn that produces zero output tokens and terminates with an upstream 5xx-class error is an admission rejection regardless of which code the upstream chose.
3. Emit the isolation decision inputs (per-code counters per account) so an operator can see why an account was or was not isolated.

## Environment

- Version 1.25.0-beta.6
- `proxy_overload_isolation_seconds=1800`, trip level 3, window 120 s, trip count 3
- 28 subscription accounts, `capacity_weighted` routing

Soju06:
Fixes #2348.

## Problem

`record_upstream_overload` is fed only by `UPSTREAM_OVERLOAD_CODES`
(`server_is_overloaded`, `overloaded_error`). When upstream refuses an admitted
turn with a bare `server_error` stream terminal instead, the account
accumulates no rejection, never trips the window, and never reaches the
isolation stage, so fresh admissions and soft sticky owners keep landing on it.

Both codes already classify as `retryable_transient` and arrive the same way:
HTTP 200, a normal `response.created` frame, then the stream dies with a bare
upstream error and no output.

On the deployment that motivated the issue, `server_error` went from tens per
day to **2,347 in a single day**, 1,483 of them in one 6-hour window. None
contributed to isolation. Isolation logs showed 17 accounts engaged while the
sticky reroute line still reported `overload_free_candidates=12`.

## Change

| | |
|---|---|
| New class | `UPSTREAM_SOFT_OVERLOAD_CODES = {"server_error"}`, `SOFT_OVERLOAD_TRIP_WEIGHT = 0.5` |
| New state | `RuntimeState.soft_overload_rejections`, same 120 s horizon as the existing window |
| Trip rule | `len(hard) + len(soft) * 0.5 >= OVERLOAD_TRIP_COUNT` |
| Gate | soft path only when the failure carries **no** upstream HTTP status |

`server_error` genuinely covers one-off upstream faults as well as sustained
capacity refusal, so it is counted at a fractional weight rather than promoted
to a full rejection: a lone fault can never trip the window, six inside the
horizon do, and the two classes combine toward the one threshold. Keeping them
in one threshold means one deadline, one exponential level and one decay path
rather than a second set of tunables.

The soft path is keyed on the **status-less terminal shape**, because that is
what makes the observation an admission rejection. An ordinary HTTP 5xx whose
body code is `server_error` stays an ordinary transient error, and an HTTP 429
carrying it keeps its existing `record_upstream_burst_rejection` branch.

Unchanged: backoff deadline, exponential level, decay, isolation trip level,
soft-reroute semantics, failure classification, failover decision, and the
status and body returned to the client.

## Scope

Replica-local runtime state only. No migration, no new `CODEX_LB_*` setting, no
`.env.example` change, no dashboard column, no new metric, no growth of a
ceiling-guarded file.

## Verification

- `openspec validate count-server-error-terminals-as-soft-overload --strict` — valid
- `pytest tests/unit/test_overload_backoff.py tests/unit/test_exhaustion_probe.py` — 91 passed
- `pytest tests/unit/test_proxy_soft_sticky_spillover.py tests/unit/test_selection_errors.py tests/unit/test_work_admission.py` — 19 passed
- `ruff check app tests/unit/test_overload_backoff.py` — clean; `ruff format --check` — clean
- Local `codex review --base origin/main` raised one [P2] (soft path also caught coded HTTP 5xx). Fixed in the second commit and covered by a regression test.

Seven unit tests added: soft-only trip, mixed trip, window pruning, weight
invariant, funnel wiring, coded-HTTP-5xx regression guard, HTTP 429 regression
guard.


Soju06:
Interaction note with #2351 (`fix/retry-output-free-overload`, @Komzpa), which touches two of the same three files, plus one problem in this PR that I found while testing the combination. Posting here so the merge is not a surprise.

## They are complementary, not overlapping

- #2351 buffers `response.created` / `response.in_progress` so an **output-free overload** can be replayed on a sibling. Its `raise_if_retryable` gate fires only for `error_code in UPSTREAM_OVERLOAD_CODES` (`server_is_overloaded`, `overloaded_error`).
- This PR changes **which accounts get deprioritized** when a bare `server_error` stream terminal arrives. `server_error` is not in `UPSTREAM_OVERLOAD_CODES`, so #2351's buffer never intercepts the frames this PR is about: they still reach `_handle_stream_error` through `_TerminalStreamError` and get counted at weight 0.5.
- In the other direction, when #2351 *does* divert a coded overload terminal into `_RetryableStreamError`, the handler at `retry.py:2759` still calls `_handle_or_defer_keyed_stream_health(account, exc.error, exc.code)` → `_handle_stream_error` with no HTTP status, so hard overload accounting (weight 1.0) is preserved and the soft gate correctly does not fire. Neither PR subsumes the other: #2351 recovers the request, this one changes the ranking for the next request.

One combined-behaviour consequence worth stating explicitly: after both land, an output-free *coded* overload is retried on a sibling, while an output-free bare `server_error` is still surfaced to the client and only counted toward the soft window. Widening #2351's buffer to `server_error` would be a separate change and should wait until the 0.5-weight evidence from this PR says the condition is really sustained capacity refusal rather than a one-off fault — that ambiguity is the whole reason for the fractional weight.

## Textual conflicts: none, in either order

Trial merges in a scratch worktree off `main` @ `789a2dd3d`:

```
main + #2351 → clean; then + #2360 → clean (auto-merging helpers.py, retry.py)
main + #2360 → clean; then + #2351 → clean (auto-merging helpers.py, retry.py)
```

The overlaps are in the same files but disjoint regions:

| file | #2351 hunks | #2360 hunks | nearest gap |
|---|---|---|---|
| `_service/streaming/helpers.py` | import `_terminal_payload_reports_output` (~L88), import `_RetryableStreamError` (~L303), new `_OutputFreeOverloadReplayBuffer` class (~L505-578) | import `UPSTREAM_SOFT_OVERLOAD_CODES` (~L73-77), `_handle_stream_error` signature + docstring (~L1097-1115), soft-overload gate (~L1158-1195) | ~15 lines between the two import hunks; git merges them |
| `_service/streaming/retry.py` | `allow_transient_retry` at the forced-refresh call (~L929-935), `allow_transient_retry` predicate (~L2220-2235) | `_finalize_terminal_settlement_after_downstream_close` signature + `_finalize` (~L793-825), `_TransientStreamError` handler (~L2371-2400), HTTP-coded failure (~L3483-3492) | ~135 lines |
| `_service/streaming/mixin.py` | rewritten relay/yield path, removes the local `_facade` | not touched | n/a |

Tests do not collide either: #2351 extends `tests/unit/test_proxy_utils.py`, this PR extends `tests/unit/test_overload_backoff.py`. On the combined tree (`main` + #2351 + #2360) `tests/unit/test_overload_backoff.py` is 56 passed.

**Recommended order: #2351 first, then this one rebased.** Not because of conflicts — there are none — but because #2351 is green today and this PR is not (below), and because #2351 gives the previously-dead `allow_transient_retry` parameter a real meaning, which is easier to review without an unrelated diff layered on it.

## This PR currently breaks 8 pre-existing tests (CI is red)

Reproduced locally on this head, and it is not a merge-order artifact — it fails at `refs/pull/2360/head` alone, and passes at its merge base:

```
tests/unit/test_proxy_utils.py
  test_stream_responses_route_keyed_owner_rewrite_settles_before_original_health[first_event|later_event|proxy_response_error]
  test_stream_responses_route_keyed_owner_rewrite_preserves_health_after_disconnect_during_settlement[first_event|later_event|proxy_response_error]
  test_stream_with_retry_keyed_empty_terminal_queue_settles_before_terminal_effect[health_confirmed|health_downstream_close]
→ TypeError: handle_stream_error() got an unexpected keyword argument 'upstream_http_status'
```

Cause: `_finalize` in `retry.py:818` now always passes `upstream_http_status=...`, including when it is `None`, and three test doubles (`tests/unit/test_proxy_utils.py:17967`, `:18081`, `:18227`) declare fixed signatures `(failed_account, error, code, http_status=None)`. Either widen those three doubles to `**kwargs` like the majority of the doubles in that file already do, or only pass the keyword when it is not `None`. I will fix it on this branch; noting it here so the failure is not mistaken for a #2351 interaction.

For completeness: #2352 (`codex/issue-2349-affinity-observation`) merges cleanly with this PR, and conflicts with #2351 in one trivial hunk at `retry.py:932`.

Soju06:
## Follow-up: the soft gate no longer *infers* "status-less"

`5de98ae` addresses the remaining weakness in the gate this PR added. `soft_overload = code in UPSTREAM_SOFT_OVERLOAD_CODES and http_status is None` reads the *absence of an argument* as evidence about upstream, which is an inference about the caller rather than a fact the caller asserted — and three call sites break it. They write account health from a settled failure or a terminal frame whose upstream HTTP status they know, but deliberately withhold it, because the positional `http_status` also drives four unrelated decisions:

- `_is_account_neutral_request_rejection` and `_is_model_scoped_rejection` both bail on `http_status is not None and http_status != 400`, so forwarding a 500 flips those requests from *unpenalized* to *penalized*
- the `elif http_status == 429` burst branch would newly arm `RuntimeState.burst_backoff_until`
- `codex_lb_upstream_reasoning_replay_400_total` would double-count a frame already counted at `_observe_terminal_stream_error_frame` (the invariant is written verbatim at `helpers.py:1113-1116`)

So `_handle_stream_error` gains a keyword-only `upstream_http_status` read by the soft-overload gate and **nothing else**, and the gate becomes `... and http_status is None and upstream_http_status is None`. One boolean changes; every other decision keeps byte-for-byte its previous inputs.

An HTTP 500 whose body has no `code` normalizes to `server_error` (`type: "server_error"` is the default error type in `app/core/errors.py`), so six of those inside 120 s would have deprioritized a healthy account.

### Sites that now report their status

| Site | Source |
|---|---|
| `streaming/retry.py` mid-stream downstream-visible renderer | `tex.status_code` when `tex` is a `ProxyResponseError` (a `_TransientStreamError` reports nothing, which is correct) |
| `streaming/retry.py` outer `except ProxyResponseError` renderer | `exc.status_code`, the same object `error_code` was parsed from |
| `websocket/mixin.py` WS/bridge settlement write | `request_state.error_http_status_override`, already parsed by the bridge from the terminal frame; stays `None` for a direct WebSocket terminal |

`_finalize_terminal_settlement_after_downstream_close` takes the keyword too, so a downstream disconnect mid-failure-frame reports the same status as the non-cancelled path — `_StreamSettlement` has no HTTP-status field of its own.

### One site that looks like a fourth offender and is not

The `ordered_settlement_required` write that follows the forced-refresh retry is reached **only** when the post-refresh stream returned normally (every path out of its `except ProxyResponseError as retry_exc:` returns, raises, or continues). Its settlement therefore came from a `_TerminalStreamError` — an HTTP 200 stream that died on a terminal frame, i.e. exactly the shape this window exists for. The only `ProxyResponseError` in scope there is the original **401** that triggered the refresh, so reporting a status would have silently disabled the feature on the whole post-refresh path. Left alone on purpose.

### Everything else, and why

The funnel has 28 call sites; 12 already forward `http_status`. Of the 16 that do not, all but the three above have a bounded code vocabulary that provably excludes `server_error` — a literal `upstream_unavailable`, `_ACCOUNT_RECOVERY_RETRY_CODES`, the quota / `server_is_overloaded` / `stream_incomplete` / model-unsupported `retry_error_code` set, or a connection-level transport code. The two remaining ones are the genuinely status-less SSE terminals this feature is for.

### Known coverage gap (pre-existing, not closed here)

`retry.py` queues its deferred penalty with a hardcoded `502` for a `_TransientStreamError`, whose `.code` can be `server_error` after wrapping a connect-phase or capacity failure. Those are genuinely status-less but the hardcoded 502 keeps them out of the window, so real coverage is narrower than the proposal's headline. Widening it needs the queue to carry the true status; left for a follow-up and written up in the proposal.

### Verification

- `pytest tests/unit/test_overload_backoff.py tests/unit/test_exhaustion_probe.py tests/unit/test_proxy_soft_sticky_spillover.py tests/unit/test_selection_errors.py` — 108 passed
- `pytest tests/integration/test_proxy_websocket_responses.py tests/integration/test_http_responses_bridge.py` — 339 passed
- `ruff check app tests` / `ruff format --check app tests` — clean; `scripts/check_proxy_architecture.py` — passes (none of the touched files is ratcheted)
- `openspec validate count-server-error-terminals-as-soft-overload --strict` — valid

Four tests added beside the existing funnel guards: an HTTP 5xx reported through the terminal-renderer shape records nothing; a bridge `error` frame carrying `status: 500` (parsed through the real `_http_error_status_from_payload`) records nothing; the evidence keyword never arms the burst cooldown; and a status-less terminal still records one soft observation, so the fix cannot over-correct into a no-op.

PR #2361, opened 2026-09-11
refactor(db): route every SQLite lock check through the shared predicate

Soju06:
> **Post-merge note.** This PR merged (`e715165e8`) while three review items were still being applied, so parts of the description below were written for a state this PR does not contain. Everything marked **→ #2382** landed in the follow-up PR instead: the ADDED requirement's explicit driver-evidence wording and its scenario, the `update_key` call-site regression test, the `_report_lock_attempt` extraction, and the `proposal.md` note recording the `api-keys` requirement. What *did* merge here is the rewiring itself, the union predicate, `sqlite_errorname` in every report, and the code-level `exc.orig is None` rule (`3104e94`, `9a3c74a3`). The OpenSpec analysis below stands as written — it is the corrected analysis, replacing two claims this description carried at merge time (see the `api-keys` and `database-backends` entries).

## Summary

PR #2339 created `app/db/sqlite_lock_retry.py` as the shared home for "is this transient SQLite write-lock contention", but three sites still hand-rolled their own substring match. This routes all three through the shared predicate, deletes the duplicated helpers, and makes every lock failure report the driver's `sqlite_errorname` so the next production/CI occurrence self-classifies as `SQLITE_BUSY_SNAPSHOT` (instant — retry is the whole fix) vs `SQLITE_BUSY` (busy-timeout expiry — a foreign writer held the slot). No call site's control flow changes.

## Type of change

- [x] `refactor:` — internal refactor (no behavior change, no API change)
- [ ] **Breaking change**

Linked issue: relates to #1949 (the shared helper's origin); no issue closed.

## OpenSpec

Change directory: `openspec/changes/unify-sqlite-lock-classification/`

**Which specs name the behaviour of the rewired sites.** Three do. All three are accounted for below — including the one an earlier revision of this description wrongly said named nothing.

**1. `scheduler-coordination` → *Requirement: Lease is released on graceful shutdown* — MODIFIED here.** Its body says a transient `` `database is locked` (or `busy`) `` on the best-effort shutdown lease writes must be swallowed at debug. Unifying the predicate widens what that site treats as transient, pins "fail fast — the next cadence *is* the retry" as deliberate rather than accidental, and adds `sqlite_errorname` to its debug report. Delta in `specs/scheduler-coordination/spec.md`.

**2. `database-backends` → ADDED here:** the shared classification itself (one predicate; matched against the driver exception the wrapper carries, never the rendered statement and parameters; the union message set; `sqlite_errorname` in every report).

> Correcting an earlier revision of this description, which said #2339's requirement was "not yet archived into `openspec/specs/`": it **is** archived now. `retry-startup-sentinel-sqlite-lock` landed in `openspec/specs/database-backends/spec.md` on `main` (archived `2026-09-11`) as *Requirement: Startup sentinel writes tolerate and name transient SQLite lock contention*, and `main` is merged into this branch. The two requirements remain disjoint: the archived one binds the **two startup sentinel writes** and their shared budget; the one added here binds the **process-wide classification and diagnostic** that every site — those two included — reaches through. Different requirement names, no overlap in what they bind, so this is still an `## ADDED Requirements` delta and not a modification of the archived one.

**3. `api-keys` → *Requirement: API Key update* — cited, no delta.** An earlier revision of this description claimed nothing in `api-keys` named this behaviour. That was wrong. The requirement's second paragraph says:

> A transient SQLite lock or snapshot conflict during the update MUST roll back and retry the complete read/build/write transaction, including rereading existing limits, before returning an error.

with the scenario *Retry API-key PATCH after a transient SQLite snapshot conflict*. `update_key` is one of the three sites rewired here, and the one whose predicate haystack this change narrows most, so that requirement was re-read against the rewiring.

**The promise does not move, so no delta is needed.** Everything the requirement binds is untouched: the `rollback()` still precedes the retry; the retry is still the recursive whole-method call that rereads existing limits and rebuilds the write; the budget is still four attempts at `0.1 * 2**attempt`; an exhausted budget still returns the error; and the scenario still passes unchanged. What changes is only *which driver failures enter that branch* — and both directions move the code **toward** the requirement, not away from it:

- `database is busy` is a transient SQLite lock (SQLITE_BUSY) that `update_key`'s now-deleted local predicate did **not** match. The old code was under-delivering a retry the requirement already demanded; it now retries.
- A non-lock failure whose lock text lived only in the rendered statement or a bound parameter is **not** "a transient SQLite lock or snapshot conflict", so the requirement never owed it a retry. The old code spent the whole budget on it anyway.

Recording either correction as a MODIFIED requirement would be wrong — the normative text does not change; only the code's fidelity to it improves. The new `database-backends` requirement is where the classification rule itself is now normative; **→ #2382** additionally carries this reasoning into the change's `proposal.md` under **Unchanged Capabilities** so the next reader does not re-litigate it from a PR description alone.

**Validation.** `openspec validate --specs --strict` → 65 passed / 0 failed; `openspec validate unify-sqlite-lock-classification --strict` → valid. Archive re-simulated against a clean copy of **current post-merge `main`**'s `openspec/`: applies with `+1 added, ~1 modified`; afterwards exactly two files differ from `main` (`database-backends/spec.md`, `scheduler-coordination/spec.md`) and the archived tree re-validates 65/65.

### Predicate coverage — before/after match set per site

The shared predicate now matches the **union** of what the four sites matched, and matches it against the **driver exception** (`exc.orig`) rather than `str(OperationalError)`.

| Site | Haystack before | Match set before | Match set after | Delta |
|---|---|---|---|---|
| `leader_election._is_locked_error` (deleted) | `str(exc.orig or exc)` | `database is locked`, `database is busy` | union | **widened**: `database table is locked`, `database schema is locked`, `busy_snapshot`. **narrowed**: the `or exc` fallback when `orig` is absent |
| `api_keys/service._is_sqlite_database_locked` (deleted) | `str(exc)` | `database is locked`, `database table is locked`, `database schema is locked`, `sqlite_busy_snapshot`, `busy_snapshot` | union | **widened**: `database is busy`. **narrowed**: lock text that appears only in the rendered SQL / bound parameters |
| `refresh_claims._is_sqlite_database_locked` (deleted) | `str(exc)` | `database is locked` | union | **widened**: `database is busy`, `database table is locked`, `database schema is locked`, `busy_snapshot`. **narrowed**: same rendered-SQL narrowing |
| `key_fingerprint` + accounts startup seed (unchanged call sites) | `str(exc.orig or exc)` | `database is locked`, `database is busy` | union | **widened**: `database table is locked`, `database schema is locked`, `busy_snapshot`. **narrowed**: the `or exc` fallback when `orig` is absent |

Union kept because each added message is the same fact: `database is locked` / `database is busy` are SQLITE_BUSY (and SQLITE_BUSY_SNAPSHOT, which reuses the prose); `database table is locked` / `database schema is locked` are SQLITE_LOCKED, the same contention reported against a table or the schema cookie rather than the file; `busy_snapshot` catches a wrapper that surfaces the extended result-code name instead of the prose. All of them mean "another writer holds the slot, try again", so no site is made to retry something it should not. `sqlite_busy_snapshot` is dropped as a separate entry only because `busy_snapshot` already subsumes it.

### The driver-evidence rule, and why the spec now states it in full

The **narrowing is a deliberate bug fix**, not a regression: `str(OperationalError)` renders the failing statement *and its bound parameters* alongside the driver message, so `update_key` with `name="database is locked"` would previously have had an unrelated `OperationalError` on that statement classified as transient and silently retried four times.

The same rule decides the `exc.orig is None` case, and this is the one the earlier revision of the ADDED requirement over-claimed: the requirement said classification "MUST inspect the DRIVER exception rather than the rendered statement and bound parameters" while the code still fell back to `str(exc)` when the wrapper carried no driver exception — which is exactly the rendered statement and parameters the requirement forbids, so a bound parameter containing the phrase was a live false positive on that path.

**Resolved by making the code match the spec** (option (a)), not by softening the requirement, because nothing real depends on the fallback. Evidence: `grep -rn "OperationalError(" app/` returns **zero** hits — no application path constructs one — and every test double in `tests/` passes a driver exception as the third argument. The only orig-less construction in the tree is the regression test added for this rule. Concretely:

- `is_sqlite_lock_error` returns `False` when `exc.orig is None` (`3104e94`, merged here), so a wrapper with no driver evidence is never contention.
- **→ #2382**: the requirement states that consequence explicitly instead of leaving it implied, and adds the scenario *A wrapper with no driver exception is not contention*.
- Pinned at the predicate here (`test_wrapper_without_a_driver_exception_is_not_a_lock_failure`); **→ #2382** adds the spec-governed product path (`test_update_key_does_not_retry_a_wrapper_without_driver_evidence`: one commit attempt, one rollback, propagates — not the four attempts a classified lock failure would spend).

### Sites judged fail-fast-by-design (no `retry_on_sqlite_lock` added)

- **`LeaderElection.release()`** — the DELETE is one-shot at shutdown. Its documented fallback is "leave the row, the lease expires after its TTL", and there is no next release to hold the retry. An in-place retry would also push the shutdown release past its deadline, which `scheduler-coordination` explicitly bounds. Kept failing fast; only `sqlite_errorname` was added to the existing debug swallow.
- **`LeaderElection._renew_lease_row()`** — the caller (release keeper / drain loop) treats `False` as "not renewed this cadence" and renews again on the next tick, at an interval no greater than TTL/3. **The next cadence *is* the retry**; retrying in place would consume the cadence it exists to protect. Kept failing fast; only `sqlite_errorname` added.

### Sites that already retried by hand (budget and backoff unchanged)

`ApiKeysService.update_key`, `enforce_limits_for_request`, `finalize_usage_reservation`, `fail_usage_reservation`, `touch_usage_reservation`, `release_usage_reservation`, and `RefreshClaimCoordinator.try_acquire` each own a 4-attempt loop with `base * 2**attempt` backoff (`base` = 0.1s for API keys, 0.05s for refresh claims). `retry_on_sqlite_lock` was **not** substituted in: its budget is a different shape (3 retries at 0.05/0.1/0.2s), so swapping it would silently change eight call sites' retry timing. Instead they call the new `should_retry_after_sqlite_lock(...)`, which owns only the predicate, the diagnostic and the backoff formula; each caller keeps its own attempt count, its own `raise`, and its own `rollback()` placement. `update_key`'s recursive retry keeps falling through to the `IntegrityError` branch exactly as before.

### Diagnostics

- Each retry logs at DEBUG, each exhausted budget at WARNING, both naming `what=` and `sqlite_errorname=` (or its absence). Neither report includes written values.
- **→ #2382**: both helpers emit through one private `_report_lock_attempt(exc, *, what, giving_up, **extra)`, so the report format has a single owner and the per-helper timing/attempt fields are appended as trailing `key=value` pairs. That is a formatting extraction only — **the budgets are deliberately not unified** (that is the behaviour change this PR avoids), and every emitted line is byte-identical.
- `sqlite_error_name()` and `is_sqlite_lock_error()` now accept `BaseException`, so leader election's `except Exception` sites no longer need their own `isinstance` narrowing — the predicate owns it.
- `asyncio` became unused in `api_keys/service.py` and `refresh_claims.py` and was dropped; the four existing tests that monkeypatched `app.modules.api_keys.service.asyncio.sleep` now patch `app.db.sqlite_lock_retry.asyncio.sleep` (same module object, same effect).

## Test plan

New: `tests/unit/test_sqlite_lock_classification.py` (15 tests) pins the union, the rejection of non-lock and non-`OperationalError` failures, the statement/parameter narrowing, the orig-less wrapper, `sqlite_errorname` extraction, and `should_retry_after_sqlite_lock`'s backoff/give-up/non-lock semantics. Counts below are from the working tree at the time of writing; the `test_api_keys_service.py` total includes the one test that landed in **#2382**.

Per site, a test that the lock is classified as before **and** that the error name reaches the log record:

- leader election — `test_best_effort_lease_writes_name_the_lock_mechanism` (parametrized over `SQLITE_BUSY_SNAPSHOT` / `SQLITE_BUSY`, asserts both swallows stay DEBUG with no warning and carry the name);
- API keys — `test_usage_writes_still_classify_every_transient_lock_message` (the four messages the deleted local predicate matched), `test_update_key_retry_names_the_lock_mechanism`, `test_touch_usage_reservation_reports_the_exhausted_lock_budget` (still raises), and **→ #2382** `test_update_key_does_not_retry_a_wrapper_without_driver_evidence` (the spec-governed PATCH path does not spend its budget without driver evidence);
- refresh claims — `test_claim_upsert_retries_a_transient_lock_and_names_the_mechanism`, `test_claim_upsert_still_raises_when_the_lock_outlives_the_budget` (still 4 attempts, still raises).

```
make lint                                              # ruff + architecture + cancellation + timing seams + settings tiers — pass
uv run ty check                                        # pass
make migration-check                                   # migration_policy=ok schema_drift=none
python3 .github/scripts/check_simplicity_budgets.py    # all within budget

uv run pytest tests/unit/test_sqlite_lock_classification.py -q     # 15 passed
uv run pytest tests/unit/test_leader_election.py -q                # 44 passed
uv run pytest tests/unit/test_api_keys_service.py -q               # 90 passed
uv run pytest tests/unit/test_refresh_claims.py -q                 # 12 passed
uv run pytest tests/unit/test_startup_sentinel_lock_retry.py -q    #  7 passed
uv run pytest tests/unit -q -x                                     # 9936 passed, 4 skipped

openspec validate --specs --strict                                 # 65 passed, 0 failed
openspec validate unify-sqlite-lock-classification --strict         # valid
# archive simulated on a clean copy of current main's openspec/:
#   +1 added, ~1 modified; only database-backends + scheduler-coordination differ; re-validates 65/65
```

## Screenshots / output

No user-visible surface. Log shape, before → after, for the leader-election renewal swallow:

```
- Leader lease renewal contended on a locked database; will retry next cadence leader_id=node-a
+ Leader lease renewal contended on a locked database; will retry next cadence leader_id=node-a sqlite_errorname=SQLITE_BUSY
```

and the new give-up report on a hand-rolled loop:

```
sqlite lock retry budget exhausted what=touch_usage_reservation sqlite_errorname=SQLITE_BUSY attempts=4
```

## Checklist

- [x] Ran the relevant `make <target>` subset locally (`make lint`, `uv run ty check`, the suites above).
- [x] If touching specs: `openspec validate --specs --strict` passes and the change validates strict; archive simulated against current `main`.
- [x] Simplicity gates reviewed: the six simplicity rules (PRINCIPLES.md P1-P6). No new setting, no README/nav/.env change; net effect is three fewer duplicated predicates and one fewer duplicated report format.


Soju06:
Follow-up: #2382 carries the three review items that were still being applied when this merged — the ADDED `database-backends` requirement's explicit driver-evidence wording plus its scenario, the `update_key` call-site regression test for the orig-less wrapper, the `_report_lock_attempt` extraction (report format only; budgets stay separate on purpose), and the `proposal.md` note recording that `api-keys` → *Requirement: API Key update* does name this retry and why its promise does not move. The description above has been corrected accordingly and marks each item.

PR #2366, opened 2026-09-11
refactor(bridge): retire the durable recovery-dispatch storage surface

Soju06:
## Summary

`drop-bridge-recovery-modes` (#2336) deleted every request-path caller of the server-owned recovery dispatch and explicitly deferred the storage cleanup: its own delta says the durable primitive "is now unreferenced by the request path and keeps its repository-level tests until a follow-up retires it together with the `recovery_dispatch_count` column semantics." This is that follow-up — it deletes the callerless surface, drops three CAS predicates that can no longer reject anything, and retires the ORM mapping while keeping the physical column one more release for rolling-upgrade safety.

## Type of change

- [x] `refactor:` — internal refactor (no behavior change, no API change)
- [ ] **Breaking change**

Linked issue: Refs #2336 — the deferral this PR follows up. Refs #2189 — the column-retention precedent (`retire-prewarm-canary-column-mappings`). **No issue is closed by this PR**; both references are context, not fixes.

## Proof: no writer can advance `recovery_dispatch_count` on `main`

Every occurrence of the column name in the tree at `origin/main`, grouped by role. Reproduce with `grep -rn recovery_dispatch_count .` (excluding `.git`/`node_modules`). The full write-up lives in `openspec/changes/retire-recovery-dispatch-storage/design.md`.

**Writes — all three are inside the callerless surface:**

| Site | Kind | Reachable? |
|---|---|---|
| `durable_bridge_repository.py:2202` `operation.recovery_dispatch_count += 1` | increment | No. Inside `claim_unknown_operation_for_recovery` (`:2157`). Its only in-repo references are `durable_bridge_coordinator.py:931` (a pass-through whose own method has zero callers) and repository tests. |
| `durable_bridge_repository.py:2250` `-= 1` | decrement | No. Guarded by `restore_recovery_dispatch_claim and recovery_dispatch_count > 0`. The single production caller of `mark_operation_unknown` (`_service/http_bridge/request_submit.py:2291`) passes only `operation_id`/`session_id`/`instance_id`/`owner_epoch`, so the flag is always `False`. And even if it were `True`, `> 0` can never hold with no increment. |
| `durable_bridge_repository.py:2260` `-= 1` | decrement | No. Same flag, same `> 0` floor. |

**Non-writes (the rest of the hits):**

- `app/db/models.py:2511` — the ORM mapping (this PR retires it).
- `app/db/alembic/versions/20260810_000000_add_http_bridge_recovery_dispatch_count.py` + the `20260812` merge revision — DDL that creates the column `NOT NULL DEFAULT 0`.
- `durable_bridge_repository.py:217`, `:4190` — the read-only `DurableBridgeOperationSnapshot` field and its mapper.
- `durable_bridge_repository.py:340`, `:2795`, `:3061` — the three `== expected_recovery_dispatch_count` **read** predicates.
- `_service/http_bridge/request_submit.py:1543` — a read: `request_state.operation_attempt_generation = getattr(operation, "recovery_dispatch_count", 0)`.
- Pass-through keyword arguments in `durable_bridge_coordinator.py`, `http_bridge_event_batcher.py`, `_service/http_bridge/upstream_events.py`.
- `tests/…`, `openspec/…` — tests and prose.

No raw SQL `text()`, no bulk `update(...).values(...)`, and no ORM `INSERT` anywhere in `app/` names the column outside the rows above. The only remaining INSERT that sets it is a deliberately legacy-shaped one in `tests/integration/test_migrations.py`.

**Consequence.** The expectation is not a hardcoded constant: `request_submit.py:1543` seeds `request_state.operation_attempt_generation` from the operation snapshot's own `recovery_dispatch_count`, and `streaming.py` only preserves it across the account-neutral replay (`:2082`, `:2117`) or copies it to the retry state (`:3859`) — where the identity is *not* preserved, a fresh `request_state` is built and reseeded from the row it then finds. Since no writer can move the column between that read and the append, and since `record_operation`'s `failed`→`submitted` rebind does not touch the counter, each `expected_recovery_dispatch_count` CAS compares a value to itself for **every row the code can reach — including rows carried over from before #2336 with a non-zero counter**. A predicate that can never reject is not a fence, and the two orphaned request-state hops that carried the value around are pure noise. Removing it is a no-op, not a latent fix and not a regression.

## OpenSpec

Change directory: `openspec/changes/retire-recovery-dispatch-storage/`

- **REMOVED** `responses-api-compat` → "Fenced one-shot recovery dispatch". The requirement MUSTs a durable one-shot replay budget "consumed atomically when a replay is claimed for dispatch"; all three of its scenarios drive the deleted claim path and its title names the deleted dispatch, so `MODIFIED` cannot express it.
- **MODIFIED** `responses-api-compat` → "Terminal append failure preserves authoritative settlement". The newer-attempt rejection now rests on operation state + persisted upstream-response identity (no durable per-attempt counter), and the requirement records the ORM-mapping retirement plus the one-release physical-column retention. Adds a rolling-upgrade scenario mirroring #2189's.
- `drop-bridge-recovery-modes` is **already archived on `main`** (`openspec/changes/archive/2026-09-11-drop-bridge-recovery-modes/`), so both blocks are transcribed from the **post-archive** `main` spec text (`openspec/specs/responses-api-compat/spec.md`) — the state a reviewer sees today. There is no longer a pending interleaving to worry about.
- Free-form rationale (the per-writer proof, why each CAS predicate is safe to drop, the column-retention reasoning) lives in `design.md` beside the proposal, per `AGENTS.md`; `proposal.md`'s `Why` is 603 characters, under the archive lint's 1000-character threshold.

Validation and archive simulation on a temp copy of current `main`'s `openspec/` (`git archive origin/main openspec`), with only this change dir copied in:

```
$ openspec validate --specs --strict            → Totals: 65 passed, 0 failed (65 items)
$ openspec validate retire-recovery-dispatch-storage --strict
                                                → Change 'retire-recovery-dispatch-storage' is valid
$ openspec archive retire-recovery-dispatch-storage --yes
  Specs to update:
    responses-api-compat: update
  Applying changes to openspec/specs/responses-api-compat/spec.md:
    ~ 1 modified
    - 1 removed
  Totals: + 0, ~ 1, - 1, → 0
  Specs updated successfully.
  Change 'retire-recovery-dispatch-storage' archived as '2026-09-11-retire-recovery-dispatch-storage'.
$ openspec validate --specs --strict   (post-archive)  → 65 passed, 0 failed
$ grep -c "Fenced one-shot recovery dispatch" openspec/specs/responses-api-compat/spec.md → 0
```

## Changes

**Deleted (callerless):**

- `DurableBridgeRepository.claim_unknown_operation_for_recovery` and its `max_recovery_dispatches` bound; the `DurableBridgeSessionCoordinator` pass-through.
- The `restore_recovery_dispatch_claim` flag and both refund branches on `mark_operation_unknown` (repository + coordinator). **`mark_operation_unknown` itself stays** — it is the live ambiguous-dispatch fence called from `request_submit`.
- `HTTPRequestState.operation_attempt_generation` (`_service/support.py`) and its propagation through the account-neutral replay and the server-owned retry hand-off in `_service/http_bridge/streaming.py`, plus the `request_submit` assignment.

**CAS arguments removed — what each fence still guards:**

| Site | Removed | Still guarded by |
|---|---|---|
| `append_terminal_operation_event` | `expected_recovery_dispatch_count` predicate | session-owner `instance_id`/`owner_epoch` `SELECT … FOR UPDATE`, `session_id`, the `state == "abandoned"` refusal, and **`spool_format == rows_v1`** in the same statement |
| `_lock_operation_for_chunk_append` (used by `append_terminal_operation_chunk`) | the whole optional parameter | session-owner fence, `session_id`, `abandoned` refusal. Its other caller (`append_operation_event_chunk`) already passed no expected count, so the parameter's only value was the tautology |
| `settle_terminal_append_failure` | `expected_recovery_dispatch_count` predicate | session-owner fence, `state != "abandoned"`, and the operation-state + persisted-response-identity `or_` |

Note on the first two rows: because the counter cannot move across a retry or a rebind (see **Consequence** above), `append_terminal_operation_event` and `_lock_operation_for_chunk_append` had **no newer-attempt discriminator before this change either** — the removed predicate never distinguished attempts, so their protection is unchanged.

The last row is the interesting fence. Its spec scenario ("Newer retry rejects delayed fallback settlement") is preserved by the **state** predicate alone: the only path left that can reset a live row to `submitted` is `record_operation`'s `failed`→`submitted` rebind, and `submitted` matches neither `state == "acknowledged"` nor `state == <terminal state>`, so the `UPDATE` touches no row and the caller sees the rejection. A comment at the predicate records this.

**Column retention (#2189 precedent):**

- `HttpBridgeOperationRecord.recovery_dispatch_count` removed from the ORM; a comment at the former site explains why the column is unmapped.
- `("http_bridge_operations", "recovery_dispatch_count")` added to `_LEGACY_EXTRA_COLUMNS` in `app/db/migrate.py` so the drift gate treats the retained column as known drift.
- **The physical column is NOT dropped.** The Helm migration Job is a `pre-upgrade` hook that runs before old replicas drain, so a replica on the previous build must still be able to `INSERT` an operation row naming the column. Because it is `NOT NULL DEFAULT 0`, this release's INSERTs simply omit it.
- Queued in `openspec/specs/deployment-installation/context.md` under the next-release entry titled **"Drop the retired bridge recovery-dispatch column"** (cited by title, not by list position, so renumbering cannot break the reference), worded the same way #2189's entry was (Alembic `drop_column` in batch mode for SQLite, `NOT NULL` re-add with `server_default 0` on downgrade, allow-list entry removed in the same PR).

## Orphan sweep from #2336

Checked and **still live** — not touched by this PR:

- `HttpBridgeRecoveryAttemptRecord` and the `recovery_attempt_*` request-state fields. This is a *different* journal (fresh-replay attempts), with live readers/writers in `_service/http_bridge/upstream_events.py` (~20 sites), `request_submit.py`, and the `http-bridge-recovery-settlement-*` shutdown drain in `app/main.py`.
- `request_state.replay_count` / `clean_close_replay_count` — client-observed replay accounting, read at 20+ sites across `request_submit.py`, `streaming.py`, `upstream_events.py`, `support.py`, and surfaced in `response_create.py`.
- `operation_replay`, `operation_created`, `operation_rebound*`, `operation_dispatched`, `operation_rebind_required`, `operation_persisted_response_id` — every one still read on the request path.
- `mark_operation_unknown`, `rollback_operation_before_dispatch`, `reset_operation_event_spool` — live cleanup primitives.

## Test plan

```
$ make lint                                            → all checks passed
$ uv run ty check                                      → All checks passed!
$ make migration-check                                 → current_revision=20260911_010000_merge_pin_index_and_affinity_heads
                                                         migration_policy=ok, schema_drift=none
$ python3 .github/scripts/check_simplicity_budgets.py  → all OK

# full touched files
$ uv run pytest tests/unit/test_bridge_ring_lifecycle.py tests/unit/test_proxy_http_bridge.py \
      tests/unit/test_db_migrate.py tests/unit/test_http_bridge_event_batcher.py -q
  1206 passed

$ uv run pytest tests/unit -q -x
  9923 passed, 4 skipped

# integration for the touched modules
$ uv run pytest tests/integration/test_migrations.py tests/integration/test_http_responses_bridge.py \
      tests/integration/test_multi_instance_bridge.py tests/integration/test_startup_bridge_cleanup.py -q
  242 passed, 9 skipped

$ codex review --base origin/main   (run on the pre-rebase head; the rebase changed
                                     only hunk offsets, the app/ + tests/ diff content
                                     is byte-identical)
  "No actionable correctness defects were found in the changes. The retired storage
   surface is removed consistently, while rolling-upgrade compatibility for the
   physical column is preserved."
```

Test changes:

- Deleted `test_unknown_operation_recovery_claim_is_atomic_and_single_use`, `test_one_shot_recovery_budget_survives_unknown_reset`, `test_pre_dispatch_recovery_claim_restores_one_shot_budget` — each exercised only the deleted claim path.
- Renamed `test_chunk_format_resets_on_failed_rebind_and_unknown_claim` → `test_chunk_format_resets_on_failed_rebind` and dropped its unknown-claim half; the failed-rebind half is unchanged.
- Rewrote the generation-fence block in `test_terminal_append_failure_settlement_is_visible_to_recovery` to prove the **state** fence rejects a prior attempt's delayed fallback settlement after `record_operation` rebinds the row to `submitted`.
- Dropped the `claim_unknown_operation_for_recovery` assertion from the abandoned-fence test; the four other late writers there already cover it.
- New `tests/unit/test_db_migrate.py::test_check_schema_drift_ignores_retired_recovery_dispatch_column_kept_for_rolling_upgrade` and `tests/integration/test_migrations.py::test_retired_recovery_dispatch_column_stays_insertable_for_legacy_replicas` — head schema still carries the column, the ORM does not map it, a legacy-style INSERT with an explicit `1` succeeds, a current-release ORM INSERT defaults to `0`, and `check_schema_drift` is clean.

## Screenshots / output

No user-visible surface. Every removed predicate was satisfied by every row the current release can write, so request/response behaviour is byte-identical at the shipped default.

## Checklist

- [x] Linked the related issue / discussion above (`Refs #2336`, `Refs #2189`; no issue is closed).
- [x] Ran `make lint`, `uv run ty check`, `make migration-check`, `check_simplicity_budgets.py`, the full touched test files, `tests/unit`, and the bridge/migration integration files locally.
- [x] If touching specs: `openspec validate --specs --strict` passes, the change validates `--strict`, and `openspec archive --yes` was simulated on a temp copy of current `main`'s `openspec/` (which already contains the archived `drop-bridge-recovery-modes`).
- [x] Simplicity gates reviewed: **net +68 lines** — `app/` −110, `tests/` −116, `openspec/` +294 (the spec delta, proposal, design notes, tasks and the deployment context entry). No new setting, no new default.

### DIFF
--- a/app/db/migrate.py
+++ b/app/db/migrate.py
@@ -115,9 +115,9 @@ _SQLITE_FLOAT_TYPE_COMPAT_COLUMNS = frozenset(
     }
 )
 # Columns the ORM no longer maps but the schema still carries. A replica running
-# the previous release keeps mapping them and renders explicit NULLs in its
-# request-log INSERTs, so the physical drop must wait one release after the
-# mapping retirement (the Helm migration Job runs before old replicas drain).
+# the previous release keeps mapping them and renders an explicit value in its
+# INSERTs, so the physical drop must wait one release after the mapping
+# retirement (the Helm migration Job runs before old replicas drain).
 _LEGACY_EXTRA_COLUMNS = frozenset(
     {
         ("request_logs", "slim_summary_json"),
@@ -125,6 +125,10 @@ _LEGACY_EXTRA_COLUMNS = frozenset(
         # revision + removal from this set is queued for the following release.
         ("request_logs", "prewarm_canary_bucket"),
         ("request_logs", "prewarm_eligible_reason"),
+        # Retired from the ORM in retire-recovery-dispatch-storage; same
+        # queued drop. The column is NOT NULL with a server default, so this
+        # release's INSERTs simply omit it.
+        ("http_bridge_operations", "recovery_dispatch_count"),
     }
 )
 
--- a/app/db/models.py
+++ b/app/db/models.py
@@ -2508,7 +2508,10 @@ class HttpBridgeOperationRecord(Base):
     request_text: Mapped[str | None] = mapped_column(Text, nullable=True)
     state: Mapped[str] = mapped_column(String(32), nullable=False, server_default=text("'submitted'"))
     response_id: Mapped[str | None] = mapped_column(Text, nullable=True)
-    recovery_dispatch_count: Mapped[int] = mapped_column(Integer, nullable=False, server_default=text("0"))
+    # ``recovery_dispatch_count`` is intentionally unmapped: the server-owned
+    # recovery dispatch it counted was deleted, so nothing reads or writes it.
+    # The physical column stays for one release so a previous-release replica
+    # can keep inserting operations (see ``_LEGACY_EXTRA_COLUMNS``).
     event_bytes: Mapped[int] = mapped_column(Integer, nullable=False, server_default=text("0"))
     event_spool_complete: Mapped[bool] = mapped_column(Boolean, nullable=False, server_default=text("false"))
     spool_format: Mapped[str] = mapped_column(
--- a/app/modules/proxy/_service/http_bridge/request_submit.py
+++ b/app/modules/proxy/_service/http_bridge/request_submit.py
@@ -1540,7 +1540,6 @@ class _HTTPBridgeRequestSubmitMixin:
                 operation, "rebound_from_parent_response_id", None
             )
             request_state.operation_persisted_response_id = getattr(operation, "response_id", None)
-            request_state.operation_attempt_generation = getattr(operation, "recovery_dispatch_count", 0)
 
         async def _cleanup_unsubmitted_recovery_claim() -> None:
             if (
--- a/app/modules/proxy/_service/http_bridge/streaming.py
+++ b/app/modules/proxy/_service/http_bridge/streaming.py
@@ -2079,9 +2079,6 @@ class _HTTPBridgeStreamingMixin:
                 else None
             )
             prior_operation_registered = request_state.operation_registered if preserve_operation_identity else False
-            prior_operation_attempt_generation = (
-                request_state.operation_attempt_generation if preserve_operation_identity else 0
-            )
             prior_operation_persisted_response_id = (
                 request_state.operation_persisted_response_id if preserve_operation_identity else None
             )
@@ -2114,7 +2111,6 @@ class _HTTPBridgeStreamingMixin:
                 request_state.operation_fingerprint = prior_operation_fingerprint
                 request_state.operation_parent_response_id = prior_operation_parent_response_id
                 request_state.operation_registered = prior_operation_registered
-                request_state.operation_attempt_generation = prior_operation_attempt_generation
                 request_state.operation_persisted_response_id = prior_operation_persisted_response_id
                 request_state.operation_rebind_required = True
             request_state.enforce_openai_sdk_contract = enforce_openai_sdk_contract
@@ -3856,7 +3852,6 @@ class _HTTPBridgeStreamingMixin:
                 retry_request_state.operation_fingerprint = request_state.operation_fingerprint
                 retry_request_state.operation_parent_response_id = request_state.operation_parent_response_id
                 retry_request_state.operation_registered = request_state.operation_registered
-                retry_request_state.operation_attempt_generation = request_state.operation_attempt_generation
                 retry_request_state.operation_persisted_response_id = request_state.operation_persisted_response_id
                 retry_request_state.operation_rebind_required = request_state.operation_rebind_required
                 # An anchored recovery replays the proxy's own anchor, so the
--- a/app/modules/proxy/_service/http_bridge/upstream_events.py
+++ b/app/modules/proxy/_service/http_bridge/upstream_events.py
@@ -484,7 +484,6 @@ async def _persist_http_bridge_operation_event(
                                 )
                             ),
                             state=terminal_state,
-                            expected_recovery_dispatch_count=request_state.operation_attempt_generation,
                             response_id=response_id,
                         ),
                         None,
@@ -534,7 +533,6 @@ async def _persist_http_bridge_operation_event(
                             owner_epoch=owner_epoch,
                             state=terminal_state,
                             expected_response_id=expected_response_id,
-                            expected_recovery_dispatch_count=request_state.operation_attempt_generation,
                             alternate_expected_response_id=alternate_expected_response_id,
                             response_id=response_id,
                         )
--- a/app/modules/proxy/_service/support.py
+++ b/app/modules/proxy/_service/support.py
@@ -1177,9 +1177,6 @@ class _WebSocketRequestState:
     operation_rebound_from_parent_response_id: str | None = None
     operation_replay: bool = False
     operation_dispatched: bool = False
-    # Immutable durable attempt generation. Recovery claims increment the
-    # operation's dispatch count before sending a replacement attempt.
-    operation_attempt_generation: int = 0
     # Last response identity successfully written to the durable operation.
     # Retry setup may clear the active response before a replacement is
     # acknowledged, but fallback settlement must still fence against this ID.
--- a/app/modules/proxy/durable_bridge_coordinator.py
+++ b/app/modules/proxy/durable_bridge_coordinator.py
@@ -770,7 +770,6 @@ class DurableBridgeSessionCoordinator:
         event_text: str,
         max_bytes: int,
         state: str,
-        expected_recovery_dispatch_count: int = 0,
         response_id: str | None = None,
     ) -> bool:
         async with self._session() as session:
@@ -782,7 +781,6 @@ class DurableBridgeSessionCoordinator:
                 event_text=event_text,
                 max_bytes=max_bytes,
                 state=state,
-                expected_recovery_dispatch_count=expected_recovery_dispatch_count,
                 response_id=response_id,
             )
 
@@ -820,7 +818,6 @@ class DurableBridgeSessionCoordinator:
         event_text: str,
         max_bytes: int,
         state: str,
-        expected_recovery_dispatch_count: int = 0,
         response_id: str | None = None,
     ) -> bool:
         async with self._session() as session:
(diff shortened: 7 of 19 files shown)


Soju06:
Flagging a collision before this merges: **#2374 is the caller this PR proves does not exist.**

The "callerless" premise is accurate for `main` today — the enumeration in the description is correct, nothing advances `recovery_dispatch_count`, and a CAS that compares a constant `0` against a column that is always `0` is not a fence. That is true *because* #2336 deleted the caller. The owner decided on 2026-09-11 to add one back: `openspec/changes/relocate-anchored-turns-across-accounts/` (#2374) restores at-least-once relocation for ambiguous eventless dispatches and uses `claim_unknown_operation_for_recovery` as its one-shot duplicate-suppression gate.

The spec collision is direct and order-independent: this PR **REMOVES** `responses-api-compat` → *Fenced one-shot recovery dispatch*, and #2374 **MODIFIES** the same requirement (adding the concrete one-dispatch bound, the zero-event precondition and the bounded ambiguity window). The interleaving check in the description covers `drop-bridge-recovery-modes`, which is genuinely disjoint; #2374 came later and is not.

## Why the caller is being added back

Measured on `codexlb-vultr` (beta.7), 7 days, requests that died with no response event while carrying a conversation:

| class | dead turns | distinct conversations | per day |
|---|---|---|---|
| ambiguous (`stream_incomplete`, `upstream_request_timeout`, `stream_idle_timeout`) | 2,352 | **993** | 336 |
| definitive (`previous_response_owner_unavailable`) | 1,883 | 367 | 269 |

The ambiguous class — the one that needs this fence — is not the minor half: it touches ~3x more distinct conversations than the definitive class. The definitive class concentrates 1,883 turns onto 367 conversations (~5.1 each: the client retries a dead anchor and collects a 502 every time), while the ambiguous class spreads 2,352 turns over 993 conversations.

Caveat on the upper bound: some eventless `stream_incomplete` rows are probably confirmed pre-dispatch failures, which belong to the definitive class and need no fence. The logs cannot separate them, so 2,352 is a ceiling — but not one that changes the ordering.

Also worth recording here: `upstream_operation_status_unknown` fired **0 times in those 7 days**. The bounded 503 that is supposed to be the fail-closed terminal for this path never occurs in production; these turns surface to the client as a broken `stream_incomplete` stream instead.

## Suggestion

Hold rather than close. Holding costs nothing — the physical column is retained by this PR anyway, the surface is inert, and the branch keeps the work. Merging and re-adding later costs a rewrite of the atomic claim (a serialized `FOR UPDATE` over the session and operation rows, which is the part that makes concurrent reconnects safe), a REMOVED→ADDED round trip on the same requirement within days, and a column re-add migration if #2374 slips past this release.

A re-scope after #2374 lands would capture most of the value without the collision: the genuinely orphaned pieces here — the `operation_attempt_generation` hops that carry a constant `0` around — are dead regardless of whether the fence gets a caller.

PR #2367, opened 2026-09-11
fix(proxy): anchor unanchored HTTP threads to a stable cache key

### LINKED ISSUE #2347: bug(proxy): derived prompt_cache_key does not anchor unanchored HTTP threads — 60% cross-account turn switching and 40pp prompt-cache loss on ~28% of traffic
## Summary

Responses requests that arrive over HTTP with **no continuity identifiers at all** (no `conversation_id` in the payload, no `session_id` header, no `x-codex-turn-state`) are not held on one account across turns. The derived `prompt_cache_key` from `_derive_prompt_cache_key` is supposed to anchor them, but in production **60% of consecutive turns of the same logical thread land on a different account**, against **8% for threads that carry `conversation_id`** measured on the same day. The result is a large, permanent prompt-cache loss.

This is not a symptom of an upstream incident. The numbers below are from a day with **zero overload isolation events**, so soft-sticky rerouting cannot explain it.

## Measurements

Production fleet, ~28 subscription accounts, `capacity_weighted` routing, `sticky_threads_enabled=true`, `openai_cache_affinity_max_age_seconds=1800`.

Threads were reconstructed conservatively: consecutive requests on the same API key, less than 10 minutes apart, with non-decreasing `input_tokens`, and `input_tokens > 5000`.

| Day | Traffic class | Cross-account switch rate between consecutive turns | Cached share of input tokens |
|---|---|---|---|
| Clean day, 0 isolation events | has `conversation_id` | **8%** | **91.0%** |
| Clean day, 0 isolation events | no continuity keys | **60%** | **51.2%** |
| Incident day, 19 accounts isolated | has `conversation_id` | 26% | 82.7% |
| Incident day, 19 accounts isolated | no continuity keys | 64% | 50.0% |

Two separable defects fall out of this table:

1. **This issue.** Unanchored HTTP threads switch accounts 60% of the time even with no isolation in play, and cache only 51% of input tokens.
2. Separately, overload isolation reroutes *soft* prompt-cache sticky owners, which tripled the switch rate for anchored threads on the incident day (8% to 26%) and cost about 8 points of cache. That is expected behaviour of the isolation stage, but its cache cost is currently invisible.

### Scale and cost

Unanchored requests are about **28% of all proxy traffic** (16,644 requests in a 10-hour window). They are not short: median `input_tokens` 49,810, p90 453,374.

On the clean day, unanchored traffic consumed **1.52 B input tokens** and cached only **0.78 B** of them. At the 91.0% ratio that anchored traffic achieved the same day, **0.61 B more input tokens per day** would have been cache hits. That is pure waste in latency and quota.

### The failure is payload-shape dependent

Cache-hit ratio for unanchored traffic, grouped by client, same 10-hour window:

| Client (UA group) | Requests | Cached share of input tokens | Median input tokens |
|---|---|---|---|
| `codex_cli_rs` (key A) | 3,318 | 46% | 140,884 |
| `OpenAI` (key B) | 1,332 | 60% | 87,460 |
| `fasthttp` (key C) | 1,045 | **87%** | 141,239 |
| `OpenAI` (key D) | 769 | 50% | 181,368 |
| `AsyncOpenAI` (key E) | 648 | **1%** | 29,714 |
| `codex_cli_rs` (key F) | 618 | 49% | 78,966 |
| (no UA) (key G) | 274 | 4% | 5,942 |
| `OpenAI` (key H) | 178 | 37% | 95,272 |

Key C reaches 87% with no continuity keys, so the derivation *can* anchor a thread. Keys E and G are near zero. The spread strongly suggests the derivation depends on payload shape rather than on thread identity.

## Suspected cause

`app/modules/proxy/affinity.py::_derive_prompt_cache_key` builds the key from:

- model class
- `api_key.id[:12]`
- `sha256(instructions[:512])`
- `sha256(instruction_input_text[:512])`
- `sha256(first_user_text[:512])`

The docstring states that "successive turns within one session get the same key (first input stays constant)". Two ways that assumption breaks:

1. **`first_user_text` is not stable.** Clients that trim or compact history, or that prepend an environment/context block containing a timestamp, cwd, or other volatile field, change the first user item between turns. Every turn then derives a different key and is treated as a brand-new thread.
2. **Empty ingredients fall back to a random key.** When `parts` ends up empty the function returns `uuid4().hex[:24]`. Any client hitting that branch is guaranteed a fresh key per request, so it can never be sticky.

Requests with `input_tokens` in the hundreds of thousands and a 1% cache ratio (key E above) look like case 2 or a per-turn-changing first item.

## Impact

- Roughly 0.6 B input tokens per day resent uncached on this deployment.
- Elevated TTFT on the affected clients, since a cache miss on a 100k+ token prefix is expensive.
- Threads scattered across accounts, which also multiplies exposure to per-account upstream problems and makes owner-bound replay failures more likely (see #2068 for the failover consequence of unanchored full resends).

## Suggested direction

1. Make the derived key independent of mutable transcript content. Anchor on a first-seen identity per (api key, model class, thread) and keep it in the existing sticky store, rather than re-deriving it from payload text every turn.
2. Never return a random key silently. If no stable ingredient exists, treat the request as unanchored and say so in diagnostics instead of minting a per-request identity that looks sticky but is not.
3. Consider deriving from a stable prefix hash of the retained transcript, which survives appends, instead of only the first user item.
4. Emit a counter for unanchored requests and their cache ratio so this regression class is visible without a database query.

## Environment

- Version 1.25.0-beta.6
- `sticky_threads_enabled=true`, `routing_strategy=capacity_weighted`, `openai_cache_affinity_max_age_seconds=1800`
- HTTP downstream transport policy `smart`, upstream stream transport `auto`
- Clients: mixture of native Codex CLI over HTTP, OpenAI/AsyncOpenAI SDKs, and fasthttp callers

Soju06:
## Summary

Responses turns that arrive over HTTP with **no continuity identifier at all** — no `conversation`, no `previous_response_id`, no Codex session/thread header, no client `prompt_cache_key` — were supposed to be held on one account by the `prompt_cache_key` the proxy derives for them. In production they were not (#2347).

Measured on this fleet (~28 pooled accounts) on a day with **zero** overload-isolation events, so soft-sticky rerouting does not explain it:

| Traffic class | Cross-account switch between consecutive turns | Cached share of input tokens |
|---|---|---|
| has `conversation_id` | **8%** | **91.0%** |
| no continuity keys | **60%** | **51.2%** |

The next morning, on a quiet upstream, keyed traffic was at 0% switching / 84.6% cache while the unkeyed class was at 38% switching and **0.1%** cache.

Size of the prize: unanchored traffic is about **28% of all proxy traffic** (16,644 requests in a 10-hour window), and it is not short — median `input_tokens` 49,810, p90 453,374. On the clean day it consumed **1.52 B** input tokens and cached only **0.78 B**; at the 91.0% ratio anchored traffic reached the same day, roughly **0.6 B input tokens per day** would have been cache hits instead of resends. Per-client cache-hit ratio inside that traffic ranges from **1% to 87%** — one `fasthttp` client reaches 87% with no continuity keys at all — which is the signature of a derivation that depends on payload shape rather than on thread identity.

The cause is the derivation. `_derive_prompt_cache_key` hashed model class, `api_key.id[:12]`, `instructions[:512]`, the system/developer input text `[:512]` and the **first user item** `[:512]`. That is stable only for a strictly append-only history:

- a client that trims or compacts history, or that writes a timestamp or cwd into its first item, minted a brand-new key **every turn**;
- two genuinely independent threads sharing their first 512 characters — for `codex_cli_rs`, every session whose first item is the same `<environment_context>` block for the same cwd — collapsed onto **one** key, pinning unrelated work to one account;
- two threads whose `instructions` differ only past 512 characters collided the same way;
- a turn with no `role == "user"` item degenerated to one global bucket per API key;
- with `api_key_auth_enabled=false` the empty-ingredient branch returned `uuid4().hex[:24]` — a sticky key that can never be hit again, yet still wrote a `sticky_sessions` row;
- `_extract_first_user_input` joined every text part of the first user item before slicing to 512 characters, so a p90 453k-token payload was concatenated on the hot path to produce 512 bytes.

All of it was invisible: the key looked sticky and was not.

## Type of change

- [x] `fix:` — bug fix
- [ ] **Breaking change**

Linked issue: Fixes #2347

## OpenSpec

- [x] This PR touches a codex-faithful path (request shape) and preserves upstream-equivalent behavior

Change directory: `openspec/changes/anchor-unanchored-http-thread-cache-keys/`

## Changes

- **A verified thread anchor replaces the content hash.** A new bounded, process-local `ThreadAnchorIndex` (`app/modules/proxy/thread_anchors.py`) records each thread's trailing item-digest window. A turn reuses that thread's key only when the recorded window, from some offset to its end, is **exactly** the head of the new turn's items. A match starting mid-window must cover at least two items; a match that consumes the recorded window from its beginning may cover one, which is the ordinary second turn of a one-item opening. That covers appends, leading-history trims, and byte-equal re-derivation of the same body (bridge-to-HTTP fallback, cross-transport replay) — and nothing else. No partial, fuzzy, truncated-prefix or summary-aware match can reuse a key, so this is strictly *fewer* false merges than today, not more.
- **Both collision classes removed.** Item digests are domain-separated by API key, model class and the **complete** `instructions`, and each item is hashed over its full canonical encoding, so neither 512-character truncation class can merge two threads. An item whose encoding exceeds the 1 MiB per-item ceiling makes the body unanchorable rather than merging on a partial digest.
- **Compaction is admitted, not bridged.** A compacted turn does not extend the recorded transcript, so it mints a new key and reports `anchor_reset`. Its upstream prefix cache is genuinely cold; heuristic suffix matching to "survive" compaction is exactly where false merges would come back.
- **The `uuid4()` fallback is gone.** A body with nothing to anchor still gets a stable derived `prompt_cache_key` forwarded upstream — the `responses-api-compat` MUST — but supplies **no** sticky routing key, so selection takes the unbound path and no single-use `sticky_sessions` row is written.
- **Serialization is bounded during encoding.** `iterencode` chunks are folded straight into the hash (256 KiB per window, 32 items, 1 MiB per item), removing the unbounded join on the hot path.
- **Observability first.** Every resolution reports `payload` / `disabled` / `anchor_hit` / `anchor_new` / `anchor_reset` / `unanchorable` on the `proxy_request_shape` trace next to `sticky_key_source` — on the streaming, compact, direct-WebSocket and HTTP-bridge paths — and on `codex_lb_prompt_cache_key_derivation_total{outcome}`. This answers the open objection on #2347 that the production percentages were reconstructed rather than correlated to resolved keys.
- **Migration.** Minted keys carry a `v2t-` prefix (`v2u-` when unanchorable). `prompt_cache` rows of the retired shape already expire through `purge_prompt_cache_before`; `sticky_thread` rows have no TTL, so the cleanup pass runs a one-shot bounded sweep over the three retired prefixes and retires itself once a pass finds nothing. Because anchored keys are per thread, the same pass also purges `v2t-` `sticky_thread` rows idle past the freshness window that expires their anchor, in bounded batches — giving the no-TTL kind the bound its sibling already had.

Derivation still runs **after** the transport decision (`retry.py`), so no request moves between HTTP and WebSocket. No new setting: the behaviour rides the existing dashboard `openai_cache_affinity` boolean, so there is no new `CODEX_LB_*` env var, dashboard column, tier entry or migration.

## Memory and failure modes, stated up front

Anchor state is process-local, positive-only and LRU/TTL-bounded: at most 2048 anchors x 32 x 8-byte item digests, plus at most 65536 reverse-index rows holding at most 8 thread-key references each. Measured at **~9 MiB** with every cap saturated; it never grows with traffic. A test asserts the caps hold.

Losing anchor state is always safe — the next turn mints a new key and says so. But it *is* lost on restart, and during a blue/green swap both colors serve with independent indexes, so one thread can hold two keys and two owners for the swap window. That is bounded by the freshness window and costs one prefix-cache miss; it is stated here rather than discovered in production.

## Simplicity

- [x] Works with zero configuration
- New setting(s): none — the anchor TTL is the existing dashboard cache-affinity freshness window
- [x] README sections / `.env.example` / dashboard nav within budget

## What this does and does not recover

Recovers: threads that resend a transcript and either append to it or trim its front — the dominant unanchored shape, and the one that accounts for the 0.6 B/day figure.

Does **not** recover, and now reports instead of pretending:

- a client that compacts every turn — each compaction is a genuine upstream cache reset (`anchor_reset`);
- a delta-only turn that shares no item with the previous one (`anchor_new` / `unanchorable`);
- an empty body, a non-list input, or a body carrying an item too large to digest exactly (`unanchorable`, no sticky row);
- anything after a process restart or an anchor eviction.

It also does nothing for `http_continuation.inferred_http_bridge_key`, which has the identical first-user-item instability without even the 512-character truncation and is preferred over `affinity.key` for bridge socket locality. Different blast radius (socket sharing), left for a follow-up so this derivation can be measured on its own first.

## Verification plan on the fleet

`codex_lb_prompt_cache_key_derivation_total{outcome}` plus the `conversation_id`-null / `cached_input_tokens` split in `request_logs` that #2347 already used — now correlatable per request against the `affinity_source` / `affinity_key_hash` columns landed by #2352. A healthy fleet shows `anchor_hit` dominating `anchor_new` + `anchor_reset` for unanchored traffic.

## Test plan

```
uv run pytest tests/unit -q
# 9910 passed, 4 skipped

uv run pytest tests/integration/test_proxy_sticky_sessions.py \
  tests/integration/test_cache_locality_fix.py \
  tests/integration/test_proxy_compact.py \
  tests/integration/test_proxy_responses.py -q
# 380 passed

uv run ruff check app tests && uv run ruff format --check app tests
# All checks passed / 1208 files already formatted

uv run python scripts/check_proxy_timing_seams.py
# proxy timing seam checks passed

openspec validate anchor-unanchored-http-thread-cache-keys --strict
# valid

codex review --base origin/main
# two P2 findings, both agreed with and fixed in this branch:
#   truncated per-item digests could merge two large threads -> items hashed in full
#   derivation_outcome missing on the WebSocket and HTTP-bridge shape logs -> plumbed
```

Coverage added: stability across a growing transcript (including past the retained window, and under a randomized 200-step append sequence over 12 interleaved threads that also asserts they never merge); stability when the client trims leading history, twice in a row; two threads from one API key staying distinct across both 512-character collision classes; a one-item tail alignment being rejected while a one-item opening turn still anchors its second turn; the anchor and index caps holding with LRU and TTL eviction; the per-item and per-window encoding ceilings; unanchorable bodies attaching a stable key twice while supplying no sticky key; compaction reporting `anchor_reset`; and the cleanup scheduler sweeping the retired shape exactly once while continuing to purge the anchored shape.

## Screenshots / output

No dashboard surface.

## Checklist

- [x] Ran the relevant `make` / pytest subset locally.
- [x] `openspec validate anchor-unanchored-http-thread-cache-keys --strict` passes.
- [x] Simplicity gates reviewed (PRINCIPLES.md P1-P5).
- [x] CHANGELOG not edited by hand.


Soju06:
## Adversarial review left three P1s unresolved — please do not merge as-is

Two independent review passes over this branch converged on the same three P1-class defects, all reproduced by execution rather than inspection. I am the author of the branch and am flagging them against my own work; they are not style points.

**P1 — the unanchorable key is promoted to a real sticky key.**
`affinity.py:718` assigns `payload.prompt_cache_key = anchor.attached_key` unconditionally, including when `sticky_key is None`. A second resolution of the *same mutated payload object* — which happens for real on the bridge → `_stream_with_retry` fallback (`http_bridge/streaming.py:1160` → `retry.py:436`) — then takes the client-supplied branch and returns the constant `v2u-<class>-<apikey12>` as a `PROMPT_CACHE` sticky key. Every unanchorable request from one API key (any inline base64 item over the 1 MiB ceiling) therefore collapses onto one sticky row and one account, and the shape log reports `source=payload` for a key the proxy minted. This violates the change's own spec MUST.

**P1 — window collapse on large trailing items.**
`thread_anchors.py:185` breaks the backward walk once `encoded_chars >= 256 KiB` and `len(digests) >= _MIN_WINDOW_ITEMS`, and `_MIN_WINDOW_ITEMS` is 1. A thread whose last item alone is ≥256 KiB records a one-item window that differs every turn, so it mints a new key every turn where `main` held one. Reviewers traced 8 turns to 8 keys; a threshold scan hit 10/10 at 96 KiB per item. That is precisely the p90 453k-token cohort this change exists to fix, and the `anchor_reset` label reads as benign compaction.

**P1 — one oversized item poisons roughly 16 turns.**
`_MAX_ITEM_ENCODED_CHARS = 1 MiB` makes the whole body unanchorable, and the reversed walk keeps reaching that item for about 16 turns, so `sticky_key` stays `None` and everything is forwarded under the same constant `v2u-` key.

**P2 — a new false-merge class, which contradicts the PR's "strictly fewer false merges" claim.**
The whole-window-consumed acceptance branch at `thread_anchors.py:279` lets thread B claim thread A's key when A's recorded window is a prefix of B's (two shared preamble items, or a one-item shared opening across parallel workers). B's `register` then overwrites A's window, so **A permanently loses its key**. Both reviewers reproduced this. Avoiding exactly this was the stated non-negotiable for the change.

**P2 — both cleanup sweeps target the wrong `StickySessionKind`.**
A derived key can never be written as `STICKY_THREAD` (that branch is only reachable when `openai_cache_affinity` is false, where `sticky_key` is `None`), so `_purge_expired_anchored_sticky_threads` can never match a `v2t-` row. Meanwhile the legacy sweep *can* match client-supplied keys and delete them from the deliberately TTL-less `sticky_thread` kind, and it never self-retires. The three tests assert only mock call shape.

Also worth noting for scheduling: this branch conflicts with #2375 in five files (`affinity.py`, `streaming/retry.py`, `compact.py`, `http_bridge/streaming.py`, `websocket/mixin.py`), and the conflict is **order-independent** — whichever lands second conflicts, because #2375 deletes the per-transport `sticky_key_source` ladders that this branch edits. Resolution is semantic (fold `derivation_outcome` into `AffinityObservation.from_policy`), so landing #2375 first and rebasing this on top is the cheaper order.

I will push fixes for the P1s and the false-merge class before asking for a merge.

Soju06:
All six defects from the two adversarial passes are fixed. Each reproduced by execution first, and each fix has a test verified to **fail on the previous tip and pass now** — I checked that direction explicitly per fix, listed below.

## Fixed

**P1 — the unanchorable key was promoted to a real sticky key.** The placeholder is gone entirely, not guarded: `_PromptCacheAnchor` now carries one key that is both forwarded and routed on, and an unanchorable turn attaches **nothing** to the payload. There is no longer a `v2u-` constant to promote. I checked what breaks by not writing it: `http_continuation_signal` falls through to `http_history_signal`, which still classifies any real transcript, and the only bodies now affected (non-list input, empty list, an oversized newest item) have nothing for upstream to cache anyway. The anchorable case still attaches, which is what makes a bridge fallback or a ring-forwarded replica reuse the identical key.
*Before:* second `_resolve_prompt_cache_key` on one payload returned `v2u-std-ak_test_0011` with `source=payload`. *After:* `None`, `source=none`. (`TestUnanchorableKeyIsNeverPromoted`, 4 tests)

**P1 — window collapse on large trailing items.** `_MIN_WINDOW_ITEMS` raised 1 → **8**: the byte ceiling may no longer end the walk before eight items are recorded. Eight leaves room for a four-item turn on top of the four-item acceptance floor.
*Before:* 10 turns → 10 keys at both 96 KiB and 256 KiB per item. *After:* 10 turns → 1 key. (`TestLargeTrailingItemsDoNotCollapseTheWindow`, parametrised at both sizes)

**P1 — one oversized item poisoned ~16 turns.** An item past `_MAX_ITEM_ENCODED_CHARS` is now a **window boundary**: the walk stops there and keeps the items after it. It is still never digested, whole or truncated.
*Before:* `build_thread_window` returned `None` and `sticky_key` stayed `None` for every turn in reach of the item. *After:* anchorable, and the thread settles on one key within a few turns. (`TestOversizedItemDoesNotPoisonTheThread`)

**P2 — the false-merge class.** New rule, stated in `thread_anchors`' matching contract and in the spec:

> accept only when at least `_MIN_OVERLAP_ITEMS` (**4**) items match at any offset, **or** the two windows are byte-equal (re-derivation of one body). Nothing else.

The floor now applies to the `offset == 0` strict-extension case too, which is the whole point: "thread B opens with everything thread A recorded" and "thread A appended a turn" are byte-identical evidence, so the only available defence is to demand a shared run too long to be a generic opening. The whole-window-consumed branch is deleted.
*Before:* thread A's 2-item opening `[P1,P2]`, then B sends `[P1,P2,b1]` → B got A's key, and B's `register` overwrote A's window so A's next identical body minted a fresh key. *After:* B gets its own key and A's re-derivation still returns A's key. (`TestSharedPreambleNeverTransfersAKey`, covering the two-item and one-item openings, "A never loses a key it owns", and an established thread holding its anchor against five siblings)

The cost is honest and documented: one extra mint per thread, on the smallest body of its life. Six existing tests that encoded the old permissive behaviour were rewritten rather than deleted, and the route-level `/v1/responses` test now covers it end to end (early reset on a one-item opening, then the account **held** once the transcript clears the floor even after usage flips to favour the other account).

**P2 — both cleanup sweeps targeted the wrong `StickySessionKind`.** Confirmed: a derived key is only ever supplied when `openai_cache_affinity` is on, which is the branch that classifies `PROMPT_CACHE`; the `STICKY_THREAD` branch is `elif sticky_threads_enabled`, where the derivation supplies no key. So the `v2t-` sweep could never match, and the legacy `std-`/`codex-`/`mini-` sweep could only match **client-supplied** keys and delete them from the deliberately TTL-less kind. Both sweeps are removed; `purge_prompt_cache_before` already retires every derived row of either shape. Replaced the three mock-shape tests with a behavioural one plus a **real-database** test.
*Before (real DB):* a client-supplied `codex-…` `sticky_thread` row was deleted by a cleanup pass. *After:* it survives, while derived `prompt_cache` rows of both shapes still expire.

**P3 — memory bound.** `~9 MiB` → `~17.6 MiB`, with the measurement method spelled out. An independent saturation run here measured 16.5 MiB RSS, consistent with the ~2x correction.

## From `codex review --base origin/main`

Four more, all agreed and fixed, all with fail-before tests:

1. **P1 (its call):** `test_v1_auto_derived_key_stable_across_turns` had to be reconciled with the overlap floor at the route level, not just updated away. Extended to four turns so it now proves both the deliberate early reset and the held account.
2. **P2:** an evicted / expired / re-registered anchor left its key in every reverse-index candidate row, where it occupied one of the 8 capped slots on a popular digest. `_forget_index_references` now runs on all three paths.
3. **P2:** `iterencode` yields a large string scalar as one already-materialised chunk, so the per-item ceiling could only fire after allocating the thing it refuses. A structural pre-check (sum of raw string lengths is a lower bound on the encoding) refuses first.
4. **P2 → a real bug underneath it:** the pre-check summed raw lengths while the encoder ran `ensure_ascii=True`. Chasing that showed the ceiling itself was wrong for this fleet: every non-ASCII code point becomes `\uXXXX`, so a Korean or Japanese transcript expands ~6x and hit the 1 MiB per-item ceiling at ~175k characters. A 240k-character Korean item was being turned into a window boundary. Encoding now uses `ensure_ascii=False`; the digests only ever compare against each other in one process, so the encoding needs to be stable, not identical to any other fingerprint in the tree.

Final `codex review --base origin/main`: *"No actionable correctness issues were identified."*

## Not fixed

Nothing from the review comment is outstanding. Two residual limits are now written down rather than left implicit:

- Two threads that diverge only **inside** an item past the per-item ceiling and then agree on the four items after it can still share a key. That needs four byte-identical items after the divergence, and it is the accepted price of not stranding the thread.
- Byte-identical bodies still resolve to one key. That is not a merge to fix — it is the same content, and upstream's prefix cache treats it as the same prefix anyway.

## Verification

`openspec validate anchor-unanchored-http-thread-cache-keys --strict` clean, with the normative text updated to match every behaviour above (acceptance rule, the window-item floor, the oversized-item boundary, attaching nothing when unanchorable, no `sticky_thread` prefix sweep, index-reference cleanup, and non-ASCII sizing) plus seven new scenarios.
`ruff check` / `ruff format --check` / `ty check app` clean.
1,536 tests across the affinity, sticky, selection, cleanup-scheduler and thread-anchor suites, plus 4,512 across the proxy / bridge / websocket / compact / stream unit suites — all passing.

PR #2370, opened 2026-09-11
feat(proxy): scope outbound thread identity per account (default off)

Soju06:
## What this does

One logical downstream thread can be served by several of the 28 pooled ChatGPT accounts. The identifiers that name that thread travel upstream unchanged today: the outbound `prompt_cache_key` is either the client's value or the one `_derive_prompt_cache_key` computed, and the session/conversation headers are not in `IGNORE_INBOUND_HEADERS`, so every header builder forwards them verbatim. Account A and account B therefore present the **same** identifier for the same thread to an upstream that scopes those resources per account.

This adds a deterministic per-account rewrite of those identifiers, behind a **default-off T3 flag**.

It is the same defect `apply_codex_installation_metadata` already fixes for the Codex installation id, governed by the existing requirement *"Codex installation metadata must be account-owned"*.

## What is scoped

`app/core/clients/account_scoped_identity.py` is a pure module: a frozen namespace UUID and a mapping with no randomness, no clock and no process state. The same `(account, original value)` pair always maps to the same output; a UUID-shaped input maps to a UUID-shaped output (`uuid5` over length-framed material), anything else to a `cxlb-`-prefixed 128-bit digest so an operator reading an upstream trace can tell a rewritten identifier from a client-authored one.

The salt is the per-seat `accounts.codex_installation_id` — **not** `chatgpt_account_id`, which is the workspace identity every seat of a Team/Business org shares and is nullable; salting with that would leave same-workspace seats presenting identical values and defeat the whole point.

Applied to:

| surface | what | where |
|---|---|---|
| headers | `session_id`, `session-id`, `x-codex-session-id`, `x-codex-conversation-id`, `thread-id`, `x-parent-session-id`, `x-opencode-session`, `x-session-id`, `x-session-affinity`, `x-codex-parent-thread-id`, `x-codex-window-id` | the three outbound builders (`proxy.py` HTTP + per-request WS, `proxy_websocket.py` handshake) |
| body | `prompt_cache_key` and the `client_metadata` mirrors of the last two headers above | `_stream_responses_with_session` chokepoint (HTTP, per-request WS, and the websocket-rejection fallback all derive from one mutation), the HTTP bridge send boundary, the direct-WebSocket send boundary, and `compact_responses` |

The mapping is deliberately **not idempotent**, so it is applied exactly once per request. In particular it is *not* inside `apply_codex_installation_headers`, which runs twice on the HTTP and per-request-WebSocket paths (`proxy.py:3680` and `:3795`) — a twice-scoped value there would differ from the once-scoped value the persistent bridge presents, and a thread migrating between transports would lose its cache.

## What is deliberately **not** scoped

- **`x-codex-turn-state`** and **`previous_response_id`**. Both are upstream-issued opaque values that must round-trip verbatim. codex-lb never decodes turn state, and already *deletes* it when a thread moves accounts ("account-local state must die with its upstream"). The existing assertions in `test_proxy_upstream_fingerprint.py` that they are forwarded unchanged still pass, and the new tests assert it directly.
- **Prompt content.** Injecting a per-account token into the user/system prompt has no correctness or caching justification — a stable outbound `prompt_cache_key` already gives per-account cache anchoring — it pollutes text the user authored, and it can leak into model output. Identifier scoping only.
- **Routing.** `_resolve_prompt_cache_key` still writes the account-neutral derived key onto the request model *before* account selection, and `_sticky_key_for_responses_request` still reads it there. The scoped value is written only onto the outbound copy, after selection, so selection cannot shatter into per-account lanes. Same for the durable bridge operation fingerprint: the bridge and direct-WebSocket rewrites are send-only and never reach `request_state.request_text`, so an account swap cannot change the operation identity and miss the spool lookup mid-recovery. Both are asserted by test.

### Deferred, stated plainly

- `/transcribe`, `/files` and the generic Codex control egress forward `x-codex-*` headers wholesale and get no per-account post-processing today. They still cross accounts. The invariant this PR establishes holds on the Responses surfaces it names, **not globally**.
- The realtime live sideband (`connect_live_websocket`) receives no salt, so it is unchanged.

## Why the flag defaults off

Enabling it changes what every pooled account presents upstream, so the first turn of each live thread re-anchors and costs exactly one cache miss. That is a fleet-wide, one-time cost an operator should schedule, and it must be reversible from the dashboard without a restart — turning the flag back off restores the previous identifiers exactly. So it cannot be a fixed default.

The off path is inert **by construction**: the flag resolves to a `None` salt, and every scoping call site is guarded on it. `test_flag_off_outbound_request_is_byte_for_byte_unchanged` runs the dispatch twice with the flag off — once normally, once with every scoping entry point monkeypatched to raise — and asserts header-by-header and byte-by-byte equality. That is the most important test here.

## Settings budget: why not a default

`[settings_fields].max` goes **96 → 97**. Per PRINCIPLES.md P2 the question is why this cannot just be a constant. It cannot, for the reason above: it is a rollout fence, not a tunable. The correct long-term value is "on", but getting there costs a synchronised cache miss across ~28 accounts, so the operator picks the moment, watches the cache-hit ratio, and rolls back from the dashboard if it regresses. A hardcoded default gives them neither the moment nor the rollback. Once it has been on in production for a release the field can be retired and the budget lowered again.

Full T3 procedure in this diff: nullable `dashboard_settings` column + alembic `20260911_000000_dashboard_account_scoped_thread_identity` (`down_revision = 20260910_020000_add_dashboard_role_mappings`), `SETTING_TIERS` entry, `DASHBOARD_SWITCH_SETTINGS` so the dispatch path reads the dashboard value off the request-bound overlay with no DB read, `GET`/`PUT /api/settings` with provenance and the tri-state update (omitted = unchanged, `null` = back to inherited, value = store), dashboard card with `en`/`ko`/`zh-CN`, regenerated `docs/reference/settings.md`. Precedence is unchanged: code default < environment < dashboard, and the environment is a fallback, never a seed and never an override.

## Relation to #2347

Related, but this PR does **not** fix the cache-hit defect in #2347. Unkeyed traffic (~28% of proxy traffic, median `input_tokens` 49,810) switches accounts 38–60% of the time and caches 0.1–51%, burning roughly 0.6 B avoidable input tokens/day. That is a *derivation* problem in `_derive_prompt_cache_key` — the payload shapes where it fails to anchor a thread — and it needs its own change. What this PR does is make the identifier each account presents correct, and give a thread a stable per-account anchor once the flag is on.

## OpenSpec

`openspec/changes/scope-outbound-thread-identity-per-account/`, `--strict` clean.

- `upstream-proxy-routing`: **ADDED** "Outbound thread identity must be account-scoped" (salt, determinism, shape, exactly-once, excluded continuity tokens, selection invariance, flag-off no-op).
- `responses-api-compat`: **MODIFIED** "Use prompt_cache_key as OpenAI cache affinity" and "Codex backend session_id preserves account affinity" — both currently mandate unchanged forwarding, which this change contradicts when the flag is on. All 2 + 11 existing scenarios were extracted programmatically and carried over verbatim; one scenario added to each.
- `configuration-tiers`: unchanged — this follows its existing procedure rather than altering it.

## Tests

- `tests/unit/test_account_scoped_identity.py` — determinism, per-account divergence, crafted-collision resistance from length framing, UUID-shape preservation, opaque form, blank/no-salt passthrough, turn-state exclusion, copy-on-write metadata.
- `tests/unit/test_account_scoped_identity_dispatch.py` — all three builders parametrised (flag-off byte-identical, vocabulary-only rewrite, turn state untouched, two accounts diverge, one account stable); the flag-off byte-for-byte dispatch test; header/body agreement; sticky-key invariance.
- `tests/unit/test_proxy_http_bridge.py` — the send-only bridge rewrite leaves the durable operation fingerprint unchanged.
- Settings round trip with provenance + dashboard-overlay reachability (`tests/integration/test_settings_api.py`), migration upgrade/downgrade (`tests/integration/test_migrations.py`), frontend card tests.

Local: `pytest tests/unit` 9943 passed; `pytest tests/integration/test_settings_api.py tests/integration/test_migrations.py` 146 passed / 8 skipped (PostgreSQL-only); `ruff check .` + `ruff format --check .` clean; `ty check` clean; proxy-architecture, cancellation-safety, timing-seam, settings-tier and simplicity-budget checks pass (`service.py` stays at 2600/2600 and `_service/streaming/mixin.py` at 1100/1100 — no line added to either); frontend lint, typecheck and vitest pass; `codex review --base origin/main` found no P1/P2.


Soju06:
## Superseded by #2376 — closing, with its coverage ported across

Closing this in favour of #2376, which takes the same problem further. The reasoning, and what was carried over, so nothing here is lost.

**Why #2376 and not this one.** The measured record is that the upstream prefix cache is **not partitioned per account**: across three runs and 26 direct calls with a fresh-nonce ~28k-token prefix, content seeded by account A produced near-full hits on four other accounts (28,032/28,168, 27,136/27,328, 28,416/28,588). `prompt_cache_key` does not gate the hit — one arm hit with a *different* key while several missed with the same one — and neither does the account: same-account repeats hit 1/5 while other accounts hit 3/8. Hits are sporadic at roughly 20–40% regardless of caller, which is what a node-local cache looks like.

That has a direct consequence for this PR. Scoping the cache key and the session headers is cache-neutral and safe, but it **isolates nothing**, because the hit is decided by the content prefix. Only #2376's third leg changes that. This PR cannot be justified under an isolation rationale; the honest remaining rationale is identifier correctness — sending account A's session name on account B's request is simply wrong — which is real but smaller, and does not on its own earn a `[settings_fields]` budget slot.

The decisive practical point is the rollout. Evaluating isolation requires trialling it on **one API key** while the fleet stays `shared`. Only #2376 has `api_keys.thread_cache_identity_override`; this PR is fleet-wide only, so it cannot run that step at all.

**What was ported.** #2376's placement was narrower than this PR's — two call sites against seven — and that gap was a real defect, not a smaller scope: the HTTP session bridge dispatches `response.create` itself, so a thread served partly by the bridge and partly by the per-turn bypass alternated between a scoped and an unscoped identity turn by turn. Both of this PR's extra surfaces are now on #2376:

- the **HTTP session bridge**, following this PR's send-only design so the scoped text never reaches `request_state.request_text` or the durable operation fingerprint;
- the **websocket-rejection HTTP fallback**, which rebuilt headers from the raw inbound set and would otherwise have sent a scoped body with unscoped headers.

The single-mutation-before-the-transport-split structure from this PR is what avoids the fallback divergence, and #2376 keeps it.

**What is still open on #2376**, so this is not a claim that the problem is solved: the direct downstream WebSocket surface remains a documented gap, and flipping the mode does not clear retained `x-codex-turn-state`, so live threads replay continuity tokens under a session name that seat has never seen. Both are recorded in its spec rather than papered over.

Thanks for the work here — it is the reason the surviving PR covers the bridge at all.

PR #2372, opened 2026-09-11
feat(dashboard): add a cross-account prompt cache isolation probe

Soju06:
## Why

Whether the upstream prompt cache is partitioned per account decides how much cross-account failover actually costs, and the answer is not the obvious one.

A hand-run experiment on this deployment sent a ~28,000-token prefix carrying a fresh nonce on one account, then the identical bytes on four siblings. The siblings came back near-fully cached — **28,032/28,168**, **27,136/27,328** and **28,416/28,588** across three runs. `prompt_cache_key` did not gate the hit (one arm hit with a *different* key, several missed with the *same* one) and neither did the account: same-account repeats hit 1/5 while other accounts hit 3/8. That is the shape of a node-local prefix cache that is not partitioned per account, and right now it is a **benefit** — a thread that fails over from A to B still hits A's warm prefix, worth ~27-28k input tokens per hit.

That experiment only existed as a shell transcript. Before enabling per-account outbound cache-identity scoping, and again afterwards, "did the sharing actually stop?" has to be answerable by measurement rather than memory. This turns the experiment into a button.

## What this does

`Settings → Advanced → Cross-account cache isolation probe` (gated on `ops:write`):

1. Builds a filler prefix of roughly 28,000 tokens keyed on a nonce **fresh for the run**, explicitly marked as an inert reference corpus to ignore.
2. Sends it on one **seed** account N times (default 3), recording `usage.input_tokens_details.cached_tokens`.
3. Sends the identical prefix once on each of up to M **other** accounts (default 4).
4. Reports a per-call table plus the headline: did any non-seed account register a hit.

### The nonce is in the first line, on purpose

Upstream matches on a *prefix*. A nonce appended after 28,000 tokens of fixed filler would leave those 28,000 tokens byte-identical to the previous run, so the second run would measure the **first run's** cache and report a spurious hit on every account. The nonce is therefore in line 1 and every filler word derives from it, so two runs share no line at all. `build_probe_prefix` is a pure function of `(nonce, target_tokens)` built from a keyed BLAKE2b stream rather than `random`, so every call of one run sends byte-identical content on every replica and interpreter — a prefix that drifts between calls can never hit and would report isolation that is not there.

### Reading the answer is asymmetric, and the UI says so

A hit on a non-seed account is **proof** the cache is not partitioned per account: that content existed nowhere before this run. The converse is much weaker — hits are sporadic (~20-40%) for *every* caller, so with four siblings a fully shared cache produces zero hits with probability 0.13-0.41. So:

| verdict | meaning |
|---|---|
| `cross_account_sharing` | a sibling hit — proof the cache spans accounts |
| `no_cross_account_hit` | the seed cached, no sibling did — **evidence, not proof**; repeat before concluding |
| `inconclusive` | the seed never cached either, or every sibling call failed — nothing can be concluded |

The card states in-line that **a miss on the seed account is expected and is not a failure of the probe**, because an operator who reads a seed miss as a broken probe will re-run it and spend quota for nothing. The `inconclusive` verdict exists for the same reason in reverse: without one same-account hit there is no evidence the prefix was cacheable at all, so silence from the siblings says nothing.

This PR makes no claim that namespacing `prompt_cache_key` isolates anything. Measurement says hits are not key-gated, so that change is cache-neutral, not isolating; the only lever that isolates is changing the content prefix per account.

## Spending real quota is gated five ways

- **Cost preview first.** `GET /api/diagnostics/cache-isolation-probe` returns the seed, every candidate sibling, the model, the planned call count and the estimated input tokens — before anything is sent.
- **Explicit confirmation.** `POST .../run` with anything but `confirm: true` is rejected `400 cache_probe_confirmation_required` without issuing a call or consuming the budget.
- **Caps.** At most 5 seed repetitions and 5 siblings. The sibling count is an *upper bound*, not a demand, so a three-account pool still runs with what it has instead of being refused.
- **Shared rate limit.** A database-backed budget of 3 runs/hour on a fixed key — deliberately shared across operators *and replicas*, because the protected resource is the pool's quota, which two operators drain just as fast as one.
- **Pool pressure refusal.** `409` before any call when the proxy is degraded, when every account breaker is open, when fewer than two accounts are active, or when ≥25% of the routable pool is rate-limited or quota-exceeded. Paused / deactivated / reauth-required accounts are operator or credential state rather than pressure and leave the denominator entirely.

## Reuse and hygiene

Calls go through the existing `AuthManager.ensure_fresh` → `resolve_upstream_route` → `stream_responses` path — the same primitives the limit warm-up uses — so the probe inherits each account's credentials, its proxy binding (including fail-closed behaviour) and the configured stream timeouts without restating any of them. No raw sockets. Calls run **sequentially, seed first**, which the experiment requires anyway and which is also the gentlest possible load on a pool serving real traffic.

Nothing generated is persisted. The nonce, the prefix and every byte of response content stay in memory; the audit event carries only the run id, model, verdict, account id and token counts. Because `stream_responses` archives at its own gate, the run is wrapped in a new context-scoped `suppress_conversation_archive()` — the archive records what Codex and the upstream said to each other, and ten locally generated 28k-token corpora are neither. The operator's `conversation_archive_enabled` setting is untouched and a request already in flight keeps being archived.

## UI

Screenshots could not be attached from this non-interactive environment. The card's rendered copy, pinned by `cache-isolation-probe-section.test.tsx`, is:

```
Cross-account cache isolation probe
Send one single-use 28k-token prefix on a seed account, then the identical bytes on
its siblings, and read back how much each one was served from cache.

Seed repetitions [3 v]        Other accounts (at most) [4 v]

Estimated cost: 7 calls, about 196,042 input tokens of real quota.        [ Run probe ]
Seed sol-01, then 4 other account(s), on gpt-5.6-codex-low.

⚠ How to read this: a hit on a non-seed account means the upstream prefix cache is not
  partitioned per account. Hits are sporadic for every caller, so a miss on the seed
  account is expected and is not a failure of the probe — and a run with no
  cross-account hit is evidence, not proof.

[Shared across accounts]  2/4 other accounts hit · 1/3 seed repeats hit
A non-seed account read back content only the seed had ever sent, so the upstream
prefix cache is not partitioned per account.

 # | Account | Role  | Cached / input tokens | Result
 1 | sol-01  | Seed  |         0 / 28,168    | Miss
 2 | sol-01  | Seed  |    28,032 / 28,168    | Hit
 3 | sol-01  | Seed  |         0 / 28,168    | Miss
 4 | sol-02  | Other |    28,032 / 28,168    | Hit
 5 | sol-03  | Other |         0 / 28,168    | Miss
 6 | sol-04  | Other |    28,416 / 28,588    | Hit
 7 | sol-05  | Other |            —          | Error (server_is_overloaded)
```

Clicking `Run probe` opens a confirm dialog ("This sends 7 real requests and spends about 196,042 input tokens of account quota. It cannot be undone."); the button is disabled outright while the pool is under pressure, with the refusal reason shown above it. Copy is in `en` / `ko` / `zh-CN`.

## Review findings addressed

`codex review --base origin/main` raised five; all five are fixed in the second commit:

- **P1** archiving would have persisted the generated corpus → scoped suppression seam at the archive's existing single gate.
- **P1** a re-raised transport failure aborted the run, dropping collected rows → `CodexTransportError` becomes a typed failed row in the sender, anything else an untyped one in the orchestrator (reported by exception *type*, never its message, which has not been proved credential-safe).
- **P1** the in-process lock cannot serialize replicas. Rather than hold a distributed lock open for minutes, the contract now says what is actually true: the DB-backed limiter already bounds spend across replicas, and two runs cannot contaminate each other's measurement because each generates its own nonce. Per-replica scope is stated in the spec with that reasoning.
- **P2** selecting five siblings against a plan computed for four understated the confirm dialog by one ~28k call → the plan returns every candidate as `availableOtherAccounts`.
- **P2** returning mid-stream left cleanup to generator finalization → `contextlib.aclosing`.

## Verification

- `tests/unit/test_cache_isolation_probe_prefix.py` — determinism within a run, no shared line across runs, first-line divergence, inert markers, the estimate not understating the prefix, clamping, empty nonce rejected.
- `tests/unit/test_cache_isolation_probe_sender.py` — cached-token reporting, the stream closed before `send` returns, a routed transport failure as a typed row, a non-terminating stream reported rather than raised.
- `tests/unit/test_cache_isolation_probe_service.py` — table shape and ordering, identical prefix across a run and distinct across runs, sub-threshold caching as a miss, all three verdicts, cap enforcement, the pressure matrix, archive suppression scoped to the run, and one account's failure not aborting it.
- `tests/integration/test_cache_isolation_probe_api.py` — plan shape, the confirmation gate (including a bodyless POST), the result table over the wire, caps, both refusal paths, the shared budget throttling at 4 runs, and an unconfirmed run not consuming it.
- `tests/unit/test_conversation_archive.py` — suppression excludes generated traffic without changing the setting, and does not reach a request already in flight.
- `tests/integration/test_dashboard_route_permission_matrix.py` — both routes pinned to `ops:write`.
- Frontend: 6 new component tests; full suite 1524 passed; `typecheck` and `lint` clean.
- `ruff check` / `ruff format --check` clean; `check_proxy_architecture` / `check_cancellation_safety` / `check_proxy_timing_seams` / `check_settings_tiers` pass; `openspec validate add-cross-account-cache-probe --strict` and `openspec validate --specs` clean.

No new setting, no schema change, no migration. The diagnostic is inert until an operator confirms a run.


Komzpa:
@codex review

Issue #2123, opened 2026-09-07
feat(proxy): subscription-exhaustion fallback to an OpenAI-compatible Model Source (design-first, supersedes #1664)


Soju06:
# feat(proxy): subscription-exhaustion fallback to an OpenAI-compatible Model Source (design-first, supersedes #1664)

> Implementation status (triaged 2026-09-10): WP-E #2124, WP-F #2125, WP-A #2165, WP-B #2176, WP-C1 #2208/#2222, and WP-C2 + WP-D #2257 are merged. The feature remains off until the separate-deployment canary and pre-flip anchor follow-up pass. WP-G, captured Codex body fixtures, canary/drills, and v1.1 compaction forwarding remain open. See the latest maintainer update for the exact pre-flip findings; this issue remains open.

> Original design-first proposal (historical): code was gated on owner design sign-off. Full design: `DESIGN.md` (v3, 2026-09-05, targets `origin/main` @ `5ad638b6a`). Section references (§) below point into that document.

## Motivation

- When every eligible ChatGPT/Codex subscription account is usage-exhausted the proxy answers `429 usage_limit_reached` and the client stops working until a window resets. Operators who also hold a paid OpenAI-compatible key want the request to overflow to that source so Codex keeps working, at their cost, only for that gap.
- #428 (closed 2026-07-16) asked for something similar; `sequential_drain`/`reset_drain` answer the *ordering* question but not the *exhaustion* one. #1664 (@oformaniuk, 2026-08-06) implemented the overflow and stalled: author unresponsive, two confirmed P1s, `ty` error, migration parented on an old alembic head, and `api.py` has since drifted 2,500+ lines (SSE parse-once verbatim relay, pure ASGI, retry-loop changes).
- Owner decision (2026-09-05): the feature IS wanted; the architecture must be defect-free on paper first. This issue supersedes #1664.

## Why #1664 could not be rebased (§17)

1. Reservation transfer via a ContextVar wrapped only the startup probe: a post-probe `usage_limit_reached` released the reservation in `_stream_with_retry`'s `finally`, then the fallback reused a released reservation.
2. Treating `response.created` as non-substantive spliced a second response lifecycle into one SSE stream and broke main's indefinite-recovery fence (`while True` never terminates).
3. Undefined return path: a thread whose transcript now holds source item ids / source reasoning ciphertext had no rule preventing it from being replayed against a subscription account on the next turn.
[...]

PR #2373, opened 2026-09-11
docs(openspec): propose walking the account pool before surfacing a 429

Soju06:
## Summary

A single pooled account's HTTP 429 is not the pool's 429, but today it is returned to the client as one.

`failover_decision` stops the walk on `candidates_remaining <= 0`, and every caller derives that number from a fixed per-transport constant — `_STREAM_MAX_ACCOUNT_ATTEMPTS = 3`, `_WEBSOCKET_MAX_ACCOUNT_ATTEMPTS = 3`, `_COMPACT_MAX_ACCOUNT_ATTEMPTS = 2` — that has no relationship to how many usable accounts the pool holds. On a 28-account fleet a request gives up after three, and the failure it then surfaces is the *first rejecting account's* verbatim upstream 429.

**Spec only. No code in this PR** — this is the OpenSpec gate for the implementation PRs that follow.

## Type of change

- [x] `docs:` — documentation only
- [ ] **Breaking change**

## What the delta says

**`account-routing`** — two ADDED, one MODIFIED:

- *A single account's rejection is not the pool's rejection.* An unbound pre-visible `rate_limit` / `quota` / `retryable_transient` failure excludes the account and reselects while candidates remain. Termination is proved three ways: the existing request deadline, a runaway ceiling, and a monotone-progress invariant (a `failover_next` that does not grow `excluded_account_ids` logs and stops). Exactly one health write per attempted account.
- *Pool-walk termination consults the exhaustion probe.* The walk's terminal failure is decided by `probe_pool_usage_exhaustion`, asked once: exhausted pool → the canonical `usage_limit_reached` 429 with `error.resets_at`; anything else, including the drain-strategy decline → the preserved last per-account failure verbatim, so drain strategies and `single_account` stay byte-identical.
- The burst-429 cooldown requirement is MODIFIED so its "MUST NOT change the status and body returned to the client" clause applies to **owner-bound** requests only. Owner-bound behaviour is unchanged; unbound requests now walk.

**`responses-api-compat`** — two ADDED, one MODIFIED:

- *Usage-limit messages classify as account rate limits.* A code-less (or non-rate-limit-code) 429 whose message asserts the usage limit is classified `rate_limit` instead of `retryable_transient`. **This is a live bug fix**: today such a rejection makes `is_upstream_burst_rejection` true, so the proxy applies 1 s / 2 s / 4 s same-account backoff to an account that is out of quota.
- *Model-capacity rejections do not exclude the account from the pool walk.* Capacity describes the model, not the account, so it must not rotate a healthy pool once the walk is unbounded.
- *Streaming Responses requests use a bounded retry budget* is MODIFIED to state that the account-attempt bound is the walk rather than a fixed per-transport constant. All existing scenarios are preserved.

## What this deliberately does **not** do

The existing requirement *Model-capacity messages are retryable transient failures* already decided that a quota or rate-limit **code** keeps its stronger classification even when the capacity message is present, and that the capacity replay wait still applies (scenarios *Quota and rate-limit codes retain their stronger classification* and *Classified quota failures still use the model-capacity replay wait*). That decision stands and is **not** reordered.

Instead the single classifier answers one more question — *is this account exhaustion?* — beside its unchanged failure class. Account selection reads that; account health keeps reading `failure_class`. `test_rate_limit_code_takes_precedence_over_capacity_message` keeps passing unchanged.

This is also why there is no second classifier: the downstream fork `aafqaq/codex-lb-enhanced` added `account_exhaustion_code_for_failover` next to the existing one and then had to call it from ~20 sites, leaving two sources of truth for the same question.

## Simplicity gates

- **No new settings.** `[settings_fields].max` stays 96. `MAX_ACCOUNT_ATTEMPTS_CEILING` is a module constant beside `BURST_SAME_ACCOUNT_MAX_RETRIES`, following the precedent those constants set ("the Settings ratchet is full and this is a transport invariant, not an operator knob").
- No README section, no `.env.example` entry, no new dashboard nav item.
- Not dashboard-visible.

## Scope boundary

Owner-bound requests still cannot move, so this change does not help them. Making an anchored conversation relocatable by rebuilding its context from the durable operation spool is a separate change (`restore-anchored-cross-account-relocation`), which will follow. The two compose: this one decides *whether* a request may walk, that one widens *which* requests can.

## Validation

`openspec validate walk-account-pool-before-surfacing-429 --strict` → valid.


Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Soju06:
Holding this gate. The premise is real and I verified all of it in code — `failover_decision` surfaces on `candidates_remaining <= 0` (`app/core/balancer/logic.py:1493`) with the number coming from `_COMPACT_MAX_ACCOUNT_ATTEMPTS = 2` / `_STREAM_MAX_ACCOUNT_ATTEMPTS = 3` / `_WEBSOCKET_MAX_ACCOUNT_ATTEMPTS = 3` (`app/modules/proxy/service.py:870-872`), and a code-less usage-limit 429 really does land in `retryable_transient` (`app/modules/proxy/helpers.py:92-119`) and therefore in `is_upstream_burst_rejection` (`helpers.py:121-125`), buying 1/2/4 s of backoff on an account that is out of quota. That part is worth landing. But the folder cannot be archived as SSOT in its current shape.

**1. The probe this design is built on no longer exists on main (blocker).**

`proposal.md:35-41` says "The pool-exhaustion answer already exists and is already specified: `probe_pool_usage_exhaustion` (`app/modules/proxy/_load_balancer/exhaustion_probe.py`)". It does not exist. `#2416` (`982125a33`, merged 2026-09-14T01:52Z, "Closes #2123") deleted it, and its commit body names it explicitly: "the read-only pool-exhaustion probe". It also deleted `tests/unit/test_exhaustion_probe.py` (750 lines) and `tests/integration/test_exhaustion_probe_integration.py` (365 lines) — the latter is still listed as a validation target at `tasks.md:105-106`.

This branch is 3 commits behind main with merge base `837964b91` (#2411), which predates the revert, so nothing conflicts textually and `#2391` still compiles on this base while importing `from app.modules.proxy._load_balancer.exhaustion_probe import AdmissionProbeService, probe_pool_usage_exhaustion` in its new `pool_terminal.py`. After a rebase onto current main that import is a `ModuleNotFoundError`.

The drain-strategy decline the delta leans on (`specs/account-routing/spec.md:91`, `context.md:120-123`) was part of the same deleted module — that behaviour has no home on main either. And no requirement in `openspec/specs/**` mentions the probe any more; `responses-api-compat/spec.md:6231` "Pool usage exhaustion is reported as a usage-limit error" predates it and is satisfied by ordinary selection.

Please pick one and say so in the folder:
- reintroduce the probe as an explicit, task-covered deliverable of *this* change (it is ~119 lines plus its tests, and it was withdrawn as part of a feature the repo decided against — resurrecting it needs its own justification), or
- re-target `resolve_pool_terminal_failure` at what main actually has: `ProxyService.check_opportunistic_admission` (`app/modules/proxy/service.py:2069`) over `pool_usage_exhaustion` (`app/core/balancer/logic.py:167`), and rewrite the "Pool-walk termination consults the exhaustion probe" requirement (`specs/account-routing/spec.md:83-113`) in those terms.

**2. The runaway fence cannot be "derived from the current candidate count" (blocker, and the open `@codex` P1).**

Four places require it — `specs/account-routing/spec.md:9`, `specs/responses-api-compat/spec.md:76`, `proposal.md:63`, `proposal.md:94-96` — plus `tasks.md:53`. But `AccountSelection` (`app/modules/proxy/load_balancer.py:216-227`) carries only account/error/lease/ownership metadata, no count, and no task in this folder adds one. `context.md:5-6` concedes that counting usable accounts "would need a new selector API" and declines to add it, while `context.md:85-88` describes the intended fix as binding "the ceiling to exceed the largest supported pool" — a fixed large constant, which is what #2391 actually ships (`MAX_ACCOUNT_ATTEMPTS_CEILING = 128`, "sized to sit far above the largest pool a deployment runs"). As written, merging this gate makes the very next PR in the stack non-compliant with it.

Fix: change the five "derived from the current candidate count" phrasings to a fixed module constant chosen to exceed the largest supported pool by a wide margin, and keep the sentence that it must never become the ordinary bound. Or add the selector-API task — but then say so in `context.md:5-6` too.

**3. Security-scope parity is asserted, not reachable (should fix before archive).**

`specs/account-routing/spec.md:87` requires the terminal probe use "the same eligibility and security-scope filtering ordinary selection applies". `select_account` has `require_security_work_authorized` (`app/modules/proxy/load_balancer.py:598,635,695`); `check_opportunistic_admission` (`service.py:2069`, `load_balancer.py:1440-1456`) takes api_key/model/lease_kind/service_tier and nothing else. Either add the task that extends the seam with the request's security-selection context, or narrow the parity claim to the scopes that exist.

**Smaller items:**
- `context.md:120-123` and the PR body still promise drain strategies stay "byte-identical". With the walk unbounded, `sequential_drain`/`reset_drain` today stop at the third attempt and return that account's body; afterwards they walk further and return a different account's body. The normative scenario at `specs/account-routing/spec.md:109-113` was already softened to "without replacing it with a canonical pool-exhaustion response" — please align the context and the description with it rather than leaving a stronger claim in the non-normative text.
- `specs/usage-refresh-policy/spec.md:5-9` makes the refresh trigger a *distinct usage-limit evidence* field on the classifier result, but the only field `tasks.md:10-18` adds to `ClassifiedFailure` is `excludes_account: bool`. Name the second field in tasks, otherwise the implementer has to invent it (and `failure_class == "rate_limit"` would widen refreshes to plain `rate_limit_exceeded` throttling, which the same requirement forbids).
- `context.md:5-6` says `load_balancer.py` "sits at 3000 of its 3021-line ceiling"; it is 3009 on main today.
- Name drift: the PR body calls the follow-up change `restore-anchored-cross-account-relocation`, `context.md:130-132` calls it `relocate-anchored-turns-across-accounts` (#2374). Pick one.

Everything else checks out and I want to record it so it does not get re-litigated: all three MODIFIED requirements exist on main (`account-routing/spec.md:1173`, `responses-api-compat/spec.md:4495`, `usage-refresh-policy/spec.md:1837`) and every one of their existing scenarios is carried over verbatim — the "all existing scenarios are preserved" claim is true, 8+9+7 in, 9+10+9 out. The earlier `@codex` P1s on deterministic-failover gating, the compact fresh-replay carve-out, narrowing quota-class exhaustion evidence, one-write-per-dispatch, strict validation and WS/compact product-path coverage are all genuinely addressed at `0343efb`.

Komzpa:
@codex review

PR #2374, opened 2026-09-11
docs(openspec): propose relocating anchored turns across accounts

Soju06:
## Summary

A Codex continuation does not resend its conversation — it sends the new turn plus `previous_response_id`, and that identifier is owned by the account that produced it. When the owner is exhausted, suspended or gone, the anchor is worthless everywhere else, so we fail the turn closed (`previous_response_owner_unavailable` 502 on the bridge, the equivalent `response.failed` on the stream path). The conversation dies while the pool still has usable accounts.

We already hold the material to rebuild it. `http_bridge_operations.request_text` + the ordered terminal SSE spool + `parent_response_id` is a complete parent chain, and **`DurableBridgeRepository.get_replayable_transcript` already walks it fail-closed with `max_turns` / `max_bytes` bounds — with zero production callers.** #2336 deleted the request-path caller and deliberately left the storage.

The payload-only half is already specified and already works (*Verified full resend can recover from selection-time owner loss*). The gap is the ordinary case: a **delta resend**, where the client sends only the new turn and the anchor.

**Spec only. No code in this PR.**

## Type of change

- [x] `docs:` — documentation only
- [ ] **Breaking change**

## ⚠️ Blocked on #2366

#2366 (`retire-recovery-dispatch-storage`) deletes `claim_unknown_operation_for_recovery`, the `recovery_dispatch_count` ORM mapping and the `expected_recovery_dispatch_count` CAS predicates **on the grounds that they are callerless**. This change is the caller.

The two cannot both land. Either #2366 is closed, or the ambiguous lane here has to re-add equivalent storage plus two migrations. This needs resolving before implementation starts — the spec delta is unaffected either way.

## The two evidence classes

**① definitive** — an upstream quota/usage-limit rejection or a confirmed pre-dispatch transport failure, with zero response events and nothing downstream-visible. Upstream provably accepted nothing, so dispatching elsewhere is a first attempt. No claim consumed; a failed rebuild rolls back.

**② ambiguous** — the frame left, the connection died before any event. The turn may have run.

② is an at-least-once semantic, and this **partially reverses `aae61f6f3` (#2336)**, whose proposal states there is "no replacement, by design". Two things from #2336 are deliberately *not* reversed: the four-valued `..._AMBIGUOUS_CONTINUATION_RECOVERY_MODE` setting stays deleted (this is one behaviour, no operator knob, `[settings_fields].max` unmoved), and the server-owned SSE recovery loop with its keepalive stream and six-attempt cap stays deleted.

## What bounds ②

The residual exposure is **one duplicated generation per operation for its whole retention** — duplicate model tokens and duplicate upstream-hosted tool execution. It is **not** duplicate client-side tool execution, for two independent reasons: relocation is refused once any output is downstream-visible, and the relocated dispatch carries the origin operation's **side-effect replay-dedupe identity**, so a repeated side-effecting call is suppressed with the dedicated terminal failure rather than executed. That contract already exists in this spec; #2336 did not account for it.

Four fences, all on existing primitives:

1. the atomic one-shot claim, bounded at one dispatch per operation across every replica and reconnect;
2. **zero spooled events and no response id** — one event proves upstream ran the turn, which makes the outcome known rather than ambiguous;
3. a bounded ambiguity window, so a dispatch old enough to be mid-execution is not duplicated;
4. the armed side-effect dedupe above.

## Delta

**ADDED (4)** — *Relocation eligibility has a single decision point*; *Anchored turns relocate on a rebuilt durable transcript* (①); *Ambiguous eventless dispatches relocate once behind the durable fence* (②); *A relocation dispatch starts from a fresh spool*.

**MODIFIED (2)** — *Durable replay is limited to ambiguous transport outcomes* gains the unfenced ① lane while keeping invalid-request rejections non-relocatable; *Fenced one-shot recovery dispatch* states the concrete bound and preconditions and gains a live caller.

**REMOVED** — none.

## One decision point, not fourteen

Relocation eligibility is answered today by **five independent implementations** (`streaming/retry.py`, `http_bridge/streaming.py` ×2, `compact.py`, `websocket/helpers.py`) that do not agree on scope — which is why the durable-transcript case is missing from all of them rather than from one.

The downstream fork `aafqaq/codex-lb-enhanced` shipped this same feature by calling its projection directly from 14 sites across 7 modules, then spent a day emitting ~20 consecutive single-line corrections to those sites. This change puts the decision in one pure module with a decision table as the regression wall, and explicitly does **not** port the fork's lenient fallback projection: a body we cannot classify stays owner-bound.

## Scope boundary

`Compact requests recover from quota-caused previous-response owner loss` carries an explicit carve-out ("the durable prefix metadata that could prove it is deliberately not consulted here"). Reversing it is the natural follow-up, but that requirement has 13 scenarios and a `MODIFIED` block must reproduce all of them — it would double this change for a second concern. Compact keeps today's behaviour.

Pairs with #2373, which decides *whether* a request may walk the pool; this one widens *which* requests can.

## Validation

`openspec validate relocate-anchored-turns-across-accounts --strict` → valid. `openspec validate --specs` → 65 passed, 0 failed.


Soju06:
**The durable recovery-dispatch fence is restored — your blocker task is discharged by #2383.**

This change's `tasks.md` opens with: *"Blocker, resolve first: PR #2366 (`retire-recovery-dispatch-storage`) deletes `claim_unknown_operation_for_recovery`, the `recovery_dispatch_count` ORM mapping and the `expected_recovery_dispatch_count` CAS predicates as callerless. This change is the caller. Close #2366 or re-scope it before implementing the fenced lane."*

#2366 merged before that could happen. Per the owner's 2026-09-11 decision, #2383 partially reverts it and restores what the fenced lane (case ②) needs:

- `DurableBridgeRepository.claim_unknown_operation_for_recovery` and its `max_recovery_dispatches` bound — restored verbatim from `53a198d33^`, so the serialized `FOR UPDATE` over the session row and the `unknown` operation row, the spool deletion and the `submitted` reset are the same code, not a re-derivation.
- `mark_operation_unknown(restore_recovery_dispatch_claim=True)` and both refund branches — the post-claim/pre-frame rollback your task list calls for.
- The coordinator entry points for both.
- The `expected_recovery_dispatch_count` CAS surface on `append_terminal_operation_event`, `append_terminal_operation_chunk`, `_lock_operation_for_chunk_append` and `settle_terminal_append_failure`, their coordinator pass-throughs and the two `HttpBridgeOperationEventBatcher` methods, plus `DurableBridgeOperationSnapshot.recovery_dispatch_count`.
- The `HttpBridgeOperationRecord.recovery_dispatch_count` ORM mapping. The physical column was never dropped, so **there is no migration for you to write** — neither a re-add nor the "two migrations" your Impact section warned about.
- The repository tests #2366 deleted, including `test_unknown_operation_recovery_claim_is_atomic_and_single_use` — the concurrent-reconnect single-claim test your `test_proxy_http_bridge.py` task asks to restore is already back at the repository level.

**Spec compatibility, checked and stated in #2383's body.** #2383 brings the requirement back as an **ADDED** block whose text is the pre-#2366 text verbatim, so your **MODIFIED** of *Fenced one-shot recovery dispatch* still has the heading it modifies and still applies on top without a rebase. Diffed character-for-character, your block adds exactly four things and changes nothing else: the "at most one dispatch per operation for the whole of that operation's retention" bound, the `together with the spool clear required by "A relocation dispatch starts from a fresh spool"` clause, the refusal paragraph pointing at *Ambiguous eventless dispatches relocate once behind the durable fence*, and the *A restored claim is reusable exactly once* scenario.

#2383 also narrows the still-pending `retire-recovery-dispatch-storage` change folder so its unapplied `REMOVED` block cannot delete the requirement out from under you at archive time — that was a live hazard, because `restore-…` sorts before `retire-…` and a bulk archive would have applied them in the wrong order. Verified by simulating `openspec archive --yes` in both orders on a copy of current `main`'s `openspec/`: the requirement survives either way, 247 requirement headings before and after.

**Two things to know when you implement.**

1. The restored `expected_recovery_dispatch_count` is **opt-in**: `int | None = None` on every signature, predicate applied only when a value is supplied. It is not the pre-#2366 `int = 0`. The first codex round on #2383 caught why — with the request-state plumbing deliberately left deleted, a literal `0` default made terminal append and fallback settlement reject any operation carried across an upgrade with `recovery_dispatch_count > 0`, leaving it `acknowledged` with an incomplete spool. So when your claim site pins the generation it observed, **pass it explicitly**; passing nothing means "no generation fence", not "generation 0".

2. `HTTPRequestState.operation_attempt_generation` and its hops (the `request_submit` seed, the two `streaming.py` carry-overs, the two `upstream_events.py` call-site arguments) stay **deleted** — they only ever moved a constant `0`. Your task list already plans to wire the expectation from the claim site, which is the right place for it; you will be adding that propagation yourself rather than inheriting it.

#2383 is open and not merged; it needs the owner. Nothing here changes runtime behaviour — with no caller, the claim still never advances the counter and no restored predicate is added to any statement.

Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Soju06:
Reviewed the delta against `origin/main` (ca243dec4). The gap is real: `DurableBridgeRepository.get_replayable_transcript` (`app/modules/proxy/durable_bridge_repository.py:2469`) has no production caller — only the coordinator wrapper at `durable_bridge_coordinator.py:715` and tests — and a delta-only anchored continuation has no recovery path today. Five things need fixing in the spec before implementation starts; three of them would put the implementation in direct conflict with requirements that are on `main` right now.

**1. The definitive class does not cover the cohort the proposal measures.** `spec.md:37` admits "an upstream quota or usage-limit rejection, or a confirmed pre-dispatch transport failure". `previous_response_owner_unavailable` is neither — it is raised by the *selector*, before any upstream call: `_service/http_bridge/mixin.py:1757` (`selection.error_code in (CONTINUITY_OWNER_UNAVAILABLE, "hard_affinity_saturated")`), `:1767` (selector returned a different account), `:2055` (required owner already excluded). An implementation can satisfy this requirement in full and leave all 1,883 measured turns exactly as they are. Name the trigger the way the existing requirement already does — "a typed pre-visible `continuity_owner_unavailable` account-selection result" (`openspec/specs/responses-api-compat/spec.md:10450`) — and state explicitly that `hard_affinity_saturated` does **not** qualify: at `mixin.py:1756` it maps to the same 502 while the owner is alive and merely saturated, and relocating those turns moves traffic off a healthy owner and throws away its prompt-cache locality.

**2. The ② lane contradicts a requirement that is not in the MODIFIED list.** `openspec/specs/responses-api-compat/spec.md:2704` ("Codex WebSocket stale-anchor failures remain recoverable by a full-context retry") carries these as normative clauses: "an eventless transport failure without an explicit stale-anchor rejection MUST NOT by itself authorize cross-account replay" (`:2753`), "an UNKNOWN recovery journal on an inactive durable owner MUST fail closed without being claimed for anchor removal or account migration" (`:2755`), "an operation-journal recovery after an ambiguous transport failure MUST NOT remove the anchor or migrate accounts" (`:2776`), and "generic eventless transport retry MUST NOT convert an anchored safe-fresh request into an unanchored replay without explicit stale-anchor rejection". "Ambiguous eventless dispatches relocate once behind the durable fence" (`spec.md:166`) mandates precisely that, on precisely that trigger. Reproduce that requirement under `## MODIFIED Requirements` (`spec.md:218`) and reconcile those four clauses. It is 8 scenarios and ~40 bullets and a MODIFIED block has to carry all of them — that cost is real, but shipping the delta without it leaves two contradictory MUSTs in the same capability at archive time. The two requirements you did modify are reproduced correctly (checked clause-by-clause against `main:5885` and `main:6667`), so this is the only structural gap of that kind.

**3. ①'s "confirmed pre-dispatch transport failure" contradicts same-account recovery.** `openspec/specs/responses-api-compat/spec.md:2884` ("Pre-dispatch Responses requests recover from local network transitions"): "the proxy MUST retry on the same account with bounded backoff until the attempt succeeds or the existing request budget expires … Recovery MUST NOT move account-owned continuation or file state to another account." Either drop the pre-dispatch clause from the definitive class (leaving owner-unavailable plus quota, which is what the measurement is about anyway), or add that requirement to MODIFIED with an explicit ordering: same-account recovery first, relocation only after its budget is exhausted.

**4. "A relocation dispatch starts from a fresh spool" has an unreachable scenario.** `spec.md:211` requires a partial non-terminal spool *and* the claim consumed in the same atomic step. But the fenced lane requires "zero spooled events" (`spec.md:168`) and its own scenario at `spec.md:186` says a spooled event declines relocation *without consuming the claim*. No implementation satisfies both. State the spool clear as an invariant of the claim (the `event_spool_complete` / `event_bytes` reset that `claim_unknown_operation_for_recovery` already performs at `durable_bridge_repository.py:2212`) rather than as behaviour over an operation the preconditions have already excluded.

**5. The rebuild is unscoped across API keys, and the spec is silent about it.** `get_replayable_transcript` walks `get_operation_by_response_id` (`durable_bridge_repository.py:2460`), which selects `HttpBridgeOperationRecord` on `response_id` alone — no `api_key_scope` predicate, unlike every alias/session lookup in the same file (`:995`, `:1013`, `:1037`) and unlike `durable_bridge_operation_fingerprint`, which is keyed by scope (`:134`). `tasks.md:79` makes `materialize_relocation` the only caller and passes only the anchor. A client that supplies another key's `previous_response_id` and then trips any relocation trigger gets that key's stored request bodies and terminal outputs rebuilt into its own dispatch, and the model's answer back. Add a normative clause: the rebuild MUST be scoped to the caller's API key at every chain hop, and a hop that leaves the scope MUST be reported as an absent transcript.

**Re-measure before implementing.** The 7-day table is from beta.7 and predates #2384/#2390/#2397, which attack the same error code. #2390's own body counts 1,480 `previous_response_owner_unavailable` rows in 36h, of which **1,361 belong to one `reauth_required` account** its sweep now retires; #2397 then retires an owner that cannot return inside the request's budget and re-sends the turn anchor-free — but deliberately only for a *proxy-injected* anchor, because "a client that supplied its own `previous_response_id` named upstream state that lived on that account". What survives is exactly the client-supplied-anchor shape this change targets, which is the right residual — at a small fraction of 269/day. #2423 moves that same cohort's terminal again. Re-run the split on beta.9/beta.10 and restate ① and ② separately: if ① is now ~10/day, the entire justification rests on ②, the at-least-once half, and that deserves re-approval on the current number rather than the pre-sweep one.

Minor: the PR body still opens with "⚠️ Blocked on #2366" while `tasks.md` correctly records it as discharged — `claim_unknown_operation_for_recovery` is back at `durable_bridge_repository.py:2212` and `openspec/changes/restore-recovery-dispatch-claim/` is on main. Update the body so a reader does not stop at the blocker.

Not blocking, for the record: CodeRabbit's `sticky-session-operations` conflict is already precedented — `Verified full resend can recover from selection-time owner loss` moves an anchored request across accounts today without a MODIFIED there, so "MUST NOT silently reroute to an account that cannot preserve continuity" is already read as a statement about preserved continuity, not about account identity.

Komzpa:
@codex review

Komzpa:
@codex review

PR #2377, opened 2026-09-11
fix(dashboard): calm the telemetry consent dialog

Soju06:
## Summary

What this instance actually transmits is unusually tame. The account count leaves as a **bucket string** (`2-5`), so do the cost, database-size and API-key counts; user-agents and model names collapse to `other` unless allowlisted; upstream errors are reduced to at most five allowlisted *codes*; and the schema is `extra="forbid"`, so a stray field is a validation error rather than a leak. No hostname, no IP, no account identity, no prompt or response content, no keys.

The disclosure did not read that way. On first dashboard entry the operator got a **blocking modal with a backdrop** — mounted globally in `AppLayout`, so it covered whatever page they landed on — containing a description, a categories paragraph, an opt-out-notification paragraph, a label, and then a scrolling dump of the raw JSON envelope, before reaching the two buttons. Dismissing with Escape or the backdrop persisted nothing, so it **returned on the next dashboard entry** and repeated the whole wall.

That is a surface that reads like "we are about to take your data" for a payload that is mostly booleans and buckets.

## Type of change

- [ ] **Breaking change**

## Screenshots

> Before/after captures are attached in the first review comment (2880x1800, dashboard entry, light theme).

**Before** — 3 paragraphs + label + a scrolling JSON wall; the dialog fills most of the viewport and the decision sits below the fold of the payload.

**After** — title, one paragraph, a collapsed `View what is sent` disclosure beside the docs link, then the two actions. Roughly a third of the previous height.

## What changed

- The dialog body is **one short paragraph**: `Anonymous usage statistics — version, deployment shape, and aggregated counts. No accounts, prompts, API keys, or IP addresses. On by default; change it any time in Settings.` The separate `categories` paragraph is folded into it; the opt-out-notification paragraph moves out of the dialog and stays on the settings card, where the toggle that triggers it lives.
- The exact envelope **stays in the dialog and stays exact**, but collapsed behind one disclosure control rather than rendered as a wall. It no longer stands between the operator and the decision.
- **Dismissal is remembered per browser** (`localStorage`), so the question is asked once instead of on every dashboard entry. A dismissed browser also skips the consent query entirely, so it stops paying for the snapshot aggregation.
- Copy rewritten short and flat across `en`, `ko` and `zh-CN` together.

## What deliberately does not change

- **Enable and disable stay equally prominent, with equal click cost.** The existing requirement forbids favouring one, and making opt-out quieter would damage the trust this change exists to repair. A new test asserts the two buttons carry identical classes.
- **Consent semantics are untouched.** `undecided` still resolves active, `CODEX_LB_TELEMETRY_ENABLED` still applies while undecided, a persisted decision still wins, and dismissal persists **no** decision — the test asserts no `PUT` fires.
- **No field is removed or hidden.** The envelope shown is still the exact one the sender would transmit, from the same shared constructor, and the settings card still exposes it on demand for any consent state.

## Known gaps, deliberately not in this PR

- `resolve_consent` treats `undecided` as active and the scheduler sends its first snapshot at startup, so a snapshot can leave before the operator has seen this dialog. Calmer copy makes that mismatch *more* visible, not less. Changing it is a consent-semantics decision rather than a copy change; the dialog continues to say plainly that telemetry is on by default.
- `docs/telemetry.md` tells the reader to "assume transmitted snapshots remain stored until a published retention policy or explicit deletion" — the most alarming sentence in the documentation. Replacing it needs the collector's actual retention period, which is not recorded in this repository.

## OpenSpec

`openspec/changes/calm-telemetry-consent-dialog/` — one MODIFIED requirement (`One-time consent dialog with exact payload preview`), no ADDED, no REMOVED. It now requires the summary paragraph, the envelope to be reachable without being expanded by default or blocking the actions, equal prominence for both actions, and that dismissal neither changes consent state nor re-presents the dialog.

## Simplicity gates

Frontend only. No API, schema, settings or payload change; `[settings_fields]` untouched. No README section, no `.env.example` entry, no new dashboard nav item.

## Validation

- `bunx vitest run src/features/settings` → **31 files, 333 tests passed**
- `bun run lint` → clean; `bun run typecheck` → clean
- `openspec validate calm-telemetry-consent-dialog --strict` → valid


Soju06:
The diagnosis behind this is right, and the shape of the fix is right. The payload really is buckets and booleans, and the old dialog really did read like a data grab: three stacked paragraphs plus a label plus a scrolling JSON dump before the operator reaches either button. `onOpenChange={(nextOpen) => setDismissed(!nextOpen)}` on main only touched component state, so every fresh page load replayed the whole wall — that is a genuine bug, not a taste complaint. Keeping the envelope exact and in the dialog rather than moving it to a docs link, keeping both decisions on the same variant (and adding a test that asserts identical `className`), persisting no consent on a dismissal, and moving all three locales together are all the right calls. The proposal is also honest about the `undecided`-resolves-active gap instead of papering over it.

Three things block it.

**1. The dialog no longer discloses that "Disable telemetry" sends a network request.** `telemetry-consent-dialog.tsx` drops the `settings.telemetry.optOutNotice` paragraph, and `telemetry-consent-dialog.test.tsx:44` pins the removal with `queryByText(...optOutNotice).not.toBeInTheDocument()`. But that button still fires an outbound, instance-identified beacon: `app/modules/telemetry/consent.py:57` resolves `undecided` to `active=True`, so `app/modules/telemetry/api.py:53-77` sees `previous.active and not consent.active` and spawns `TelemetrySender().send_opt_out(identity, ...)`, which POSTs `/v1/optout` carrying `identity.instance_id` (`sender.py:62-89`).

Failure scenario: fresh install, consent `undecided`. Operator lands on `/dashboard`, reads a paragraph that says "No accounts, prompts, API keys, or IP addresses", clicks **Disable telemetry** precisely to stop transmissions — and the instance immediately transmits an instance-identified opt-out. main disclosed that before the click, in the dialog, one line above the button. The proposal's rationale is that the notice "stays where it is actionable, on the settings card" (`telemetry-settings.tsx:62`), but a first-entry operator who opts out at the modal never goes to `/settings`; for that operator the notice has moved to a page they will never see. This is not a spec breach — `openspec/specs/telemetry/spec.md` requires the notification to exist (line 251) but never requires it be disclosed in the dialog; it was task 3.1 of `openspec/changes/archive/2026-09-09-add-telemetry-optout-signal`. It is still a disclosure regression in a change whose entire purpose is trust repair, and it is the one deletion here that removes information rather than compressing it.

**2. The dismissal flag suppresses the dialog for installs that were never asked.** `CONSENT_DIALOG_DISMISSED_STORAGE_KEY = "codex-lb.telemetry-consent-dismissed"` (`telemetry-consent-dialog.tsx:26`) is per-origin; consent is per-installation. `useState(readDismissed)` at `:50` plus `enabled: canWrite && !dismissed` at `:56` means a browser that once pressed Escape does not even issue `GET /api/settings/telemetry`, so the client can never discover that consent went back to `undecided`.

Failure scenario: operator presses Escape once on `http://host:8000`. The DB is later reset, or the host is re-provisioned, or the deployment is upgraded — exactly the empty-DB first run the browser smoke models. Consent is `undecided`, `resolve_consent` treats that as active, and the scheduler transmits at startup. The dialog never appears, and no request is made that could reveal it should. That installation transmits without anyone having been asked on it. The delta's "in the same browser" wording describes the implementation, but `openspec/specs/telemetry/spec.md:45-49` — "Upgrade treats existing users as undecided ... the migrated consent state is `undecided` and the one-time consent dialog is shown on next dashboard entry" — is a requirement this delta does **not** modify, and per-browser suppression contradicts it directly.

**3. CI is red and it is this PR's doing; the fix is one line.** `frontend/browser-smoke/dashboard.spec.ts:155` still does `await expect(consentDialog.getByText('"instance_id"').first()).toBeVisible();`. `frontend/src/components/ui/collapsible.tsx` wraps Radix `CollapsibleContent` with no `forceMount`, so the envelope is unmounted while closed — your own vitest case asserts exactly that. Job 103163980789 shows 3 failed / 5 passed: `:155` "element(s) not found", then `dashboard.spec.ts:85` twice ("Settings heading not visible"). The latter two are a cascade, not separate bugs — the smoke run shares one backend and one DB, the test at `:96` is the only one that persists a decision, it now dies before `acceptTelemetryConsent`, so consent stays `undecided` and the modal's `aria-hidden` on the page content hides the `Settings` heading from role queries.

On why CI let 1 and 2 through: the unit and integration suites were updated in the same commit to encode the new behaviour, so the opt-out deletion is asserted rather than caught; `openspec validate --strict` only checks the delta the change declares, and this delta MODIFIES only "One-time consent dialog with exact payload preview", so the conflicting unmodified scenario at `spec.md:45-49` is never cross-checked; and the one test that drives the real dialog against a real empty-DB backend lives in `frontend/browser-smoke/`, outside every path this PR touches, so it failed instead of covering.

To make this mergeable:

1. `frontend/browser-smoke/dashboard.spec.ts:155` — insert `await consentDialog.getByRole("button", { name: "View what is sent" }).click();` before the assertion. That should take all three failures green.
2. Restore the opt-out-send disclosure inside the dialog: either fold a clause into `settings.telemetry.consentDialog.description` across `en`/`ko`/`zh-CN`, or keep `optOutNotice` adjacent to the Disable button. Revert the `not.toBeInTheDocument()` assertion at `telemetry-consent-dialog.test.tsx:44` to a positive one.
3. Pick one for the dismissal, and say which in the proposal: (a) drop the persistence from this PR and ship collapse + copy only, removing the dismissal clause from the openspec delta as well; (b) key the flag to the envelope's `instance_id` so a reset DB or a new install asks again, and keep issuing the consent query so a changed instance identity is observable — i.e. stop gating `enabled:` on `dismissed`; or (c) if per-browser suppression is what you want, the delta has to also MODIFY the "Consent state and default activation" requirement whose upgrade scenario it contradicts, and that is an owner call rather than something the delta can leave implicit.
4. While in `dashboard.spec.ts`: move `await acceptTelemetryConsentIfShown(page);` above the `Settings` heading assertion in `openLongSettingsPage` (lines 83-86). With a modal up the heading is unreachable, so that helper can never run today — that ordering is what turned one failure into three.
5. Tick the boxes in `openspec/changes/calm-telemetry-consent-dialog/tasks.md`, and attach the before/after screenshots the body says are in the first review comment (there are no review comments on this PR yet).

Nit, not blocking: the docs link went from `text-primary` to `text-xs text-muted-foreground`. The two decisions keep equal prominence as required, but the single path to "learn what is collected and why" got quieter in a change arguing for transparency — `text-sm` would hold the line.

Sequencing: #2294 is still open and rewrites this same component, ships its own MODIFIED block for the same requirement, and already implements a server-side answer to "stop re-asking" (`telemetry_notice_version`, `acknowledge_notice()`/`notice_version()` in `app/modules/telemetry/consent.py`). Since a MODIFIED block replaces the whole requirement on archive, landing this and then #2294 would silently revert the calm-dialog clauses with no revert-shaped diff. It also deletes `consentDialog.categories` and `.payloadLabel` from all three locales while #2294's component still calls both — the locale lines conflict, the `t()` calls do not, so a mechanical resolution ships raw key strings in the modal. Whichever lands first, the other needs a hand-merge. If #2294 goes first, option (b)/(c) above collapses into simply using its server-side acknowledgement instead of a localStorage key.

PR #2384, opened 2026-09-11
fix(proxy): attribute refused cross-account continuity replays

Soju06:
## Why

`previous_response_owner_unavailable` is the user-visible face of #1707: an existing Codex thread whose upstream owner account became unroutable fails with 502 on every resume while healthy accounts sit idle in the pool.

The proxy already contains the cure. `switch_to_account_neutral_replay` projects reasoning out of the history and re-sends plaintext context on a fresh account. Its gate, `durable_full_resend_allows_account_neutral_replay`, is a conjunction of six independent proofs — and it returns a bare `bool`.

That makes the failure unattributable. A 36-hour window on a production deployment (`1.25.0-beta.7`, 24 accounts) recorded:

| Observation | Value |
|---|---|
| `previous_response_owner_unavailable` request-log rows | **1,480** / 36h |
| … by owner account status | `reauth_required` 1,361; `paused` 22; `rate_limited` 3; owner now `active` again ~94 |
| … by `transport` | `websocket` 1,491 / 48h; `http` **0** |
| `owner_unavailable_fresh_resend` bridge events | **0** / 24h |
| `dead_owner_fresh_resend` bridge events | **0** / 24h |
| Bridge 502s present in `request_logs` over the last 6h | **0** of 14,555 rows |

Two readings matter:

1. **The recovery path has never fired in production.** Zero `owner_unavailable_fresh_resend` events against 1,480 failures means the refusal is systematic, not incidental — and nothing on the failing request says which of the six proofs refused it. The 2026-09-08 status note on #1707 asks reporters for exactly this attribution; the current logs cannot supply it.

2. **The bridge surface's failures are invisible where operators look.** Every logged row carries `transport="websocket"` — the direct-WebSocket surface logs the code itself. Not one carries `transport="http"`: the HTTP session bridge raises the identical envelope from session creation, before any request-log write, and `app/modules/proxy/_service/http_bridge/` contains no request-log write at all (`_write_request_log` appears only as an unused protocol declaration). The dashboard shows nothing, so a bridge incident is only reachable by reading a rotating container log (3×50MB, ~8h retention) — which is how the 2026-09-04 recurrence of this same class had to be reconstructed.

## What changes

- **Typed rejection reason.** `durable_full_resend_allows_account_neutral_replay` becomes a thin `is None` wrapper over a new `classify_account_neutral_replay_rejection()` returning `None` when a cross-account replay is allowed, else one of a closed set: `file_bound`, `no_durable_lookup`, `payload_not_full_resend`, `anchor_metadata_missing`, `prefix_fingerprint_mismatch`, `input_not_itemized`, `missing_prior_output`, `account_scoped_input`.

  The stored-anchor half moves to a new module-level, pure `_durable_full_resend_anchor_rejection(payload, lookup, *, payload_looks_like_full_resend)` — the seam the unit tests drive — so the three causes `classify_durable_full_resend` collapses today (`payload_not_full_resend`, `anchor_metadata_missing`, `prefix_fingerprint_mismatch`) stop being indistinguishable. **The predicate's truth value is unchanged at every existing call site**; evaluation order is preserved from the conjunction, so the reported reason is the first proof that refused.

- **One emission point.** `owner_unavailable_allows_account_neutral_replay` emits the new bridge event `owner_unavailable_replay_rejected` (WARNING, `detail=reason=<reason>`). It already gates all three recovery attempt sites (session creation, owner forward, capacity retry), so "exactly one reason per refusal" is structural rather than a convention each future raise site must remember. The `required_continuity_owner_missing` exit has no exception to classify and calls the classifier directly.

- **Metric.** `codex_lb_continuity_replay_rejected_total{surface, reason}` with the usual `None` fallback, recorded via `_record_continuity_replay_rejected` beside the existing `_record_continuity_owner_resolution` / `_record_continuity_fail_closed`.

- **Request-log parity for the bridge surface.** `_stream_http_bridge_or_retry` writes one `_write_stream_preflight_error` row before re-raising a pre-submit `ProxyResponseError` coded `previous_response_owner_unavailable` or `continuity_owner_conflict`, mirroring what `retry.py` already does at its own owner-unavailable exit. Both sites resolve the same `ensure_request_id()` contextvar, so the container log line and the row are joinable on that id.

  The write sits **inside the propagating branch only**: when the bridge instead hands the turn to the raw-HTTP upstream, that path settles its own outcome, and an error row for a turn that then succeeded would corrupt the very error rate this row exists to expose.

## Why observability first

The follow-up work in #1707 is a behavior change: retiring an unroutable continuity owner so the thread rebinds. Which mechanism to build depends on which proof is actually refusing the existing recovery in production — `no_durable_lookup` (owner pinned from the request-log index with no anchor context) and `prefix_fingerprint_mismatch` (client body diverged from the stored turn) point at different fixes. Landing the attribution first makes that choice evidence-based and gives the later PRs a before/after signal that is not a rotating container log.

## Scope and safety

No new `CODEX_LB_*` setting, no migration, no `.env.example` change, no README growth. No routing, selection, or failover decision changes — the classifier is a pure refactor of an existing conjunction, and the counter and log line are observational.

None of the four ceiling-guarded proxy files is touched: `service.py` (2600/2600), `load_balancer.py` (3000/3021), `_service/http_bridge/mixin.py` (2435/2436), `_service/streaming/mixin.py` (1097/1100). The recorder deliberately skips a `service.py` re-export — `_service_global` falls back to the module global, and that file sits exactly at its ratchet.

Two closed-set notes:
- `input_not_itemized` is a **guard, not an outcome**: a non-list body can never satisfy the stored-prefix proof, so the prefix check always refuses first. `test_string_input_refuses_on_the_prefix_proof` pins that ordering so a future caller admitting a non-list body past the prefix proof shows up as a test change rather than a silent mislabel.
- Both counter labels are closed (`surface` is `http_bridge`; `reason` is the eight-value literal), asserted in `test_closed_reason_set_matches_the_literal` and `tests/unit/test_metrics.py`.

## Test plan

Local, all green:

| Gate | Result |
|---|---|
| `make lint` (incl. `check_proxy_architecture.py`, `check_cancellation_safety.py`, `check_proxy_timing_seams.py`, `check_settings_tiers.py`) | pass |
| `make typecheck` (`ty`) | pass |
| `make test-unit` | **9,984 passed**, 6 skipped, 1 xfailed |
| `make test-integration-bridge` | **341 passed** |
| `make test-integration-core-1` | **889 passed**, 332 skipped |
| `make test-integration-core-2` | **873 passed**, 31 skipped |
| `make test-integration-core-3` | **774 passed**, 34 skipped |
| `uv run pytest tests/e2e` | **27 passed**, 1 skipped |
| `openspec validate trace-continuity-owner-replay-rejection --strict` | valid |
| `codex review --base origin/main` | no findings |

New coverage:
- `tests/unit/test_http_bridge_replay_rejection.py` (new) — the anchor-proof reasons via the pure module-level seam, the allowed path reporting no reason, the three previously-collapsed causes reported distinctly, `input_not_itemized` pinned as unreachable behind the prefix proof, the counter's closed label set, the recorder's no-prometheus no-op, and the event's WARNING level asserted through `caplog`.
- `tests/integration/test_http_responses_bridge.py` — a bridged resume whose owner is not selectable returns the unchanged 502 envelope, logs exactly one `owner_unavailable_replay_rejected` with a reason from the closed set, and leaves exactly one `request_logs` row with `error_code="previous_response_owner_unavailable"`. A sibling case proves a capacity-coded pre-submit failure still writes no row.
- `tests/unit/test_metrics.py` — name/label assertion beside the sibling continuity counters.

## OpenSpec

`openspec/changes/trace-continuity-owner-replay-rejection/` — proposal, context, tasks, and a `proxy-runtime-observability` delta with two ADDED requirements (rejection attribution; bridge continuity-owner failures reach the request log).

Partial fix for #1707 — it does not by itself let any thread recover.

> Shards 2 and 3 additionally reported sqlite `unable to open database file` (444 occurrences) on files unrelated to this change — `test_v1_reset_credit`, `test_subscription_overflow_*`, `test_usage_*`, `test_proxy_api_extended`, `test_proxy_images`, `test_proxy_realtime_live`, `test_proxy_sticky_sessions`. The host root partition was at 96% with several test suites running concurrently. All of those files pass in isolation on this branch (103 + 247 tests).

### DIFF
--- a/app/core/metrics/prometheus.py
+++ b/app/core/metrics/prometheus.py
@@ -224,6 +224,12 @@ if PROMETHEUS_AVAILABLE:
         ["surface", "source", "outcome"],
         registry=REGISTRY,
     )
+    continuity_replay_rejected_total = Counter(
+        "codex_lb_continuity_replay_rejected_total",
+        "Total cross-account continuity replays refused by surface and refusing proof",
+        ["surface", "reason"],
+        registry=REGISTRY,
+    )
     continuity_fail_closed_total = Counter(
         "codex_lb_continuity_fail_closed_total",
         "Total continuity fail-closed or masked retryable outcomes by surface and reason",
@@ -508,6 +514,7 @@ else:
     bridge_forward_latency_seconds: HistogramLike | None = None
     bridge_public_contract_error_total: CounterLike | None = None
     continuity_owner_resolution_total: CounterLike | None = None
+    continuity_replay_rejected_total: CounterLike | None = None
     continuity_fail_closed_total: CounterLike | None = None
     account_lease_acquired_total: CounterLike | None = None
     account_lease_released_total: CounterLike | None = None
@@ -587,6 +594,7 @@ __all__ = [
     "event_loop_lag_seconds",
     "event_loop_lag_warnings_total",
     "continuity_owner_resolution_total",
+    "continuity_replay_rejected_total",
     "http_bridge_prewarm_total",
     "http_bridge_retry_circuit_total",
     "http_bridge_spool_cleanup_backlog_likely",
--- a/app/modules/proxy/_service/http_bridge/helpers.py
+++ b/app/modules/proxy/_service/http_bridge/helpers.py
@@ -121,6 +121,9 @@ from app.modules.proxy._service.observability import (
     _record_continuity_owner_resolution as _record_continuity_owner_resolution,
 )
 from app.modules.proxy._service.observability import (
+    _record_continuity_replay_rejected as _record_continuity_replay_rejected,
+)
+from app.modules.proxy._service.observability import (
     _summarize_input as _summarize_input,
 )
 from app.modules.proxy._service.observability import (
@@ -3837,6 +3840,7 @@ def _log_http_bridge_event(
         "capacity_exhausted_active_sessions",
         "owner_mismatch",
         "owner_forward_fail",
+        "owner_unavailable_replay_rejected",
         "missing_response_created_timeout",
         "prompt_cache_locality_miss",
         "reallocation_orphan",
--- a/app/modules/proxy/_service/http_bridge/streaming.py
+++ b/app/modules/proxy/_service/http_bridge/streaming.py
@@ -6,7 +6,7 @@ import json
 import logging
 import math
 from collections.abc import AsyncGenerator, Callable
-from typing import Any, AsyncIterator, Mapping, TypeVar, cast
+from typing import Any, AsyncIterator, Literal, Mapping, TypeVar, cast, get_args
 from uuid import uuid4
 
 import anyio
@@ -116,6 +116,7 @@ from app.modules.proxy._service.http_bridge.helpers import (
     _proxy_admission_wait_timeout_seconds,
     _record_bridge_reattach,
     _record_continuity_fail_closed,
+    _record_continuity_replay_rejected,
     _release_http_bridge_denied_anchor_fences,
     _release_http_bridge_unanchored_handoff,
     _release_http_bridge_unanchored_handoffs_for_request,
@@ -197,6 +198,7 @@ from app.modules.proxy._service.support import (
     _HTTPBridgeSession,
     _HTTPBridgeSessionKey,
     _is_local_account_cap_code,
+    _request_log_client_fields,
     _signal_propagated_capacity_startup_ready,
     _signal_propagated_capacity_startup_wait,
     _signal_propagated_responses_service_cleanup_ready,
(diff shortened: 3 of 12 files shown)


Soju06:
Addressed the CodeRabbit nitpick in a502ca0f5.

`capacity_exhausted_active_sessions` is in `LOCAL_OVERLOAD_CODES` (`app/core/resilience/overload.py:21`), so `selection_failure_response` gives it the stable 429 `rate_limit_error` contract. The assertion now pins 429 rather than "not 200", with a comment recording why — a broad check would also have passed if the request had failed for an unrelated reason, which is exactly the kind of false green a no-row negative test must not have.

Komzpa:
@codex review

PR #2386, opened 2026-09-11
fix(proxy): narrow the input-image upstream transport pin

### LINKED ISSUE #2363: bug: input_image requests are forced off the HTTP bridge onto direct HTTP upstream, where ~52% end as server_is_overloaded (vs 2.7% bridged)
## Summary

Any Responses request whose payload contains an `input_image` is excluded from the HTTP session bridge and pinned to the direct HTTP upstream, regardless of the dashboard `upstream_stream_transport` value. While upstream is rejecting fresh admissions with `server_is_overloaded` (the #2166 pattern), that path fails far more often than the bridge: on our deployment **~52% of image-bearing requests ended as errors (nearly all `server_is_overloaded`) vs 2.7% for bridged requests in the same window**.

Codex keeps earlier images in the conversation input, so a single pasted screenshot moves **every later turn of that thread** onto the degraded path. Users see `server_is_overloaded: Our servers are currently overloaded. Please try again later.` repeatedly in that thread only; starting a new thread without the image recovers immediately.

## Where

`app/modules/proxy/_service/http_bridge/streaming.py` (v1.25.0-beta.7, around L990):

```python
image_request = _responses_request_contains_input_image(payload)
image_generation_request = _responses_request_uses_image_generation(payload)
force_upstream_stream_transport = "http" if image_request else None
if runtime_config.enabled and (image_request or image_generation_request):
    record_http_bridge_routing(stage="bypass", reason="image")
```

`force_upstream_stream_transport="http"` is forwarded as `upstream_stream_transport_override`, so it wins over `upstream_stream_transport=auto` and `=websocket` alike; there is no setting to route image requests through the bridge.

Log signature per request:

```
stream_responses bypassing http bridge for image-capable request input_image=True image_generation=False request_id=...
http_downstream_transport_decision policy=explicit override_applied=False sticky=True upstream_stream_transport=http request_id=...
```

## Evidence (v1.25.0-beta.7, 2026-09-11 01:21–03:01 UTC, 100 min)

| Path | Requests | Final error rate (`request_logs`) |
|---|---|---|
| Bridge (upstream WebSocket) | 7,402 | **2.7%** |
| Image bypass → direct HTTP | 249 matched\* | **52.2%** (124 of 130 errors are `server_is_overloaded`) |

\* Proxy log request ids differ from `request_logs.request_id`, so bypassed requests (694 in the log) were matched to `request_logs` rows by api key + timestamp (±2 s) + `upstream_transport='http'`; 249 matched.

Errors on the bypassed requests are spread across all ~15 accounts in the pool (no single-account concentration). The errors are consistent with the output-free `response.created → response.failed(server_is_overloaded)` shape that the direct SSE path does not fail over (the situation #2351 addresses), though we have not captured raw streams for these specific requests.

Why we think the path (not the image payload) is the main factor: before #2247 (v1.25.0-beta.5), native Codex HTTP traffic *without* images was also forced onto direct HTTP. For the same accounts in the same minutes, direct HTTP failed 95.5% vs 42.9% over the bridge, and pinning `upstream_stream_transport=websocket` (which skipped the native-Codex downgrade) took overall errors from 40.9% to ~2% within five minutes. #2247 fixed that for non-image traffic; image-bearing traffic is now the only Codex traffic still forced onto that path.

## Questions / possible fixes

1. Is the image bypass still required, i.e. does the upstream WebSocket path reject or mishandle `input_image`? If the reason is historical (stream slot accounting, payload size), the separate `payload_size` bypass right above it already covers oversized payloads.
2. If it must stay, letting output-free overloads on this path fail over to a sibling account (#2351) would remove most of the user-visible failures.

## Environment

- codex-lb `1.25.0-beta.7` (official `ghcr.io/soju06/codex-lb` image, Docker Compose, PostgreSQL)
- Dashboard: `upstream_stream_transport=auto`, `http_downstream_transport_policy=smart`, HTTP responses session bridge enabled
- ~17 ChatGPT Pro accounts; clients: Codex Desktop, codex-tui, codex_vscode (HTTP and WebSocket downstream)

Soju06:
## Summary

An `input_image` anywhere in a Responses payload pinned the request to the direct HTTP upstream. On the reporting deployment **~52% of image-bearing requests ended in error** (nearly all `server_is_overloaded`) against **2.7% for bridged requests in the same window** — and because Codex keeps earlier images in the conversation input, a single pasted screenshot moved *every later turn of that thread* onto the degraded path (#2363).

The pin was never needed for that. It came from bcd63c827 (#903), whose stated rationale is about **bridge pending slots**, not about the HTTP-vs-WebSocket upstream transport — the bridge bypass on the next line already delivers it. Inline `data:` images ride the upstream websocket unchanged: `_slim_response_create_payload_for_upstream` (the only websocket-specific mutation of image payloads) runs strictly above the size budget, and native downstream websocket clients already send inline images over the upstream websocket today with no image bypass at all.

## Type of change

- [ ] **Breaking change**

Linked issue: Fixes #2363

## OpenSpec

- [x] This PR touches a codex-faithful path (image pipeline, request shape) and preserves upstream-equivalent behavior

Change directory: `openspec/changes/narrow-input-image-upstream-transport-pin/`

## Changes

There are **two** pin sites and narrowing either one alone ships a no-op:

- `http_bridge/streaming.py` — `force_upstream_stream_transport = "http" if image_request else None` becomes the narrow predicate. The bridge bypass, its `record_http_bridge_routing(stage="bypass", reason="image")` counter and its log are left exactly as they were.
- `streaming/retry.py` — `input_image` was folded into `image_bypass` and passed as `has_image_generation_tool=`, so `_resolve_stream_transport` returned `"http"` **before** the visible `if` on the next line ever ran, inside a function where no log or counter records the decision. `has_image_generation_tool` now means exactly that, and `input_image` is pinned only by the shared predicate.
- New shared predicate `_input_image_request_requires_http_upstream` in `_service/response_create.py`, reached from both sites (via the service façade re-export and the `service_stubs` seam, matching how the existing image predicates are wired). One predicate, so the two sites cannot drift apart again — which is exactly how this reached production.

It keeps the pin for the two hazards that are real:

1. **Over the websocket frame budget** (`_ws_transport_payload_budget_bytes()`, 14 MiB), where `_prepare_websocket_response_create_payload` would replace historical inline images with `[codex-lb omitted historical inline image to fit upstream websocket budget]`. The 14 MiB frame budget is used rather than the 15 MiB `upstream_response_create_max_bytes`, because `_resolve_stream_transport` already diverts above 14 MiB — gating on 15 would add a second threshold that never fires.
2. **An external `http(s)` image URL still present** after `_inline_content_images` gave up on a failed fetch (non-https, SSRF-blocked host, non-200, over 8 MiB, or the 8 s timeout). The evidence that the upstream websocket rejects or hangs on one is a code comment at `http_bridge/request_submit.py:872-879` rather than a captured trace; that is recorded as such in the change context.

## Simplicity

- [x] New behaviour works with zero configuration
- New settings: none
- [x] README / `.env.example` / dashboard nav untouched

## Test plan

```
uv run ruff check app tests                       # All checks passed!
uv run ty check app                               # All checks passed!
uv run openspec validate narrow-input-image-upstream-transport-pin --strict

uv run pytest tests/unit/test_proxy_utils.py                                    # 1400 passed
uv run pytest tests/unit/test_proxy_http_bridge.py tests/unit/test_websocket_transport_fallback.py \
             tests/unit/test_http_bridge_safe_continuity.py tests/unit/test_http_continuation.py \
             tests/unit/test_proxy_api_responses_contract.py                    # 1205 passed
uv run pytest tests/integration/test_proxy_responses.py \
             tests/integration/test_http_promotion_accounting.py                # 108 passed
uv run pytest tests/integration/test_http_responses_bridge.py                   # 185 passed
```

**Fails-before-the-fix proof.** With the `app/` diff reverse-applied and the tests left in place, 7 of the 15 node ids fail, including the route-level `test_historical_image_does_not_pin_later_turns_to_http` — the reported production shape, where the *first* turn carried the screenshot and a later text-only turn is still pinned.

**Mutation proof on the residual clause.** `test_oversized_image_request_still_pins_http` originally passed even with the predicate's size clause disabled, because `_resolve_stream_transport`'s own 14 MiB gate produced `"http"` anyway — it was passing for a neighbouring reason. It now runs under an explicit `upstream_stream_transport = "websocket"`, which short-circuits that gate and leaves the predicate as the only thing that can produce `"http"`; re-running the same mutation afterwards makes it fail.

## Known bounds, recorded in the change context

- The residual pin still overrides an explicit operator `upstream_stream_transport = "websocket"`, as it does today. Removing that would turn an oversized image payload into a local 400 `payload_too_large` that operators do not get now.
- Site B gates its residual clauses behind `not explicit_transport`, so a request that never reaches the bridge routing decision (a `/v1/chat/completions` request whose bridge admission declined) resolves by ordinary precedence under an explicit websocket pin. The spec text is scoped to say so rather than claiming a precedence that does not hold everywhere.
- An `input_image` nested deeper than one `content` level — inside a `function_call_output` output array, say — is invisible to both `_count_external_image_urls` and `_inline_input_image_urls`. That blind spot is identical to the existing HTTP bridge guard's, so no deployment gains a hazard the bridge path did not already have. Deliberately not widened here.
- The 52% figure is not verifiable from this repository: `request_logs` carries no image marker. Post-deploy this is verifiable as a ratio shift on `codex_lb_upstream_transport_decisions_total` with `codex_lb_http_bridge_routing_total{reason="image"}` flat.

## Checklist

- [x] Added regression coverage at the externally failing product path.
- [x] Ran the relevant `make` / pytest subset locally.
- [x] `openspec validate --strict` passes.
- [x] Simplicity gates reviewed (PRINCIPLES.md P1-P6).

### DIFF
--- a/app/modules/proxy/_service/http_bridge/service_stubs.py
+++ b/app/modules/proxy/_service/http_bridge/service_stubs.py
@@ -223,6 +223,10 @@ def _responses_request_uses_image_generation(*args: Any, **kwargs: Any) -> Any:
     return _service_global("_responses_request_uses_image_generation")(*args, **kwargs)
 
 
+def _input_image_request_requires_http_upstream(*args: Any, **kwargs: Any) -> Any:
+    return _service_global("_input_image_request_requires_http_upstream")(*args, **kwargs)
+
+
 def _input_prefix_matches_stored_context(*args: Any, **kwargs: Any) -> Any:
     return _service_global("_input_prefix_matches_stored_context")(*args, **kwargs)
 
--- a/app/modules/proxy/_service/http_bridge/streaming.py
+++ b/app/modules/proxy/_service/http_bridge/streaming.py
@@ -141,6 +141,7 @@ from app.modules.proxy._service.http_bridge.service_stubs import (
     _header_value_case_insensitive,
     _http_bridge_startup_keepalive_grace_seconds,
     _inject_missing_interrupted_function_call_outputs,
+    _input_image_request_requires_http_upstream,
     _input_prefix_matches_stored_context,
     _is_previous_response_not_found_error,
     _maybe_log_proxy_request_payload,
@@ -963,7 +964,20 @@ class _HTTPBridgeStreamingMixin:
             runtime_config = dataclasses.replace(runtime_config, enabled=False)
         image_request = _responses_request_contains_input_image(payload)
         image_generation_request = _responses_request_uses_image_generation(payload)
-        force_upstream_stream_transport = "http" if image_request else None
+        # The bridge bypass below is about bridge pending slots; it must not also
+        # decide the upstream transport. An inline ``data:`` image rides the
+        # upstream websocket unchanged, so only the two websocket-specific
+        # hazards keep the pin (#2363). ``image_request`` gates the predicate
+        # because the predicate's own first check is that same walk of the whole
+        # input, and this runs on every request on the hot path.
+        force_upstream_stream_transport = (
+            "http"
+            if image_request
+            and _input_image_request_requires_http_upstream(
+                payload, payload_size_estimate_bytes=payload_size_estimate_bytes
+            )
+            else None
+        )
         if runtime_config.enabled and (image_request or image_generation_request):
             record_http_bridge_routing(stage="bypass", reason="image")
             logger.info(
--- a/app/modules/proxy/_service/response_create.py
+++ b/app/modules/proxy/_service/response_create.py
@@ -194,7 +194,11 @@ def _json_value_contains_external_input_image(value: JsonValue) -> bool:
     """
     if _input_part_is_image(value):
         image_url = value.get("image_url") if is_json_mapping(value) else None
-        return isinstance(image_url, str) and image_url.startswith(("http://", "https://"))
+        # URL schemes are case-insensitive, and this decision is the fail-safe
+        # direction: missing one sends a raw external URL to a websocket that
+        # only accepts ``data:``. ``_count_external_image_urls`` still matches
+        # case-sensitively; that is the bridge's own guard and out of scope here.
+        return isinstance(image_url, str) and image_url.lower().startswith(("http://", "https://"))
     if isinstance(value, list):
         return any(_json_value_contains_external_input_image(item) for item in value)
     if is_json_mapping(value):
--- a/app/modules/proxy/_service/response_create.py
+++ b/app/modules/proxy/_service/response_create.py
@@ -181,6 +181,27 @@ def _responses_request_contains_input_image(payload: ResponsesRequest) -> bool:
     return any(_json_value_contains_input_image_part(item) for item in input_value)
 
 
+def _json_value_contains_external_input_image(value: JsonValue) -> bool:
+    """Whether ``value`` holds an ``input_image`` part that still names an external URL.
+
+    Recurses the same way :func:`_json_value_contains_input_image_part` does,
+    rather than walking the two shapes ``_count_external_image_urls`` knows
+    (a top-level item and its ``content`` array). The transport decision has to
+    see every external URL the payload carries, including one nested in a
+    ``function_call_output`` output array, because an image the URL inliner
+    never visits is precisely the one that is still external when the request
+    reaches the upstream websocket.
+    """
+    if _input_part_is_image(value):
+        image_url = value.get("image_url") if is_json_mapping(value) else None
+        return isinstance(image_url, str) and image_url.startswith(("http://", "https://"))
+    if isinstance(value, list):
+        return any(_json_value_contains_external_input_image(item) for item in value)
+    if is_json_mapping(value):
+        return any(_json_value_contains_external_input_image(child) for child in value.values())
+    return False
+
+
 def _input_image_request_requires_http_upstream(
     payload: ResponsesRequest,
     *,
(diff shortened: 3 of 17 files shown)


Soju06:
Good catch — fixed in bf2727b00.

You are right that this PR introduces the exposure. Before it, *any* `input_image` pinned upstream HTTP, so a nested external URL never reached the websocket. Narrowing the pin to "external URL present" while reusing `_count_external_image_urls` — whose traversal stops at a top-level item and one level of `content` — left that shape unpinned. A `function_call_output` output array is a routine Codex tool-result form, so this was reachable, not theoretical.

I scoped the fix slightly differently from your suggestion, for one reason.

**Applied:** the transport decision gets its own recursive check, `_json_value_contains_external_input_image`, which recurses the same way `_json_value_contains_input_image_part` already does and filters on the `http(s)` scheme. So the clause stays narrow — an inline `data:` image nested just as deeply still does not pin, which matters because those are the overwhelming majority of image turns and pinning them is the defect this PR exists to fix. Proved by mutation: with the check reverted to the shallow counter, the new `test_external_image_url_inside_a_tool_output_still_pins_http` fails and `test_inline_image_inside_a_tool_output_does_not_pin_http` still passes.

**Not applied here:** widening `_inline_input_image_urls` and the bridge's post-inline guard. That is a real gap — a *bridged* request carrying an external URL nested deeper is neither inlined nor rejected by the 400 `image_download_failed` guard, and reaches the upstream websocket raw. But it is pre-existing and independent of this PR's concern, and changing the guard's traversal turns requests that pass today into 400s on the bridge path. It is now recorded as a known bound in the change context rather than left silent, and it deserves its own PR.

The spec scenario and the capability context were updated to match: external-URL detection for the transport decision now says "anywhere in the input — including inside a tool-output array, which the image inliner never rewrites", instead of naming the two shallow shapes.

Soju06:
Both addressed in the latest push.

**Case-sensitive scheme match (minor)** — right, and it matters more here than in the helper it mirrors: this is the fail-safe direction, so missing `HTTPS://` hands the upstream websocket a URL it does not accept. The transport predicate now lowercases before matching, with a parametrised regression over `https` / `HTTPS` / `HtTp`. I left `_count_external_image_urls` case-sensitive on purpose — that is the bridge's own post-inline 400 guard, changing it turns requests that pass today into 400s on the bridge path, and it belongs with the traversal widening in a separate PR.

**Explicit-override precedence (major)** — correct, the scenario and the precedence text disagreed. Precedence item 2 already says the residual `input_image` pins are deliberately evaluated ahead of an explicit `"websocket"` override wherever the request passes through the bridge routing decision, because that override short-circuits the size gate and an oversized image payload would otherwise fail locally with `400 payload_too_large`. The *scenario* still said "under any policy" with no carve-out, so it read as contradicting that. It is now scoped to a request that trips none of the item 2 bypasses, and there is a new scenario — `external image URL still forces HTTP under an explicit websocket override` — plus a route-level regression for it. `openspec validate --strict` and `--specs` both pass.

PR #2387, opened 2026-09-11
fix(proxy): deliver the denied-anchor refusal to native clients

### LINKED ISSUE #2364: bug: denied proxy-injected anchor fails closed with a 502 raised after the SSE response started — client gets HTTP 200 + empty stream ("stream closed before response.completed")
## Summary

When upstream rejects a **proxy-injected** `previous_response_id` (`previous_response_not_found`, typically right after the previous turn was killed by `server_is_overloaded` / `stream_incomplete`), the next request on that bridge session is refused before dispatch with `denied_proxy_anchor_before_dispatch`. The refusal is a `ProxyResponseError(502, stream_incomplete, "The previous response anchor was rejected upstream; retry the request.")`, but it is raised **after the downstream streaming response has already started**:

- the client receives **HTTP 200** and then the stream ends with **0 bytes and no terminal event**;
- Codex shows `stream disconnected before completion: stream closed before response.completed`;
- the intended 502 / "retry" message never reaches the client, and the ASGI layer logs an unhandled exception with a full traceback.

Users report several of these in a row in the same thread; pressing continue does not help, a new thread does.

## Where

`app/modules/proxy/_service/http_bridge/request_submit.py` (v1.25.0-beta.7, around L2232–2254):

```python
_record_continuity_fail_closed(
    surface="http_bridge",
    reason="denied_proxy_anchor_before_dispatch",
    previous_response_id=request_state.previous_response_id,
    session_id=request_state.session_id,
    upstream_error_code="previous_response_not_found",
)
raise ProxyResponseError(
    502,
    openai_error(
        "stream_incomplete",
        "The previous response anchor was rejected upstream; retry the request.",
    ),
)
```

It propagates out of the stream body iterator (`_stream_http_bridge_session_events` streaming.py:5227 → `_stream_via_http_bridge_impl` streaming.py:3247 → `downstream_delivery`) instead of being turned into an error response.

## Log sequence (one occurrence; account ids redacted)

```
01:58:47Z WARNING http_bridge_event event=terminal_error bridge_kind=thread_header ... model=gpt-5.6-sol detail=server_is_overloaded
01:58:48Z WARNING continuity_fail_closed surface=websocket_stream reason=previous_response_not_found previous_response_id=sha256:72aac0a1dcd6 session_id=sha256:64fdeb3ba12b upstream_error_code=previous_response_not_found diagnostics=previous_response_source=proxy_injected fresh_replay_available=false ...
01:58:48Z WARNING http_bridge_event event=terminal_error bridge_kind=thread_header ... detail=stream_incomplete
01:58:49Z WARNING continuity_fail_closed surface=http_bridge reason=denied_proxy_anchor_before_dispatch previous_response_id=sha256:72aac0a1dcd6 session_id=sha256:64fdeb3ba12b upstream_error_code=previous_response_not_found diagnostics=None
01:58:49Z INFO    "POST /backend-api/codex/responses HTTP/1.1" 200 OK
01:58:49Z WARNING responses_stream_terminal_delivery request_id=... surface=responses outcome=exception_before_terminal terminal=None chunks=0 bytes=0 exc=ProxyResponseError
01:58:49Z ERROR   unhandled_error_response request_id=None method=POST path=/backend-api/codex/responses status=500 code="server_error" message="Proxy response error (502)"
01:58:49Z ERROR   Exception in ASGI application   (traceback follows)
```

(The access log shows only `200 OK` for `/backend-api/codex/responses` at every one of these timestamps; the `status=500` exists only in the exception handler log.)

## Frequency (v1.25.0-beta.7, 100 min)

- 20 occurrences; every one is 1:1 with a preceding `continuity_fail_closed surface=websocket_stream reason=previous_response_not_found` on the same session.
- All 20 carry `previous_response_source=proxy_injected fresh_replay_available=false`.
- 20 distinct anchors (not one stuck anchor), but the same session can hit it repeatedly as successive turns are killed by upstream overload.

## Expected

Since the anchor was proxy-injected and `fresh_replay_available=false`, failing closed seems right — but the decision should reach the client as an error it can act on, before the SSE response starts. Ideally the same shape the Codex client already recovers from (`400 previous_response_not_found`, param `previous_response_id`), which makes it resend the full request; at minimum a real 5xx instead of `200` + an empty stream, so the unhandled-exception traceback also goes away.

Related: #1879 (retire denied anchors without redispatch).

## Environment

- codex-lb `1.25.0-beta.7` (official `ghcr.io/soju06/codex-lb` image, Docker Compose, PostgreSQL)
- Dashboard: `upstream_stream_transport=auto`, HTTP responses session bridge enabled
- Clients: Codex Desktop / codex-tui over HTTP downstream (`/backend-api/codex/responses`)

Soju06:
## Summary

When upstream rejects a **proxy-injected** `previous_response_id`, the HTTP bridge refuses the next request on that session before dispatch (`denied_proxy_anchor_before_dispatch`) with `ProxyResponseError(502, stream_incomplete, "The previous response anchor was rejected upstream; retry the request.")`. A native Codex client received **HTTP 200 and a zero-byte stream with no terminal event**: `stream disconnected before completion: stream closed before response.completed`. The retry message never arrived and the ASGI layer logged an unhandled traceback (#2364).

## Type of change

- [ ] `feat:` / `refactor:` / `docs:` / `chore:` / `test:`
- [ ] **Breaking change**

Linked issue: Fixes #2364

## OpenSpec

- [x] This PR touches a codex-faithful path (SSE framing, error envelopes) and preserves upstream-equivalent behavior

Change directory: `openspec/changes/deliver-http-bridge-denied-anchor-refusal-to-native-clients/`

## Root cause

This is not "the response happened to have started already" — the api layer classifies the failure **purely by error code**. `_stream_responses` computes

```python
native_transport_startup_failure = (
    preserve_native_failure_lifecycle
    and startup_error_code in {"stream_incomplete", "stream_idle_timeout",
                               "upstream_request_timeout", "upstream_unavailable"}
)
```

and, when true, deliberately re-raises an **already probe-caught** error back into the committed body instead of returning an error response; `_stream_response_error_events` applies the identical set post-commit. `preserve_native_failure_lifecycle` is native-Codex-only, and native clients also skip heartbeat and keepalive injection while the bridge suppresses its own pre-`response.created` keepalive — so the body is literally 0 bytes. Non-native clients were unaffected: they got a JSON 502 or a terminal `response.failed`.

So shortening the window could never have fixed it. Nor could re-coding the refusal to `previous_response_not_found`: that flips `explicit_previous_response_rejection` / `should_attempt_previous_response_recovery`, turning a deliberate fail-closed into a local recovery attempt, against the spec.

## Changes

The public contract is unchanged — still `502`, still `stream_incomplete`, same message, as the spec mandates. What changes is that the exception now carries **provenance**, so the api layer stops mistaking a proxy-side pre-dispatch refusal for an observed upstream transport failure.

- `ProxyResponseError` gains `local_pre_dispatch_refusal`. The denied-anchor fence sets it — provably pre-dispatch, since the raise sits above both the pending-request enqueue and the upstream send.
- `_stream_responses` excludes a flagged error from `native_transport_startup_failure`, so it returns a real HTTP 502 JSON body in the common fast-failure case (inside the 2 s startup probe window).
- `_stream_response_error_events` excludes it from **both** the re-raise gate and the `synthetic_transport_failure_event` marker gate, so the probe-overrun case yields an *unmarked* terminal `response.failed`. Both gates are load-bearing: changing only the first leaves the marker to convert the terminal straight back into a silent abort.
- `_OwnerForwardRequestError` propagates the flag, since that wrapper rebuilds the exception from a fixed field list.

### Cross-hop fix found during review

The in-process fix alone is not enough on a multi-instance ring. Adversarial review built a two-instance reproduction (durable session owned by instance-a, real owner-forward client, faked transport) and got **`STATUS 200 BODY ''` one hop out**: the origin rebuilds the error from the owner's status and body, and the body cannot express the difference between "the owner refused before sending anything" and "the owner observed a transport failure" — both are `stream_incomplete`.

So the provenance travels: the owner stamps `x-codex-bridge-local-pre-dispatch-refusal` on its error response (only for forwarded requests — the header belongs to the internal bridge contract and is asserted never to reach an external client), and `HTTPBridgeOwnerClient` reads it back into the rebuilt `ProxyResponseError`.

## Test plan

```
uv run ruff check app tests                       # All checks passed!
uv run ty check app                               # All checks passed!
uv run openspec validate deliver-http-bridge-denied-anchor-refusal-to-native-clients --strict
uv run openspec validate --specs                  # 65 passed, 0 failed

uv run pytest tests/unit                                              # 9968 passed, 4 skipped
uv run pytest tests/integration/test_http_responses_bridge.py \
             tests/unit/test_http_bridge_forwarding.py                # 244 passed
uv run pytest tests/integration/{test_multi_instance_bridge,test_proxy_responses,test_codex_client_compat,\
             test_native_sse_egress,test_native_routed_egress,test_proxy_affinity_local_recovery,\
             test_proxy_api_extended}.py                              # 216 passed, 320 skipped
uv run pytest tests/integration/{test_proxy_websocket_responses,test_proxy_sticky_sessions,test_proxy_compact,\
             test_routed_stream_release,test_startup_bridge_cleanup,test_proxy_websocket_overflow,\
             test_proxy_chat_completions}.py                          # 284 passed
```

Regression coverage sits at the product path on **both halves of the hop**, and every test asserts the `continuity_fail_closed reason=denied_proxy_anchor_before_dispatch` record so a mis-orchestrated test fails loudly instead of silently exercising the ordinary upstream-denial path (which returns a visible 502 and would pass in both the buggy and the fixed world).

**Fails-before-fix proof**, patch reverse-applied:
- owner half — `AssertionError: assert None == '1' where None = get('x-codex-bridge-local-pre-dispatch-refusal')`
- origin half — `AssertionError: native client received 200 with body ''`

## Known bounds, recorded in the proposal

- **Scope is one raise site.** The ~20 sibling local refusals in `_service/http_bridge/` that also use codes from the native-abort set are enumerated by name in `proposal.md` but deliberately untouched — some of them are *not* safely pre-dispatch, so a blanket sweep would be wrong. Follow-up issue below.
- **Rolling-upgrade window for ring deployments.** An upgraded owner answers the internal forward with a real 502 where it previously aborted a committed 200; an origin that predates this change ignores the marker and applies its own native lifecycle, so a forwarded refusal reaches the client as an empty 200 until every replica is on this build. Single-instance deployments — the default — never take that path.
- **The anchor-poisoning loop is untouched.** When the durable clear returns a clean no-match the tombstone persists and the durable row re-injects the same id, so the user now sees repeated *visible* 502s rather than repeated silent truncations. Fixing the loop itself is a separate change.

## Checklist

- [x] Added regression coverage at the externally failing product path.
- [x] Ran the relevant pytest subsets locally.
- [x] `openspec validate --strict` passes.
- [x] Simplicity gates reviewed (PRINCIPLES.md P1-P6) — no new setting, no new nav, defaults unchanged.

### DIFF
--- a/app/core/clients/proxy.py
+++ b/app/core/clients/proxy.py
@@ -590,6 +590,7 @@ class ProxyResponseError(Exception):
         retry_after_seconds: int | None = None,
         retry_after_header: str | None = None,
         reservation_released: bool = False,
+        local_pre_dispatch_refusal: bool = False,
     ) -> None:
         super().__init__(f"Proxy response error ({status_code})")
         self.status_code = status_code
@@ -604,6 +605,14 @@ class ProxyResponseError(Exception):
         self.retry_after_seconds = retry_after_seconds
         self.retry_after_header = retry_after_header
         self.reservation_released = reservation_released
+        # True when the proxy refused the request itself before any upstream
+        # frame was sent, so the failure is not an observed upstream transport
+        # failure: the native Codex transport-failure lifecycle, which ends the
+        # body without a terminal event, must not be applied to it (issue
+        # #2364, where such a refusal reached the client as an empty 200). The
+        # refusing instance may be another replica: an internal bridge forward
+        # carries the provenance back so the origin reaches the same verdict.
+        self.local_pre_dispatch_refusal = local_pre_dispatch_refusal
 
 
 def _safe_retry_after_header(headers: Mapping[str, object] | None) -> str | None:
--- a/app/modules/proxy/_service/http_bridge/owner_forwarding.py
+++ b/app/modules/proxy/_service/http_bridge/owner_forwarding.py
@@ -185,6 +185,7 @@ class _OwnerForwardRequestError(ProxyResponseError):
             upstream_status_code=source.upstream_status_code,
             upstream_error_code=source.upstream_error_code,
             failed_session=source.failed_session,
+            local_pre_dispatch_refusal=source.local_pre_dispatch_refusal,
         )
         self.outcome = outcome
 
--- a/app/modules/proxy/_service/http_bridge/request_submit.py
+++ b/app/modules/proxy/_service/http_bridge/request_submit.py
@@ -2062,12 +2062,20 @@ class _HTTPBridgeRequestSubmitMixin:
                             session_id=request_state.session_id,
                             upstream_error_code="previous_response_not_found",
                         )
+                        # The fence sits above the pending-request enqueue and
+                        # the upstream send, so this refusal is proven to be
+                        # ours rather than an upstream transport failure. Say
+                        # so on the exception: without the flag the native
+                        # Codex lifecycle aborts the committed body with no
+                        # terminal event and the client sees an empty 200
+                        # (issue #2364).
                         raise ProxyResponseError(
                             502,
                             openai_error(
                                 "stream_incomplete",
                                 "The previous response anchor was rejected upstream; retry the request.",
                             ),
+                            local_pre_dispatch_refusal=True,
                         )
                     if (
                         request_state.verified_stale_anchor_replay
--- a/app/modules/proxy/api.py
+++ b/app/modules/proxy/api.py
@@ -278,7 +278,10 @@ from app.modules.proxy.api_key_usage import estimate_api_key_request_usage
 from app.modules.proxy.capability_routing import required_capability_metadata_values
 from app.modules.proxy.downstream_delivery import DeliveryTracedStreamingResponse
 from app.modules.proxy.helpers import _openai_error_param, _parse_openai_error, _rate_limit_details
-from app.modules.proxy.http_bridge_forwarding import parse_forwarded_request
+from app.modules.proxy.http_bridge_forwarding import (
+    HTTP_BRIDGE_LOCAL_PRE_DISPATCH_REFUSAL_HEADER,
+    parse_forwarded_request,
+)
 from app.modules.proxy.images_observability import (
     IMAGE_ROUTE_MODEL_STATE,
     IMAGE_ROUTE_STARTED_AT_STATE,
@@ -6812,6 +6815,11 @@ async def _stream_responses(
             and isinstance(startup_error, ProxyResponseError)
             and startup_error_code
             in {"stream_incomplete", "stream_idle_timeout", "upstream_request_timeout", "upstream_unavailable"}
+            # A refusal the proxy raised itself before any upstream frame left
+            # the process is not a transport failure, so it keeps the ordinary
+            # error response instead of being replayed into the committed body
+            # as a terminated stream (issue #2364).
+            and not startup_error.local_pre_dispatch_refusal
         )
         if native_transport_startup_failure:
             assert isinstance(startup_error, ProxyResponseError)
@@ -6834,7 +6842,10 @@ async def _stream_responses(
             return _stream_startup_error_response(
                 request,
                 startup_error,
-                headers=rate_limit_headers,
+                headers={
+                    **rate_limit_headers,
+                    **_owner_forward_local_refusal_headers(startup_error, forwarded_request=forwarded_request),
+                },
             )
     stream = _normalize_public_responses_stream(
         _stream_response_error_events(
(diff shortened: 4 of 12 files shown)


Soju06:
Addressed in the latest push — you are right that the code alone is a weak assertion. The regression that matters here is that the client is told something actionable, and a change that kept the terminal while dropping or rewording the retry instruction would have left it exactly where #2364 did. The message is now pinned at the unit seam and on both the pre-commit 502 and the post-commit terminal.

PR #2390, opened 2026-09-11
fix(proxy): retire durably unroutable bridge continuity owners

Soju06:
> Stacked on #2384 (`fix/observe-owner-replay-rejection`). Review the top commit; the base merges first.

## Why

A durable HTTP bridge row names the account that owns a Codex thread, and **nothing ever takes that name back**. `accounts/repository.py::_close_http_bridge_sessions_for_account` runs only on the `DEACTIVATED` transition — `PAUSED`, `REAUTH_REQUIRED`, `RATE_LIMITED` and `QUOTA_EXCEEDED` all leave `account_id` in place forever. Every resume re-reads that name, requires it because `thread_header` is hard by kind, and fails closed with 502 `previous_response_owner_unavailable` while healthy accounts sit idle.

`StickySession` already solved exactly this, and `sticky_repository.py:683-703` explains why deleting the row is not the answer: a delete is indistinguishable from "this key was never seen", so the anchored path keeps failing closed even after the owner recovers, because nothing on that path can re-create the row it needs. A tombstone instead proves the owner was deliberately abandoned, which authorizes picking a fresh one. **`HttpBridgeSessionRecord` has no counterpart** — `account_id`'s `ON DELETE SET NULL` is its only release, so only deleting the account row frees a thread.

Production, 36 hours on `1.25.0-beta.7` with 24 accounts:

| Observation | Value |
|---|---|
| `previous_response_owner_unavailable` request-log rows | 1,480 |
| … owned by one `reauth_required` account | **1,361** |
| … owned by `paused` accounts | 22 |
| `draining` rows pinned to unroutable owners, frozen since the previous deploy | 6 |
| … owning operation rows, so unreachable by `purge_abandoned_before` | **6 of 6** |

`reauth_required` is the largest single source and is in **neither** cleanup path: not the `DEACTIVATED` detach, not `_HARD_STICKY_UNAVAILABLE_STATUSES`. And that `~exists(operation)` guard is the same one that made the 2026-09-04 incident permanent — every poisoned row happened to own operations.

## What changes

- **Schema.** `http_bridge_sessions` gains `continuity_abandoned_at` and `continuity_abandonment_scope`, mirroring `sticky_sessions` in name and meaning so a reviewer reads one pattern, not two. Nullable, no backfill: a replica on the previous build keeps treating `account_id` as hard ownership through a rolling deploy.

- **Retirement sweep.** `retire_stale_unavailable_bridge_owners` runs on the existing sticky cleanup pass with the existing grace constant (`_STALE_HARD_CODEX_SESSION_UNAVAILABLE_SECONDS`, 6h) and **no new setting**. Phase 1 tombstones a row whose owner sat in `HARD_OWNER_UNAVAILABLE_STATUSES` past its reset horizon while the row itself went untouched for the window; phase 2 deletes a tombstone that then sat another full window unclaimed.

- **The grace clock is `last_seen_at`, not a refreshed marker.** The sticky design needs `_refresh_hard_sticky_outage_grace` *plus* a once-per-database `runtime_sentinels` backfill, because `StickySession.updated_at` moves for reasons unrelated to serving. The bridge row already records when the thread last actually worked, so the window is "unroutable owner **and** no successful turn for 6h" — no hook, no sentinel, and correct on the first sweep after deploy without a backfill.

- **`~exists(operation)` no longer blocks retirement.** It still guards phase 2, where deleting the row would `CASCADE` the recovery ledger (`durable_bridge_repository.py:3000-3003`). Phase 1 writes two columns and deletes nothing, so the guard protects nothing there — and keeping it would reproduce 2026-09-04 exactly.

- **Retirement is applied once, in `_to_lookup`.** Every coordinator path funnels through it, so **all four lookups are covered** — the detached-row filter it sits beside is applied at call sites and reaches only two, leaving both `find_session_by_latest_*` fallbacks uncovered. A retired lookup keeps its identity (callers still need the session id to claim the row) and surrenders every piece of continuity evidence: owner, response anchor, turn state, pending tool calls. Dropping the anchors with the owner is what makes it coherent — an anchor without its account is upstream state no replacement can read.

- **A retired owner is not a missing one.** `streaming.py`'s `durable_owner_missing` exempts retired rows. Failing closed is right when a row *should* name an account and does not — that is lost state — but a marker is positive proof the owner was abandoned. **Without this exemption, retiring an owner would convert one 502 into a different 502**, which is why the end-to-end test matters more than the unit layer here.

- **Retirement is reversible.** A successful `claim_session` clears both markers, the same way a sticky rebind clears its own.

- **`HARD_OWNER_UNAVAILABLE_STATUSES`** joins `ROUTABLE_STATUSES` in `account_eligibility.py`, documented as SQL-usable and safe only when paired with a long inactivity grace: `reauth_access_token_is_expired` needs a decrypted token, so `REAUTH_REQUIRED` earns its place there only because every consumer pairs it with the 6h window. Anything deciding routability *now* still uses `ROUTABLE_STATUSES` plus the expiry check.

No new `CODEX_LB_*` setting, no `.env.example` change, no README growth, no dashboard change. None of the four ceiling-guarded proxy files grows.

## Not in this change

- **Immediate retirement on the request path.** A thread still waits out the grace window. The evidence-based variant — retire now when the owner cannot return before this request's own deadline, using `accounts.reset_at` as the proof — is the follow-up, and it needs the request-path CAS writer with the `SELECT … FOR UPDATE` on owner status that `abandon_legacy_session_header_owner_if_unavailable` demonstrates.
- **Moving a turn whose body cannot leave its account.** Retirement stops the re-welding; it does not make an account-bound history replayable. A client-supplied `previous_response_id` with no resolvable owner still fails closed — the same outcome as today, reached by a different branch.
- **`_HARD_STICKY_UNAVAILABLE_STATUSES` omitting `REAUTH_REQUIRED`.** The same gap exists on the sticky side and deserves its own change rather than a drive-by edit from a bridge PR.

## ⚠️ Migration coordination

This PR adds an Alembic revision on `20260911_010000_merge_pin_index_and_affinity_heads`. Two migration PRs merged back to back on 09-11 produced multiple heads and turned `main` red (fixed by #2368). Please avoid merging another migration PR alongside this one.

## Test plan

| Gate | Result |
|---|---|
| `make lint` (incl. architecture / cancellation / timing-seam / settings-tier checks) | pass |
| `make typecheck` (`ty`) | pass |
| `make migration-check` | `migration_policy=ok`, `schema_drift=none`, single head |
| `make test-unit` | **10,049 passed**, 6 skipped, 1 xfailed |
| `make test-integration-bridge` | **342 passed** |
| `make test-integration-core-1` / `-2` / `-3` | **869** / **910** / **1,036 passed** |
| `uv run pytest tests/e2e` | **27 passed**, 1 skipped |
| `openspec validate retire-unroutable-bridge-continuity-owners --strict` | valid |
| `codex review --base origin/main` | 1 finding fixed, 2 rebutted (below) |

New coverage:
- `tests/unit/test_durable_bridge_owner_retirement.py` (new, 15 cases) — retired past the window; not retired inside it; never for a healthy owner; rate-limited with a future vs. past reset; `reauth_required`; operation rows survive phase 1; phase 2 deletes without a ledger and keeps with one; idempotent; a fresh claim clears it; ownerless rows left to the existing purges; a retired row names no owning replica; a client-supplied anchor is still refused; the status set is closed.
- `tests/integration/test_repositories.py` — the `reauth_required` contract end to end at the repository layer, including phase 2 on the real dialect (the correlated `NOT EXISTS` in a `DELETE` is where PostgreSQL and SQLite are most likely to diverge). Added to `POSTGRES_PYTEST_TARGETS`.
- `tests/integration/test_http_responses_bridge.py` — **the user-visible proof**: a bridged thread whose owner is paused and swept returns **200** on a healthy account and never re-selects the retired owner. It polls for the durable row rather than assuming it landed before the response returned; the first draft failed once for exactly that race.
- `tests/integration/test_migrations.py` + `POSTGRES_PYTEST_TARGETS` — upgrade/downgrade, nullability, single head.
- `tests/integration/test_affinity_invite_migration.py` — one unrelated repair this revision forces: that suite downgrades to the affinity parent and then asserted global `check_schema_drift(url) == ()`, which no revision above the topmost merge can satisfy. It already derives `_current_head` "so later merge revisions do not have to edit these suites"; the post-downgrade assertion is now scoped to the tables it governs in that same spirit, and the two assertions at head stay strict.
- `tests/unit/test_sticky_session_cleanup_scheduler.py` — the sweep runs on the same pass with the same grace window.

## Local codex review

- **Fixed** — a retired lookup still carried `owner_instance_id` / `lease_expires_at`, so
  `_durable_bridge_lookup_active_owner` would still name the replica that owned it. Not reachable
  today (the sweep needs 6h of inactivity, far beyond any bridge lease TTL), but leaving them set
  makes the invariant depend on a TTL setting nobody would think to check before changing.
  `_to_lookup` now masks both; `owner_epoch` is fencing state, not continuity evidence, so it
  survives. Regression: `test_a_retired_row_names_no_owning_replica`.
- **Rebutted** — *"allow retired owners past the previous-response guard."* A client-supplied
  `previous_response_id` names upstream state that lived on the retired account; no replacement
  can read it, so failing closed is correct. It is also the same outcome as before this change,
  reached by a different branch, so it is not a regression. What *is* wrong is the error text —
  502 "retry later" invites a retry that can never succeed — and replacing it is the follow-up.
  Pinned by `test_a_retired_row_still_refuses_a_client_supplied_anchor`.
- **Rebutted** — *"exclude the retired account from replacement selection."* Retirement already
  stripped the anchors, so selecting the recovered former owner starts a fresh turn exactly as any
  other account would — while being the account most likely to still hold this conversation's
  upstream prompt cache. Excluding it would force a cache miss for no safety gain, and while the
  account is still unroutable the selector skips it anyway.

Partial fix for #1707.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,6 +43,8 @@ POSTGRES_PYTEST_TARGETS := \
 	tests/integration/test_repositories.py::test_upsert_account_slot_discards_pending_downgrade_evidence_on_reimport \
 	tests/integration/test_migrations.py::test_account_plan_downgrade_observations_migration_upgrade_and_downgrade \
 	tests/integration/test_migrations.py::test_account_pending_deletion_migration_upgrade_and_downgrade \
+	tests/integration/test_migrations.py::test_bridge_continuity_abandonment_migration_upgrade_and_downgrade \
+	tests/integration/test_repositories.py::test_retire_stale_unavailable_bridge_owners_frees_a_reauth_pinned_thread \
 	tests/integration/test_usage_repository.py::test_bulk_history_since_primary_query_plan_is_index_only_postgresql \
 	tests/integration/test_usage_repository.py::test_bulk_history_since_cutoff_query_plan_is_index_only_postgresql \
 	tests/integration/test_usage_repository.py::test_bulk_history_since_secondary_query_plan_is_index_only_postgresql \
--- a/app/db/alembic/versions/20260911_060000_add_bridge_session_continuity_abandonment.py
+++ b/app/db/alembic/versions/20260911_060000_add_bridge_session_continuity_abandonment.py
@@ -1,0 +1,40 @@
+"""Add continuity-abandonment markers to durable HTTP bridge sessions.
+
+Revision ID: 20260911_060000_add_bridge_session_continuity_abandonment
+Revises: 20260911_030000_add_local_login_policy
+Create Date: 2026-09-11
+"""
+
+from __future__ import annotations
+
+import sqlalchemy as sa
+from alembic import op
+
+revision = "20260911_060000_add_bridge_session_continuity_abandonment"
+down_revision = "20260911_030000_add_local_login_policy"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # Mirrors sticky_sessions' pair exactly: a scope-only marker is the new
+    # reader's tombstone, while a timestamp with NULL scope abandons globally.
+    # Both stay NULL on every existing row, so a rolling deploy keeps treating
+    # account_id as hard ownership until a writer retires a specific owner.
+    bind = op.get_bind()
+    columns = {column["name"] for column in sa.inspect(bind).get_columns("http_bridge_sessions")}
+    with op.batch_alter_table("http_bridge_sessions") as batch_op:
+        if "continuity_abandoned_at" not in columns:
+            batch_op.add_column(sa.Column("continuity_abandoned_at", sa.DateTime(timezone=True), nullable=True))
+        if "continuity_abandonment_scope" not in columns:
+            batch_op.add_column(sa.Column("continuity_abandonment_scope", sa.String(32), nullable=True))
+
+
+def downgrade() -> None:
+    bind = op.get_bind()
+    columns = {column["name"] for column in sa.inspect(bind).get_columns("http_bridge_sessions")}
+    with op.batch_alter_table("http_bridge_sessions") as batch_op:
+        if "continuity_abandonment_scope" in columns:
+            batch_op.drop_column("continuity_abandonment_scope")
+        if "continuity_abandoned_at" in columns:
+            batch_op.drop_column("continuity_abandoned_at")
--- a/app/db/models.py
+++ b/app/db/models.py
@@ -2452,6 +2452,14 @@ class HttpBridgeSessionRecord(Base):
     latest_input_item_count: Mapped[int | None] = mapped_column(Integer, nullable=True)
     latest_input_full_fingerprint: Mapped[str | None] = mapped_column(String(64), nullable=True)
     latest_pending_tool_calls_json: Mapped[str | None] = mapped_column(Text, nullable=True)
+    # Continuity-owner retirement, mirroring sticky_sessions' pair: a non-NULL
+    # scope retires ownership only for the matching typed source, while a
+    # non-NULL timestamp with NULL scope retires it globally. Deleting the row
+    # instead would be indistinguishable from "never seen" and would leave the
+    # lookup failing closed forever; a marker says the owner was deliberately
+    # abandoned, so picking a fresh one is authorized.
+    continuity_abandoned_at: Mapped[datetime | None] = mapped_column(DateTime(timezone=True), nullable=True)
+    continuity_abandonment_scope: Mapped[str | None] = mapped_column(String(32), nullable=True)
     created_at: Mapped[datetime] = mapped_column(
         DateTime(timezone=True),
         nullable=False,
--- a/app/modules/proxy/_service/http_bridge/streaming.py
+++ b/app/modules/proxy/_service/http_bridge/streaming.py
@@ -2012,13 +2012,24 @@ class _HTTPBridgeStreamingMixin:
                 and durable_lookup.latest_turn_state is not None
             )
         )
+        # A retired owner is not a missing one. Failing closed is right when a
+        # row should name an account and does not — that is lost state. But a
+        # retirement marker is positive proof that the owner was deliberately
+        # abandoned because it could not come back, so the correct response is
+        # to pick a fresh owner, exactly as the sticky selector does when it
+        # sees its own tombstone. Without this exemption, retiring an owner
+        # would convert one 502 into a different 502.
         durable_owner_missing = (
-            durable_lookup is not None and durable_lookup_requires_owner and durable_lookup.account_id is None
+            durable_lookup is not None
+            and durable_lookup_requires_owner
+            and durable_lookup.account_id is None
+            and not durable_lookup.continuity_abandoned
         )
         model_transition_owner_missing = (
             durable_model_transition_lookup is not None
             and durable_model_transition_requires_owner
             and durable_model_transition_lookup.account_id is None
+            and not durable_model_transition_lookup.continuity_abandoned
         )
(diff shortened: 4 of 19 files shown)


Soju06:
Both review findings addressed in `45b079c9e`.

**P2 — "Clear continuity when reclaiming a retired owner": fixed.** This was a real bug I introduced. `_to_lookup` masked the anchors while the marker stood, so the request that triggered the claim was planned as an unanchored fresh start — but clearing only the marker left `latest_response_id`, turn state, fingerprints, pending calls and aliases behind, and the next lookup would have served that abandoned anchor as ordinary continuity. It bites hardest exactly where you said: when the retired account itself recovers and is selected again, `account_changed` is False. Reclaiming a retired row now discards continuity exactly as an account change does, aliases included. Regression: `test_reclaiming_the_recovered_former_owner_discards_the_abandoned_anchor`.

**P1 — "Start the grace period when the owner becomes unavailable": spec corrected, code kept.** You are right that the code and the change spec disagreed, and that the spec is authoritative — so I fixed the disagreement, on the spec side, deliberately.

The sticky sweep needs an outage clock because `StickySession` has **no activity signal at all**: `updated_at` moves on every pin refresh, so without the hook it cannot tell "owner died long ago" from "owner died just now". A bridge row records when the thread last actually *served*. That answers the question the grace is really asking — is this thread live enough for its anchor to be worth protecting? — without a second clock, a status-transition hook, and the `runtime_sentinels` backfill that hook then requires for accounts already unavailable at deploy time.

So the contract is now stated as: a session that served a turn inside the window keeps its owner **however long** that owner has been down; a session that served nothing across the window may be retired as soon as its owner is unroutable, **however recently** that became true.

The cost of that choice is bounded and I would rather state it than hide it: a thread idle for six hours whose owner pauses for two minutes loses its response anchor and re-sends its history on the next resume. The alternative is that the same thread returns 502 for as long as the pause lasts. An anchor nobody has touched in six hours is worth less than the thread being resumable at all — and that is the entire point of #1707.

Both halves are pinned: `test_a_dormant_thread_is_retired_on_a_fresh_outage` (healthy owner + 6h idle ⇒ not retired; owner pauses ⇒ retired) and `test_recent_activity_keeps_an_unroutable_owner` (paused owner + recent turn ⇒ kept). Rationale recorded in the change's `context.md`.

Re-verified after both changes: `make lint`, `make typecheck`, `make test-unit` (10,051 passed), `make test-integration-bridge` (342 passed).

Komzpa:
@codex review

Soju06:
Strengthened the end-to-end test in `f8574a33a` — self-review caught that it was not proving what this PR claims.

It ran over `/v1/responses` with only a `prompt_cache_key`, which produces a **soft** `prompt_cache` bridge key. A soft key falls back to a healthy account on its own, so the test would have passed without any of this change. It now goes through `/backend-api/codex/responses` with a thread identity, which produces a hard `thread_header` key — the shape that actually welds a thread to its owner.

Verified the strengthened test is load-bearing: removing the `durable_lookup.continuity_abandoned` exemption in `streaming.py` makes it fail, and restoring it makes it pass. That exemption is the part of this PR that would otherwise have swapped one 502 for another, so it is exactly what needed covering.

`make test-integration-bridge` re-run: 342 passed.

PR #2391, opened 2026-09-11
feat(proxy): shape the failover decision around the pool and render its terminal

Soju06:
Stacked on #2373, which carries the OpenSpec change this implements. **Review that one first; merge it first.**

## Summary

`failover_decision` stops a walk on `candidates_remaining <= 0`, and every caller derives that number from a fixed per-transport constant — 3, 3 and 2 — unrelated to how many usable accounts the pool holds. This branch replaces the counter with a predicate, gives the walk a terminal renderer, and records which bound ended it.

**This is a library addition. Nothing calls it yet**, by design: converting the per-transport attempt loops is the follow-up, and it is the change that removes lines from `app/modules/proxy/service.py`, which sits at 2600/2600.

## Type of change

- [ ] **Breaking change**

## What changed

- `failover_decision` takes `more_candidates_possible: bool` instead of `candidates_remaining: int`, and the `non_retryable` check moves ahead of the candidate check. `candidates_remaining` stays as a deprecated keyword shim for one release. All four call sites are converted mechanically; behaviour for every existing input is unchanged.
- `MAX_ACCOUNT_ATTEMPTS_CEILING`, a module constant beside `BURST_SAME_ACCOUNT_MAX_RETRIES`. **Its value exceeds the largest supported pool.** A ceiling below the fleet size is not a runaway fence — it is the fixed attempt cap this work removes, wearing another name.
- `failover_outcome` / `PoolWalkBound`: the walk's terminating bound is now a value rather than five different paths that all return `"surface"`. The reorder above is unobservable without it.
- New `app/modules/proxy/pool_terminal.py`: `resolve_pool_terminal_failure` asks `probe_pool_usage_exhaustion` at most once and renders either the canonical `usage_limit_reached` 429 or the preserved last per-account failure, verbatim. It skips the probe for the two bounds whose answer the pool's state cannot give — a `non_retryable` failure surfaces as itself, an exhausted budget yields `upstream_request_timeout`.

## Verification

Measured on a fresh worktree at this head, not asserted:

| check | result |
|---|---|
| `pytest test_failover_foundation test_pool_terminal test_proxy_utils test_websocket_transport_fallback` | **1568 passed** |
| `pytest tests/unit` vs the same command on the base | **0 new failures** (exact set equality) |
| `ruff check` / `ruff format --check` | clean |
| `check_proxy_architecture.py` | passed — `service.py` 2600/2600, `streaming/mixin.py` 1097/1100, `http_bridge/mixin.py` 2435/2436, all untouched |
| `check_settings_tiers.py` | `settings fields: 96/96; ok` |
| `git diff base...head -- openspec/` | empty |

Two rounds of adversarial review ran against this branch. The second round's reviewers injected off-by-one errors at all four converted call sites to check the suites could see them, and the test file's `api_key` forwarding was strengthened after the first round showed every case passing `None` — so the assertion that the probe receives the request's own eligibility reduced to `None == None`.

## Simplicity gates

No new settings (`[settings_fields]` stays 96/96 — verified by the gate script, not by reading). No README section, no `.env.example` entry, no dashboard nav item. Not dashboard-visible.

## Known follow-up, not in this PR

`resolve_pool_terminal_failure` guarantees the probe is asked once per *call*; "once per request" needs the request-scoped memo that lands with the loop conversion, since nothing calls this yet.


Soju06:
Reviewed against `origin/main` and against the spec this branch is stacked on. The predicate refactor is clean — I checked all four converted call sites (`compact.py:2077`, `retry.py:2633`, `retry.py:3344`, `websocket/mixin.py:4654`) and each new expression is exactly the old one (`max - attempt - 1 > 0` ⇔ `attempt < max - 1`; `max_attempts - attempt <= 0` ⇔ `not (attempt < max_attempts)`), and hoisting the `non_retryable` check above the candidate check moves no action for any input. The `deadline` triple (502 / `upstream_request_timeout` / "Proxy request budget exhausted") really is what `_is_proxy_budget_exhausted_error` matches at `app/modules/proxy/service.py:2455-2461`. Four things block it as written.

**1. `pool_terminal.py:54` imports a module that no longer exists on `main`.**

```python
from app.modules.proxy._load_balancer.exhaustion_probe import AdmissionProbeService, probe_pool_usage_exhaustion
```

`app/modules/proxy/_load_balancer/exhaustion_probe.py` was deleted by merged commit `982125a33` — *revert(overflow): remove the subscription-exhaustion model-source fallback (#2416)* — along with the `pool_exhaustion_probe_declined_total` counter the probe increments. This branch's merge-base with main is `837964b91` (#2411), which predates that revert, and the PR targets `feat/pool-walk-anchored-relocation`, whose tree still carries the file — which is the only reason CI is green. The moment the stack rebases onto current main, `pool_terminal.py` raises `ModuleNotFoundError` and `tests/unit/test_pool_terminal.py:27` errors at collection. Open #2422 (drop the withdrawn subscription-overflow schema) shows the retirement is still ongoing, so this will not resolve itself. The probe needs a new home (it is a ~119-line read-only module; re-landing it under `app/modules/proxy/` as part of *this* change, with its declined-counter either restored or dropped, is probably the smallest path) before any of this can reach main.

**2. `logic.py:1494` — the ceiling is fixed, the spec now says it must scale.**

`a6bb9b590` ("docs(openspec): reconcile pool-walk terminal semantics", 09-13) rewrote `account-routing/spec.md` L9 to require *"a runaway ceiling **derived from the current candidate count**, so it scales with any valid pool size and cannot become the ordinary bound."* `MAX_ACCOUNT_ATTEMPTS_CEILING = 128` is a module constant; on a deployment with more than 128 accounts it becomes exactly the fixed attempt cap the constant's own comment says it must never be. `tests/unit/test_failover_foundation.py:671` asserts only `> LARGEST_DEPLOYED_ACCOUNT_POOL` (28), which encodes the earlier `60a79e5b4` wording that the parent branch has since replaced. Either make the fence a function of the candidate count at walk start (e.g. `runaway_attempt_ceiling(candidate_count)` returning something like `2 * candidate_count + 8`) and pin *that* in the test, or amend the parent spec back — but code and spec cannot disagree here.

**3. `pool_terminal.py:108` — the terminal accepts no walk evidence.** (codex's P2 on L163 is correct.)

`933ff50af` added the requirement *"A walk proves exhaustion from its own attempts"*: *"The walk therefore MUST carry its own per-account evidence to the terminal decision"*, with the explicit note that the health write persists status but not the usage sample, so the persisted-state probe cannot see it. `resolve_pool_terminal_failure` has no evidence parameter and falls straight through to `last_account_failure` on `exhaustion is None` (L163). Concretely: accounts A, B and C each reject with coded `usage_limit_reached`; the probe's `check_opportunistic_admission(observe_only=True)` does not yet answer `usage_limit_reached`; the client gets account C's verbatim upstream 429 body — which the spec's own scenario forbids (*"AND the client does not receive the first rejecting account's verbatim upstream body"*). Add a `walk_evidence` argument (the per-account exclusion reasons the walk accumulated) and treat "every attempted candidate excluded on usage-window exhaustion evidence" as exhaustion, gated as the spec scopes it: coded `usage_limit_reached`, message-derived usage limit, or quota evidence with the same structured meaning — not `usage_not_included` / `insufficient_quota`.

**4. `pool_terminal.py:72` — no retry hint when `resets_at` is unknown.** (codex's second P2 is also correct.)

`account-routing/spec.md` L93: *"when no authoritative reset timestamp is available for `error.resets_at`, the response MUST carry a retry hint instead."* `PoolUsageLimit` is `(status_code, payload, resets_at: int | None)` with no hint field, and the module docstring at L27 asserts the opposite as intent ("never a `Retry-After` header"). With `selection.resets_at=None`, `selection_failure_response` (`app/modules/proxy/selection_errors.py:30-38`) emits a 429 with neither `error.resets_at` nor `Retry-After`. Either add a bounded fallback hint to `PoolUsageLimit` for the `resets_at is None` case, or take the requirement back out of the parent spec — the docstring currently argues against a MUST that is on the branch it is stacked on.

**Process note.** `git grep 'MAX_ACCOUNT_ATTEMPTS_CEILING\|failover_outcome\|pool_terminal' -- app/` finds only the definitions; every other hit is under `tests/` or `openspec/`. That is by design per the PR body, but it means none of (2)–(4) can be caught by a product-path test until the loop conversion lands, and (1) is invisible until the rebase. Given the four spec amendments that postdate commit `c65057095` (09-11 08:34), it is probably cheaper to rebase this onto the current parent head, fix all four points in one pass, and re-run the adversarial round against the *current* spec text than to land the library as-is. Also note #2403 is a sibling on the same base and also edits `tests/unit/test_failover_foundation.py` and `tests/unit/test_proxy_utils.py` — whichever goes first, the other needs a rebase.

Komzpa:
@codex review

PR #2392, opened 2026-09-11
feat(api-keys): add per-key account usage shares

### LINKED ISSUE #631: Account limit restriction
Use case: a person wants to share their account, but only 50% of the limit. This would work to maybe keep a bit of cloud usage leeway.

BridgetWorker:
## Summary

Adds an optional per-API-key local account usage share for account-scoped keys. This lets two keys assigned to the same upstream account split its locally attributable usage, for example with `50%` configured on each key.

The policy is distinct from the open per-account operator-cap work in #631: it limits only the configured key, while other keys can continue using the account.

## Implementation

- Adds nullable `account_usage_percent` (1-100) to API-key persistence, API, dashboard forms, and localized UI.
- Enforces the share during central account selection for assigned-account keys.
- Uses successful request-log costs for the key and account, projecting a comparable local USD budget from the live secondary usage snapshot rather than mixing USD with abstract plan credits.
- Preserves existing behavior when usage attribution, capacity, snapshot, or local aggregation is unavailable.
- Adds migration, OpenSpec requirements, API/service/routing/migration tests, and frontend form tests.
- Documents eventual consistency because request-log writes settle asynchronously; this is not a cross-request atomic reservation.

## Test plan

- [x] 227 focused Python tests passed
- [x] 30 API-key frontend tests passed
- [x] TypeScript build passed
- [x] Ruff lint and format checks passed
- [x] Python syntax compilation passed
- [x] Migration tests passed
- [x] `git diff --check` passed

Fixes #631


BridgetWorker:
Closing this PR because the contribution should target Cloud-121/codex-lb rather than the upstream repository.

PR #2396, opened 2026-09-12
Surface upstream overload to native Codex clients as a retryable named error

gabrielbryk:
## Summary

Native Codex clients hit a give-up path where codex-lb re-raised an upstream capacity/overload error *after* the 200 had already been sent, closing the SSE/WS stream with zero bytes ("stream closed before response.completed"). Codex retried blind. The cause was also dropped at the denied-anchor fail-closed rewrite, so it never reached the give-up path in the first place.

## Type of change

## OpenSpec

Change directory: `openspec/specs/responses-api-compat/spec.md` ("Native Codex preserves upstream failure lifecycle")

## Changes

- Emit one terminal `response.failed` (`error.code = rate_limit_exceeded`, message "Please try again in `<N>`s.") at both native-client give-up gates in `app/modules/proxy/api.py` (`_stream_response_error_events` / `_normalize_public_responses_stream`), instead of re-raising after the stream opened.
- Relabel `server_is_overloaded` / `slow_down` WS terminal frames the same way in `app/modules/proxy/_service/websocket/helpers.py` (`_sanitize_websocket_terminal_error_fields`).
- `app/modules/proxy/_service/http_bridge/request_submit.py` now preserves the upstream cause on the denied-anchor fail-closed raise instead of dropping it.
- `rate_limit_exceeded` (not `server_is_overloaded`/`slow_down`) because codex-rs treats the overload codes as terminal with zero retries and only parses a retry delay from `rate_limit_exceeded` messages (`codex-rs/codex-api/src/sse/responses.rs`).
- Non-native/OpenAI-compatible clients are unaffected.

## Simplicity

- No new settings added.

## Test plan

```
uv run pytest tests/unit/test_proxy_utils.py -q
uv run pytest tests/unit/test_proxy_http_bridge.py -q
```

New/updated tests:
- `test_native_codex_stream_emits_retryable_terminal_event_instead_of_reraising`
- `test_native_codex_stream_error_events_default_retry_delay_when_unknown`
- `test_native_codex_stream_error_events_unaffected_for_non_giveup_codes`
- `test_sanitize_websocket_terminal_error_relabels_overload_codes_as_retryable`
- `test_submit_rejects_a_denied_proxy_anchor_before_upstream_dispatch` (extended to assert the preserved `upstream_error_code`)

Full suite: 2461 unit tests green. `ruff check` / `ty` clean.

## Side effect

Denied-anchor failures in `request_submit.py` now record `upstream_error_code=previous_response_not_found` in RequestLog telemetry (previously `stream_incomplete`).

https://claude.ai/code/session_01EAT21bAfhFwxeYHbK5Wr6M

### DIFF
--- a/app/core/errors.py
+++ b/app/core/errors.py
@@ -119,6 +119,32 @@ SYNTHETIC_TRANSPORT_FAILURE_MARKER = "_codex_lb_synthetic_transport_failure"
 SYNTHETIC_TRANSPORT_FAILURE_CODES = frozenset(
     {"stream_incomplete", "stream_idle_timeout", "upstream_request_timeout", "upstream_unavailable"}
 )
+# The Codex CLI treats ``server_is_overloaded``/``slow_down`` as terminal
+# (zero retries), so once codex-lb gives up retrying internally for a
+# native Codex client, the give-up path must relabel the cause as the one
+# code Codex both retries on *and* parses a delay out of.
+NATIVE_GIVEUP_RETRYABLE_CODE = "rate_limit_exceeded"
+DEFAULT_NATIVE_GIVEUP_RETRY_AFTER_SECONDS = 5
+
+
+def native_giveup_retryable_message(
+    upstream_error_code: str | None,
+    message: str | None,
+    retry_after_seconds: float | int | None,
+) -> str:
+    """Build the ``rate_limit_exceeded`` message codex-rs's reconnect parses.
+
+    codex-rs matches ``(?i)try again in\\s*(\\d+(?:\\.\\d+)?)\\s*(s|ms|seconds?)``
+    against this message to size its reconnect backoff, so the trailing
+    ``Please try again in <N>s.`` clause is load-bearing and must stay in
+    that exact shape.
+    """
+    delay = retry_after_seconds if retry_after_seconds and retry_after_seconds > 0 else None
+    seconds = delay if delay is not None else DEFAULT_NATIVE_GIVEUP_RETRY_AFTER_SECONDS
+    seconds_text = str(int(seconds)) if float(seconds).is_integer() else str(seconds)
+    original = (message or "Upstream error").strip().rstrip(".")
+    code_text = upstream_error_code or "unavailable"
+    return f"codex-lb: upstream {code_text}; {original}. Please try again in {seconds_text}s."
 
 
 def openai_error(
--- a/app/modules/proxy/_service/http_bridge/request_submit.py
+++ b/app/modules/proxy/_service/http_bridge/request_submit.py
@@ -2068,6 +2068,14 @@ class _HTTPBridgeRequestSubmitMixin:
                                 "stream_incomplete",
                                 "The previous response anchor was rejected upstream; retry the request.",
                             ),
+                            # Fail-closed rewrite: preserve the upstream cause
+                            # just recorded above instead of dropping it. The
+                            # native give-up path (api.py
+                            # ``_stream_response_error_events``) reads
+                            # ``upstream_error_code``/``retry_after_seconds``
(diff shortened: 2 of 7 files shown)


Soju06:
Triage review, read-only, verified against `origin/main` @ `ca243dec4`. The motivation is real — a native Codex client that receives the 200 and then zero bytes retries blind — but this cannot land in its current shape. Four concrete things we checked in the code:

**1. The `request_submit.py` hunk is superseded.** #2387 already landed this case, by a different mechanism: on main, `app/modules/proxy/_service/http_bridge/request_submit.py:2132` raises the denied-anchor refusal with `local_pre_dispatch_refusal=True`, and `app/modules/proxy/api.py:8249` (`and not local_refusal`) plus `api.py:8285` consume that flag so the client now gets a real terminal `response.failed` instead of an empty 200 (issue #2364). Adding `upstream_error_code="previous_response_not_found"` to that same raise is both a textual conflict and no longer the mechanism this path uses. Please drop the hunk.

**2. The WebSocket relabel is not scoped to native clients.** `_sanitize_websocket_terminal_error_fields` (`app/modules/proxy/_service/websocket/helpers.py:1764`) has no native-mode predicate, and both call sites — `websocket/mixin.py:7013` (terminal cleanup, iterating every remaining pending request) and `websocket/mixin.py:7220` (`_emit_websocket_terminal_error`) — invoke it for every client kind. As written, an OpenAI-SDK/compatible WS client that sees `server_is_overloaded` today would start seeing `rate_limit_exceeded`. CodeRabbit flagged this and it reproduces by reading the call sites; the branch needs the same native guard the HTTP path uses.

**3. `rate_limit_exceeded` is a load-bearing quota classifier inside codex-lb, not a free label.** It is a member of `_ACCOUNT_RECOVERY_RETRY_CODES` (`app/modules/proxy/service.py:829`), `_WEBSOCKET_TRANSPARENT_REPLAY_ERROR_CODES` (`service.py:875`), `_MODEL_CAPACITY_LIMIT_CODES` (`_service/streaming/helpers.py:513`), `RATE_LIMIT_CODES` (`app/modules/request_logs/mappers.py:16`) and `_QUOTA_ERROR_CODES` (`app/modules/limit_warmup/service.py:36`). Emitting it for `stream_idle_timeout` / `upstream_request_timeout` / `upstream_unavailable` means every consumer that classifies on the code — including a codex-lb instance chained behind another codex-lb — reads a transport blip as a quota event and takes the account out of rotation. Please either pick a code that is not already a quota signal, or state in the proposal what happens to those classifiers.

**4. The OpenSpec artifact is missing.** This edits `openspec/specs/responses-api-compat/spec.md` in place with no `openspec/changes/<slug>/` folder. `.github/scripts/validate_changed_openspec.py:35` only strict-validates active change folders, so nothing validated this text, and AGENTS.md makes the change folder a hard gate for behavior changes. Compare #2423, which ships `openspec/changes/reject-a-dead-client-anchor/{proposal,tasks,context}.md` plus a spec delta for an adjacent fix in the same module. The new paragraph also contradicts what it leaves standing: both gates you edit are post-exhaustion by construction, so "Native Codex preserves upstream failure lifecycle" (`spec.md:7161`) and its scenario (`:7170`) become dead letters in the same file. If the intent is to retire that requirement, write it as a `MODIFIED`/`REMOVED` delta so the owner can decide explicitly — reversing a deliberate codex-faithful contract is a policy call, not a `fix:`.

Suggested path: rebase onto main, drop hunk 1, scope hunk 2 to native clients, and open an `openspec/changes/` proposal that states the code choice and the classifier impact. Worth coordinating with #2423, which is landing a 404 answer for a proven-dead client anchor in the same area.

PR #2408, opened 2026-09-12
fix(balancer): retain warm soft sticky owners through overload isolation

Soju06:
## Why

Releasing a soft sticky owner during overload isolation is a **rebind**, not a one-time move, and nothing ever returns the thread when isolation lifts — so one conversation accumulates one permanent new account per isolation episode it touches.

Measured on this deployment's `request_logs`, for conversations with ≥3 requests, the mean distinct accounts per conversation was **2.29** on a healthy day (09-09, 1,929 conversations), **3.45** during the 09-10 upstream incident (2,674), and **1.02** when upstream was quiet (09-11, 326). The factor tracks upstream pressure, and the rebind is what makes each episode's cost permanent.

The rebind buys nothing: the request goes to a healthy sibling either way, only the bookkeeping differs. And the usual justification — "the sibling is cold, commit to it" — does not hold here. Direct upstream probing (3 runs, 26 calls, ~28k-token prefixes with a fresh nonce) showed content seeded by one account producing near-full cache hits on four *other* accounts (28,032/28,168 · 27,136/27,328 · 28,416/28,588). The prefix cache is node-local and not partitioned per account, so a thread served by a sibling can still hit its owner's warm prefix; keeping the mapping reclaims that locality when the owner recovers.

## What changes

- **The isolation release becomes request-local.** The turn still goes to an overload-free sibling — availability is unchanged — but the sticky row is not deleted or rebound, so the thread returns to its warm owner on the first turn after isolation lifts. This is the rule cap spillover and retry-exclusion pressure already apply; isolation stops being the exception.
- **A retained mapping does not die of silence.** The default affinity TTL and the default isolation window are both 1800 s, so writing nothing let a retained row expire mid-episode and the next turn persisted the substitute as a new owner. Each retained turn now rewrites the mapping onto its own owner — from the pinned path *and* from the path where a cap or this request's exclusion list removed the owner before selection ran.
- **The substitute is stable across turns**, which is the correctness condition: the replacement is re-picked every turn, so a weighted draw would bounce the thread across siblings, worse than the single rebind it replaces. The thread's identity becomes a seed the selector spends *inside itself* (`select_account(selection_seed=...)`): the pick is made by keyed hash among the candidates the configured strategy was about to draw from, after every filter and narrowing it applies — budget tier, health tier, routing policy, quota, cooldown, error backoff, planner cost, reset preference, positive draw weight. Stability therefore cannot buy eligibility.
- **Drain strategies and `fill_first` keep their own answer**, and a seeded pick skips recovery-probe admission, whose whole purpose is to move between probing accounts.
- **Explicit reallocation still wins**, and retention is bounded by recoverability and by *why* the mapping was preserved: an ambiguous conversation owner may be out of the request's scope and is never retained, refreshed, or given a seeded substitute.
- **`sticky_owner_overload_isolation_reroute` gains `mapping=retained|rebound`**, covering both halves of what isolation does to a conversation, emitted only once a replacement is actually selected. No account identifiers.

No new setting, no schema change, no migration; the settings ratchet is untouched.

## Expected effect

This removes the *accumulation*, not the *episode*: a thread touching N episodes should converge from `1 + N` accounts toward ~2 (owner + its substitute). Against the incident-day 3.45 that is the dominant term; against the healthy-day 2.29 it is smaller, since most of that is ordinary failover. It does not target the quiet-upstream case, already at 1.02.

## Verification

- `openspec validate retain-warm-sticky-owners-through-isolation --strict`
- `ruff check` · `ruff format --check` · `ty check` · `make architecture-check`
- Full unit suite: **10,353 passed**, 4 skipped
- `codex review --base origin/main` iterated to a clean pass. Every finding it raised was reproduced before being fixed, and each fix carries a regression test verified to fail against the pre-fix code: budget filtering on all three usage axes, per-strategy narrowing (planner cost, reset bucket, zero draw weight), top-k tie-break stability across admitted turns, the exhausted-pool fallback, recovery-probe rotation (through the public `LoadBalancer.select_account` path), TTL freshness on both retention paths, ambiguous-owner scope, and both diagnostic gaps.

One finding was *not* applied: a status bound on the cap-preservation arm for paused/deactivated owners. `_selectable_accounts` removes both before states are built, so such an owner cannot reach that predicate; the behavior it asked for is nonetheless pinned end-to-end by `test_capped_prompt_cache_owner_that_is_gone_is_rebound`.

## Known gaps (recorded in the spec)

- Substitute convergence across replicas is best-effort: each replica has its own overload view, so the same thread can hold a different substitute per replica. This caps fan-out at one substitute per `(replica, pool)` rather than eliminating it.
- Validation cannot be a day-over-day mean — 09-11 already reads 1.02 with no code change. It needs an isolation-event-scoped metric or a held-back cohort.
- 33–38 % of requests carry neither `conversation_id` nor `session_id` and are invisible to this metric; #2347 owns that traffic.


Komzpa:
@codex review

Komzpa:
@codex review

Komzpa:
@codex review

Soju06:
@komzpa — thanks, I merged `ad94886` rather than force-pushing over it. Two of your changes I took as-is and one I reconciled; recording the reasoning so it is reviewable rather than silent:

**Taken as-is**

- `caller_requested_reallocation = reallocate_sticky and sticky_kind != STICKY_THREAD`. You are right and I had this wrong: `affinity.py:805` sets the flag on *every* sticky-thread policy, so it carries no per-request intent, and reading it as "the caller asked to retire this mapping" exempted a whole kind from retention while the two either side of it stopped accumulating. I had recorded it as a documented limitation; your reading is better. I added a comment at the site saying why the exception exists, since the comment above it still promised the opposite.
- The `persist_fallback` form of the rebound-diagnostic guard, which is more precise than the `preserve_existing_mapping_on_fallback` check I had written for the same case.

**Reconciled: the relative-availability floor**

`ad94886` disables `RELATIVE_AVAILABILITY_MIN_WEIGHT_FRACTION` for seeded callers; I kept it and dropped the top-`k` *rank* cut instead, so the seeded membership rule is:

- rank cut: **gone** for a seeded caller — it reorders whenever any account's usage refreshes, which moved a retained thread off a substitute that never became ineligible (100/100 threads in `test_a_seeded_pick_does_not_follow_a_reshuffled_ranking`);
- fraction floor: **kept** — it is an eligibility bound rather than a ranking, the same kind as the budget threshold a seeded pick already cannot cross, and without it a retained conversation can be parked on an account at ~10 % of the best remaining availability until it hard-fails.

That makes `test_relative_availability_seeded_top_k_spreads_and_survives_live_weight_changes` degrade its substitute to 40 % used rather than 90 %: still a live weight change the pick must survive, but inside the floor. Both of its assertions (spread across seeds, stability for one seed) are unchanged. If you would rather a retained thread hold its substitute all the way down, say so and I will drop the floor instead — it is a one-line change either way, and the disagreement is only about where "no longer eligible" sits.

Full unit suite on the merged tree: 10,363 passed, 4 skipped.

Soju06:
Addressed in `eb5ed1d` (the open thread on the RATE_LIMITED carve-out predates it). The off-pool block no longer copies the pinned branch's `status != RATE_LIMITED` clause: that exclusion exists on the pinned path because such an owner takes the grace-retry path instead of budget reallocation, and an owner that never reached selection has no grace path to take. `test_off_pool_budget_pressured_isolated_owner_uses_the_secondary_budget_filter` is now parametrized over ACTIVE and RATE_LIMITED and runs twelve sticky keys, since a single key can land on the safe sibling by luck through the seeded pick — it fails on the `rate_limited` case against the pre-fix code.

One related gap stays open and is recorded in the proposal: a *retry-excluded* isolated owner is removed from `selection_inputs.accounts` before states are built (`load_balancer.py:693`), so no `AccountState` exists for it anywhere downstream and its secondary-window pressure cannot be read at all. The cap sub-case gets the filter; the exclusion sub-case keeps the primary-budget filter only.

PR #2415, opened 2026-09-13
fix(overflow): validate the request fields a portable verdict forwards

Soju06:
## Summary

The overflow portability view closes the set of top-level field **names** that may reach a third-party model source. Nothing closed the **values**, so six fields rode out to that provider unread on a `portable` verdict. This PR closes the value side: a classification table that covers the view allowlist exactly, a field-value step in the verdict, and an overflow-only egress that rewrites what must not leave as the client wrote it.

Scope note, stated up front: **this PR relaxes no existing rejection.** The false *rejections* the same sweep found (`tool_search.execution`/`parameters`, the real `web_search` declaration fields, input-item ids) are a separate owner decision and are untouched.

## Type of change

Linked issue: Part of #2123

## OpenSpec

- [x] This PR touches a codex-faithful path (request shape) and preserves upstream-equivalent behavior — on **direct source routing**, which is untouched. The overflow direction deliberately diverges, and the divergence is what the new requirement records.

Change directory: `openspec/changes/validate-overflow-forwarded-request-fields/`

## The five leaks, with the reproduction for each

All five were reproduced on a body the predicate called **PORTABLE**, in the 2026-09-13 capture sweep of 60 real `codex-cli 0.154.0` bodies through the exact fixtures-gate path (`REPORT.md` §6).

| # | Leak | Reproduction |
|---|---|---|
| 1 | **`include` never validated** | `include: ["reasoning.encrypted_content"]` — on **all 60** captured bodies — yielded PORTABLE. So did `["file_search_call.results", "message.input_image.image_url"]`, `["totally.made.up"]`, and `[{"nested": "object"}]` (a non-string element did not fail closed). `include` was in `OVERFLOW_VIEW_FIELDS` *and* in `_RESPONSES_PAYLOAD_FIELDS_WITH_DEDICATED_VALIDATION`, but `responses_payload_is_account_neutral_fresh_replay` never read it. |
| 2 | **`prompt_cache_key` forwarded verbatim** | Real value is a Codex UUIDv7 on all 60 bodies; `prompt_cache_key: "acct_someone-elses-namespace"` also yielded PORTABLE. `strip_source_telemetry`'s own docstring said "forwarded untouched". |
| 3 | **`metadata`, `user`, `safety_identifier`, `prompt_cache_retention` never validated** | `user: "operator@example.com"` → PORTABLE. `safety_identifier: "operator@example.com"` → PORTABLE. `metadata: {"chatgpt_account_id": "org-123"}` → PORTABLE. `prompt_cache_retention: "24h"` → PORTABLE. `_PORTABILITY_VIEW_ONLY_FIELDS` described the group as "provider-neutral generation knobs with no account-scoped state", which is false for two end-user identifiers. |
| 4 | **Operator filesystem paths reach the third party** | In 0.154.0 the environment context left `instructions`: `input[0]` is a developer message carrying `<skills_instructions>` with a skill-roots table of absolute paths under the real `CODEX_HOME`, then `<permissions instructions>` with `sandbox_mode`, the network restriction and the approval policy; `input[1]` opens `# AGENTS.md instructions for <absolute workspace path>`. `ResponsesRequest`'s before-validator folds that developer message back into `instructions` — checked on the committed capture, where `input[0]` disappears and `instructions` grows from 24 to 98 characters. |
| 5 | **`store: true` passes** | An otherwise portable body with `store: true` → PORTABLE. Codex always sends `store: false`, so this is latent. |

**Honesty about leak 1's production reach.** The request model already closes the `include` vocabulary to seven values and to `list[str]`, so `totally.made.up` and `{"nested": "object"}` are 400s before the overflow path — the sweep reached them by injecting into the view. What was live in production is the seven-value vocabulary itself: `reasoning.encrypted_content` on every Codex request, plus three values naming hosted-tool artifacts this module rejects everywhere else. The rules for the unknown/non-string cases are fail-closed backstops for a hand-built view, and `test_every_include_value_the_request_model_admits_has_a_rule_here` pins the two closures against each other so an eighth value cannot arrive without a rule.

## The rule chosen per leak, and why

Argued from the evidence, not from symmetry — and they differ.

**1. `include` → neutralise one value, decline the rest.**
`reasoning.encrypted_content` is **removed on the egress**, not declined. Declining on it would make overflow unreachable for 100% of real Codex traffic. Removing it is free and, more than that, *correct*: `include` requests extra **response** artifacts, so removing an entry can only subtract from what comes back and never changes the generation — and the artifact it asks the destination to mint is the one `_HISTORY_ITEM_TYPES` and `_mapping_has_account_scoped_reference` reject in every other position. Not asking for it reaches the state the predicate already requires: no encrypted reasoning minted at the source, so the client's **next** turn stays portable instead of being poisoned by a blob the destination's key encrypted.
Every other value is a closed allowlist with a single principle: an `include` is portable exactly when the item class it names is one a portable body may itself carry. Portable: `message.input_image.image_url` (an echo of our own input), `message.output_text.logprobs` (a generation detail), `web_search_call.action.sources` (`web_search` is already in the portable declarable set). Declined: `code_interpreter_call.outputs`, `computer_call_output.output.image_url`, `file_search_call.results` — hosted-tool artifacts whose declarations *and* items are rejected everywhere else, so asking for them is incoherent: the client would echo the result into a next turn that is then unportable. Unknown strings, non-string elements and a non-list `include` fail closed.

**2. `prompt_cache_key` → neutralise (replace).**
A cache key is a **namespace at the destination**, and this repo has *measured* that prompt caches are not isolated between accounts upstream (2026-09-11 cross-account cache probe). A value one tenant chooses, replayed to a source other tenants share, is the cross-tenant cache-namespace collision class. Declining kills the feature (60/60 carry one). Dropping loses the cache. So the forwarded key becomes `codexlb-` + 32 hex, a pure function of `(requesting API key, token domain, client value)`: **stable per tenant** so the source's prefix cache still hits across a tenant's turns, **disjoint across tenants** so two tenants cannot name one namespace. As a bonus the client's session-shaped UUIDv7 no longer leaves.

**3a. `user` / `safety_identifier` → neutralise (replace), same token.**
These are end-user identifiers by the OpenAI specification — in practice emails. They are *not* "provider-neutral generation knobs"; that comment is corrected. Dropping them would cost the source its per-end-user abuse tooling; forwarding them leaks an operator identity to a third party. A stable opaque token is precisely what the vendor documents sending instead, so both fields get one, in a shared `end_user` token domain (so the same end user is one subject across both fields) and a domain separate from the cache key (so a cache key equal to a user id does not alias it).

**3b. `metadata` → declare safe, with a shape rule.**
Client-authored bookkeeping echoed back on the response object, and **strictly less exposing than the transcript it accompanies** — dropping it would silently break an echo an SDK client relies on, and declining it would refuse overflow for a field that cannot move the answer. What it may not be is a *reference*: it must be the API's own `Map<string, string>` and must name none of the account-scoped reference keys this module already keys on. `metadata: {"chatgpt_account_id": "org-123"}` stays portable on purpose — it is an opaque label the client chose, not a handle this proxy can resolve.

**3c. `prompt_cache_retention` → declare safe, closed enum** (`in_memory`, `24h`). The client's own retention choice about its own data at a source the operator designated, in the same class as `store`.

**4. `instructions` (the operator's paths) → declare, out of scope, recorded.**
The predicate is a check for source-side **state**, not a privacy check, and that scope is defensible: the same absolute paths already ride in the transcript's `function_call_output`s and file contents, so a rule that redacts `instructions` while the paths travel in every tool output is theatre — and stripping the environment context would change how the model behaves (it would no longer know the sandbox is read-only or what the approval policy is). The honest answer is to **state** it rather than close it: `instructions` is classified `forwarded` in the table with the evidence in the comment, the corpus README says so, and the spec requirement records it. The decision that a designated source may read the operator's conversations is made once, when the operator designates it.

**5. `store` → declare safe, no new rejection.**
Not symmetry: the anchored-pin lifecycle exists **precisely** for a client that did not send `store: false` and expects a `previous_response_id` chain to resolve at the source (`anchor_requested` is `client_store_intent(payload) is not False`). Declining or forcing `store` would break a feature the design already ships, and the request model already constrains the type. So this one gets a recorded classification and no new check — and the PR says so rather than adding a rule that does nothing.

### Neutralised vs declined, at a glance

| Field | Rule | What leaves |
|---|---|---|
| `include: ["reasoning.encrypted_content"]` | **neutralised** | entry removed; `include` dropped once emptied |
| `include`, other values | **declined** (3 hosted) / forwarded (3 portable) | — |
| `prompt_cache_key` | **neutralised** | `codexlb-<32 hex>`, scoped to the API key |
| `user`, `safety_identifier` | **neutralised** | `codexlb-<32 hex>`, shared `end_user` domain |
| `metadata` | **validated**, forwarded | as sent, if a string map naming no reference |
| `prompt_cache_retention`, `truncation` | **validated**, forwarded | as sent, if in the closed set |
| `store`, `instructions`, `stream`, `parallel_tool_calls`, `max_output_tokens`, `temperature`, `top_p` | **declared** `forwarded` | as sent |
| `model` | **rewritten** by the dispatch | the source's model id |

The top-level field set is now closed on both sides: `OVERFLOW_FIELD_CLASSIFICATION` covers `OVERFLOW_VIEW_FIELDS` exactly (pinned in both directions), and an unrecognised **value** fails closed with `not_portable_unknown_field` — the same posture the view already takes for an unrecognised **name**.

## Where the neutralisation happens, and why it is provable

`overflow_source_wire_body` is the one overflow-only shaping step. Both the body the verdict classifies (`overflow._source_body`) and the body the route helper puts on the wire (`api._source_responses_response`, in the `if overflow is not None` branch that already restored `store`) go through it, so what was judged is what leaves. The verdict then **requires** the rewritten shape — `prompt_cache_key`/`user`/`safety_identifier` must be absent or a proxy-minted token, and `reasoning.encrypted_content` in `include` declines — so a bypassed rewrite fails closed instead of leaking (mutants M6/M9 below).

Direct source routing is untouched: it still forwards `include`, `prompt_cache_key`, `user` and `safety_identifier` exactly as the client sent them, and `test_the_wire_body_carries_the_neutralised_identifiers_and_not_the_clients` asserts that control in the same test.

## Corpus grounding and the drift guard

The reason this drifted is that the predicate was written against imagined shapes and gated against two hand-written bodies.

- **The gate now classifies what production classifies.** `_stripped()` ends in `neutralize_overflow_egress`, because `overflow._source_body` does. The recorded verdicts in `provenance.json` are unchanged by this (both captured bodies still decline at their configuration-class reason; the synthetic first-turn body is still portable — it is portable *because* its `include` is neutralised, which is the point).
- **Drift guard.** `test_every_field_a_fixture_carries_is_classified_somewhere_in_production` fails **by name** when a captured body carries a top-level field production neither strips before the view, nor classifies in the table. Run on the committed body *and* on the overflow egress body, so a field that only appears after normalisation is covered too.
- **Vocabulary guard.** `test_every_include_value_the_request_model_admits_has_a_rule_here` pins the portable / neutralised / declined `include` sets against `_RESPONSES_INCLUDE_ALLOWLIST`.
- **Egress proof on real bytes.** `test_the_identifiers_that_reach_the_source_are_proxy_minted_and_scoped_per_key` (integration) sends the *same* client-chosen `prompt_cache_key` and the same end-user identifier under two API keys and asserts on the JSON a stub source received: neither value appears anywhere, no `include` arrives, and the two forwarded tokens differ.

No new fixture body is committed. The two captured bodies already carry every shape these rules read (`include`, `prompt_cache_key` before `strip_source_telemetry`, the tool declarations, the bundle `id`), and the corpus policy of one HTTP pair per family is deliberate — the sweep's websocket and Tier-D bodies are evidence, referenced in the proposal, not corpus.

## Mutant table

Each mutation was planted, the named test run, and the mutation reverted.

| # | Mutation | Test that fired | Message |
|---|---|---|---|
| M1 | `include` clause deleted from `_unportable_forwarded_field` | `test_an_include_naming_a_hosted_tool_artifact_declines` (3 cases) + `test_an_unrecognised_include_fails_closed` (4 cases) — 7 failures | `AssertionError: assert PortabilityVe..., detail=None) == PortabilityVe...call.outputs')` → `portable: True != False` |
| M2 | identifier clause deleted | `test_the_clients_prompt_cache_namespace_never_reaches_a_source` | `AssertionError: assert PortabilityVe..., detail=None) == PortabilityVe...pt_cache_key')` → `portable: True != False` |
| M3 | `prompt_cache_retention` clause deleted | `test_prompt_cache_retention_is_a_closed_set[forever-False]` | `AssertionError: assert PortabilityVe..., detail=None) == PortabilityVe...he_retention')` |
| M4 | `truncation` clause deleted | `test_truncation_is_a_closed_set_in_this_module_too[middle-out-False]` | `AssertionError: assert PortabilityVe..., detail=None) == PortabilityVe...='truncation')` |
| M5 | `metadata` clause deleted | `test_metadata_is_validated_as_a_string_map_that_names_no_reference[file-id]` | `AssertionError: assert PortabilityVe..., detail=None) == PortabilityVe...il='metadata')` |
| M6 | field-value step not wired into the verdict | `test_the_request_for_encrypted_reasoning_is_removed_on_the_way_out_not_declined` | `AssertionError: assert PortabilityVe..., detail=None) == PortabilityVe...pted_content')` |
| M7 | `NEUTRALIZED_INCLUDE_VALUES` emptied (egress stops removing it) | `test_the_egress_rewrites_exactly_the_four_overflow_fields_and_nothing_else` | `AssertionError: assert ['reasoning.e...ext.logprobs'] == ['message.out...ext.logprobs']` — `At index 0 diff: 'reasoning.encrypted_content' != 'message.output_text.logprobs'` |
| M8 | identifier rewrite deleted from `neutralize_overflow_egress` | same | `AssertionError: assert (True and False)` — `False = '01a099e8-25c2-7e30-bd5d-b1e522c07985'.startswith('codexlb-')` |
| M9 | `overflow_source_wire_body` bypasses the neutralisation | `test_the_wire_body_carries_the_neutralised_identifiers_and_not_the_clients[None]` | `AssertionError: assert 'include' not in {'model': 'gpt-5.4', 'instructions': 'You are a test.', ...}` |
| M10 | egress namespace is a constant, not the API key | `test_the_identifiers_that_reach_the_source_are_proxy_minted_and_scoped_per_key` (integration) | `AssertionError: assert 'codexlb-5741...7471c9682ddfe' == 'codexlb-1ee3...fea8aa02c04fd'` |
| M11 | `"background": true` added to a committed fixture | `test_every_field_a_fixture_carries_is_classified_somewhere_in_production[gpt55_standard_first_turn.json]` | `AssertionError: gpt55_standard_first_turn.json [synthetic, slug=gpt-5.5, codex=n/a, transport=http]: nothing in production classifies ['background'] in the committed body; add it to replay_safety.OVERFLOW_FIELD_CLASSIFICATION with the rule that decides it` |
| M12 | `"metadata"` removed from the classification table | `test_the_classification_table_answers_for_exactly_the_fields_the_view_admits` | `AssertionError: assert {'conversatio... 'model', ...} == {'conversatio...etadata', ...}` — `Extra items in the right set: 'metadata'` |

Every new rejection fires on a body that passes without it (M1–M6); every neutralisation is observable on the egress payload (M7–M10); the drift guard fails naming the field (M11).

## The three doc corrections

All three were found by the sweep and are now **asserted against the committed bytes** by `test_the_declaration_extras_the_corpus_documents_are_the_ones_the_bodies_carry`, not merely described:

1. The real `tool_search` declaration carries **two** fields beyond `_STATELESS_TOOL_DECLARATION_FIELDS` — `execution` **and `parameters`** — not one. Admitting `execution` alone would not be enough. (`README.md` §Recorded verdicts item 1, `provenance.json` capture note.)
2. `web_search`'s extras are **profile-dependent**: `gpt-5.4`/`gpt-5.5` send `external_web_access` **and** `search_content_types`; the fallback surface the CLI uses for a slug it has no bundled `model_info` for sends `external_web_access` alone. Both are outside the allowlist, so the verdict is unchanged — but "two extra fields" is not universal. (Both committed captures are standard slugs, so only the two-field form is in the corpus; the profile-dependence is recorded as coming from the sweep.)
3. The `additional_tools` bundle item carries an `id` key, outside `_ACCOUNT_NEUTRAL_INPUT_ITEM_FIELDS["additional_tools"]` (`{role, tools, type}`). Not load-bearing today (the Lite decline fires first), but "drop the Lite bundle" is not sufficient on its own either.

Plus the fourth: the **README divergence table** described the pre-0.154.0 `instructions` layout (`<environment_context>` inside `instructions`). Corrected to what 0.154.0 sends, with the fold back into `instructions` and what that means for overflow spelled out next to it.

## Changes

- `app/modules/model_sources/projection.py` — `NEUTRALIZED_INCLUDE_VALUES`, `NEUTRALIZED_IDENTIFIER_FIELDS`, `overflow_opaque_value`, `neutralize_overflow_egress`; corrected the two stale "forwarded verbatim / untouched" comments to name the overflow carve-out.
- `app/modules/proxy/replay_safety.py` — `OverflowFieldClass` / `OVERFLOW_FIELD_CLASSIFICATION`, `_unportable_forwarded_field`, `_is_proxy_minted_opaque_value`, `_metadata_is_provider_neutral`; the new step 8 in `responses_payload_is_provider_portable` and its place in the documented evaluation order; corrected `_PORTABILITY_VIEW_ONLY_FIELDS`'s comment.
- `app/modules/proxy/overflow.py` — `overflow_source_wire_body` (exported), `_source_body(payload, *, namespace)`, three call sites scoped to `decision.api_key_id`.
- `app/modules/proxy/api.py` — the overflow branch calls `overflow_source_wire_body` instead of `restore_client_store` (one import swap; `test_subscription_overflow_inert` pins the route file's import list, updated).
- `tests/fixtures/codex_bodies/{README.md,provenance.json}` — the four corrections above.
- Tests: 4 unit modules + 1 integration module, listed in the test plan.

## Simplicity

No new setting, no new env var, no README section, no dashboard surface, no default changed.

## Test plan

```
# gates
ruff check                                    # All checks passed!
ruff format --check                           # 1304 files already formatted
ty check                                      # All checks passed!
scripts/check_proxy_architecture.py           # proxy architecture checks passed
scripts/check_cancellation_safety.py          # cancellation safety checks passed
scripts/check_proxy_timing_seams.py           # proxy timing seam checks passed
scripts/check_settings_tiers.py               # settings fields: 96/96; ok
scripts/check_migration_topology.py           # head: 20260912_010000...; +0 revisions on this checkout
alembic heads                                 # ['20260912_010000_drop_legacy_dashboard_credentials']  (1 head; no revision added)
openspec validate validate-overflow-forwarded-request-fields --strict   # valid

# affected modules
pytest -p no:cacheprovider -q \
  tests/unit/test_codex_body_fixtures.py tests/unit/test_codex_body_sanitizer.py \
  tests/unit/test_model_sources_projection.py tests/unit/test_overflow_drill_coverage.py \
  tests/unit/test_overflow_pin_executor.py tests/unit/test_passthrough_request_fields.py \
  tests/unit/test_proxy_websocket_overflow.py tests/unit/test_replay_safety.py \
  tests/unit/test_replay_safety_portability.py tests/unit/test_replay_safety_strip_ids.py \
  tests/unit/test_subscription_overflow_dashboard_summary.py tests/unit/test_subscription_overflow_decision.py \
  tests/unit/test_subscription_overflow_inert.py tests/unit/test_subscription_overflow_spec_delta.py \
  tests/unit/test_subscription_overflow_store_intent.py tests/unit/test_request_log_source_parity.py \
  tests/unit/test_proxy_api_responses_contract.py tests/unit/test_metrics.py
# 1092 passed

pytest -p no:cacheprovider -q tests/integration/test_subscription_overflow_routing.py
```

## Screenshots / output

N/A — no dashboard-visible surface. The proxy-behaviour evidence is the egress payload itself, and it is asserted rather than pasted: `test_the_identifiers_that_reach_the_source_are_proxy_minted_and_scoped_per_key` reads the JSON body a stub OpenAI-compatible source received and asserts the client's cache key, its end-user identifier and its `include` are absent, and that two API keys get two different tokens.

## Known limitations

- **The pinned/anchored dispatch still forwards a body the verdict never classified.** By design (design §7.2 P16: "a pin overrides body portability except for what the source model cannot see"), so a pinned conversation can carry a top-level field outside `OVERFLOW_VIEW_FIELDS` to the source. The egress neutralisation *does* run for pinned dispatches, so the four rewritten fields are covered there too; the classification table is not consulted. The pin was established by a fresh dispatch that did pass the check.
- **Two of the new rules are fail-closed backstops, not active declines.** The identifier rule can only fire if the egress rewrite is bypassed, and `truncation` is already closed by the request model. They are load-bearing (M2, M4, M9) but they will not fire in production, and this PR says so rather than presenting them as protections that trigger.
- **`metadata`, `prompt_cache_retention` and the `user`/`safety_identifier` leak are latent for Codex**, which sends none of them. They are live for any SDK client on `/v1/responses`, which is why they are closed anyway.
- **The `prompt_cache_key` namespace is the API key, not the end user.** Two clients sharing one API key still share a cache namespace at the source — the same trust boundary every other per-key control in this proxy uses. A request with no API key (auth disabled) uses the empty namespace.
- **Threads pinned before this change take one prompt-cache miss.** The forwarded cache key changes shape, so the first turn after deploy will not hit the source's existing prefix cache.
- **Leak 4 is declared, not closed.** A `portable` verdict still ships the operator's `cwd`, workspace roots, skill roots and sandbox policy to the designated third-party provider, inside `instructions`. If the owner wants that closed, it is a different kind of change (a privacy projection over prompt text) and a different PR.
- **The false rejections are untouched.** `tool_search.execution`/`parameters`, the real `web_search` declaration fields and the input-item ids still decline every real Codex body — the same sweep measured 0/60 portable, needing ≥3 clauses relaxed per body. That decision stays with the owner; `openspec/.../tasks.md` §5.1 records it as open.

## Checklist

- [x] Ran the relevant gate subset locally (see Test plan). `uv run` was unavailable in this environment (root disk pressure); the venv interpreter was used directly.
- [x] If touching specs: `openspec validate validate-overflow-forwarded-request-fields --strict` passes.
- [x] Simplicity gates reviewed (P1–P6): no setting, no config, no new surface.
- [x] CHANGELOG is **not** edited by hand.


Soju06:
Closing: the owner has decided to **roll back the subscription-exhaustion overflow feature entirely** (#2123), so the path these validations protect is being removed.

The evidence behind the decision, from the 60-body capture sweep:

- **0 of 60** real `codex-cli 0.154.0` bodies are portable, across both transports, 7 model slugs and 5 prompt classes.
- **No clause is ever the sole blocker.** Every body needs at least three independent clauses relaxed at once; the single fix the corpus docs singled out (admitting `execution` on `tool_search`) portable-ises none.
- WebSocket traffic carries `previous_response_id` on every turn frame, so it is structurally non-portable regardless of what is relaxed.

Making native Codex traffic overflowable would mean translating request bodies per provider — rebuilding what LiteLLM does — which is out of scope for this repository.

The three leaks this PR fixed do not survive the feature: they exist only on the overflow egress path. The one with a live analogue, `prompt_cache_key` namespacing, is already handled for the subscription path by `thread_cache_identity` (#2376), so nothing is left behind.

The findings are preserved in the sweep report and in this PR's history. The capture tooling that produced them is being kept.

PR #2423, opened 2026-09-14
fix(proxy): reject a dead client anchor instead of asking for a retry

Soju06:
## Why

The three changes in `1.25.0-beta.8` removed ~99% of `previous_response_owner_unavailable` from production. Measured on the live deployment: **10 occurrences in the 20 hours after the cutover, 11 in 42 hours**, against roughly a thousand for an equivalent window before.

The attribution that #2384 added says what the remainder is, and it is **not** what the plan predicted:

| `owner_unavailable_replay_rejected` reason | count / 20h |
|---|---|
| `account_scoped_input` | 9 |
| `payload_not_full_resend` | 1 |
| `no_durable_lookup` | **0** |

Every one of the nine carries `event=fresh_reattach_full_resend_preserved` one line earlier: the client sent a full resend **with its own `previous_response_id`**, and the projected body still holds account-scoped state. `retire_continuity_owner_if_unavailable` deliberately skips that shape, because the anchor names upstream state the proxy may not silently discard — so the turn fails, correctly.

It fails as 502 `previous_response_owner_unavailable` though, which says *retry later* about an account that is paused and not coming back. The client believes it: production shows one thread retrying **nine times in five seconds**.

## What changes

When the bridge fails on an unavailable continuity owner, the anchor came from the client, and that owner cannot return before the request's own budget expires, the proxy retires the owner and answers **404 `bridge_previous_response_not_found`** instead of the retryable 502.

- **Why that code, not `previous_response_not_found`.** The API layer masks the bare code into a retryable `stream_incomplete` on purpose (`api.py::_mask_previous_response_not_found_error`): most stale anchors here are the proxy's *own* injected bookkeeping, and surfacing them as an invalid `previous_response_id` would blame the client for the proxy's state. The prefixed code is this module's existing public answer for a proven-dead anchor it will not retry — `responses-api-compat` already carries a scenario for it — and it reaches the client unmasked, telling it to resend the history or start a new conversation. This reuses that contract rather than punching a hole in the masking rule.

- **The retirement is what closes the loop.** The rejection alone would be a politer dead end. Retiring the owner means the client's natural recovery — an anchor-free resend — lands on a healthy account through the machinery #2397 already shipped.

- **The retirement call now answers "is this owner retired now".** It previously answered "did *this call* retire it", so the loser of a race between duplicate requests fell back to the retryable failure the winner had just replaced — and this failure mode arrives as a rapid retry burst, so the race is the normal case, not a corner. This also fixes the same race on #2397's injected-anchor path. Found by `codex review`.

- **Deliberately not marked as a pre-submit failure.** That provenance is what admits a failure to the raw-HTTP replay. The first version of this change set it, which sent the turn to the plain upstream, where it failed with `stream_incomplete` and hid the rejection entirely — the end-to-end test caught it.

Same horizon rule as the injected-anchor path: an owner returning inside the budget is waited for. One retirement attempt per request, shared between both paths. No schema change, no new setting, no ceiling-guarded file grows.

## Not in this change

- **Making an account-scoped body portable.** Encrypted reasoning and uploaded files are bound to their account. This change stops pretending otherwise; it does not move them.
- **The `no_durable_lookup` case.** Production reports it zero times, so the planned work to decouple the replay gate from the durable lookup is dropped as unjustified.

## Test plan

| Gate | Result |
|---|---|
| `make lint` / `make typecheck` | pass |
| `make test-unit` | **10,838 passed**, 6 skipped, 1 xfailed |
| `make test-integration-bridge` | **351 passed** |
| `openspec validate reject-a-dead-client-anchor --strict` | valid |
| `codex review --base origin/main` | 1 finding, fixed |

After the race fix I re-ran the directly affected suites — `test_durable_bridge_owner_retirement`, `test_durable_bridge_sessions`, `test_sticky_session_cleanup_scheduler`, `test_repositories` (169 passed) and the three owner-retirement end-to-end tests (3 passed). The core shards could not complete locally: the host root filesystem is at 99% and a fixture hit the faulthandler timeout, unrelated to this change. CI runs the full matrix.

New coverage:
- `tests/integration/test_http_responses_bridge.py` — a client-anchored resume on a paused owner is rejected **404 `bridge_previous_response_not_found`** with an actionable message and exactly one `dead_anchor_owner_retired`; the anchor-free resend that follows is served **200** on a healthy account. Verified load-bearing by disabling the call and watching it fail.
- `tests/unit/test_durable_bridge_owner_retirement.py` — a racing duplicate is told the owner is retired; a recovered owner and a missing session are not. Updates #2397's idempotence test, which encoded the old contract.

Partial fix for #1707.


Soju06:
The diagnosis here is right and the tracing behind it is good. I checked the premise on `origin/main` rather than taking the description: `retire_unavailable_continuity_owner` really does hard-return for a client-supplied anchor, `_http_bridge_unanchored_fork_can_spill_on_cap` is gated on `previous_response_id is None`, and `_http_bridge_account_capacity_wait_seconds` produces no wait plan for the fixed "retry later" message — so today this shape does fall straight through to a bare retryable 502, and a client answering that with a burst can never succeed. The choice of `bridge_previous_response_not_found` is also correct for the masking question: it is not matched by `is_previous_response_not_found_public_shape` (`app/core/errors.py:232-247`), so it reaches the client unmasked, and deliberately omitting `_HTTP_BRIDGE_PRE_SUBMIT_FAILURE_ATTR` is what keeps the turn off the raw-HTTP replay. The repository change is sound for the right reason: both racers serialize on the same `FOR UPDATE` account lock, the post-CAS read is fenced on `id` + `account_id` so a rebound row cannot be misreported as retired, and the "kept, not retired" early returns are preserved.

The blocker is that the same error code carries a second meaning the PR did not intend, and on the owner-forward path that meaning wins.

**Blocking: on a multi-instance deployment the 404 never reaches the client, but the retirement still happens.**

Given a bridged thread whose durable row is owned by instance B, an owner account in `PAUSED`, and a client resuming with its own `previous_response_id` and an account-scoped body (the 9-of-10 shape this PR measures):

1. Origin A forwards to B.
2. B runs the new `retire_dead_client_anchor` (`streaming.py:2322` on head), writes the `continuity_abandonment_scope='request_path'` marker, and raises the 404 at `streaming.py:2620-2626`. On the internal path that leaves B as a real HTTP 404 with the new code in the body — `_mask_previous_response_not_found_error` passes the prefixed code through with `default_status=404`.
3. A's forward client rebuilds the error at `http_bridge_forwarding.py:216-236`; `_owner_forward_error_payload` (`:786`) returns B's JSON verbatim, so the code survives the hop, and `on_response_rejected` sets `RECEIVER_REJECTED` (`owner_forwarding.py:474-477`).
4. `_owner_forward_failure_allows_local_recovery` returns True for `RECEIVER_REJECTED` (`owner_forwarding.py:193-199`), so A does not propagate.
5. `bridge_previous_response_not_found` is in the True set of `_http_bridge_should_attempt_local_previous_response_recovery` (`helpers.py:3485-3491`), so A sets `should_attempt_previous_response_recovery` and rebinds locally (`streaming.py:2803-2804` on main).
6. `recovery_payload = effective_payload`, and the anchor-injection block is gated on `previous_response_id is None`, so the client's own anchor is re-dispatched onto a fresh account that has never seen it. Upstream answers `previous_response_not_found`, which the API layer masks into a retryable 502 `stream_incomplete`.

Net: the client gets a retryable 502 instead of the terminal 404, the burst this PR exists to stop is not stopped, and the durable owner has been retired anyway by an instance whose verdict never reached the decision point. Worse, it does not self-correct — the owner is now gone, so subsequent turns route to a healthy account still carrying the client's anchor and keep producing the masked 502, where `main` at least returned a consistent `previous_response_owner_unavailable`. This is a new routing: on `main` B's 502 is not in that recoverable set, so A propagates.

This is not hypothetical for this repo. `http_responses_session_bridge_instance_id` defaults to the hostname (`app/core/config/settings.py:381`, factory at `:188`), so each container is its own bridge instance and owner-forward is live on any multi-replica deployment and during the overlap window of a blue/green swap.

CI did not catch it because the new test is single-instance: `test_v1_responses_http_bridge_reports_a_client_anchor_on_a_dead_owner_as_not_found` never constructs a `_HTTPBridgeOwnerForward`, and its asserting POST also omits `"stream": true`, so neither the hop nor native-streaming delivery of the new terminal code is exercised.

**Blocking: this silently removes the request-log row #2384 added for exactly this failure family.**

`streaming.py:1194-1196` writes the `request_logs` error row only when the propagating code is in `_HTTP_BRIDGE_CONTINUITY_OWNER_ERROR_CODES`, and that set (`streaming.py:304-309`) is exactly `{previous_response_owner_unavailable, continuity_owner_conflict}`. `63314a1b2` (#2384) added it two days ago with the note that bridge session creation is "the one proxy surface with no request-log write of its own". The new code is not in the set, and the new raise site also omits the pre-submit attr, so the diverted turns write nothing: `_logged_error_json_response` only calls `log_error_response`, and no middleware persists request rows. The nine-per-20h `account_scoped_input` turns go from one row each to zero, so the dashboard error rate for this family reads 0 while clients still fail — a false close on #1707. Do not simply add the code to the frozenset: `streaming.py:5298` raises the same code post-dispatch and would start double-logging. Tag the error at the new raise site and check that tag alongside the set.

**Blocking: an active OpenSpec delta now states the opposite of what ships.**

`openspec/changes/retire-continuity-owner-on-the-request-path/specs/sticky-session-operations/spec.md` is still unarchived on current `main`. Its requirement says retirement "MUST NOT be attempted when the client supplied its own `previous_response_id` ... and the turn MUST keep failing closed", with a "Client-supplied anchor" scenario reading "no retirement is attempted". This PR retires exactly that case and only adds requirements under `responses-api-compat`, so `openspec validate` passes while two active deltas contradict each other, and whoever archives that change writes a false requirement into the baseline.

**Note, not blocking:** `retire_continuity_owner_if_unavailable` applies `HARD_OWNER_UNAVAILABLE_STATUSES`, which includes `REAUTH_REQUIRED` (still in `ROUTABLE_STATUSES`) and `RATE_LIMITED`/`QUOTA_EXCEEDED` whose `reset_at` is routinely hours past a request budget. That is inherited from the previous change, not introduced here, but this PR raises its consequence from a silent rebind to a client-visible permanent severing: a ten-minute operator pause now terminally rejects every client-anchored thread that takes a turn in the window. The alternative is a retry storm, so this looks like the right trade — worth stating explicitly in `context.md` so the next reader does not rediscover it.

**To make this mergeable:**

1. Stop the owner-forward path from swallowing this rejection. Either raise a distinct code from `retire_dead_client_anchor` that is absent from the `_http_bridge_should_attempt_local_previous_response_recovery` set (and masked the same way at the API layer), or attach provenance that survives the hop and have `_owner_forward_failure_allows_local_recovery` return False for it. Note that `local_pre_dispatch_refusal` as it stands is only carried (`owner_forwarding.py:188`) and never read by any bridge forward-failure decision, so it needs a consumer if you go that route.
2. Alternatively, defer the retirement to the instance that can actually deliver the answer, so the irreversible write never happens on a hop whose verdict is discarded.
3. Add an owner-forward integration test: durable row owned by a peer instance, paused owner, client-supplied anchor — assert the client receives 404 `bridge_previous_response_not_found` and that no local rebind event is logged.
4. Add a `"stream": true` variant of the new single-instance test so native-streaming delivery of the terminal code is covered.
5. Restore the request-log row for the diverted shape without double-logging the post-dispatch raise at `streaming.py:5298`: set a dedicated attr on the error built in `retire_dead_client_anchor`, check it alongside `_HTTP_BRIDGE_CONTINUITY_OWNER_ERROR_CODES` at `streaming.py:1194`, and assert exactly one `request_logs` row with `error_code == "bridge_previous_response_not_found"` using the `drain_persistence_tasks` + `select(RequestLog)` pattern already in this test file.
6. Amend the `retire-continuity-owner-on-the-request-path` delta in this PR: narrow the MUST-NOT to the silent rebind, which does stay injected-anchor-only, and rewrite the "Client-supplied anchor" scenario to point at the new terminal rejection.
7. Reconcile `openspec/changes/reject-a-dead-client-anchor/tasks.md:23-27` — 3.1-3.4 are unchecked while the PR body reports all four as passing.

Issue #1450, opened 2026-07-22
bug(proxy): compacted/replayed Codex histories can enter permanent 400 loops on item_* ids and namespace


yjx-git001:
### Pre-flight checklist

- [x] I searched existing issues for the exact upstream errors, `item_*` replay IDs, `namespace`, remote compact, and compaction failures.
- [x] This is not the already-fixed `0 from 0 compaction output items` shape from #998 / #971.

### codex-lb version

Observed on `main` at `e5b101f20299` (Docker image built from that commit).

The production deployment has since moved to `ed017d52876a`. A source comparison through the current `main` head found no targeted normalization change for the input fields described below, but I am not claiming a clean deterministic reproduction on the newer commit yet.

### Deployment method

Docker Compose, with an OpenAI-compatible gateway in front of codex-lb.

### Client used against codex-lb

Codex app / native Codex Responses traffic.

### ChatGPT account plan(s) involved

Plus and Team account pool.

### Model(s) involved

- `gpt-5.6-sol`
- `gpt-5.6-luna`
- `gpt-5.6-terra`

### What happened?

Some users report that context compaction fails and the conversation cannot recover afterward. The server-side failure is not the old ?expected exactly one compaction output item, got 0? response-shape bug. Instead, a long/replayed history is rejected upstream because one historical input item carries either an invalid local-looking ID or an unsupported `namespace` field.

Sanitized exact errors:

```text
Invalid 'input[48].id': 'item_<redacted>'. Expected an ID that begins with 'rs'.
Invalid 'input[173].id': 'item_<redacted>'. Expected an ID that begins with 'rs'.
Invalid 'input[26].id': 'item_<redacted>'. Expected an ID that begins with 'rs'.
Invalid 'input[221].id': 'item_<redacted>'. Expected an ID that begins with 'rs'.
Unknown parameter: 'input[171].namespace'.
Unknown parameter: 'input[179].namespace'.
```

Aggregated codex-lb `request_logs` evidence:

| Failure shape | Attempts | Models | Behavior |
|---|---:|---|---|
| `input[48].id = item_*` | 23 | gpt-5.6-sol | same item rejected repeatedly for ~3.5 min |
| `input[173].id = item_*` | 14 | gpt-5.6-sol | same item rejected repeatedly for ~2 min |
| other `item_*` IDs | 3 | gpt-5.6-terra | indices 26 and 221 |
[...]

PR #2424, opened 2026-09-14
fix(proxy): preserve replayed tool call namespaces

HulianBuligon:
## Summary

Preserve the (namespace, name) executor identity of replayed ChatGPT
Responses tool calls across standard, Lite/compact, HTTP bridge, and
Responses WebSocket paths.

The shared serializer previously removed namespace globally. A replayed
mcp__cua_repl.js could therefore arrive as the unrelated default
functions.js executor after reconnect or compaction. OpenAI-compatible model
sources keep their existing compatibility projection and still receive
recognized replay items without namespace.

## Type of change

- [x] fix: — bug fix
- [ ] feat: — new user-facing feature or capability
- [ ] refactor: — internal refactor
- [ ] docs: — documentation only
- [ ] chore: / ci: / build: — tooling, CI, packaging
- [ ] test: — test-only change
- [ ] Breaking change

Linked issue: none — regression fix from executor-routing diagnosis.

## OpenSpec

- [x] This PR touches a codex-faithful path and preserves the intended
      upstream executor identity

Change directory: openspec/changes/preserve-replayed-tool-namespaces/

## Changes

- Disable namespace stripping only for ChatGPT-facing standard and
  compact/Lite serializers and replay-safety projections.
- Keep the existing explicit namespace sanitizer at the configured
  OpenAI-compatible model-source dispatch boundary.
- Preserve top-level namespace tool declarations and all remaining call
  fields.
- Add serializer, compact, HTTP bridge, and WebSocket regression coverage,
  including repeated call IDs and malformed permissive input.
- Document the boundary contract and rollback shape in OpenSpec.

## Behavior example

Input history:

~~~
{"type":"custom_tool_call","namespace":"mcp__cua_repl","name":"js","call_id":"c1","input":"pwd"}
~~~

ChatGPT standard/Lite/compact and WebSocket payloads retain
namespace: "mcp__cua_repl". A configured OpenAI-compatible source receives
the existing sanitized shape without that field; no other call field is
removed by the namespace projection.

## Test plan

~~~text
uv run --no-sync pytest -q tests/unit/test_openai_requests.py tests/unit/test_proxy_utils.py
# 1593 passed, 1 warning

uv run --no-sync pytest -q tests/integration/test_proxy_websocket_responses.py
# 154 passed, 1 warning

uv run --no-sync ruff check app/core/openai/requests.py tests/unit/test_openai_requests.py tests/unit/test_proxy_utils.py tests/integration/test_proxy_websocket_responses.py
uv run --no-sync ruff format --check app/core/openai/requests.py tests/unit/test_openai_requests.py tests/unit/test_proxy_utils.py tests/integration/test_proxy_websocket_responses.py
uv run --no-sync ty check
# all passed

npx --yes @fission-ai/openspec@1.11.0 validate --specs --strict
npx --yes @fission-ai/openspec@1.11.0 validate preserve-replayed-tool-namespaces --strict
# 65 specs passed; change is valid

git diff --check
# clean
~~~

The consolidated local bundle was also exercised with 2748 affected unit
tests, 555 affected integration tests, frontend settings tests/typecheck/lint,
and migration policy/schema-drift checks. No production deploy or runtime
configuration change was performed.

The broader `make test-unit` run completed with 10313 passed, 99 skipped, and
1 xfailed, plus the pre-existing environment-sensitive failure
`tests/unit/test_codex_body_capture_guards.py::test_output_under_a_temporary_filesystem_is_refused`.
That test is unchanged by this PR and fails on the current `main` baseline as
well; the focused patch and bridge gates above pass.

The first GitHub Actions run passed every required backend, migration, build,
and static gate, but its Linux `Dashboard browser smoke (Playwright)` job had
one transient responsive overflow observation (`scrollWidth=1488`, expected
`<=1440`). The same smoke passed locally (8/8) and then passed on a second
GitHub Actions run (8/8) for the unchanged commit; this PR has no frontend
diff.

## Screenshots / output

No dashboard UI is changed. The regression output is the serialized request
boundary described above; no live model continuation is claimed by this PR.

## Checklist

- [x] Added OpenSpec proposal, context, design, requirements, and task
      checklist.
- [x] Added regression tests at serializer, bridge, compact, and WebSocket
      boundaries.
- [x] Kept model-source compatibility sanitization explicit and isolated.
- [x] Ran the relevant test, lint, type, diff, and OpenSpec gates.
- [x] No production deploy or runtime configuration change was performed.
- [x] CHANGELOG was not edited by hand.


Soju06:
Thanks for the detailed write-up, but I have to close this: the stated premise is inverted, and merging it would re-open a reproduced production outage.

**The sanitizer was never a "generic model-source compatibility projection".** It was added by #1578 (commit `7c456df73`) specifically for the **ChatGPT** wire, because the ChatGPT Responses upstream rejects the field:

```
{"type":"error","error":{"type":"invalid_request_error","message":"Unknown parameter: 'input[1].namespace'."}}
```

Issue #1450 recorded 44 upstream 400s from real Codex clients (`useragent_group=Codex`, `upstream_transport=http`, `upstream_error_code=unknown_parameter`) on `input[171].namespace` / `input[179].namespace`, with the same deep history item rejected on every retry — a permanent, unrecoverable conversation. #1578 included a live before/after on the WebSocket route showing the same request going from that 400 to `response.completed`. The explicit `strip_replayed_tool_call_namespaces_from_payload(source_payload)` call in `app/modules/proxy/api.py:5281` was the **additional** site, added only because `_shape_source_responses_payload` builds from `model_dump_for_forwarding()` and therefore bypasses `to_payload()`. So this PR keeps the boundary that never had the problem and removes the one that did.

**Blast radius.** `to_payload()` is the serializer for every ChatGPT egress path, not a narrow one: `http_bridge/request_submit.py:709,775`, `http_bridge/streaming.py:1002,1323`, `websocket/mixin.py:3207,3211`, `_service/compact.py:622`, `api.py:6382`. Flipping the flag at `app/core/openai/requests.py:768` and `:1018` re-exposes all of them to the #1450 failure. After the change, no caller in `app/` uses the `True` default any more (`requests.py:768`, `:773`, `:1018` all pass `False`), which is the giveaway that this is a revert rather than a boundary separation.

**The local-identity motivation is already satisfied on main.** #1578 also added `to_replay_safety_payload()` (`requests.py:772-773`) with stripping disabled, exactly so that replay-safety classification, ownership and dedupe see the namespaced history — see `http_bridge/streaming.py:570`, `streaming/retry.py:193,921,937`, `affinity.py:521`. The validated `request.input` is likewise never mutated, and top-level `type: "namespace"` tool declarations are forwarded byte-identically (covered by `test_responses_to_payload_preserves_reserved_namespace_tool_byte_identical`). Your diff's change to `to_replay_safety_payload` is pure reformatting.

**The claimed defect is unverified.** There is no linked issue, no request log, and no upstream transcript showing a replayed `mcp__cua_repl.js` being executed as `functions.js`; the PR body states "no live model continuation is claimed by this PR" and "no production deploy or runtime configuration change was performed". Green CI here does not help, because the change works by inverting the #1578 guards (`tests/integration/test_proxy_websocket_responses.py:2818`, `tests/unit/test_proxy_utils.py:12277`, `:28379`, `:28473`) rather than by demonstrating upstream acceptance.

If you believe the executor-routing problem is real, here is what would make this reviewable:

1. An upstream transcript on a current model (e.g. `gpt-6-astra`) showing a replayed history item with `namespace` accepted with 200 — plus an explanation of why #1450's `unknown_parameter` 400s no longer occur. A single `response.completed` on one account is not enough; #1450 showed the rejection is contract-dependent.
2. Evidence of the misrouting itself: a request log or upstream frame where a namespaced replayed call is executed under the wrong executor. Note that the *tool declarations* (top-level `type: "namespace"` entries, or the `additional_tools` item in Responses-Lite) are forwarded untouched, so the upstream registry still has the namespace — please show concretely where the history item's `namespace` changes dispatch.
3. If both turn out to be true, the fix must be conditional, not a blanket revert — e.g. preserve `namespace` only when the matching top-level `namespace` tool declaration is present in the same request, behind a setting that defaults to today's stripping behaviour, so a `400 Unknown parameter` cannot come back as the default.

Closing for now; happy to reopen the discussion on a new PR carrying that evidence.

PR #2439, opened 2026-09-15
fix(proxy): preserve upstream quota reset metadata on terminal errors

HulianBuligon:
## Summary

Small standalone follow-up to #2207, requested in @Soju06's closing feedback on that PR. #2207 was closed without being merged, so this PR carries only the narrow piece that was explicitly asked to be extracted from it — no setting, no migration, and no affinity or continuity changes.

Two related terminal-error paths lost upstream quota information:

1. **Streaming Responses terminal errors** dropped the upstream `resets_at` value when a retryable stream error was surfaced as `response.failed`.
2. **HTTP bridge** rewrote the original upstream terminal into a synthetic `502 stream_incomplete` whenever a pre-created replay was staged and then failed — including when the original upstream terminal was a `429` quota response.

This PR keeps the original terminal (status + reset metadata) in both cases. `resets_in_seconds` is not part of `response_failed_event`; only `resets_at` is carried, matching the existing event contract and the closing feedback on #2207.

## Type of change

## Changes

- `app/modules/proxy/_service/streaming/retry.py`: add `_response_failed_event_from_upstream_error`, which builds the terminal `response.failed` event while forwarding the upstream `resets_at` (coerced defensively: booleans and non-numeric values become `None`). Both exhaust sites that previously inlined `response_failed_event(...)` now use it. The existing `error_message` locals keep their exact previous values and are still computed the same way.
- `app/modules/proxy/_service/http_bridge/upstream_events.py`: when a pre-created replay was **staged/attempted and failed**, preserve the upstream response if the payload status is `429` **and** the retry error code is one of `rate_limit_exceeded`, `usage_limit_reached`, `insufficient_quota`, `usage_not_included`, `quota_exceeded`. In that case the sticky error overrides are cleared via the existing `_clear_websocket_request_error_overrides`, so the original 429 plus reset metadata reaches the client.
- Every other failed-replay case is unchanged: it still fails closed as a synthetic `502 stream_incomplete`.

## Fail-closed behavior preserved

The preserve branch is deliberately narrow. It requires **both** a `status`/`status_code` of `429` in the upstream payload **and** a recognized quota error code. Replay failures without a status-bearing quota terminal (transport errors, malformed payloads, non-quota codes) keep the previous fail-closed `502 stream_incomplete` terminal.

## Scope / non-changes

- **No settings changes** — no new or modified setting, no change to `check_settings_tiers` (reported `96/96` fields, tiers unchanged).
- **No migrations** — no schema or `alembic` change.
- **No general failover behavior change** — routing, account selection, retry budgets, delays, affinity/sticky, continuity, and failover decisions are untouched. The only change is which terminal event/payload is forwarded downstream once a failure is already terminal.
- No `CHANGELOG.md` edit (release-please owns it).

## Test plan

```
uv run pytest -q \
  tests/unit/test_proxy_utils.py::test_response_failed_event_from_upstream_error_preserves_reset_metadata \
  tests/integration/test_http_responses_bridge.py::test_v1_responses_http_bridge_preserves_rate_limit_after_failed_precreated_retry \
  tests/integration/test_http_responses_bridge.py::test_v1_responses_http_bridge_preserves_rate_limit_metadata_in_429 \
  tests/integration/test_proxy_transient_retry.py::test_stream_rate_limit_on_last_attempt_returns_actual_error
# 4 passed
```

Focused tests added/updated:

- `tests/unit/test_proxy_utils.py::test_response_failed_event_from_upstream_error_preserves_reset_metadata` — asserts the terminal event carries `resets_at` and does not invent `resets_in_seconds`.
- `tests/integration/test_http_responses_bridge.py::test_v1_responses_http_bridge_preserves_rate_limit_after_failed_precreated_retry` — two-account scenario: the first upstream raises a 429 quota error, a replacement retry is staged and then fails (503). Asserts the client still receives the original `429` `rate_limit_exceeded` envelope with `plan_type` and `resets_at` (not a `502 stream_incomplete`), and that exactly two upstream connects happened.
- `tests/integration/test_proxy_transient_retry.py::test_stream_rate_limit_on_last_attempt_returns_actual_error` — extended fixture with `resets_at` and asserts the client-visible error keeps it.

Full touched suites also green:

```
uv run pytest -q tests/integration/test_http_responses_bridge.py   # 197 passed
uv run pytest -q tests/integration/test_proxy_transient_retry.py   # 49 passed
uv run pytest -q tests/unit/test_proxy_utils.py                    # 1409 passed
```

Other gates:

```
git diff --check          # clean
uv run ruff check <5 files>        # All checks passed
uv run ruff format --check <5 files>  # 5 files already formatted
uv run ty check           # All checks passed
make architecture-check    # proxy architecture / settings tiers / cancellation safety / timing seams all passed
```

## Verification of the regression

The two new/extended assertions were confirmed to reproduce the bug: temporarily reverting only the two source files (checksum-verified restore) makes `test_v1_responses_http_bridge_preserves_rate_limit_after_failed_precreated_retry` fail with `status=502 code="stream_incomplete"`. With the fix, it returns the original 429 envelope.

## OpenSpec

- [x] This PR touches a codex-faithful path (SSE terminal framing) and preserves upstream-equivalent behavior — it restores the upstream status/reset metadata rather than synthesizing a replacement.

## Simplicity

No new setting, no README section, no dashboard nav item, and no default changed. Sections above record this explicitly.


HulianBuligon:
Addressed in d4a3280d.

1. **Missing post-refresh terminal**: `app/modules/proxy/_service/streaming/retry.py` (~1151) now renders through `_response_failed_event_from_upstream_error`, so the exhausted same-account retry keeps the upstream `resets_at`. It was the only one of the three post-refresh terminals still calling `response_failed_event(..., error_message)`.

2. **Stale unit contract**: `test_process_http_bridge_upstream_text_masks_failed_replay_usage_limit` was asserting the pre-change behavior and failing CI. It is now `..._preserves_failed_replay_usage_limit`, asserting the upstream 429 code, type, `resets_at`, `resets_in_seconds`, and `error_http_status_override == 429` with no residual code/message override. A new guard, `..._masks_failed_replay_without_quota_status`, pins that replays without a status-bearing quota terminal still fail closed as 502 `stream_incomplete`.

All three new/updated tests fail against their corresponding mutants, so none is vacuous. Verification: `tests/unit/test_proxy_http_bridge.py` 1065 passed, `tests/unit/test_proxy_utils.py` 1410 passed, `tests/integration/test_http_responses_bridge.py` 197 passed, `tests/integration/test_proxy_transient_retry.py` passed, plus `ruff check`, `ruff format --check`, `ty check`, `git diff --check` clean. The contract change is documented in `openspec/changes/preserve-upstream-quota-terminals/` (`openspec validate --specs`: 65 passed; strict change validation: valid).

Komzpa:
@codex review

PR #2449, opened 2026-09-16
fix(proxy): read the usage-limit rejection off the frame upstream actually sends

Soju06:
## Summary

When an account's window is spent, upstream says so in two shapes: an HTTP body carrying `usage_limit_reached`, and a serialized `response.failed` frame that carries **no error code at all**. Both the retry decision and the **account-health write** gate on a pure code table, so the second shape is never recognised.

Measured on three accounts inside one request:

```
coded frame:      {a: rate_limited, b: rate_limited, c: rate_limited}
code-less frame:  {a: rate_limited, b: rate_limited, c: ACTIVE}
```

**The account that just said its window was spent comes out the pool's healthiest, and is selected first next time.** It bites the last account of a walk — `allow_retry_flag = attempt < max_attempts - 1` — and, separately, any account whose frame arrives after a visible event.

## Type of change

- [ ] **Breaking change**

## Every transport, because the same table is consulted three times

- **HTTP streaming** — `_should_retry_stream_error` and `_should_penalize_stream_error` now fall back to the message predicate when the code table has no opinion. The remaining call sites all had an upstream-authored message in scope, so the predicate cannot answer differently per path.
- **WebSocket, account health** — same gate, same fix.
- **WebSocket, transparent replay** — found while covering the above, and the larger half of the defect. The replay decision was a *second* pure code-table lookup, duplicated across `_websocket_precreated_retry_error_code` and `_websocket_owner_pinned_quota_error_code`: a code-less usage-limit frame was surfaced on the spent account instead of retiring the socket and releasing the turn. Both now route through one predicate.
- **Usage refresh** — triggered on the rejection rather than the literal code. Without this the transports disagreed: WebSocket canonicalizes the replay code to `usage_limit_reached` and would refresh, HTTP reaches the health write as `upstream_error` and would not.

## Scope, deliberately narrow

The message override applies **only** to an envelope that carries no code. A code with its own classification decision keeps it. `invalid_request_error` is excluded on purpose: that envelope is how upstream rejects a request, and a rejection frequently quotes the request back — reading the phrase out of client-supplied content would bench a healthy account, and benching is the expensive direction.

Nothing here touches account selection, the pool walk or the exclusion decision. This is about *where* the classification is consulted, not what it decides.

## Verification

| check | result |
|---|---|
| `tests/unit` | `10385 passed, 4 skipped` |
| touched integration files | `217 passed` (`test_proxy_transient_retry.py` + `test_proxy_websocket_responses.py`) |
| `tests/unit/test_proxy_utils.py` | `1415 passed` |
| `ruff check` / `format --check` | clean |
| line ceilings | held exactly — `service.py` 2599/2600, `streaming/mixin.py` 1100/1100 |

A full `tests/integration` run had two non-passes and one error, all triaged and none from this diff: two pass in isolation (one is a load-induced pytest-timeout whose dump shows only idle worker threads) and one fails on `built dashboard assets missing; run cd frontend && bun run build`.

**Seven mutations, all caught**, including reverting each gate to its pure code table and deleting the passive `"usage limit has been reached"` marker — the wording this repository's own fixtures carry 20+ times, which an earlier attempt at this predicate missed entirely.

The unit coverage needed a new harness: the rest of `test_proxy_utils.py` stubs `_stream_once`, which is exactly where both gates live, so this behaviour could only ever be caught by the integration suite. `_run_stream_terminal_frame_walk` drives the real `_stream_once` through `core_stream_responses` and puts the frame on the wire.

## OpenSpec

`openspec/changes/classify-usage-limit-terminal-frames/` — the requirement being modified is the delivery-form clause: the match must not depend on the HTTP status, because upstream sends the same rejection as an HTTP body and as a status-less frame. The previous wording said that and only half the delivery forms implemented it.

## Known gap

The HTTP bridge's last-account-of-a-walk position is not covered by this change; the requirement's scenario names it, and the two transports here satisfy it. Recorded rather than left implicit.


Komzpa:
@codex review

PR #2460, opened 2026-09-18
fix(dashboard-users): take the owner row before the owned-key cascade reads it

Soju06:
## The defect

On PostgreSQL an API key could survive **active on a disabled owner**. That is the
`owner_disabled` cascade — the thing that stops a leaver's proxy credentials the
moment their account is turned off, and the mechanism the SCIM deprovisioning path
(#2431) delegates to.

Two writers, taking the same two rows in opposite orders:

| | reads | then writes |
|---|---|---|
| disable (`PATCH /api/dashboard-users/{id}` `status=disabled`, SCIM `active:false`, `DELETE`) | `api_keys` — *which keys are active* | those keys, then `dashboard_users` |
| key page (`PATCH /api/api-keys/{id}` `isActive: true`) | `dashboard_users` — the owner's status, `FOR UPDATE` | the key row |

Under READ COMMITTED neither read is repeated. The cascade's scan takes its own
statement snapshot, so a key the PATCH is about to switch on is still `is_active =
false` and matches nothing — and because the row does not match, PostgreSQL never
locks it and never re-evaluates. The PATCH's `FOR UPDATE` on the owner is honest but
too late: it only makes the *disable's* `UPDATE dashboard_users` wait, which happens
after the cascade has already run. Both commit, and the end state is a disabled owner
with a live key.

Reduced to raw SQL on a throwaway `postgres:16`:

```
T_disable                                        T_patch
BEGIN;
UPDATE api_keys SET is_active=false
  WHERE owner_id='bob' AND is_active IS TRUE;    -- UPDATE 0 (key is off right now)
                                                 BEGIN;
                                                 SELECT status FROM owners
                                                   WHERE id='bob' FOR UPDATE;   -- 'active'
UPDATE owners SET status='disabled' ...          -- waits
                                                 UPDATE keys SET is_active=true ...; COMMIT;
COMMIT;
=> owner_status | key_active
   disabled     | t
```

SQLite hides it: `BEGIN IMMEDIATE` serialises the writers, so the two never interleave.

**Pre-existing on `main`**, found while reviewing an unrelated PR and confirmed with
`app/` checked out at `origin/main` against a fresh PostgreSQL database:
`tests/integration/test_dashboard_users_api.py::test_key_reactivation_never_leaves_active_keys_on_a_disabled_owner`
fails at the API-key half (`assert (await _key(key.id)).is_active is False` →
`assert True is False`) and passes on SQLite.

## Why CI missed it

`tests/integration/test_dashboard_users_api.py` is in no `POSTGRES_PYTEST_TARGETS`
entry, and the test is green on SQLite — so the integration job and the PostgreSQL
job were both green on a defect neither of them ran in the configuration that shows it.

## The fix

The campaign rule for this class is that a risk gate takes its write intent **before
its first read** and does not commit between the count and the write. The
key-reactivation path diverged from it in the direction nobody looks at: the
*cascade's* read of `api_keys`, not the restore.

- `DashboardUsersRepository._lock_owner()` takes the owner's row (`FOR UPDATE` on
  PostgreSQL, a no-op on SQLite, where the callers already hold the database write
  lock), and `deactivate_owned_keys()` calls it before its scan. Now both writers
  take the owner row first: the loser of that row reads the winner's committed state,
  so either the PATCH is refused with `409 owner_disabled` or the cascade's fresh
  snapshot sees the key it just turned on and turns it back off.
- `reactivate_owner_disabled_keys()` takes the same row. Its conditional `UPDATE ...
  WHERE EXISTS (owner active)` was already there and was not enough on its own: the
  predicate is evaluated against a snapshot that a disable committing a moment later
  is invisible to.
- The API-key gate moves ahead of the first field assignment in
  `ApiKeysRepository.update()`. Any assignment makes the key row dirty, and the next
  statement autoflushes its `UPDATE` — which would take the key row *before* the owner
  row. With the cascade now taking the owner first, that inversion is a deadlock
  rather than a refusal, and it reproduces: racing `DELETE /api/dashboard-users/{id}`
  (which clears `owner_user_id` on every owned key, inactive ones included) against
  `{"isActive": true, "name": ...}` gives `asyncpg.exceptions.DeadlockDetectedError`
  without the hoist and passes with it. The gate itself is unchanged.

A conditional `UPDATE` was preferred where one fits — the restore keeps its
`EXISTS (owner active)` predicate — but it cannot carry this race by itself: the
losing side's row simply is not visible as a candidate, so there is nothing for
PostgreSQL to lock or re-check. The owner row is the only object both writers touch.

## Sibling audit

| path | verdict |
|---|---|
| `POST /api/dashboard-users/{id}/reactivate-keys` | already correct — holds `acquire_write_intent()` from before its first read through the restore's commit, so it is serialised with every account mutation. This is the discipline the other paths were missing. |
| `PATCH /api/api-keys/{id}` `isActive: true` | **the defect.** Fixed by ordering the cascade behind the owner row, plus the gate hoist above. |
| SCIM `PATCH /scim/v2/Users/{id}` `active: true` (`reactivate_user`) | **same class.** It commits the status flip and then restores the keys in a *second* transaction, releasing the write intent in between, so the restore's `EXISTS` could read an owner a concurrent disable was about to turn off. Fixed by the `_lock_owner()` in `reactivate_owner_disabled_keys()`. Its window is one statement wide; 36 gathered rounds did not land on it, so it gets no test of its own — it shares the primitive the dashboard-route test does reproduce. |
| owner cascade on disable (`update_user`), SCIM `active:false` (`deactivate_user`), `DELETE /api/dashboard-users/{id}` | all route through `deactivate_owned_keys()` and are covered by the same lock. |
| key creation | never assigns `owner_user_id`; not affected. |

## Spec

No change directory: the guarantee already exists and this restores it.

- `openspec/changes/manage-dashboard-users-and-invites/specs/api-keys/spec.md` — "the
  owner's status MUST be read under a lock in the same transaction as the write
  (SQLite write-intent transaction, PostgreSQL `FOR UPDATE` on the owner row) **so a
  concurrent disable cannot leave an active key on a disabled account**".
- `openspec/changes/manage-dashboard-users-and-invites/specs/dashboard-users/spec.md`
  — *Scenario: a disable racing a key reactivation*: "the end state is consistent:
  either the owner is active with its keys restored, or the owner is disabled and
  every owned key is inactive with reason `owner_disabled`".

`uv run openspec validate --specs` → 66 passed, 0 failed.

## CI hole

Three node ids from `tests/integration/test_dashboard_users_api.py` join
`POSTGRES_PYTEST_TARGETS`: the two key-race tests and
`test_concurrent_admin_mutations_keep_exactly_one_admin` (the other PostgreSQL-sensitive
concurrency test in the file, which CI also never ran on PostgreSQL).

Node ids rather than the whole file, for cost: the file is **173s** on PostgreSQL for
42 tests, ~19% on top of a job that takes 925s; the three node ids are **~19s of test
time (27s with fixtures), ~2%**. The file's other 39 tests are dialect-independent
gate and audit assertions already covered by the integration job. This matches how the
rest of `POSTGRES_PYTEST_TARGETS` is written — mostly node ids, whole files only where
the file is PostgreSQL-specific throughout.

## Verification

- `test_key_reactivation_never_leaves_active_keys_on_a_disabled_owner`: fails on
  `main` under PostgreSQL, passes here — 5 consecutive runs green.
- `uv run pytest tests/integration/test_dashboard_users_api.py` → 42 passed on
  PostgreSQL (173s) and 42 passed on SQLite (88s).
- `make test-postgres` with the addition in place, against a throwaway `postgres:16`
  → **248 passed in 924.87s**.
- Adjacent suites on SQLite: API keys + key secret responses + break-glass policy
  (139 passed), SCIM users (47 passed), dashboard-users schema + operator/viewer
  presets + permission gates + audit actor attribution + role-mapping re-evaluation
  (65 passed).
- `uv run ruff check . && uv run ruff format --check . && uv run ty check` → clean.
- `uv run python scripts/check_migration_topology.py` → passed (no migration; 261
  revisions, +0 on this checkout).


Soju06:
Heads-up on the three red checks — **Lint (ruff)**, **Migration check (alembic)** and **Migration check (alembic, PostgreSQL)** are `main`'s, not this PR's.

`main` at [`9637bdee3`](https://github.com/Soju06/codex-lb/commit/9637bdee3) (#2422) has a forked alembic graph, and the same three checks are red on that commit itself:

```
ERROR: alembic_head_count_invalid expected=1 actual=2
  heads=20260914_000000_add_scim_tokens,20260914_000000_drop_subscription_overflow_schema
ERROR: alembic_timestamp_prefix_collision prefix=20260914_000000 count=2
```

Both revisions carry `down_revision = "20260913_000000_add_oidc_provider_flow"` — #2431 and #2422 were authored in parallel and took the same timestamp slot. CI builds the merge commit, so this branch inherits it.

This PR adds no revision: `scripts/check_migration_topology.py` on the branch reports `revisions: 261; head: 20260914_000000_add_scim_tokens; base-ref check: origin/main (+0 revision(s) on this checkout)` and `make lint` is clean locally. The fix belongs on `main` (re-stamp the newer revision and re-point its `down_revision`, per `scripts/rewrite_alembic_revisions.py`), and these checks should go green here once it lands.

Soju06:
Evidence for the point above, from this run:

- **Tests (pytest, PostgreSQL)** — `20 failed, 228 passed`. Every one of the 20 is `alembic.util.exc.CommandError: Multiple head revisions are present` / `ValueError: too many values to unpack (expected 1)` in a migration test. The three node ids this PR adds to `POSTGRES_PYTEST_TARGETS` are **not** among them — they are in the 228 that passed.
- **Tests (pytest, unit)** — `36 failed, 10391 passed`, all `MultipleHeads` in `tests/unit/test_db_migrate.py`, `test_db_rate_limiter.py`, `test_telemetry_migration.py`.
- **Lint (ruff)** — fails in `make lint` at `scripts/check_migration_topology.py`, which is the check that exists to catch exactly this fork.

Locally, against a throwaway `postgres:16` with this branch's single-head graph, `make test-postgres` is **248 passed in 924.87s**.

Soju06:
Full accounting now that every job has reported — **all 8 red checks share the one cause** (`main`'s forked alembic graph), and none of the failures is in code this PR touches:

| check | failures | cause |
|---|---|---|
| Lint (ruff) | `make lint` → `check_migration_topology.py` | `alembic_head_count_invalid`, `alembic_timestamp_prefix_collision` |
| Migration check (alembic) / (PostgreSQL) | `make migration-check` | `upgrade head` → MultipleHeads |
| Tests (pytest, unit) | 36 of 10427 | all MultipleHeads in `test_db_migrate` / `test_db_rate_limiter` / `test_telemetry_migration` |
| Tests (pytest, PostgreSQL) | 20 of 248 | all MultipleHeads; the 3 node ids this PR adds all passed |
| Tests (pytest, integration-core-2) | 56 of 881 | 52 `CommandError: Multiple head revisions`, 1 `ValueError: too many values to unpack (expected 1)` (the single-head unpack), 3 `AssertionError` that are the same fact — `test_the_credential_revisions_are_on_the_single_head_path` compares head lists, and the two `test_migration_serialization` CLI cases assert on a stdout the failed `upgrade head` never produced |
| Tests (pytest, integration-core-3) | 9 of 1088 | same, plus `test_overflow_transport_merge_is_the_only_head_with_both_original_parents` → `assert 2 == 1` |
| Dashboard browser smoke / Helm smoke | app startup migrates a database | MultipleHeads |

Nothing in `app/modules/dashboard_users/`, `app/modules/api_keys/` or `tests/integration/test_dashboard_users_api.py` fails. This PR adds no revision. Re-running will not help until `main`'s graph is single-headed again.

PR #2462, opened 2026-09-18
fix(db): repair September migration lineage

aacarcrash:
## Summary

- restamp the later, unreleased overflow-retirement migration
- chain it after the SCIM migration to restore one Alembic head
- add a SQLite upgrade, downgrade, and re-upgrade regression
- update both OpenSpec changes with the repaired revision identity

## Root cause

The SCIM migration and overflow-retirement migration used the same timestamp slot and the same parent. Alembic therefore reported MultipleHeads during startup. The repository topology guard also rejects a merge revision for this unreleased collision, so the later migration must be restamped and reparented.

## Verification

- migration topology guard: passed
- Ruff: passed
- strict OpenSpec validation: passed
- focused migration tests: 5 passed
- full OpenSpec spec validation: 66 passed
- disposable production-sized SQLite copy: upgraded to 20260918_171648_drop_subscription_overflow_schema
- database policy and schema drift check: passed
- live startup on port 2456: health, dashboard HTML, and dashboard JavaScript returned HTTP 200

No production database was stamped or edited manually.


Soju06:
Thank you for this — the diagnosis was correct and it was on the tracker within four hours of the break, which is faster than the incident deserved.

The repair landed as #2461 instead: a standard `alembic merge heads` revision (`20260918_000000_merge_scim_and_overflow_heads`) plus a topology-checker amendment so a collision that has *since been converged* stops being reported as a live fork. `main` is at `09a140fa9` and the graph has one head again.

Why the merge revision rather than the restamp, concretely:

- Both revision ids are already on `main`, and as of #2461 the merge revision's `down_revision` tuple names them directly. Restamping `20260914_000000_drop_subscription_overflow_schema` now would leave that tuple pointing at an id no longer in the graph, so the repair would have to be undone first.
- The restamp is only safe under the premise in your `context.md` — "no production database stamp or row needs manual repair." That holds for anything that upgraded through `head`, which has been refusing with `MultipleHeads` since the fork. It does not hold for an install that targeted the drop revision explicitly, and there is no way to establish from here that none exists. `alembic merge heads` needs no such premise: every already-stamped install, on either branch, walks forward.

None of that is a criticism of the branch — with the timestamp-slot rule as written, restamping is the documented authoring-time remedy. It is the wrong remedy specifically *after* the id is published, which is the distinction #2461's checker change now encodes.

**One piece here is worth keeping and is not in `main`:** `test_scim_and_overflow_retirement_lineage_is_single_and_round_trips`. #2461 shipped no regression coverage for the convergence, so a focused PR on current `main` that asserts a single head and round-trips the merge revision across both branches would be a real addition. Re-anchor it at `20260918_000000_merge_scim_and_overflow_heads` and drop the restamp and the lineage change folder, and I will review it.

Closing as superseded by #2461. Sorry it sat for six days.

